PR #12, opened 2018-04-06
vendor: Add

### LINKED ISSUE #5: [go] Establish vendor dependencies for go
I am trying to build manager and other binaries from source code but it seems that we do not import all vendors. I think we should use dep or glide to establish vendor dependencies.

gaocegege:
Close #5 

Using dep to install all vendor dependencies.

/assign @YujiOshima


Gopkg.toml
```diff
@@ -0,0 +1,74 @@
...
+#   source = "github.com/myfork/project2"
+#
+# [[override]]
+#  name = "github.com/x/y"
+#  version = "2.4.0"
+
+
+[[constraint]]
+  name = "github.com/docker/docker"
+  version = "1.13.1"
```
AkihiroSuda: Note: this version is no longer maintained. I'd suggest syncing with the master.

gaocegege: @AkihiroSuda Thanks for your advice, I will update it.

gaocegege:
@AkihiroSuda 

Addressed. PTAL

Gopkg.toml
```diff
@@ -0,0 +1,74 @@
...
+#   source = "github.com/myfork/project2"
+#
+# [[override]]
+#  name = "github.com/x/y"
+#  version = "2.4.0"
+
+
+[[constraint]]
+  name = "github.com/docker/docker"
+  branch = "master"
```
AkihiroSuda: nit: I guess you don't need to retain this constraint entry for the master branch

YujiOshima:
Thank you for great work!

Each docker image is still using `go get` packages.
We should update Dockerfiles to use the vendored packages.

YujiOshima:
@gaocegege Absolutely, It can reduce image size so much and accelerate images build.

gaocegege:
I opened an issue to keep track of the process https://github.com/kubeflow/hp-tuning/issues/14

YujiOshima:
Thanks! 
/lgtm

PR #23, opened 2018-04-09
Dockerfile: Use multiple stage builds

gaocegege:
- Reduce the time spent on building (80% faster in China)
- Reduce the size of the image （891 MB -> 16.7 MB (2% of the original image))

/assign @YujiOshima


gaocegege:
The CI takes too long time thus I merged it manually first.

PR #25, opened 2018-04-09
Dockerfile: Support multiple stage build in manager and cli

gaocegege:
Same as #23 

I am not sure if we should copy the CLI from the docker container. But I follow the convention to use centos instead of alpine to get the runnable binary from the container.

/assign @YujiOshima

### DIFF
--- a/cli/Dockerfile
+++ b/cli/Dockerfile
@@ -1,22 +1,9 @@
-FROM golang
-RUN : && \
-    go get k8s.io/api/apps/v1beta1 && \
-    go get k8s.io/api/core/v1 && \
-    go get k8s.io/apimachinery/pkg/apis/meta/v1 && \
-    go get k8s.io/client-go/kubernetes && \
-    go get k8s.io/client-go/tools/clientcmd && \
-    go get k8s.io/client-go/util/homedir && \
-    go get k8s.io/client-go/util/retry && \
-    go get k8s.io/client-go/rest && \
-    go get k8s.io/apimachinery/pkg/runtime/serializer && \
-    go get database/sql && \
-    go get github.com/golang/protobuf/jsonpb && \
-    go get github.com/go-sql-driver/mysql && \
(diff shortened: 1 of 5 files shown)


gaocegege:
PTAL @YujiOshima

cli/Dockerfile
```diff
@@ -1,22 +1,9 @@
...
-    go get github.com/golang/protobuf/jsonpb && \
-    go get github.com/go-sql-driver/mysql && \
-    go get github.com/mattn/go-sqlite3 && \
-    go get google.golang.org/grpc && \
-    go get gopkg.in/yaml.v2 && \
-    :
-ADD api $GOPATH/src/github.com/kubeflow/hp-tuning/api
-ADD cli $GOPATH/src/github.com/kubeflow/hp-tuning/cli
-WORKDIR $GOPATH/src/github.com/kubeflow/hp-tuning/cli
+FROM golang:1.8.2 AS build-env
```
YujiOshima: Why you use not `golang:alpine` but `golang:1.8.2` ?

gaocegege: Because you can not build the binary in alpine and use it in ubuntu or centos.

https://stackoverflow.com/questions/36279253/go-compiled-binary-wont-run-in-an-alpine-docker-container-on-ubuntu-host?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

I tried to copy the binary build in alpine and call it in ubuntu it returns:

```
$ ls -l ./bin
-rwxr-xr-x 1 ist ist 12733320 4月  10 10:51 katib-cli
$ ./bin/katib-cli
zsh: no such file or directory: ./bin/katib-cli
```

Thus we have to build it in a normal go image. And I think we could avoid it by installing the binary from another way.

YujiOshima: Ah, OK. I understood. And I agree we should look for another way.
Anyway, LGTM for this PR.

PR #49, opened 2018-04-16
improve test script

YujiOshima:
Not to use a const value for the number of pods.

### DIFF
--- a/test/scripts/run-tests.sh
+++ b/test/scripts/run-tests.sh
@@ -54,11 +54,13 @@ cd ${GO_DIR}
 ./deploy.sh
 
 TIMEOUT=60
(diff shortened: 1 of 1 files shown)


test/scripts/run-tests.sh
```diff
@@ -54,11 +54,13 @@ cd ${GO_DIR}
 ./deploy.sh
 
 TIMEOUT=60
-until kubectl get pods -n katib | grep 1/1 | [[ $(wc -l) -eq 7 ]]; do
+PODNUM=$(kubectl get pods -n katib | grep -v NAME | wc -l)
```
jose5918: You can avoid grep like this but I don't think it matters much `kubectl get pods -o jsonpath='{.items[*].metadata.name}' | wc -w`

YujiOshima:
@jose5918 Thanks!

PR #55, opened 2018-04-19
main.go: Fix style

gaocegege:
- Remove the flag `init` since it is not used in the code
- Remove the function `spawn-worker` since it is not used in the code
- Fix some variable names

/assign @YujiOshima 


gaocegege:
It is weird, the tide is still pending although we approved the PR.

PR #67, opened 2018-04-20
add CMA-ES algorithm

libbyandhelen:
This is an algorithm called Covariance Matrix Adaptation Evolution Strategy, which can compete or even beat Bayesian algorithm in long run. This is more suitable for the case with continuous parameters and having more budget.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/67)


gaocegege:
/ok-to-test

Does it WIP or finished?

YujiOshima:
@libbyandhelen Please dockerize and add kubernetes manifest.
If possible, please add unit-test for this suggestion to test script.

gaocegege:
Hi, we refactored the structure of the code base, please rebase master :-)

libbyandhelen:
@YujiOshima 
I am not sure whether this is what you mean by "unit test"
https://github.com/kubeflow/katib/pull/67/commits/40acf5a69024dfdf056bd7dc033a40bc395956cb

libbyandhelen:
@gaocegege 
the code is ready for review

YujiOshima:
@libbyandhelen Thank you for your contribution.
We are trying to refine the APIs.
See #74 #72

There are some changes in Suggestions service API.
For example,
* Rename GenerateTrials -> GetSuggesitons
* You cannot use Running/CompletedTrials in the request arg. You need to use DB or Manager API.
* You can save intermediate information to DB by SetSuggestionParametes.

Please look at the refined API and Please tell me your request and the improvements.
As Baysian optimization and CMA-ES are important algorithm, I want to support them.

libbyandhelen:
@YujiOshima 
Thank you! I will take a look:)

gaocegege:
Sorry that our CI has some problems and I am working on it.

gaocegege:
Please rebase the master, and I think it will fix the issue about CI.

libbyandhelen:
@YujiOshima 
please take a look:)

pkg/suggestion/cma_unit_test/main.py
```diff
@@ -0,0 +1,184 @@
...
+from pkg.suggestion.cma_unit_test.db_init import connect_db
+from pkg.suggestion.cma_unit_test.internface import create_study, get_study_config, get_suggestion_param, set_suggestion_param, \
+    update_suggestion_param, create_trial, \
+    get_suggestion_param_list, get_worker_list, get_worker, get_worker_logs, create_worker
+from pkg.suggestion.cma_unit_test.worker import spawn_worker
+
+_ONE_DAY_IN_SECONDS = 60 * 60 * 24
+
+
+class ManagerService(api_pb2_grpc.ManagerServicer):
```
YujiOshima: Why do you implement a new ManagerService? 
You should use [this manager](https://github.com/kubeflow/katib/blob/master/cmd/manager/main.go) .

pkg/api/python/api.proto
```diff
@@ -0,0 +1,407 @@
+syntax = "proto3";
```
YujiOshima: You should use the existing proto file. https://github.com/kubeflow/katib/blob/master/pkg/api/api.proto

pkg/suggestion/cma_service.py
```diff
@@ -0,0 +1,213 @@
...
+
+_ONE_DAY_IN_SECONDS = 60 * 60 * 24
+
+
+class CMAService(api_pb2_grpc.SuggestionServicer):
+    def __init__(self):
+        channel = grpc.insecure_channel(MANAGER)
+        self.stub = api_pb2_grpc.ManagerStub(channel)
+
+    def GetSuggestions(self, request, context):
```
YujiOshima: What happens when you call GetSuggestion before previous Trials are completed.

pkg/suggestion/cma_unit_test/main.py
libbyandhelen: Yes, I am using the manager implemented by you. This one is just for (unit) testing.
Speaking about testing, it seems that I need to follow the `getting-start.md` to deploy the katib on k8 clusters. Otherwise, it will stuck on the steps relevant to kubenetes.

pkg/api/python/api.proto
libbyandhelen: Yes, the current `api_pb2.py` and `api_pb2_grpc.py` is generated using the original proto file mentioned above. Let me delete the unused one:)

pkg/suggestion/cma_service.py
libbyandhelen: So if the previous trials are not complete, I assume that when I call `GetMetrics`, no `metircs_log_sets` will be returned. Then I will return an grpc error saying that all the trials in previous population should be evaluated.
like this:
https://github.com/libbyandhelen/hp-tuning-1/blob/b98dccb5cbc70e85004c848cc7cd7ea403a39577/pkg/suggestion/cma_service.py#L95Is my assumption correct or not?

pkg/suggestion/cma_unit_test/main.py
YujiOshima: OK, I understand! Thanks.

pkg/suggestion/cma_service.py
libbyandhelen: @YujiOshima 
Then can I check the status of the worker after calling `GetWorkers` method to see whether the worker is completed?
From your code, I see that after the client calls `RunTrial`, it will create a worker in db (Pending state) and spawn a worker (Running state), but where is code that updates the worker state to `completed` after the evaluation?

YujiOshima: @libbyandhelen When you call `GetMetrics`, the status of workers is updated.
So you can get the latest status to call `GetWorker` after `GetMetrics`.
If you think it troubles, we can add `status` item to `GetMetricsReply`.
WDYT?

libbyandhelen: @YujiOshima 
Thanks! The additional `status` item may be more natural for me. Could you please add it?

libbyandhelen:
@YujiOshima 
So now after I call `GetMetrics` method, I will loop through all the `metrics_log_set`. And if any `worker_status` in it is not completed, a grpc error is thrown, so that to ensure all the workers are finished before the final objective value is calculated.
WDYT?

YujiOshima:
@libbyandhelen That is a little difficult problem.
I think there are three types for the suggestion service.

* Not depend on the past results
* Depend on the past results and has no explicit end of the suggestion.
* Depend on the past results and has an explicit end.

Random and grid are the 1st type. Only need request number. They can return parameter lists regardless of the state of other workers. 

BO is the 2nd type. It needs to wait for the completion of the workers of the past suggestion. It requires 2 kinds of return. Parameter lists and error(the past workers are not completed).

HyperBand is the 3rd type. It also needs for the completion of the past workers. And it has an explicit end of the algorithm. It needs three type of return.  Parameter lists, error(the past workers are not completed), and the algorithm is completed.

There is two way to solve the problem.
*Handling a type of Error.
	* Param lists and no error
	* No param lists and error(Wating)
	* No param lists,  and error(Completed)

* Introduce completed flag to the `GetSuggestionsReply`
	* Param lists, no error, and completed:false
	* No param lists,  error and  completed:false
	* No param lists, no error, and  completed:true

It looks simple to handle the type of error. But I don't know it is possible across golang and python or others.
WDYT?

libbyandhelen:
@YujiOshima 
I am not sure whether I get your question right, but as for the type of error, I think we can use the error message and error code in gRPC, which should be portable across different languages.
http://avi.im/grpc-errors/
But I think using `GetSuggestionsReply` is also nice:))

libbyandhelen:
@YujiOshima 
How about now? Is this OK?

test/scripts/build-suggestion-bo.sh
```diff
@@ -39,3 +39,6 @@ cd ${GO_DIR}
 
 cp cmd/suggestion/bayesianoptimization/Dockerfile .
 gcloud container builds submit . --tag=${REGISTRY}/${REPO_NAME}/suggestion-bayesianoptimization:${VERSION} --project=${PROJECT}
+
+cp cmd/suggestion/cma/Dockerfile .
```
YujiOshima: Please create a new script `build-suggestion-cma.sh` and register it to `/test/workflows/components/workflows.libsonnet`.

pkg/suggestion/test_cma_client.py
```diff
@@ -0,0 +1,81 @@
+import grpc
```
YujiOshima: Please add unit test for CMA to CI (/test/scripts/run-tests.sh) .

YujiOshima:
@libbyandhelen Could you add e2e test for CMA?
In e2e test, we use /test/e2e/test-client.go to get suggestion and run trial test.

jlewi:
@libbyandhelen This looks like a great contribution. I know its a lot of work but if you could update the tests and make the other changes it would be great to get this added.

jlewi:
@libbyandhelen Any plans to try to push this forward?

c-bata:
Hi! Can I work on this issue?
https://github.com/kubeflow/katib/issues/1100

johnugeorge:
@c-bata 
Thanks for the interest. That would be great.

johnugeorge:
This PR is obsolete now as interfaces are no more valid

c-bata:
Hi, I guess we can close this because #1131 is merged.

andreyvelich:
@c-bata Sure.
/close

PR #70, opened 2018-04-23
worker: Rename worker_interface to worker

gaocegege:
worker_interface is a little misunderstanding, since it is not just a package for interface.

/assign @YujiOshima 


YujiOshima:
@gaocegege The worker is an instance of evaluating a trial.
So the worker is a k8s job, a TF Job etc.
How about `worker-controller`?

gaocegege:
I think worker suits it, because it represents the worker to run these trials :thinking: 

WDYT @ddysher

ddysher:
conceptual-wise, i agree with @YujiOshima that `worker` is more like a controller; however, using worker_controller is quite verbose IMO; so my suggestion is to simply use `worker` as proposed in this PR.

cmd/manager/main.go
```diff
@@ -41,7 +41,7 @@ type studyCh struct {
 	addMetricsCh chan string
 }
 type server struct {
-	wIF         worker_interface.WorkerInterface
+	wIF         worker.WorkerInterface
```
ddysher: Can we change `WorkerInterface` to `Interface`, per
https://github.com/golang/go/wiki/CodeReviewComments#package-names

example:
https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/cloud.go

ddysher:
LGTM

@YujiOshima do you want to take another look?

YujiOshima:
@ddysher Thanks! LGTM the name worker.Interface !
/approve

gaocegege:
I am not sure why tide does not merge it automatically. Same as https://github.com/kubeflow/tf-operator/pull/557

gaocegege:
It is too hack.. We can not merge a PR without lgtm although we have already added approve.

PR #75, opened 2018-05-02
workflows.libsonnet: Fix the image name

gaocegege:
Ref https://github.com/kubeflow/tf-operator/issues/567

Signed-off-by: Ce Gao <gaoce@caicloud.io>

PR #76, opened 2018-05-02
Move away from mlkube-testing to kubeflow-ci

ankushagarwal:
Related https://github.com/kubeflow/testing/issues/90

/cc @gaocegege @YujiOshima


gaocegege:
Thanks, while we already had a PR to fix it: #75

PR #81, opened 2018-05-10
add random forest prior to Bayesian Optimization

libbyandhelen:
This PR add random forest as prior to Bayesian Optimization algorithm. RF is more suitable for discrete and categorical parameters, while gaussian process if more suitable for categorical parameters.


YujiOshima:
@libbyandhelen Thank you! 
For validation of logic, can you add any citation for this algorithm?
If there is any difference to the citation, please specify the difference.

libbyandhelen:
@YujiOshima 
The reference paper is 
>Frank Hutter, Holger H Hoos, and Kevin Leyton-Brown. 2011. Sequential modelbased
>optimization for general algorithm configuration.
[https://www.cs.ubc.ca/~hutter/papers/10-TR-SMAC.pdf](url)

shall I put this into the code?

gaocegege:
I do not think we should put the paper citation in the code. Maybe we could maintain a doc about the algorithms that we support and list the citations there. WDYT

YujiOshima:
Yes, I agree with @gaocegege . 
It should be put in docs. 
You can add BO to [README](https://github.com/kubeflow/katib/blob/master/README.md#suggestion) and cite here.

libbyandhelen:
@YujiOshima @gaocegege 
The citation is added to the doc, which is for the whole BO algorithm in general, and it mentions the use of random forest prior.

gaocegege:
@libbyandhelen Please rebase the upstream/master and have a try :smile:

gaocegege:
@YujiOshima PTAL

YujiOshima:
I'm going to take a look.

Issue #50, opened 2018-04-17
Reduce time it takes to build all images


jose5918:
Takes a long time to build all images.  There are several ways to speed up the e2e tests.

The easiest way, with the least benefit, would be to add a `.glcoudignore` file like in the other repos. This will prevent uploading the entire vendor directory every time you submit a docker build to gcloud.

The second would be to run container builds in parallel  using argo rather than sequentially.

Any other ideas?

gaocegege:
> @gaocegege a few of things that can be done here. Pull the image `golang:alpine` and push it to gcr.io to reduce number of pull failures. Retry on build failures, and add a `.gcloudignore` to your repository. Even better would be to add a retry and run your container builds in parallel to reduce the points of failure.

gaocegege:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_katib/83/kubeflow-katib-presubmit/39/
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-katib-presubmit-e2e-83-84b74b0-39-869b?tab=workflow

We spend 40 mins in building and it causes time out in Prow. We need to fix it ASAP.

YujiOshima:
I will address this right now.

gaocegege:
Thanks :+1: 

:smile:

PR #74, opened 2018-04-26
Refine API

YujiOshima:
Related #72 #66 #68 
* Add low-level API and refine APIs
* Split Trial to Worker(an instance of evaluating processes) and Trial (a parameter set).
* Migrate dkl worker to kubernetes worker.
* Save suggestion and Early stopping configs to DB
* Add MinikubeDemo

HyperBand and Bayes optimization suggestion are not supported in this PR.
Random and Grid Support for now.

@gaocegege 
@ddysher 

Signed-off-by: YujiOshima <yuji.oshima0x3fd@gmail.com>

PR #85, opened 2018-05-15
Build parallel

YujiOshima:
For more robust parallel build, separate build containers with components(manager, suggestions, and earlystopping)
Depends on #74


gaocegege:
Close #50

PR #100, opened 2018-06-04
Add manifests/dlk in deploy.sh

vinaykakade:
While working through the getting started instructions, noticed that manifests/dlk is not part of deploy.sh, and so the examples do not work and the workers are not spawned. Verified that the workers are spawned correctly after adding manifests/dlk to deploy.sh

cc @mayankjuneja

### DIFF
--- a/docs/getting-start.md
+++ b/docs/getting-start.md
@@ -32,6 +32,8 @@ $ ./scripts/deploy.sh
 
 ## Use CLI
 
+vizier-core is a service of type NodePort, with port 30678.
+
 Check which node the vizier-core was deployed.
 Then access vizier API.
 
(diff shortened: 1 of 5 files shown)


vinaykakade:
@YujiOshima - are you able to take a quick look at this change?

YujiOshima:
@vinaykakade The dlk is already deprecated after #74.
We don't need dlk deployment in the latest code.
We will release a new version soon. Please use it.
@gaocegege Could you release v0.1.2-alpha with the current master?

gaocegege:
@YujiOshima Done

@vinaykakade Thanks for your contribution while I think maybe we do not need to add dlk anymore, could you please close the issue?

gaocegege:
And, if you are interested in the development of katib, you could get start with these issues: https://github.com/kubeflow/katib/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

 :smile:

vinaykakade:
Thanks, closing this. BTW, since dlk is no longer used, the references to it should be deleted from the top level manifests, but that is a topic of separate issue!

gaocegege:
@YujiOshima 

Should we remove dlk from the repository? Maybe we could open an issue.

YujiOshima:
@gaocegege @vinaykakade 
Yes, we should remove. I'm sorry to forget to remove it.
@vinaykakade Thank you!!

PR #107, opened 2018-06-06
Remove dlk from manifests

vinaykakade:
- verified that the pods are starting up correctly after this change.


YujiOshima:
@vinaykakade Thank you!
/lgtm
/approve

PR #111, opened 2018-06-09
use kubectl port-forward in demos

YujiOshima:
In the getting-start demo, using `kubectl port-forward` is probably easier way to access to katib services.
/assign @gaocegege 


gaocegege:
/lgtm
/approve

Thanks for your work!

PR #121, opened 2018-06-19
Fix indentation to use spaces (instead of a mix of tabs and spaces)

vinaykakade:
Let me know in case you prefer to use tabs everywhere. But, in general, it is better to use one consistently rather than a mix of both.

### DIFF
--- a/pkg/api/api.proto
+++ b/pkg/api/api.proto
@@ -3,10 +3,10 @@ syntax = "proto3";
 package api;
 
 service Manager {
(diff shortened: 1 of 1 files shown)


gaocegege:
Thanks for your contribution!

Personally, I like spaces.

YujiOshima:
@vinaykakade Thanks! I also prefer space. 
As the proto files were not formatted by `go fmt`, their style wouldn't be consistent.

PR #125, opened 2018-06-22
Fix broken client-example.go

vinaykakade:
With the recent changes in the repo (https://github.com/kubeflow/katib/pull/110), I found that client-example.go no longer works. There are two issues:
(1) With having 'matrics\n - accuracy' in study-config, the CreateStudy request no longer is successful. Until the underlying cause is found/fixed, removing it from study-config. Let me know in case replacing it with something else works.
(2) Even after the workers terminate successfully, the client-example.go is no longer able to detect their termination and keeps on running in an infinite loop. Reverted to the original terminal condition change which works correctly.

/assign @YujiOshima


YujiOshima:
I have no error on client-example.go.
Maybe there is some mismatch between client and server.
Which docker image do you use for vizier-core?

lluunn:
I was also able to run client-example.go (last week)

vinaykakade:
Thanks, I'll check for the mismatch between client and server.

jlewi:
@vinaykakade Were you successful?

vinaykakade:
Yes, was able to run it long ago. Apologies for not reporting this sooner. Will close this PR.

wukong1992:
@vinaykakade How did you solve it?

vinaykakade:
The issue was mismatched images in dockerhub and the tip of master. I had to rebuild the images, upload them to an image repository we use, and point the manifest for vizier-core to pull the image from our repository instead of dockerhub.

PR #141, opened 2018-07-26
Studyctl crd

YujiOshima:
Add StudyController
CRD: studycontroller.kubeflow.org
Operator: StudyController

Update examples.
This implementation is polling workers status in go process of StudyController.
Though I understand this is not an elegant implementation, this is the least impact to existing codes. 

Next step we should make worker CRD and its controller and support multi-type jobs (k8s, TF-Job..).
Assign @gaocegege

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/141)


YujiOshima:
@gaocegege It looks so many files are changed. But almost are vendoring pkg.
The main changes for the controller are under `pkg/apis`, `pkg/controller`.
The get-start doc is `https://github.com/YujiOshima/hp-tuning/blob/studyctlCRD/examples/MinikubeDemo.md`.
Please take a look.

gaocegege:
@YujiOshima Awesome work!

Could you please split the commit? I think we could put all hand written code in one commit. Then it is friendly  to reviewers.

BTW, I will travel to Japan in 8.10-17. :smile: Then I will be offline in that time.

YujiOshima:
@gaocegege I split it!

>travel to Japan 

Great! Have a nice trip!

cmd/studycontroller/main.go
```diff
@@ -0,0 +1,64 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"log"
```
gaocegege: We recommend logrus or zap if you like.

examples/MinikubeDemo/studycontroller/crd.yml
```diff
@@ -0,0 +1,11 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: studycontroller.kubeflow.org
+spec:
+  group: kubeflow.org
+  version: v1alpha1
+  names:
+    kind: StudyController
+    singular: studycontroller
+    plural: studycontroller
```
gaocegege: I think the plural should be studycontrollers

examples/MinikubeDemo/studycontroller/studycontroller.yml
```diff
@@ -0,0 +1,24 @@
...
+    matchLabels:
+      app: study-controller
+  template:
+    metadata:
+      labels:
+        app: study-controller
+    spec:
+      serviceAccountName: study-controller
+      imagePullSecrets:
+        - name: gitlabregcred
```
gaocegege: Is it necessary for the example?

examples/hypb-example.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: "kubeflow.org/v1alpha1"
+kind: StudyController
```
gaocegege: Could we name it Study?
I think studycontroller will confuse users. Our controller is also study controller.

manifests/studycontroller/crd.yml
```diff
@@ -0,0 +1,11 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: studycontroller.kubeflow.org
+spec:
+  group: kubeflow.org
+  version: v1alpha1
+  names:
+    kind: StudyController
+    singular: studycontroller
+    plural: studycontroller
```
gaocegege: Could we maintain one copy for the crd yaml? I think we have two now.

pkg/apis/apis.go
```diff
@@ -0,0 +1,32 @@
+/*
```
gaocegege: Could we name the package to api? I think it is a convension in k8s community. WDYT

pkg/apis/studycontroller/v1alpha1/studycontroller_types.go
```diff
@@ -0,0 +1,172 @@
...
+	// be set in happens-before order across separate operations.
+	// It is represented in RFC3339 form and is in UTC.
+	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
+
+	// Represents last time when the TFJob was reconciled. It is not guaranteed to
+	// be set in happens-before order across separate operations.
+	// It is represented in RFC3339 form and is in UTC.
+	LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`
+
+	State   State      `json:"state,omitempty"`
```
gaocegege: In k8s community I think we prefer conditions instead of phase. PTAL https://github.com/kubernetes/kubernetes/issues/7856

pkg/apis/studycontroller/v1alpha1/studycontroller_types.go
```diff
@@ -0,0 +1,172 @@
...
+const (
+	OptimizationTypeUnknown  OptimizationType = ""
+	OptimizationTypeMinimize OptimizationType = "minimize"
+	OptimizationTypeMaximize OptimizationType = "maximize"
+)
+
+type WorkerSpec struct {
+	Image      string    `json:"image,omitempty"`
+	Command    []string  `json:"command,omitempty"`
+	Gpu        int       `json:"gpu,omitempty"`
```
gaocegege: GPU is better, IMO

pkg/controller/studycontroller/studycontroller_controller.go
```diff
@@ -0,0 +1,480 @@
...
+				} else {
+					return false
+				}
+			}
+		}
+	}
+	return false
+}
+
+func (r *ReconcileStudyController) controllerloop(instance *katibv1alpha1.StudyController) {
```
gaocegege: I think we could add some comments for the func, since it is the main control loop.

examples/MinikubeDemo/studycontroller/studycontroller.yml
YujiOshima: I will fix it.

examples/hypb-example.yaml
YujiOshima: I want to distinguish between resource and process.
The Study and Trial are resources that should be saved to DB persistently.
The Worker and StudyController(now) are a process of Trial and Study each other.
Since the process is ephemeral, I want to make them CRD.

I agree StudyController is confusing. We should rename it.
Do you have any idea for the name?
For example, Experiments, StudyJob.

pkg/apis/studycontroller/v1alpha1/studycontroller_types.go
YujiOshima: Thanks, I will fix it.

pkg/apis/apis.go
YujiOshima: The kubebuilder automatically generate apis dir. I will fix it.

examples/hypb-example.yaml
YujiOshima: I renamed StudyController to StudyJob and StudyJob-Controller.

cmd/studycontroller/main.go
YujiOshima: I want to separate the change introducing `zap` from this PR.
I will use zap for all components.

YujiOshima:
@gaocegege Thank you for your review. I updated. PTAL.

YujiOshima:
@gaocegege I updated. Could you review it?

gaocegege:
I will do it today or tomorrow. 
Thanks for your contribution!

jlewi:
Woo Hoo!

A couple high level questions. I'll take a look at the code to see if I can answer these but it would be good to include the answers directly in the PR description as this would help speed up the review process. (There are so many files its crashing reviewable and I'm having a hard time looking at the code)

1. Are you using a kit like kubebuilder (https://github.com/kubernetes-sigs/kubebuilder) to help build the CRD and handle code gen?

2.Will each new job type require explicit support in the controller?

3. What is the interface between the CRD controller and the user's code? For example, suppose I want to use Katib to optimize learning rate and suppose that corresponds to a command line argument "--learning_rate", how do I tell Katib that?

4. What if my job requires certain PVs to be mounted or certain environment variables to be set. How do I tell that?

My expectation is that the user provides the following

1. A template (e.g. jinja/go/helm) 
2. A list of parameters taken by the template that correspond to hyperparameters.

The CRD controller should create the job by getting a set of hyperparameters and substituting them into the template.

So the user can use any K8s resource they want and use any features they want (e.g. PVs, environment variables, resource specs) just by supplying the appropriate template.

jlewi:
Looking at your example
https://github.com/YujiOshima/hp-tuning/blob/6ad60bcdec83e75183ab61a3f9cc0537c3793105/examples/MinikubeDemo.md

It looks like the WorkerSpec is how user specifies the job to launch. This seems limited

1. User can only control things exposed in WorkerSpec
2. Users can not reuse existing templates for their jobs
3. WorkerSpec must explicitly support new K8s resource types.
4. CRD makes strong assumptions about how HP parameters are passed to the job
   
    * e.g. you assume parameters are passed as command line arguments; what if user uses environment variables?

Why not

1. Take a template as a file path
    * We can start with whichever template you want and is easy to implement (e.g. go templates)
2. User can use PV/ConfigMap or other means to make the template accessible

3. CRD creates the spec and resource by substituting in the hyperparameters into the template
  
If we start with a config map then the CRD could just use the K8s client libraries to fetch the config and we don't need to worry about mounting PVs.

What do you think?

gaocegege:
@jlewi We can review from https://github.com/kubeflow/katib/pull/141/commits/5ac22f96162df11b157c36563d4fc7601e864f52. The first two PRs are auto-generated.

YujiOshima:
@jlewi Thanks for your comment.
>1. Are you using a kit like kubebuilder (https://github.com/kubernetes-sigs/kubebuilder) to help build the CRD and handle code gen?

1) Yes, I used kubebuilder.

The workerSpec limitation you pointed is right. And I agree it is inconvenient.
But I'm not sure about your suggestion.
Your expectation is:
User create ConfigMap for worker template(e.g. pod Spec)
Then Controller read the ConfigMap, embed hyperparameters to the template, and create resources from it.
Is my understanding correct?

jlewi:
@YujiOshima 

Your CRD is defining a new type of object "WorkerSpec"

  Worker Spec:
    Command:
      python
      /mxnet/example/image-classification/train_mnist.py
      --batch-size=64
    Image:  katib/mxnet-mnist-example
    Mountconf:

This is limiting and in my opinion limiting I'd like to be able to specify the fully spec for any K8s resource and the CRD controller should just create it.

So the "value" of WorkerSpec should just be a K8s resource. The CRD can inspect metadata to figure out how to create it (i.e. API Endpoint).

However, I need to provide a template not a fully specified K8s resource because we need to be able to substitute in the hyperparameters.

So why I want to supply to the CRD controller is a template for K8s resource which will be filled in with the actual parameters.

Right now you are basically defining your own template engine that is very limited. i.e. it makes assumptions about what type of object and only allows certain places for the parameters to be supplied.

Those restrictions are unnecessary. We already have a variety of really powerful and well understood template engines (jinja, go templates, etc...)

So I think the CRD should let users pick a supported template engine and provide a template that is parameterized with a set of named parameters. The CRD can then substitute in the parameters to create
the actual object.

I don't care how exactly you provide the template. A ConfigMap was one suggestion.

You could also just store the template in the spec as a string field.

YujiOshima:
@jlewi Thanks. OK, I understand.
SGTM about using a powerful template engine since it is flexible.
But I wonder how to collect metrics from user-defined workers.
Currently, Katib uses k8s log API from pods to collect metrics.
How about we allow only k8s job or  Kubeflow-job (TF-Job, Cainer-Job..) for Worker.
Users need to upload templates for  k8s job or  Kubeflow-job.
When the worker is TF-Job of other Kubeflow-Jobs, I don't know the best way to collect metrics but it depends on operators. 
What do you think?

jlewi:
@YujiOshima Can we use the same idea to generalize metrics collection? i.e as part of the StudyCrd controller I specify a resource (or service) to collect metrics. 

For example, for a TensorFlow job we have a couple options

1. After the training job finishes. We could run an evaluation job that would then report metrics back to Katib using the Katib API.
   * This job might just open up the TFEvents file, get metrics, and report them back to Katib

2. Distributed TF jobs have eval workers; so in the eval workers the user could run code that reports metrics to Katib.

I think a good starting point would be add a metricsCollector resource to the CRD. This could be a job that will get launched after the training job. This should be a job that reports metrics to Katib using the Katib API.

YujiOshima:
@jlewi I try to implement the first step as below.

* Add Report metrics API. (for generalization. All job operator can report metrics in the same way)
* Add MetricsCollector as a CronJob. It will be created by Studyjobcontroller.
* As a default worker, users can push k8s job config as a template of a worker. the MetricsCollector will collect log from stdout of the job.

Why CronJob: Katib should collect metrics during training for Early stopping not only the evaluation step.

After that, we can add more flexible metrics collecter for TF-Job or other specific Job.
WDYT?

jlewi:
This sounds good to me but let me make sure I understand correctly.

> Add Report metrics API. (for generalization. All job operator can report metrics in the same way)

A Report metrics API sounds great. Is that already in this PR? Can you provide a link to the API definition.

Side note; I don't know that metrics should be reported by the operator. That would require us to build metrics reporting into the operator and that might be too limiting.

> Add MetricsCollector as a CronJob. It will be created by Studyjobcontroller.

Why is it a CronJob? What is MetricsCollector doing? Is this just collecting metrics from stdout and reporting them via the API?

Is the idea that StudJobController creates a MetricsCollector cron job for each HP tuning job. This job periodically collects the metrics and reports them via the ReportMetrics APi?

YujiOshima:
@jlewi 
> Is that already in this PR? 

Not yet. I will update soon.

>Is the idea that StudJobController creates a MetricsCollector cron job for each HP tuning job. This job periodically collects the metrics and reports them via the ReportMetrics APi?

Yes. CRD controller runs event-driven. I want to collect and report the metrics even no events happened.

jlewi:
Sounds good.

My suggestion would be to update the examples or instructions first to reflect how you think it will work from the user perspective. That way we can make sure we are all on the same page. So for example, I'd like to be able to look [minikube.md](https://github.com/kubeflow/katib/blob/6ad60bcdec83e75183ab61a3f9cc0537c3793105/examples/MinikubeDemo.md) and understand

1. How users will be able to specify a template per discussion above
1. How users will report metrics (e.g. by running a job that outputs to stdout).

jlewi:
@YujiOshima How's this coming?

YujiOshima:
@jlewi I'm working [here](https://github.com/YujiOshima/hp-tuning/tree/studycrd-dev) (I cut a new branch since it is not ok-to-test now)
I already
* Add Report metrics API.
* Use go template for creating workers.
* Add CronJob for metrics collector.

I should implement logic for management events of the worker(create, deletion, update).
Then, I will push it to this PR.

YujiOshima:
@jlewi OK, I pushed. It is almost functional. But there is no docs. I'll update soon.

jlewi:
@YujiOshima That's fantastic. I think its fine to submit code incrementally to master; i.e. its fine to submit code to master that makes docs on master obsolete. 

Its just easier if everyone is working on master because then we can all work on the same code base.

jlewi:
Let me know when this is ready for another look.

YujiOshima:
@jlewi I will update doc today. Please check tomorrow in your time-zone.

YujiOshima:
@jlewi I updated the documents. https://github.com/YujiOshima/hp-tuning/blob/studyctlCRD/examples/MinikubeDemo.md
The architecture design is shown in the design doc https://docs.google.com/document/d/1ZEKhou4z1utFTOgjzhSsnvysJFNEJmygllgDCBnYvm8/edit#heading=h.1mtrlouglg6w .
PTAL

jlewi:
@YujiOshima Thanks. Can you please exclude vendor from this PR and rebase? The 7000 files makes it really difficult to review this PR. If we want to check in vendor lets do it in a follow on PR.

jlewi:
Where is the CRD actually defined? I thought it would be studyjob_controller.go but I don't see it there.
https://github.com/YujiOshima/hp-tuning/blob/studyctlCRD/pkg/controller/studyjobcontroller/studyjob_controller.go

jlewi:
I found the CRD spec;
https://github.com/YujiOshima/hp-tuning/blob/d1980b4bd942e6439428051ad38d64bbb6752936/pkg/api/operators/apis/studyjob/v1alpha1/studycontroller_types.go#L102

It looks like WorkerSpec isn't using a template. Am I missing something?

At any rate; anything we can do to split this PR up into smaller PRs would be a huge help in reviewing it and make things go faster. Removing vendor should be a big help.

YujiOshima:
@jlewi I merged commits other than vendor.
Could you look this commit https://github.com/kubeflow/katib/pull/141/commits/2ae19b1e308e1f696515f2c9d66454606ad3464e ? 
We can't exclude vendor from this PR because of CI.

> It looks like WorkerSpec isn't using a template. Am I missing something?

Workers are created from template defined ConfigMap. I use go template.
 https://github.com/kubeflow/katib/blob/2ae19b1e308e1f696515f2c9d66454606ad3464e/pkg/controller/studyjobcontroller/studyjob_controller.go#L602

YujiOshima:
@jlewi I updated.
* WorkerSpec contains the only path for a template. https://github.com/kubeflow/katib/blob/002501049c046a3d29a4cc95c2b645b60b306251/pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go#L116
* add comments

WorkerType is for specifying the Object type of worker(Job, TF-Job...). The controller needs to register the object type for watching https://github.com/kubeflow/katib/blob/002501049c046a3d29a4cc95c2b645b60b306251/pkg/controller/studyjobcontroller/studyjob_controller.go#L88 .
And then, it will get the instance of the object https://github.com/kubeflow/katib/blob/002501049c046a3d29a4cc95c2b645b60b306251/pkg/controller/studyjobcontroller/studyjob_controller.go#L332 .
The controller needs to know which kind of object should be watched in this study.



> What's a controller in this context? Is it the StudyJobController? Why is it added to the Manager?
This comment was generated by kubebuilder. In kubebuilder concept, we can register multiple controllers to a manager.  The manager is a process that manages the controllers.

jlewi:
Why do you need WorkerType? Can't the controller generate the resource from the template and then get the object type from the metadata?  I would be ok with filing an issue and doing it in a follow up PR.

cmd/manager/main.go
```diff
@@ -6,32 +6,24 @@ import (
...
-	namespace                = "katib"
-	port                     = "0.0.0.0:6789"
-	defaultEarlyStopInterval = 60
-	defaultSaveInterval      = 30
+	port = "0.0.0.0:6789"
 )
 
-var workerType = flag.String("w", "kubernetes", "Worker Type")
-var ingressHost = flag.String("i", "kube-cluster.example.net", "Ingress host for TensorBoard visualize")
+var ingressHost = flag.String("i", "kube-cluster.example.net", "Ingress host")
```
jlewi: What is this ingress used for?

cmd/manager/main.go
```diff
@@ -196,16 +165,70 @@ func (s *server) GetMetrics(ctx context.Context, in *pb.GetMetricsRequest) (*pb.
...
+					Value: l.Value,
+					Time:  l.Time.UTC().Format(time.RFC3339Nano),
+				}
 			}
 		}
 	}
 	return &pb.GetMetricsReply{MetricsLogSets: mls}, nil
 }
 
+func (s *server) ReportMetrics(ctx context.Context, in *pb.ReportMetricsRequest) (*pb.ReportMetricsReply, error) {
```
jlewi: Is this RPC new in this PR?

pkg/api/api.proto
```diff
@@ -212,14 +205,11 @@ message GetTrialsReply {
     repeated Trial trials = 1;
 }
 
-message RunTrialRequest {
-    string study_id = 1;
-    string trial_id = 2;
-    string runtime = 3;
-    WorkerConfig worker_config = 4;
+message CreateWorkerReauest {
```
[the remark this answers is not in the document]
jlewi: Filed https://github.com/kubeflow/katib/issues/150
Would be good to document the API.

pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go
```diff
@@ -0,0 +1,182 @@
...
+	//runtime "k8s.io/apimachinery/pkg/runtime"
+
+	pb "github.com/kubeflow/katib/pkg/api"
+)
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// StudyJobSpec defines the desired state of StudyJob
+type StudyJobSpec struct {
+	StudySpec            *StudySpec            `json:"studySpec,omitempty"`
```
jlewi: I think I left a comment on a previous commit  about the naming of StudySpec.
See this comment
https://github.com/kubeflow/katib/commit/2ae19b1e308e1f696515f2c9d66454606ad3464e#r30104814

jlewi: The name StudySpec seems confusing; I would think the whole thing is the StudySpec.

What's the distinction between StudySpec and SuggestionSpec?

cmd/manager/main.go
```diff
@@ -196,16 +165,70 @@ func (s *server) GetMetrics(ctx context.Context, in *pb.GetMetricsRequest) (*pb.
...
+	sc, err := dbIf.GetStudyConfig(in.StudyId)
+	if err != nil {
+		return &pb.ReportMetricsReply{}, err
+	}
+	for _, mls := range in.MetricsLogSets {
```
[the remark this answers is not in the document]
jlewi: Looked at the API looks like this is not the logs.

jlewi:
I reviewed the commits
https://github.com/kubeflow/katib/pull/141/files/5429a4dd121ccd08d3989edd90476e2cbf954c41..842ee429692cbdced85f7a6f47e295cd60f113ec

And left some comments.

The 7727 files still make it really difficult to review everything; (e.g. because I can't use reviewable to track review threads).

Should we just go ahead and submit this and fix issues in follow on PRs?

YujiOshima:
@jlewi 
>Why do you need WorkerType?
Hmm, I thought It needs to specify the object type before loading the template. But it may not be necessary. I'm going to try.

cmd/manager/main.go
YujiOshima: It was used when the ingress was used in example demo. 
We can delete it now.

cmd/manager/main.go
YujiOshima: Yes

cmd/manager/main.go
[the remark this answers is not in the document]
YujiOshima: Katib can store every log of metrics.
I would change  the name of RPC to `ReportMetricsLogs`.
WDYT?

pkg/api/api.proto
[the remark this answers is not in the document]
YujiOshima: It can store the worker info to DB and generate workerId by CreateWorker rpc. 
I would change rename the RPC to `RegisterWorker`.
> Would be good to document the API.

OK.

pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go
YujiOshima: SuggestionSpec is parameters that related to only suggestion service.
How about merge StudySpec to StudyJobSpec like below.

```
type StudyJobSpec struct {
	Name               string            `json:"name,omitempty"`
	Owner              string            `json:"owner,omitempty"`
	OptimizationType   OptimizationType  `json:"optimizationtype,omitempty"`
	OptimizationGoal   *float64          `json:"optimizationgoal,omitempty"`
   ....
	WorkerSpec           *WorkerSpec           `json:"workerSpec,omitempty"`
	SuggestionSpec       *SuggestionSpec       `json:"suggestionSpec,omitempty"`
	EarlyStoppingSpec    *EarlyStoppingSpec    `json:"earlyStoppingSpec,omitempty"`
	MetricsCollectorSpec *MetricsCollectorSpec `json:"metricsCollectorSpec,omitempty"`
}
```

YujiOshima:
@jlewi 
> Should we just go ahead and submit this and fix issues in follow on PRs?

Do you mean we should delete vendored code from this repo as @vinaykakade said?
https://github.com/kubeflow/katib/pull/142#issuecomment-410673179

Let's open new issue about this.

pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go
jlewi: This looks good to me.

pkg/api/api.proto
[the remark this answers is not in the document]
jlewi: Thanks for the explanation. I don't think this needs to change in this PR.  I'd suggest filing an issue if you think its worth fixing.

cmd/manager/main.go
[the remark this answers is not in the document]
jlewi: Thanks for the explanation. Lets not change it in this PR. If you think its worth changing lets file an issue and do it in a follow on PR.

jlewi:
@YujiOshima I'm pretty happy with this PR as is. Is there any work that you think should be accomplished in this PR as opposed to a follow on PR? e.g. should we try wait for a follow on PR to remove the type?

This PR is already pretty massive so I think its better if we defer additional changes to a follow on PR. Hopefully, once we have the initial PR is committed; maybe we can split it up the remaining work more easily?

/lgtm
/approve

/hold

YujiOshima:
I will work on a new PR.
/hold cancel

vinaykakade:
@YujiOshima - given that the public API will change with this PR, should we create a 0.2 branch from current master which will have stable 0.2 code, or check this in a 0.3 branch, or any other proposal you may have in mind to minimize the migration pain for current customers? Overall, would be great if we are thinking about customers who are dependent on current public API while working through backwards incompatible changes.

gaocegege:
https://github.com/kubeflow/katib/tree/0.2

I created one, thanks for the advice

jlewi:
I'll merge this manually. It looks like the reviewable status check is choking on the large PR size and that's blocking the merge.

PR #146, opened 2018-08-18
Corrected typos in hyperband example yml

shibuiwilliam:
Corrected typos in hyperband example yml, from 'hayperband' to 'hyperband' in order to make it possible to deploy the pods and services.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/146)

PR #148, opened 2018-08-19
fix imagepull not found issue

cheyang:
```
# docker pull katib/suggestion-hayperband
Using default tag: latest
Error response from daemon: repository katib/suggestion-hayperband not found: does not exist or no pull access
```


YujiOshima:
@cheyang Thanks, but it is duplicate of #146

PR #152, opened 2018-08-22
refactor studyjob CRD controller

YujiOshima:
This PR for remained task of #141 
* Delete `WorkerType` from `WorkerSpec` (get kind of worker from template)
* Rename `GetMetrics` to `GetMetricsLogs`
* Rename `CreateWorker` to `RegisterWorker`
* Merge `StudySpec` to `StudyJobSpec`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/152)


YujiOshima:
@jlewi @gaocegege Please take a look.

ankushagarwal:
Hi @YujiOshima Is there a way I can build the studyjob controller locally and test this?

ankushagarwal:
`examples/MinikubeDemo/deploy.sh` seems to be broken

1. The manifests paths should be `kubectl apply -f ../../manifests/...` instead of `kubectl apply -f ../../../manifests...`
2. `kubectl apply -f ../../manifests/studycontroller/studycontroller.yaml` - this is no longer valid

YujiOshima:
@ankushagarwal Thanks.

For testing, please do in this way.
* Set up your docker registry
* dep ensure
* In script/build.sh, set your registry name to PREFIX and run the script
* push images and rewrite manifests for using your registry.

I update deploy.sh.

ankushagarwal:
Thanks I'll try that out

ankushagarwal:
This looks great! I was successfully able to launch a StudyJob.

Here is my primary concern

In StudyJob crd, the workerSpec takes a path to a gotemplate which is mounted from a configmap into the controller. That seems a bit odd to me because we are coupling the StudyJob CRD controller and the template.

The controller should not know anything about the templates. The templates should be specified as part of the CRD.

For example my CRD would contain the following section

```
workerSpec:
  goTemplate : |-
    apiVersion: batch/v1
    kind: Job
    metadata:
      name: {{.WorkerId}}
      namespace: katib
    spec:
      template:
        spec:
          containers:
          - name: {{.WorkerId}}
            image: alpine
          restartPolicy: Never
```

examples/gpu-example.yaml
```diff
@@ -6,32 +6,31 @@ metadata:
...
-            min: "2"
-            max: "3"
-        - name: --optimizer
-          parametertype: categorical
-          feasible:
-            list:
-            - sgd
-            - adam
-            - ftrl
+  studyName: gpu-example
```
ankushagarwal: `studyName` seems redundant to me. Why not use `metadata.name` as the name?

examples/gpu-example.yaml
```diff
@@ -6,32 +6,31 @@ metadata:
...
+      feasible:
+        min: "2"
+        max: "3"
+    - name: --optimizer
+      parametertype: categorical
+      feasible:
+        list:
+        - sgd
+        - adam
+        - ftrl
```
ankushagarwal: should `workerSpec.workerType` be deleted from these examples?

YujiOshima: Yes. I will delete.

examples/gpu-example.yaml
YujiOshima: StudyJobName and StudyName are different. You can create multiple StudyJob for the same Study.
When you do so, the results of these StudyJobs are plotted in the same graph.

YujiOshima:
@ankushagarwal Thank you for your review. 
Shince the template for worker spec can be shared among StudyJobs, I use ConfigMap.
But including WorkerSpec template to StudyJob Manifest helps easy to start.
How about we make users can select to use ConfigMap or include to CRD?

```
WorkerSpec:
    GoTemplate:
        RawTemplate:
        TemplatePath:
```

Using RawTemplate
```
workerSpec:
  goTemplate :
     rawTemplate: |-
    apiVersion: batch/v1
    kind: Job
    metadata:
      name: {{.WorkerId}}
      namespace: katib
    spec:
      template:
        spec:
          containers:
          - name: {{.WorkerId}}
            image: alpine
          restartPolicy: Never
```

Using TemplatePath
```
workerSpec:
  goTemplate :
     templatePath: "/worker-template/cpuWorkerTemplate.yaml"
```

YujiOshima:
I update WorkerSpec. https://github.com/kubeflow/katib/pull/152/commits/d7e42d2a842a9c7359f71046fe8bffc12460d7e9#diff-9c75f45e5ba1d29b3bc4c85a9662f936

ankushagarwal:
@YujiOshima : Thanks for your changes.

I think this looks good for now. This gives the user flexibility of using a configmap v/s defining the template in the spec itself.

For the future direction, I think that mounting the configmap into the controller is a bad idea. We are coupling these two together and making the controller deployment depend on this configmap.

Ideally, the controller should be able to read the configmap specified instead in the worker template using k8s API instead of trying to read it from its filesystem. 

For example if a user wants to define a configmap(`worker-template`) with their template, they should be able to specify the name of the configmap in the `workerSpec` and the controller can read the configmap and the `templatePath` in the configmap using k8s api.

Here we don't have to deal with mounting the configmap into the controller and don't have to know about the configmap when creating the controller deployment.

```
workerSpec:
  goTemplate :
     configMap: "worker-template"
     templatePath: "cpuWorkerTemplate.yaml"
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: worker-template
  namespace: katib
data:
  cpuWorkerTemplate.yaml : |-
    apiVersion: batch/v1
    kind: Job
    metadata:
      name: {{.WorkerId}}
      namespace: katib
    spec:
      ...
```

ankushagarwal:
Feel free to self-approve and merge when ready

YujiOshima:
@ankushagarwal Thank you for your comments!
/approve

PR #170, opened 2018-09-16
add mutex to studyjob controller

YujiOshima:
To avoid `Fail to check status Suggestion count mismatched. May be duplicate suggestion request` error, introduce the mutex to studyjob controller.


gaocegege:
Generally LGTM, I can review it after hold is canceled.

YujiOshima:
@gaocegege Thanks! I think it is completed, please review it.

/hold cancel

PR #178, opened 2018-09-25
Add pv example to katibDB

### LINKED ISSUE #137: Katib needs to use durable storage that outlives the pod.
Katib uses two databases
- MySQL (for Katib)
- MongoDB (for ModelDB)

It doesn't look like we are using any sort of durable storage for either one.

Here's the MySQL deployment it doesn't look like its using a persistent volume
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/katib/vizier.libsonnet#L185

Here's the MongoDB deployment:
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/katib/modeldb.libsonnet#L39

@YujiOshima @gaocegege Is this accurate or am I missing something?

YujiOshima:
For persistent data, add pv example for katib.
Related to #137 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/178)

### DIFF
--- a/manifests/pv/pv.yaml
+++ b/manifests/pv/pv.yaml
@@ -10,3 +10,5 @@ spec:
     storage: 10Gi
   accessModes:
     - ReadWriteOnce
+  hostPath:
+    path: /data/katib
--- a/manifests/pv/pv.yaml
+++ b/manifests/pv/pv.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: PersistentVolume
 metadata:
-  name: katib-volume
+  name: katib-mysql
   labels:
     type: local
     app: katib
@@ -9,6 +9,4 @@ spec:
   capacity:
     storage: 10Gi
   accessModes:
-    - ReadWriteMany
-  hostPath:
-    path: /data/katib
+    - ReadWriteOnce
--- a/manifests/pv/pv.yaml
+++ b/manifests/pv/pv.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: katib-volume
+  labels:
+    type: local
+    app: katib
+spec:
+  capacity:
+    storage: 10Gi
+  accessModes:
+    - ReadWriteMany
+  hostPath:
+    path: /data/katib
--- a/manifests/pv/pvc.yaml
+++ b/manifests/pv/pvc.yaml
@@ -1,17 +1,16 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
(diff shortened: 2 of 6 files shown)


jlewi:
@YujiOshima do you plan on making these changes to the Katib component as well?
https://github.com/kubeflow/kubeflow/tree/master/kubeflow/katib

manifests/pv/pv.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: katib-volume
+  labels:
+    type: local
+    app: katib
+spec:
+  capacity:
+    storage: 10Gi
+  accessModes:
+    - ReadWriteMany
```
inc0: do we need ReadWriteMany? It's hard to maintain on non-minikube env

manifests/pv/pv.yaml
```diff
@@ -0,0 +1,14 @@
...
+  hostPath:
+    path: /data/katib
```
inc0: why do we need hostpath?

manifests/pv/pvc.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: katib-volume
+  namespace: katib
+  labels:
+    app: katib
+spec:
+  accessModes:
```
inc0: again, why not RWO?

manifests/pv/pvc.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: ""
```
inc0: leave out storageClass, let's assume users have default

manifests/pv/pvc.yaml
```diff
@@ -0,0 +1,18 @@
...
+  selector:
+    matchLabels:
```
inc0: no need for that if users have working storageclass. By the way, minikube comes with storageclass by default so it's unnecessary.

manifests/vizier/db/deployment.yaml
```diff
@@ -28,10 +28,10 @@ spec:
...
           containerPort: 3306
-#        resources:
-#          requests:
-#            cpu: 500m
-#            memory: 500M
-#          limits:
-#            cpu: 500m
-#            memory: 500M
+        volumeMounts:
+        - name: katib-volume
```
inc0: I think better name would be katib-mysql or katib-db. We might need more volumes later

YujiOshima:
@inc0 Thanks! I updated.

PR #187, opened 2018-09-30
make job name contain the name of studyjob

wukong1992:
The original job name is random, which is not conducive to further monitoring of the job, so make the job name support the name of the studyjob.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/187)


YujiOshima:
@wukong1992 since the length of StudyName is not limited, it is not suitable for ID.
We should separate readable name and ID.
Why not show readable name (study name+ID) on UI and use ID as a unique key internally.

wukong1992:
@YujiOshima It is not about UI. We need to check the usage index and other indicators according to the name of the job. However, the job‘name here has no connection with the user-specified studyjob’name. Can you consider limiting the length of the studyjob‘name and splicing it with the job’ID as the job‘s name? I think it is more reasonable than the jobs’name in the cluster are all random numbers.

PR #201, opened 2018-10-10
Fix deadlock condition in ReconcileStudyJobController#Reconcile

hnakagawa:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/201)


YujiOshima:
@hnakagawa Thanks!
/lgtm
/approve

PR #205, opened 2018-10-14
GA

shibuiwilliam:
What:
Made genetic algorithm logic in golang.
Not yet integrated to Katib workflow.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/205)

### DIFF
--- a/pkg/suggestion/genetic_algorithm/gene.go
+++ b/pkg/suggestion/genetic_algorithm/gene.go
@@ -1,0 +1,144 @@
+package genetic_algorithm
+
+// for gene with continuous variables in float64s
+type ContinuousGeneType struct {
+	min float64
+	max float64
+}
+
+func NewContinuousGeneType(min, max float64) *ContinuousGeneType {
+	return &ContinuousGeneType{min, max}
+}
+
+func (cg *ContinuousGeneType) generateRandom() float64 {
+	return getContinuousRandom(cg.min, cg.max)
+}
+
+func (cg *ContinuousGeneType) setGeneRandomValue(g *Gene) {
+	g.setGeneValue(cg.generateRandom())
+}
+
+// for gene with discrete variables in int
+type DiscreteGeneType struct {
+	min int
+	max int
+}
+
+func NewDiscreteGeneType(min, max int) *DiscreteGeneType {
+	return &DiscreteGeneType{min, max}
+}
+
+func (dg *DiscreteGeneType) generateRandom() int {
+	return getDiscreteRandom(dg.min, dg.max)
+}
+
+func (dg *DiscreteGeneType) setGeneRandomValue(g *Gene) {
+	g.setGeneValue(dg.generateRandom())
+}
+
+// for gene with categorical variables in string
+type CategoricalGeneType struct {
+	category []string
+}
+
+func NewCategoricalGeneType(category []string) *CategoricalGeneType {
+	return &CategoricalGeneType{category}
+}
+
+func (cg *CategoricalGeneType) generateRandom() int {
+	return getCategoricalRandom(cg.category)
+}
+
+func (cg *CategoricalGeneType) setGeneRandomValue(g *Gene) {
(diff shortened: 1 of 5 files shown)


pkg/suggestion/genetic_algorithm/gene.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (g *Gene) setRandomGeneValue(s setGeneRandomValueInterface) {
+	s.setGeneRandomValue(g)
+}
+
+// set random value depending on the gene type
+func setRandomGeneValueWithType(gene *Gene, geneTypeMap GeneTypeMap) {
+	switch s := geneTypeMap[gene.geneName].(type) {
+	case CategoricalGeneType:
```
YujiOshima: Why you use Type Switch?
Why not make an interface for GeneType and only call its func.

pkg/suggestion/genetic_algorithm/gene.go
```diff
@@ -0,0 +1,144 @@
...
+	for i := 0; i < len(geneNames); i++ {
+		geneTypeMap[geneNames[i]] = geneTypes[i]
+	}
+	return geneTypeMap
+}
+
+// gene struct; one name and one value per one gene
+type Gene struct {
+	geneName  string      // name of the gene
+	geneValue interface{} // value of the gene; must fit with the gene type
```
YujiOshima: Why not use a GeneType interface instead of interce{}?

pkg/suggestion/genetic_algorithm/genetic_algorithm.go
```diff
@@ -0,0 +1,313 @@
...
+		bestOffspring = generation.offsprings[k]
+	}
+
+	// returns the best optimized offspring and its score
+	return bestOffspring, bestScore
+}
+
+// a set of evaluation result for each offspring
+type GAResult struct {
+	result map[int]float64
```
YujiOshima: I think we should use a list of a struct that has a Sorter interface.
At `GetBestScores` you can sort and get a slice.
It is better for both performance and readable.

pkg/suggestion/genetic_algorithm/gene.go
shibuiwilliam: The geneValue takes the actual value for the gene, while GeneType defines type for the gene.

pkg/suggestion/genetic_algorithm/gene.go
shibuiwilliam: I have the interface for the geneType on lines 56 to 58 as like below.
 
type setGeneRandomValueInterface interface {
	setGeneRandomValue(g *Gene)
}

pkg/suggestion/genetic_algorithm/genetic_algorithm.go
shibuiwilliam: Oh, that looks better. I'll make it this weekend.

pkg/suggestion/genetic_algorithm/genetic_algorithm.go
```diff
@@ -0,0 +1,323 @@
...
+		for i, o := range generation.offsprings {
+			gaResult.recordResult(i, evaluateFunc(o))
+		}
+		bestOffsprings = ga.Select(gaResult, generation)
+	}
+
+	// best genes after optimization
+	bestScoreMap := ga.GetBestScores(gaResult, 1)
+	for k, v := range bestScoreMap {
+		bestScore = v
```
YujiOshima: Is it OK?
The iteration order of golang map is not consistent. 
( from go1.3 https://golang.org/doc/go1.3#map
 https://github.com/golang/go/blob/release-branch.go1.11/src/runtime/map.go#L763)
So we won't know which value will be store in `bestScore`.

gaocegege:
/close

Thanks for the PR. Now we use the mature python packages to implement the algorithms. I think it is easy to maintain since we do not need to maintain the algorithm itself, we just need to maintain the intermediate layer to use these packages.

PR #225, opened 2018-10-24
More db tests

toshiiw:


YujiOshima:
@toshiiw Thank!!
/ok-to-test

PR #229, opened 2018-10-29
Only retry an INSERT operation on unique constraint violation

toshiiw:
The retry logic is used to generate an unique ID, but if there is
another error the DB code can fall into an infinite loop.

Signed-off-by: IWAMOTO Toshihiro <iwamoto@valinux.co.jp>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/229)


YujiOshima:
Thanks!
/lgtm
/approve

PR #253, opened 2018-11-20
Set MYSQL_ROOT_PASSWORD via Secret

lkpdn:
For security reasons, we better to (a). factor out hard-coded mysql root password and use Secret instead + (b). set dynamically generated random password.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/253)

### DIFF
--- a/manifests/vizier/db/secret.yaml
+++ b/manifests/vizier/db/secret.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: vizier-db-secrets
+  namespace: katib
+data:
+  MYSQL_ROOT_PASSWORD: test
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -22,7 +22,23 @@ set -o xtrace
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
 
 cd ${SCRIPT_ROOT}
+# Dedicated namespace has to be present beforehand.
 kubectl apply -f manifests/0-namespace.yaml
+
+# Generate Secret with dynamically initialized data, so as to keep Go codebase
+# simple as possible i.e., without client-go ClientSet.
+cat <<EOF | kubectl create -f -
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: vizier-db-secrets
+  namespace: katib
+data:
+  MYSQL_ROOT_PASSWORD: $(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c24 | base64)
+EOF
+
+# Rest of the static manifests.
 kubectl apply -f manifests/pv
 kubectl apply -f manifests/vizier/db
 kubectl apply -f manifests/vizier/core
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -22,23 +22,7 @@ set -o xtrace
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
 
 cd ${SCRIPT_ROOT}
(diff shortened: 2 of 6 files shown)


scripts/deploy.sh
```diff
@@ -22,7 +22,23 @@ set -o xtrace
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
 
 cd ${SCRIPT_ROOT}
+# Dedicated namespace has to be present beforehand.
 kubectl apply -f manifests/0-namespace.yaml
+
+# Generate Secret with dynamically initialized data, so as to keep Go codebase
+# simple as possible i.e., without client-go ClientSet.
+cat <<EOF | kubectl create -f -
```
YujiOshima: Please create a db-secret.yml not generating a key in deploy.sh.
The deploy.sh should be a simple support script.

lkpdn: Thanks for the comment. If we still keep the password randomised as my original patch does, I may choose either of:

- (1). make it as a go template, add a lightweight cli tool to do `template.Execute()`
- (2). choose a popular template engine, such as jinja2

Does the option (1) sound ok to you? If I misunderstood your point, please correct me.

YujiOshima: Do we really need to randomize?
Why not to make secret manifest and set a default password. Then users will edit it when they need.

lkpdn:
To googlebot, abt. CLA:
I signed it!

lkpdn:
@YujiOshima please take another look.

YujiOshima:
@lkpdn Thank you! It gets simple!
/ok-to-test

manifests/vizier/db/secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: vizier-db-secrets
+  namespace: katib
+data:
+  MYSQL_ROOT_PASSWORD: test
```
toshiiw: I think we should take a "secure by default" approach and the value should be randomized, while its implementation can be a bit messy. Also it has to be initialized before the mysql pod starts up.
At least there should be an issue so that we don't forget to randomize before GA. ;-)

lkpdn: IMHO editing manifest for the secret before running deploy.sh seems not much burden.

lkpdn:
I'm just guessing but the test failure seems not relating to this PR. Looking into it...

manifests/vizier/db/secret.yaml
toshiiw: To do that, users must be aware of the secret.yaml file. Randomization must happen automatically or the majority of deployments would be in an insecure state.

lkpdn:
I realized that `sed -i -e "31,37d" manifests/vizier/db/deployment.yaml` in `test/scripts/run-tests.sh` had to be updated as well. I pushed a fix https://github.com/kubeflow/katib/pull/253/commits/f1ed1c1baca9173e0e059135e455bb34438dacd3 . Please let me know if these should be fixed-up.

YujiOshima:
Great! Thank you!
/lgtm
/approve

lkpdn:
Thanks for reviewing.

manifests/vizier/db/secret.yaml
lkpdn: I agree with your point "secure by default". How about dynamically generating manifests when executing `./scripts/build.sh`, or introduce a new script, e.g. `./scripts/generate-manifests.sh`? Other projects tend to have users do that sort of thing, such as `$ make manifests`. I think it's much better than doing it in `./scripts/deploy.sh`

PR #254, opened 2018-11-20
Refactor studyjobcontroller

richardsliu:
Refactor studyjobcontroller into 4 files:
- studyjobcontroller.go: main logic
- katib_api_util:  wrappers for calling katib APIs
- model_db_util: wrapper for modelDB operations
- manifest_parser: parsers worker manifests

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/254)

### DIFF
--- a/pkg/controller/studyjobcontroller/model_db_util.go
+++ b/pkg/controller/studyjobcontroller/model_db_util.go
@@ -1,0 +1,95 @@
+// Copyright 2018 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


johnugeorge:
How about renaming the folder 'studyjobcontroller' ->'studyjob'? As it is already in controller folder, it looks redundant.

pkg/controller/studyjobcontroller/model_db_util.go
```diff
@@ -0,0 +1,94 @@
+// Copyright 2018 The Kubeflow Authors
```
YujiOshima: This is not only for ModelDB.
It will be an interface for the model management systems including ModelDB.
So I want to rename it to model_util.go.

YujiOshima:
@richardsliu Thank you! Generally LGTM. One comment, please check.
And I agree with @johnugeorge . If rename the folder, it's better to rename package of studyjobcontroller to studyjob.
@richardsliu WDYT?

richardsliu:
@YujiOshima @johnugeorge Done.

YujiOshima:
@richardsliu Thank you!
/approve

PR #256, opened 2018-11-21
Downgrade kubernetes dependency to 1.10.1

richardsliu:
This is to ensure that tf-operator can be imported to katib. Katib uses controller-runtime which is compatible with k8s 1.10.1. Tf-operator uses 1.10.0 (upgrading to 1.11.2 breaks a lot of things).

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/256)

### DIFF
--- a/pkg/controller/studyjob/studyjob_controller.go
+++ b/pkg/controller/studyjob/studyjob_controller.go
@@ -289,7 +289,7 @@ func (r *ReconcileStudyJobController) checkStatus(instance *katibv1alpha1.StudyJ
 						job := &batchv1.Job{}
 						joberr := r.Client.Get(context.TODO(), nname, job)
 						if joberr == nil {
-							if err := r.Delete(context.TODO(), job, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+							if err := r.Delete(context.TODO(), job); err != nil {
 								return false, err
 							}
 						}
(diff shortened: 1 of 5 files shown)


pkg/controller/studyjob/studyjob_controller.go
```diff
@@ -289,7 +289,7 @@ func (r *ReconcileStudyJobController) checkStatus(instance *katibv1alpha1.StudyJ
 						job := &batchv1.Job{}
 						joberr := r.Client.Get(context.TODO(), nname, job)
 						if joberr == nil {
-							if err := r.Delete(context.TODO(), job, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
```
YujiOshima: If you remove `DeletePropagationForeground`, the pods generated by Job or CronJob remain after the studyjob was deleted.
You should delete them explicitly after a Job or a CronJob is deleted.

YujiOshima:
Removing `DeletePropagationForeground` has a big side effect.
We need to be careful.
/hold

richardsliu:
@YujiOshima Please take another look at this change. This is a short term solution until we figure out a better solution for managing dependencies across repos. 

The flaws of this solution are:
- Pods created by the batch jobs must be deleted manually;
- Need to allow studyjob-controller to delete pods in the namespace (unsafe but no worse than what the metrics collector is already doing).

Let me know what you think.

YujiOshima:
@richardsliu Thank you for your hard work!
/hold cancel

PR #260, opened 2018-11-26
Fix incorrectly set namespace

lkpdn:
Commit b6f8e07d26a ("Update manifests (#246)") has just changed the
namespace as a whole. This new manifest should be updated as well.

Fixes: 67e94c7697b ("Set MYSQL_ROOT_PASSWORD via Secret (#253)")
Signed-off-by: Koichiro Den <den@valinux.co.jp>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/260)

PR #261, opened 2018-11-26
Prevent pod restarts caused by slow db boot

lkpdn:
When we run `./scripts/deploy.sh` for the first time, `vizier-core` is highly likely to undergo a few times of "Pod Restart". This is because `vizier-db` takes much longer time to become ready. This PR intends to have `vizier-core` wait for DB to become ready, so that we no longer see any pod Restart if all components are properly deployed.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/261)

### DIFF
--- a/manifests/vizier/db/deployment.yaml
+++ b/manifests/vizier/db/deployment.yaml
@@ -31,6 +31,15 @@ spec:
         ports:
         - name: dbapi
           containerPort: 3306
+        readinessProbe:
+          exec:
+            command:
+            - "/bin/bash"
+            - "-c"
+            - "mysql -D $$MYSQL_DATABASE -p$$MYSQL_ROOT_PASSWORD -e 'SELECT 1'"
+          initialDelaySeconds: 5
+          periodSeconds: 2
+          timeoutSeconds: 1
         volumeMounts:
         - name: katib-mysql
           mountPath: /var/lib/mysql
--- a/pkg/db/interface.go
+++ b/pkg/db/interface.go
@@ -22,6 +22,9 @@ const (
 	dbDriver     = "mysql"
 	dbNameTmpl   = "root:%s@tcp(vizier-db:3306)/vizier"
 	mysqlTimeFmt = "2006-01-02 15:04:05.999999"
+
+	connectInterval = 5 * time.Second
+	connectTimeout  = 60 * time.Second
 )
 
 type GetWorkerLogOpts struct {
@@ -90,6 +93,26 @@ func getDbName() string {
 	return fmt.Sprintf(dbNameTmpl, dbPass)
 }
 
+func openSQLConn(driverName string, dataSourceName string, interval time.Duration,
+	timeout time.Duration) (*sql.DB, error) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	timeoutC := time.After(timeout)
+	for {
+		select {
+		case <-ticker.C:
+			if db, err := sql.Open(driverName, dataSourceName); err == nil {
+				if err = db.Ping(); err == nil {
+					return db, nil
+				}
+			}
+		case <-timeoutC:
+			return nil, fmt.Errorf("Timeout waiting for DB conn successfully opened.")
+		}
+	}
+}
+
 func NewWithSQLConn(db *sql.DB) VizierDBInterface {
 	d := new(dbConn)
 	d.db = db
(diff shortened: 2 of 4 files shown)


lkpdn:
https://github.com/kubeflow/katib/pull/260 has been merged so let me rebase this (branch: prevent-pod-restarts-caused-by-slow-db-boot) to master.

pkg/db/interface.go
```diff
@@ -105,10 +150,17 @@ func NewWithSQLConn(db *sql.DB) VizierDBInterface {
 }
 
 func New() VizierDBInterface {
+	timeToWait := time.Duration(30) * time.Second
+	err := waitForDbReady(timeToWait)
```
toshiiw: readiness probe has its own merit but I think it is enough to try to connect the mysql in a loop here.
The extra function just adds unnecessary complexity.

lkpdn: When the pod for vizier-db literally not ready or failing, we'd better to wait for it being ready, perhaps infinitely (though we should configure redinessProbe for vizier-core as well in that case). I set it to 30 sec since normally DB seems to get ready in 30 sec if it's properly configured, while Pod Restart due to mis-configuration or something like that is okay. OTOH, when the pod literally is ready to serve, I think we don't need retrying at all in that case, though retrying to connect with its timeout + interval set to some reasonable value might be okay.

Simple loop as you commented seems okay, while I think this PR diff is small enough. Just a minor con is that vizier-db's log would be dirtied. I can replace https://github.com/kubeflow/katib/pull/261/commits/a0d4d063b53596d39e7480532d59c922a0f27cee shortly.

lkpdn:
Please take another look: https://github.com/kubeflow/katib/pull/261/commits/dc7bb07553b1c131c304cf94bd7152d93e7de2d7
Thanks.

pkg/db/interface.go
```diff
@@ -105,11 +128,12 @@ func NewWithSQLConn(db *sql.DB) VizierDBInterface {
 }
 
 func New() VizierDBInterface {
-	db, err := sql.Open(dbDriver, getDbName())
-	if err != nil {
+	if db, err := openSQLConn(dbDriver, getDbName(), connectInterval, connectTimeout); err != nil {
 		log.Fatalf("DB open failed: %v", err)
+		return nil
+	} else {
```
toshiiw: AFAIK, go style avoids this extra else{} block.

manifests/vizier/db/deployment.yaml
```diff
@@ -31,6 +31,15 @@ spec:
         ports:
         - name: dbapi
           containerPort: 3306
+        readinessProbe:
```
toshiiw: It would be nice if vizier-core can get readinessprobe or readinessgate, too.

pkg/db/interface.go
```diff
@@ -90,6 +93,26 @@ func getDbName() string {
...
+	startTime := time.Now()
+	for {
+		select {
+		case <-ticker.C:
+			if db, err := sql.Open(driverName, dataSourceName); err == nil {
+				if err = db.Ping(); err == nil {
+					return db, nil
+				}
+			}
+		case <-time.After(timeout - time.Since(startTime)):
```
toshiiw: I think a new timer instance is created on every connection failure and they are leaked. What do you think?

pkg/db/interface.go
lkpdn: Just avoided declaration of db outside if statement. I'm not sure which is better,

manifests/vizier/db/deployment.yaml
lkpdn: I think so too, will do in another PR maybe.

pkg/db/interface.go
lkpdn: If connection attempt fails consecutively, I think just one instance lives until return. There seems no chance of entering this loop in a short period

pkg/db/interface.go
toshiiw: How about retaining the old code style and just replacing the sql.Open call with openSQLConn? :-)

manifests/vizier/db/deployment.yaml
toshiiw: Yup. We can do it later.

lkpdn:
@toshiiw thanks for reviewing, PTAL: https://github.com/kubeflow/katib/commit/2ae5826cef0b859466855f2aaaaa51e33bec97b3

manifests/vizier/db/deployment.yaml
lkpdn: Opened an issue: https://github.com/kubeflow/katib/issues/268

YujiOshima:
@lkpdn Thanks, great!
/lgtm
/approve

lkpdn:
@YujiOshima thanks :)

Issue #268, opened 2018-11-29
Add readiness/liveness probes to vizier-core


lkpdn:
Currently we don't make much use of readiness/liveness probe, though there are major cases where if those are to be installed, it would help cluster admins or users recognize and diagnose what is going on when some component goes bad. Let me write some down as my first thought.

- [ ] 1). When vizier-db takes tens of seconds to boot up, which is quite normal especially if the components are cleanly deployed for the first time, vizier-core is unavailable in the meantime. We can tell `vizier-core` is ready once Katib manager successfully starts to listen.
- [ ] 2). When vizier-db gets unhealthy, much of Katib manager methods starts to fail. So it might help if we regard `vizier-core` as being healthy iff. healthcheck probe against vizier-db succeeds.
- [ ] 3). When vizier-suggestion-XXX gets unhealthy, `vizier-core` cannot get suggestions.
- [ ] 4). When vizier-earlystopping-XXX gets unhealthy, `vizier-core` cannot serve GetShouldStopWorkers.

I think at least we should catch (1) and (2), while (3) and (4) is kind of brainstorming.

PR #270, opened 2018-12-03
Implement gRPC Health Checking Protocol + add readiness/liveness probes to vizier-core

lkpdn:
(1) and (2) listed in Ref: #268. Plus, set timeout (https://github.com/go-sql-driver/mysql/#timeout) parameter so that vizier-core never get stuck in a fruitless and harmful manner.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/270)


YujiOshima:
@lkpdn Thank you for create this PR.
Could you implement [gRPC Health Checking Protocol]( https://github.com/grpc/grpc/blob/master/doc/health-checking.md) and check healthy with [grpc_health_probe](https://github.com/grpc-ecosystem/grpc-health-probe) ?

lkpdn:
@YujiOshima OK, I'll push a commit later. Thanks.

lkpdn:
@YujiOshima patch size gets quite heavier so I split it up into seven logical commits. Please take another look, thanks.

cmd/manager/main.go
```diff
@@ -217,88 +219,120 @@ func (s *server) SetEarlyStoppingParameters(ctx context.Context, in *pb.SetEarly
...
+	}
+
+	// We only accept optional service name only if it's set to suggested format.
+	if in != nil && in.Service != "grpc.health.v1.Health" {
+		resp.Status = health_pb.HealthCheckResponse_UNKNOWN
+		return &resp, fmt.Errorf("grpc.health.v1.Health can only be accepted if you specify service name.")
+	}
+
+	// Check if we can connect to vizier-db since otherwise manager could not serve most of its methods.
+	dbIf, err := kdb.New()
```
YujiOshima: You should check dbIf is already initialized or not since another APIs will be called asynchronously.
You should not rewrite dbIf when it has been used in another function.

cmd/manager/Dockerfile
```diff
@@ -10,3 +10,6 @@ COPY --from=build-env /go/src/github.com/kubeflow/katib/cmd/manager/vizier-manag
 COPY --from=build-env /go/src/github.com/kubeflow/katib/pkg/manager/visualise /
 ENTRYPOINT ["./vizier-manager"]
 CMD ["-w", "kubernetes"]
+RUN GRPC_HEALTH_PROBE_VERSION=v0.2.0 && \
```
YujiOshima: Please move up this line between `WORKDIR /app` and `COPY --from=build-env...`.
Then we can use a cache even when there are some changes in the manager.

lkpdn: I'll push a fix, thanks.

cmd/manager/main.go
lkpdn: Indeed, thanks! Will push a fix shortly.

lkpdn:
Sorry about repeated forced push. Now all fixed, though I've just done testrun on my local minikube.
Thanks for reviewing, PTAL.

YujiOshima:
Thanks!!
/ok-to-tesr

YujiOshima:
Thanks!!
/ok-to-test

lkpdn:
Sorry I forgot to update some test codebase. Pushed a commit.: https://github.com/kubeflow/katib/pull/270/commits/25012d46e6c12b2dff0df8325978ca597168b661

YujiOshima:
@lkpdn Thank you!! This is very helpful.
/lgtm
/approve

lkpdn:
Thanks for the review!

Issue #272, opened 2018-12-03
CLI for Katib-NAS system


andreyvelich:
We should create CLI for Katib-NAS system.
Related to https://github.com/kubeflow/katib/issues/257

YujiOshima:
What kind of CLI? Please tell me a detail of the CLI.

andreyvelich:
@YujiOshima We want to create simple CLI for users who don't want to use our UI. 
In first release we want to add this commands to CLI: create, delete NAS job from yaml file, list of all NAS jobs, describe current NAS job.
That will be binary file which you need to install in your cluster.
Pretty much the same what you did here [Katib CLI](https://github.com/kubeflow/katib/tree/master/cmd/cli). 
And what we have in [Argo CLI](https://github.com/argoproj/argo/tree/master/cmd/argo).
I am using cobra framework for this as well.

gaocegege:
Could we define NAS Job as a CRD to make use of kubectl as our CLI?

@andreyvelich

andreyvelich:
@gaocegege 
Right now, yes we can. 
But, if we want to implement some other functionality to NAS Job, that kubectl can't provide, we should create our own CLI.

Issue #257, opened 2018-11-21
[Feature] AutoML/NAS support in Katib


johnugeorge:
We have an early design arch of AutoML/NAS support in Katib.  

Doc link: https://docs.google.com/document/d/1fE5yKG4d1eOC8vXvxNoAiTHErLfkVbvC-i_cArZp9Jo/edit

Please review the design

Related issues: 
https://github.com/kubeflow/kubeflow/issues/1074
#24 

Contributed by @Akado2009 @DeeperMind @garganubhav @andreyvelich @tremblerz

/cc @YujiOshima @gaocegege @jlewi

DeeperMind:
A presentation of our work cab be found in https://docs.google.com/presentation/d/1SfcKy7fzHFJZMChodCR1uBN44yX6p6QEkb0gRO0w_JI/edit?usp=sharing

andreyvelich:
A final Katib-NAS design doc can be found here 
https://docs.google.com/document/d/1qGWy-C5XSQmh82XYoMcJ_JWLHwmyvdMRjCkFMfkO0vE/edit

gaocegege:
https://github.com/Microsoft/nni/blob/b6cf7cee0e72671672b7845ab24fcdc5aed9ed48/docs/en_US/GeneralNasInterfaces.md#example-enas-macro-search-space

NNI developed by MSRA has a document about their design decisions on NAS. It is really helpful for us.

richardsliu:
Any more remaining work here?

andreyvelich:
@richardsliu I think we can close this issue, since we support NAS in v1alpha1 and v1alpha2. 
For any other NAS improvements we can create another issue.

PR #279, opened 2018-12-08
[WIP] Add CLI for NAS-Katib system

andreyvelich:
Related to: https://github.com/kubeflow/katib/issues/257.
Fixes: https://github.com/kubeflow/katib/issues/272.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/279)


YujiOshima:
@andreyvelich Why do we need this CLI? Why not use the only kubectl?

andreyvelich:
@YujiOshima In this CLI we can add additional functional. 
For example, monitoring status of all NAS jobs, show results of NAS job in cmd if you don't want to check it in the UI.

YujiOshima:
It looks similar to this cli https://github.com/kubeflow/katib/tree/master/cmd/cli .
We plan to remove this in the next version since we can do create, delete job by `kubectl` and don't want to manage CLI in this repo.
Do you think we should have our own CLI?

andreyvelich:
@YujiOshima We will make other parts of this system and after that will think about CLI. 
Maybe we will add to this CLI some functionality that kubectl can't provide. 
If not, I will delete this CLI as well. Is it sounds good?

YujiOshima:
@andreyvelich Sounds good! Thank you!

Issue #39, opened 2018-04-13
[feature] Integrate with tf-operator, pytorch-operator


gaocegege:
1. take parameters that control distributed training (e.g. number of replicas)
2. create the appropriate CRD as opposed to K8s job controller

https://kubeflow.slack.com/archives/C9ZLKR73L/p1523553472000001

gaocegege:
#dibs 

I will try to integration katib and tf-operator.

lluunn:
cc @ankushagarwal

ankushagarwal:
@gaocegege Any ideas on how we would go about doing this?

gaocegege:
@ankushagarwal We need to create workers for the operator. Then use the new worker instead of k8s worker.

jlewi:
Hopefully the requested design doc (#138) will make this clear.

@gaocegege Not sure I follow what you mean by creating workers for the operator. I'd hope the design would be flexibly enough that we wouldn't have to write new code to support different types of resources.  I think #138 will go a long way to clearing thi sup.

Related issues:
#66 Manager & Worker API 
#87 StudyController - Running HP Tuning without writing code
kubeflow/examples#162 - HP Controller should create resources correctly.

gaocegege:
@jlewi Is the doc clear?

jlewi:
@gaocegege The [doc](
https://docs.google.com/document/d/1ZEKhou4z1utFTOgjzhSsnvysJFNEJmygllgDCBnYvm8/edit#heading=h.4xeaz4l3pt2y) still doesn't answer my questions.

The section on TF-Job worker still isn't filled out.

In particular,

1. Is it the study controller that launches/waits for the jobs?
1. Can we allow users to use any type of job?
   
     * Can users provide the study controller with a template (helm/jinja/ksonnet/go template) which specifies any K8s resource 
     * StudyController provides parameters to the template to get a K8s job which it then creates
     * StudyController monitors the K8s resource

So only requirement on K8s resource is that it provides certain conditions so that the StudyController can monitor it.

jlewi:
@ankushagarwal @YujiOshima can we get this done in time for 0.3? I think the CRD now supports templates so adding support for TFJob should be doable.

Here is sample code for using the K8s rest API to create/get/delete K8s objects.
https://github.com/kubeflow/tf-operator/blob/db5c875455bc323a8215785053e792d3206c4c4a/pkg/util/k8sutil/tf_job_client.go

[...]

PR #283, opened 2018-12-10
Support Pytorch job in Katib

johnugeorge:
This PR tracks

1. Integration of Pytorch job in Katib Studyjob
2. Support multiple worker kinds in the default Metric Collector

Related:  #39 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/283)

### DIFF
--- a/examples/pytorchjob-example.yaml
+++ b/examples/pytorchjob-example.yaml
@@ -43,7 +43,7 @@ spec:
                 spec:
                   containers:
                     - name: pytorch
-                      image: johnugeorge/pytorch-mnist-with-summary:0.4
+                      image: gcr.io/kubeflow-ci/pytorch-mnist-with-summary:0.4
                       imagePullPolicy: Always
                       command:
                         - "python"
--- a/examples/pytorchjob-example.yaml
+++ b/examples/pytorchjob-example.yaml
@@ -1,0 +1,58 @@
+apiVersion: "kubeflow.org/v1alpha1"
+kind: StudyJob
+metadata:
+  namespace: kubeflow
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: pytorchjob-example
+spec:
+  studyName: pytorchjob-example
+  owner: crd
+  optimizationtype: maximize
+  objectivevaluename: accuracy
+  optimizationgoal: 0.99
+  requestcount: 4
+  metricsnames:
+    - accuracy
+    - loss
+  parameterconfigs:
+    - name: --lr
+      parametertype: double
+      feasible:
+        min: "0.01"
+        max: "0.05"
+    - name: --momentum
+      parametertype: double
+      feasible:
+        min: "0.5"
+        max: "0.9"
+  workerSpec:
+    retain: true
+    goTemplate:
+        rawTemplate: |-
+          apiVersion: "kubeflow.org/v1beta1"
+          kind: PyTorchJob
+          metadata:
+            name: {{.WorkerID}}
+            namespace: kubeflow
+          spec:
+           pytorchReplicaSpecs:
+            Master:
+              replicas: 1
+              restartPolicy: Never
+              template:
+                spec:
+                  containers:
+                    - name: pytorch
+                      image: johnugeorge/pytorch-mnist-with-summary:0.4
+                      command:
+                        - "python"
+                        - "/opt/pytorch_dist_mnist/mnist_with_summary.py"
+                        {{- with .HyperParameters}}
+                        {{- range .}}
+                        - "{{.Name}}={{.Value}}"
+                        {{- end}}
+                        {{- end}}
+  suggestionSpec:
+    suggestionAlgorithm: "random"
+    requestNumber: 3
--- a/examples/pytorchjob-example.yaml
+++ b/examples/pytorchjob-example.yaml
@@ -14,7 +14,6 @@ spec:
   requestcount: 4
   metricsnames:
     - accuracy
-    - loss
   parameterconfigs:
     - name: --lr
       parametertype: double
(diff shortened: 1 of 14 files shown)


examples/pytorchjob-example.yaml
```diff
@@ -0,0 +1,93 @@
...
+          spec:
+           pytorchReplicaSpecs:
+            Master:
+              replicas: 1
+              restartPolicy: Never
+              template:
+                spec:
+                  containers:
+                    - name: pytorch
+                      image: johnugeorge/pytorch-mnist-with-summary:0.4
```
richardsliu: Can you check in the source code for this? Also this should be on gcr.io.

examples/pytorchjob-example.yaml
```diff
@@ -0,0 +1,93 @@
...
+          successfulJobsHistoryLimit: 15
+          failedJobsHistoryLimit: 15
+          jobTemplate:
+            spec:
+              template:
+                spec:
+                  serviceAccountName: metrics-collector
+                  containers:
+                  - name: {{.WorkerID}}
+                    image: johnugeorge/metrics-collector
```
richardsliu: Same with this.

johnugeorge: New metric-collector image has to be created once this current PR is merged

examples/pytorchjob-example.yaml
```diff
@@ -0,0 +1,93 @@
...
+                      imagePullPolicy: Always
+                      command:
+                        - "python"
+                        - "/opt/pytorch_dist_mnist/mnist_with_summary.py"
+                        {{- with .HyperParameters}}
+                        {{- range .}}
+                        - "{{.Name}}={{.Value}}"
+                        {{- end}}
+                        {{- end}}
+  metricsCollectorSpec:
```
richardsliu: Do you also need to configure PV here?

johnugeorge: @richardsliu Currently, this example issuing the default metric collector which parses the stdout logs. It doesn't need PV. I will add one more example the uses the tf event metric collector.

examples/pytorchjob-example.yaml
richardsliu: Can you update the image?

pkg/controller/studyjob/const.go
```diff
@@ -13,5 +13,6 @@ package studyjob
 
 const (
 	DefaultJobWorker = "Job"
-	TFJobWorker = "TFJob"
+	TFJobWorker      = "TFJob"
+	PytorchJobWorker = "PyTorchJob"
```
richardsliu: PyTorchJobWorker?

examples/pytorchjob-example.yaml
johnugeorge: I will change the image name once the new image is created after this PR merge.

examples/pytorchjob-example.yaml
richardsliu: @johnugeorge Done. Pushed as gcr.io/kubeflow-ci/pytorch-mnist-with-summary:0.4.

johnugeorge: updated

pkg/controller/studyjob/const.go
johnugeorge: updated

pkg/manager/metricscollector/meticscollector.go
```diff
@@ -34,8 +35,16 @@ func NewMetricsCollector() (*MetricsCollector, error) {
...
+func (d *MetricsCollector) CollectWorkerLog(wID string, wkind string, objectiveValueName string, metrics []string, namespace string) (*api.MetricsLogSet, error) {
+	var labelName string
+	if wkind == studyjob.TFJobWorker {
+		labelName = "tf_job_name"
+	} else if wkind == studyjob.PyTorchJobWorker {
+		labelName = "pytorch_job_name"
+	} else {
+		labelName = "job-name"
+	}
+	pl, _ := d.clientset.CoreV1().Pods(namespace).List(metav1.ListOptions{LabelSelector: labelName + "=" + wID, IncludeUninitialized: true})
```
YujiOshima: Whitch pod will be watched in Pytorch and TFJob Job by `pytorch_job_name = wID` and `tf_job_name  = wID` ?

richardsliu:
@johnugeorge Need to rebase the PR.

pkg/manager/metricscollector/meticscollector.go
johnugeorge: Pods spawn by the pytorch job and tf job will have this label key whose value is set to job name https://github.com/kubeflow/tf-operator/blob/master/pkg/common/jobcontroller/jobcontroller.go#L190

johnugeorge:
@richardsliu rebased

richardsliu:
/lgtm
Will wait for @YujiOshima for approval.

pkg/manager/metricscollector/meticscollector.go
YujiOshima: Yes, my question is pytorch job and tf job will create several pods, and metrics collector will get logs from one pod. 
Does metrics collector get logs from which one of the pods created by a pytorch job?

johnugeorge: Currently, it uses only 1 worker. If there are more workers,  Master should take responsibility to emit logs.

YujiOshima:
I wrote a comment to metrics collector. 
Though the pytoch job and tf job will create several pods, the default metrics collector will collect one pod of them.
I wonder which pod should we choice, manager, parameter-server, worker or anyone is OK.
If it is difficult to define, the default metrics collector may be deprecated when you use tf job or pytorch job.
WDYT @johnugeorge @richardsliu

johnugeorge:
@YujiOshima Shall we merge this and start a separate discussion for distributed jobs? I will create one issue

YujiOshima:
@johnugeorge OK, Thanks.
/approve

PR #288, opened 2018-12-12
Extend studyjob client API 

andreyvelich:
I added namespace parameter to studyJob client API functions.
After this we can use the API not only inside containers.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/288)


andreyvelich:
@YujiOshima

YujiOshima:
@andreyvelich Thank you. Could you split getting namespace process to another function like `func (s *StudyjobClient)getNameSpace(string ns) string`?

andreyvelich:
@YujiOshima Done

johnugeorge:
@andreyvelich What would be the scenario for multiple namespaces in the function call?

Issue #289, opened 2018-12-13
confusing studyJob.status.suggestionCount value


hougangliu:
studyJob.status.suggestionCount should record how many times this studyJobs had called suggestion, but now it seems this filed doesn't match what it should record as below: 

**$ kubectl get studyjob random-example3 -o yaml -n kubeflow**
apiVersion: kubeflow.org/v1alpha1
kind: StudyJob
metadata:
  creationTimestamp: 2018-12-12T13:52:18Z
  generation: 1
  labels:
    controller-tools.k8s.io: "1.0"
  name: random-example3
  namespace: kubeflow
  resourceVersion: "13797889"
  selfLink: /apis/kubeflow.org/v1alpha1/namespaces/kubeflow/studyjobs/random-example3
  uid: 23b263a2-fe15-11e8-af31-005056ad997c
spec:
  metricsnames:
  - accuracy
  objectivevaluename: Validation-accuracy
  optimizationgoal: 0.99
  optimizationtype: maximize
  owner: crd
  parameterconfigs:
  - feasible:
      max: "0.03"
      min: "0.01"
    name: --lr
    parametertype: double
  - feasible:
      max: "5"
      min: "2"
    name: --num-layers
    parametertype: int
  - feasible:
      list:
      - sgd
      - adam
      - ftrl
    name: --optimizer
    parametertype: categorical
  **requestcount: 4**
  studyName: random-example3
  suggestionSpec:
    requestNumber: 3
    suggestionAlgorithm: random
    suggestionParameters:
    - name: SuggestionCount
      value: "0"
  workerSpec:
    goTemplate:
      rawTemplate: |-
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: {{.WorkerID}}
          namespace: kubeflow
        spec:
          template:
            spec:
              containers:
              - name: {{.WorkerID}}
                image: katib/mxnet-mnist-example
                command:
                - "python"
                - "/mxnet/example/image-classification/train_mnist.py"
                - "--batch-size=64"
                {{- with .HyperParameters}}
                {{- range .}}
                - "{{.Name}}={{.Value}}"
                {{- end}}
                {{- end}}
              restartPolicy: Never
status:
  bestObjectiveValue: 0.980593
  completionTime: 2018-12-12T14:08:21Z
  conditon: Completed
  earlyStoppingParameterId: ""
  lastReconcileTime: 2018-12-12T14:08:21Z
  startTime: 2018-12-12T13:52:16Z
[...]

PR #290, opened 2018-12-13
fix studyJob status suggestionCount mismatch error

hougangliu:
Fixes: #289

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/290)

### DIFF
--- a/pkg/controller/studyjob/katib_api_util.go
+++ b/pkg/controller/studyjob/katib_api_util.go
@@ -73,7 +73,6 @@ func initializeStudy(instance *katibv1alpha1.StudyJob, ns string) error {
 		return err
 	}
 	instance.Status.SuggestionParameterID = sPID
-	instance.Status.SuggestionCount += 1
 	instance.Status.Condition = katibv1alpha1.ConditionRunning
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/studyjob/katib_api_util.go
```diff
@@ -73,7 +73,6 @@ func initializeStudy(instance *katibv1alpha1.StudyJob, ns string) error {
 		return err
 	}
 	instance.Status.SuggestionParameterID = sPID
-	instance.Status.SuggestionCount += 1
```
johnugeorge: why is this removed?

hougangliu: SuggestionCount should record how many times the studyJob had called suggestion, and this value should **+= 1** every time after calling suggestion.
Here (initialize studyJob instance and no suggestion called before) instance.Status.SuggestionCount should be 0 (0 by default).

richardsliu: @YujiOshima Is this the intended behavior? Should we update the API documentation instead?

hougangliu: For now, SuggestionParam["SuggestionCount"] in db change path is [0, 2, 3, ..., requestcount+1].
after this patch its path is [0,1,2,...,requestcount]

YujiOshima: @hougangliu Thanks.
> SuggestionParam["SuggestionCount"] in db change path is [0, 2, 3, ..., requestcount+1].
This is not intended behavior, we may need to fix. I'm reviewing and testing this PR.

johnugeorge: LGTM

YujiOshima:
@hougangliu Thank you! /approve

Issue #301, opened 2018-12-19
studyjob cannot work if its namespace is not kubeflow


hougangliu:
When I create a studyjob with default namespace, the job will be stuck in `Running`
```
# cat random-example.yaml 
apiVersion: "kubeflow.org/v1alpha1"
kind: StudyJob
metadata:
  namespace: default 
  labels:
    controller-tools.k8s.io: "1.0"
  name: random-example3
spec:
  studyName: random-example3
  owner: crd
  optimizationtype: maximize
  objectivevaluename: Validation-accuracy
  optimizationgoal: 0.99
  requestcount: 1
  metricsnames:
    - accuracy
  parameterconfigs:
    - name: --lr
      parametertype: double
      feasible:
        min: "0.01"
        max: "0.03"
    - name: --num-layers
      parametertype: int
      feasible:
        min: "2"
        max: "5"
    - name: --optimizer
      parametertype: categorical
      feasible:
        list:
        - sgd
        - adam
        - ftrl
  workerSpec:
    goTemplate:
        rawTemplate: |-
          apiVersion: batch/v1
          kind: Job
          metadata:
            name: {{.WorkerID}}
            namespace: {{.NameSpace}}
          spec:
            template:
              spec:
                containers:
                - name: {{.WorkerID}}
                  image: katib/mxnet-mnist-example
                  command:
                  - "python"
                  - "/mxnet/example/image-classification/train_mnist.py"
                  - "--batch-size=64"
                  {{- with .HyperParameters}}
                  {{- range .}}
                  - "{{.Name}}={{.Value}}"
                  {{- end}}
                  {{- end}}
                restartPolicy: Never
  suggestionSpec:
    suggestionAlgorithm: "random"
    requestNumber: 3
```

PR #302, opened 2018-12-19
studyJob with non-kubeflow namespace cannot work

hougangliu:
Fixes: #301

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/302)

### DIFF
--- a/pkg/types.go
+++ b/pkg/types.go
@@ -1,5 +1,5 @@
 package pkg
 
 const (
-	ManagerAddr = "vizier-core:6789"
+	ManagerAddr = "vizier-core.kubeflow:6789"
 )
--- a/manifests/studyjobcontroller/metricsControllerConfigMap.yaml
+++ b/manifests/studyjobcontroller/metricsControllerConfigMap.yaml
@@ -9,7 +9,7 @@ data:
     kind: CronJob
     metadata:
       name: {{.WorkerID}}
-      namespace: kubeflow
+      namespace: {{.NameSpace}}
     spec:
       schedule: "*/1 * * * *"
       successfulJobsHistoryLimit: 0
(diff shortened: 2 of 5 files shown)


hougangliu:
Before create a non-kubeflow (saying `nonkubeflow`) namespace studyJob, we should doc that user must run below command:

```
$ cat mcrbac.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: metrics-collector
  namespace: nonkubeflow
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: metrics-collector
  namespace: nonkubeflow
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: metrics-collector
subjects:
- kind: ServiceAccount
  name: metrics-collector
  namespace: nonkubeflow
$ kubeflow apply -f  mcrbac.yaml
```

pkg/types.go
```diff
@@ -1,5 +1,5 @@
 package pkg
 
 const (
-	ManagerAddr = "vizier-core:6789"
+	ManagerAddr = "vizier-core.kubeflow:6789"
```
johnugeorge: Why is this change added ?

hougangliu: kube-dns will assign k8s services a DNS A record for a name of the form my-svc.my-namespace.svc.cluster.local and  my-svc.my-namespace. This resolves to the cluster IP of the Service.
vizier-core is service name, and if we don't add namespace, when Pod in non-kubeflow namespace cannot resolve vizier-core to its cluster IP (it tries to resolve vizier-core.non-kubeflow).

johnugeorge: Yes. I mean to say that vizier core can run in any namespace

PR #310, opened 2018-12-21
add hougangliu as a reviewer

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/310)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -4,8 +4,9 @@ approvers:
     - gaocegege
     - richardsliu
 reviewers:
-    - ddysher
-    - jose5918
     - libbyandhelen
     - texasmichelle
     - johnugeorge
+    - hougangliu
+    - ddysher
+    - jose5918


OWNERS
```diff
@@ -9,3 +9,4 @@ reviewers:
     - libbyandhelen
```
jlewi: Can you fix the indentation; these items should not be indented relative to jose5918.

hougangliu:
@jlewi It is very strange that all reviewers keep same indentation in my local repo, but in GitHub review page it looks these items are indented relative to jose5918.
When I cut down one indentation for all items. It seems that GitHub review page show rightly

jlewi:
Is there a mix of tabs and spaces?

hougangliu:
@jlewi I verified that all indentations are spaces. When I `dd` the first two lines and `yy` in the last line as change set. github page shows well now.
Maybe you can git clone and verify it if you're interested :)

PR #320, opened 2019-01-11
Add xgboost example using Bayesian optimization

richardsliu:
This example uses the image built from https://github.com/kubeflow/examples/tree/master/xgboost_ames_housing.

The hyperparameters being tuned are:
- Number of boost trees
- learning rate

We are trying to minimize the mean_absolute_error.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/320)

### DIFF
--- a/examples/xgboost-bayesian-example.yaml
+++ b/examples/xgboost-bayesian-example.yaml
@@ -1,0 +1,70 @@
+apiVersion: "kubeflow.org/v1alpha1"
+kind: StudyJob
+metadata:
+  namespace: kubeflow
+  labels:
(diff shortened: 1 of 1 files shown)


johnugeorge:
Is it using default metric collector?

richardsliu:
@johnugeorge Yes, the default metric collector works here (assuming that the training worker outputs logs in the expected format).

richardsliu:
@hougangliu The example docker image is being fixed in this PR: https://github.com/kubeflow/examples/pull/476. I added comments clarifying the prerequisites.

Issue #293, opened 2018-12-13
Extend studyjob structure for NAS jobs


andreyvelich:
We should extend current studyjob structure to parse yaml file for NAS jobs. 
Current yaml file for NAS job should be like this: https://github.com/DeeperMind/katib/blob/patch-1/examples/NAS-example.yaml. 
Related to: https://github.com/kubeflow/katib/issues/257.

andreyvelich:
This is new Yaml file structure 
https://github.com/andreyvelich/katib/blob/293-extend-sj-structure/examples/nasjob-example-RL.yaml

PR #327, opened 2019-01-17
Extend Katib API for NAS jobs

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/293
Related to: https://github.com/kubeflow/katib/issues/257


cmd/manager/main.go
```diff
@@ -29,34 +29,59 @@ type server struct {
...
+		}
+		s.SaveStudy(ctx, &api_pb.SaveStudyRequest{
+			StudyName:   in.StudyConfig.Name,
+			Owner:       in.StudyConfig.Owner,
+			Description: "StudyID: " + studyID,
+		})
+
+	} else {
+		//If it is a NAS job
+		studyID = "11111111111"
```
Akado2009: Should it really be "11111111"?

cmd/suggestion/nasrl/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3
+
+ADD . /usr/src/app/github.com/kubeflow/katib
+WORKDIR /usr/src/app/github.com/kubeflow/katib/cmd/suggestion/nasrl
+RUN pip install --no-cache-dir -r requirements.txt
+ENV PYTHONPATH /usr/src/app/github.com/kubeflow/katib:/usr/src/app/github.com/kubeflow/katib/pkg/api/python
+
+ENTRYPOINT ["python", "-u", "main.py"]
```
Akado2009: Empty line needed

cmd/suggestion/nasrl/main.py
```diff
@@ -0,0 +1,29 @@
...
+    print("Listening...")
+    server.start()
+    try:
+        while True:
+            time.sleep(_ONE_DAY_IN_SECONDS)
+    except KeyboardInterrupt:
+        server.stop(0)
+
+if __name__ == "__main__":
+    serve()
```
Akado2009: Empty line needed

examples/nasjob-example-RL.yaml
```diff
@@ -0,0 +1,142 @@
...
+      - name: "embedding_dim"
+        value: "20"
+      - name: "accuracy_beta"
+        value: "0.8"
+      - name: "clip_rewards"
+        value: "0"
+      - name: "embedding_dim"
+        value: "20"
+      - name: "discount_factor"
+        value: "0.95"
```
Akado2009: Empty line needed I believe

pkg/suggestion/nasrl_service.py
```diff
@@ -0,0 +1,28 @@
...
+        trials = []
+        study_conf = self.getStudyConfig(request.study_id)
+        print(study_conf)
+        return api_pb2.GetSuggestionsReply(trials=trials)
+    
+    def getStudyConfig(self, studyID):
+        channel = grpc.beta.implementations.insecure_channel(self.manager_addr, self.manager_port)
+        with api_pb2.beta_create_Manager_stub(channel) as client:
+            gsrep = client.GetStudy(api_pb2.GetStudyRequest(study_id=studyID), 10)
+            return gsrep.study_config
```
Akado2009: Empty line needed

andreyvelich:
Is this PR:
1) I extend API for structure of NAS yaml file.
2) Make blank NAS Reinforcement Learning service suggestion for future improvement
3) Extend manager with JobType functionality. We have to create our own tables for NAS config and NAS suggestion. We have to divide HP jobs and NAS jobs in the manager.
4) I am using the same StudyConfig to parse information from StudyJob. I extend StudyConfig structure and add new functionality inside getStudyConf function in katib_api_util. 
5) Add new type in ParameterType 
Please, review it.
@YujiOshima @hougangliu @jlewi @johnugeorge @xyhuang @ddutta

cmd/manager/main.go
andreyvelich: Change it to "1"

gaocegege:
Is it ready to review?

andreyvelich:
@gaocegege Yes, you can review it, please.

cmd/manager/main.go
```diff
@@ -29,34 +29,59 @@ type server struct {
...
+		}
+		s.SaveStudy(ctx, &api_pb.SaveStudyRequest{
+			StudyName:   in.StudyConfig.Name,
+			Owner:       in.StudyConfig.Owner,
+			Description: "StudyID: " + studyID,
+		})
+
+	} else {
+		//If it is a NAS job
+		studyID = "1"
```
gaocegege: Why should we set the study id to 1 here, I am not sure if I understand it.

cmd/manager/main.go
```diff
@@ -29,34 +29,59 @@ type server struct {
...
-		return &api_pb.DeleteStudyReply{}, err
+	if in.JobType != "nas" {
+		//If it is a HP job
+		err := dbIf.DeleteStudy(in.StudyId)
+		if err != nil {
+			return &api_pb.DeleteStudyReply{}, err
+		}
+	} else {
+		//If it is a NAS job
+		log.Printf("NAS job is deleted, id= %v", in.StudyId)
```
gaocegege: We do not need to delete the study, could you please give me more details here?

Akado2009:
Everyone and @gaocegege  in particular, sorry for the inconvenience, but this PR should be labeled as WIP as it heavily depends: https://github.com/kubeflow/katib/pull/328, which actually introduces additional APIs to delete NAS job/create them.

cmd/manager/main.go
Akado2009: See the commentary :)

cmd/manager/main.go
Akado2009: See the commentary :)

gaocegege:
OK, gotcha. Thanks for the explanation, I will go to review that PR :smile:

hougangliu:
> Everyone and @gaocegege in particular, sorry for the inconvenience, but this PR should be labeled as WIP as it heavily depends: #328, which actually introduces additional APIs to delete NAS job/create them. So we insist this PR should be merged after this one: #328

I think #328 and #327 should be in same PR since they depend on each other.

Akado2009:
@hougangliu  @YujiOshima isn't that gonna be too much code for a review?

hougangliu:
> @hougangliu @YujiOshima isn't that gonna be too much code for a review?

Anyway #328 can never pass kubeflow-katib-presubmit since it depends on change of this PR

gaocegege:
Personally, I think there is no need to merge these two PRs to one. We can review and merge one PR, then we deal with the other which depends on the first one.

andreyvelich:
@gaocegege @hougangliu Thank you for your review. I think better way to divide these 2 PRs in that way:
1) PR https://github.com/kubeflow/katib/pull/327 will contains only StudyJobController changes.
2) PR https://github.com/kubeflow/katib/pull/328 will contains only Manager (vizier-core) changes.
I will remove changes in Manager (vizier-core) in that PR https://github.com/kubeflow/katib/pull/327.
After this, you can merge first PR https://github.com/kubeflow/katib/pull/327 and, after that, merge PR https://github.com/kubeflow/katib/pull/328.
WDYT about it?

andreyvelich:
@hougangliu 
I can put GraphConfig and Operations under NasConfig, but I don't think we should avoid jobType field. 
We use this field to divide functionality in manager and exchange with DB.

andreyvelich:
@hougangliu Do you think is it a good idea to breake getStudyConf in two methods? We have some fields that NAS and HP configs include both. We should copy this parsing in two methods.

hougangliu:
> @hougangliu
> I can put GraphConfig and Operations under NasConfig, but I don't think we should avoid jobType field.
> We use this field to divide functionality in manager and exchange with DB.

we can get jobtype like below:
```
func getJobType(studyjobInstance) string {
    if studyjobInstance.NasConfig != nil {
        return "nas"
    } else {
        return "hp"
   }
}

I think we should do our best to reduce complexity of studyJob type for better UE
```

hougangliu:
> @hougangliu Do you think is it a good idea to breake getStudyConf in two methods? We have some fields that NAS and HP configs include both. We should copy this parsing in two methods.

I mean 
```
func getStudyConf(instance *katibv1alpha1.StudyJob) (*katibapi.StudyConfig, error) {
     // existing common part
   if getJobType(instance) == "nas" {
       populateConfigForNas(sconf, instance)
   } else {
      populateConfigForHp(sconf, instance)
   }
}
```

andreyvelich:
@hougangliu Is it good, if I add field jobType inside StudyJobConfig depending on NasConfig object? 
If I don't do it, I should transmit whole object here https://github.com/kubeflow/katib/pull/327/files#diff-89a3fff23e4189a95ec9434f963f6ab1R477, not just job_type.

johnugeorge:
I feel it is better we can move NAS specific utilities to separate file ?

Akado2009:
@johnugeorge What do you mean by "NAS-specific utilities"? 
If you mean protobuf and related stuff - then no, since they all come form StudyConfig.
If you mean functions in controller/DB API - then also no, since they are divided into HP/NAS at least now, so there is no point of doing that :)
Otherwise we don't know which utilities you are talking about

johnugeorge:
i mean, instead of having both HP and NAS utility functions  in katib_api_util.go, we can split into katib_hp_util and katib_nas_util for better code organization and readability. eg: populateConfigForNAS and populateConfigForHP 
Just a nitpick

andreyvelich:
@johnugeorge Do you think we still need to split this functionality? 
We have only one different between HP and NAS, that is functions populateConfigForNAS and populateConfigForHP.

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -79,12 +78,22 @@ func initializeStudy(instance *katibv1alpha1.StudyJob, ns string) error {
...
+		return populateConfigForNAS(instance)
 	}
+	return populateConfigForHP(instance)
+}
+
+func getJobType(instance *katibv1alpha1.StudyJob) string {
+	if instance.Spec.NasConfig != nil {
+		return "NAS"
+	}
+	return "HP"
```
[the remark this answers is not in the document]
andreyvelich: Thanks, done

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -79,12 +78,22 @@ func initializeStudy(instance *katibv1alpha1.StudyJob, ns string) error {
...
+}
+
+func populateCommonConfigFields(instance *katibv1alpha1.StudyJob, sconf *katibapi.StudyConfig) *katibapi.StudyConfig {
```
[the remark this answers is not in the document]
andreyvelich: Thanks, done

hougangliu:
@andreyvelich from this PR, I have no idea about:
1. what trials will be like based on nasConfig (if you have no plan to implement nas service in this PR, can you please add related comment to show an example of **nasConfig Input => trials Output** so that we can better understand it?)
2. can you provide an e2e example of nas studyjob, that is use a nas worker in nasjob-example-RL.yaml?

pkg/api/api.proto
```diff
@@ -232,6 +232,7 @@ enum ParameterType {
     INT = 2; /// Int type. Use "Max/Min".
     DISCRETE = 3; /// Discrete number type. Use "List" as float.
     CATEGORICAL = 4; /// Categorical type. Use "List" as string.
+    RANGE = 5; /// Range type. Use "Max/Min" and "Step" parameter.
```
YujiOshima: Do we really need Range ParameterType?
Only add step field, We can merge it to INT and Double.
If the range is not set, the default range is 1 for INT, 0 for Double.

andreyvelich:
@hougangliu 
1. We have plans to integrate NAS service in a separate PR. In this PR I have just initial version of the service. We have first version of our suggestion here: 
https://github.com/CiscoAI/katib/tree/master/pkg/suggestion. We need to modify this suggestion little bit, to correctly parse NAS config and suggestion parameters. 
We will create trials using same functionality as here: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/bayesian_service.py#L83,
but in ```value``` we will send encoding model using special way, that we will describe later. 
2. After we will create our training container and finish suggestion service, I can provide e2e example with correct NAS worker and current nasjob-example-RL.yaml.

examples/nasjob-example-RL.yaml
```diff
@@ -0,0 +1,142 @@
...
+  studyName: nas-rl-example
+  owner: crd
+  optimizationtype: maximize
+  objectivevaluename: Validation-accuracy
+  optimizationgoal: 0.99
+  requestcount: 4
+  metricsnames:
+    - accuracy
+  nasConfig:
+    graphConfig:
```
YujiOshima: Please explain about graphConfig and operations.
I believe The graphConfig is general information for NAS Job and Operations are task-specific info e.g. CNN.
Why not merge them? The graphConfig is able to encode to parameterconfig.

hougangliu:
> @hougangliu
> 
> 1. We have plans to integrate NAS service in a separate PR. In this PR I have just initial version of the service. We have first version of our suggestion here:
>    https://github.com/CiscoAI/katib/tree/master/pkg/suggestion. We need to modify this suggestion little bit, to correctly parse NAS config and suggestion parameters.
>    We will create trials using same functionality as here: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/bayesian_service.py#L83,
>    but in `value` we will send encoding model using special way, that we will describe later.
> 2. After we will create our training container and finish suggestion service, I can provide e2e example with correct NAS worker and current nasjob-example-RL.yaml.

@andreyvelich  Thanks! It sounds good. Look forward to e2e example

pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go
```diff
@@ -182,6 +185,25 @@ type StudyJobList struct {
 	Items           []StudyJob `json:"items"`
 }
 
+// NasConfig contains config for NAS job
+type NasConfig struct {
```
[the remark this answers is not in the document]
DeeperMind: The StudyJob config we are building here works universally for all NAS algorithms, including evolutionary methods proposed by _Regularized Evolution for Image Classiﬁer Architecture Search_.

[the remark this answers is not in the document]
hougangliu: Cool!

pkg/api/api.proto
DeeperMind: The incentive for adding Range ParameterType is that, to the best of our knowledge, none of the NAS algorithms supports continuous search space. All the search space must be discretized. 
Yes, what you said makes good sense. We will delete Range and add step in int and double.

examples/nasjob-example-RL.yaml
andreyvelich: Yes, GraphConfig is a initial general information for the Directed Acyclic Graph (DAG). We can't merge them, because for each StudyJob in NAS GraphConfig should be single and Operations can be as many as you want. Also, right now structure of Parameter config doesn't match with GraphConfig, we have to extend ParameterConfig as well in that case.

andreyvelich:
I made some changes inside metrics-collector and add function "GetStudyJobType" inside protobuf, please check it @hougangliu @YujiOshima @gaocegege
https://github.com/kubeflow/katib/pull/327/files#diff-82bfd7279d5e08ab134ad69029e09925R73

YujiOshima:
@andreyvelich Thank you!! LGTM except for @hougangliu 's comment.

andreyvelich:
@hougangliu @YujiOshima 
Thanks! I changed it.

PR #328, opened 2019-01-17
Database APIs for NAS

Akado2009:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/328)


Akado2009:
This PR introduced a distinct set of API call to save/get NAS jobs into the mysql databases. For now it creates a separate table, but after new update we're gonna use studies table, just increase the number of columns in that one. This one should be merged before: https://github.com/kubeflow/katib/pull/327 to introduce APIs into katib the latter misses.

pkg/db/db_init.go
```diff
@@ -15,9 +23,13 @@ func (d *dbConn) DBInit() {
 		optimization_goal DOUBLE,
 		parameter_configs TEXT,
 		tags TEXT,
+		trials TEXT,
```
YujiOshima: Please add a timestamp fields to Trial table instead of adding this to Study.

pkg/db/interface.go
```diff
@@ -341,6 +343,201 @@ func (d *dbConn) DeleteStudy(id string) error {
 	return err
 }
 
+func (d *dbConn) CreateNAS(in *api.StudyConfig) (string, error) {
```
YujiOshima: The level is different between CreateNAS and CreateStudy. Please change them to an equal level e.g. CreateNASStudy, CreateHPStudy.

pkg/db/interface.go
```diff
@@ -341,6 +343,201 @@ func (d *dbConn) DeleteStudy(id string) error {
...
+	if !isin {
+		in.Metrics = append(in.Metrics, in.ObjectiveValueName)
+	}
+
+	var studyID string
+	var trials string
+	var configs string
+
+	i := 3
+	for true {
```
YujiOshima: It looks the same process with CreateStudy. Please merge them and split only different process.

pkg/db/interface.go
```diff
@@ -46,10 +46,19 @@ type VizierDBInterface interface {
...
-	UpdateStudy(string, *api.StudyConfig) error
+
+	GetHPStudyConfig(string) (*api.StudyConfig, error)
+	CreateHPStudy(*api.StudyConfig) (string, error)
+	UpdateHPStudy(string, *api.StudyConfig) error
+
+	/* APIs for NAS */
+	GetNASStudyConfig(string) (*api.StudyConfig, error)
+	CreateNASStudy(*api.StudyConfig) (string, error)
+	UpdateNASStudy(string, *api.StudyConfig) error
```
hougangliu: I think we should drop NAS and HP interface here.  It is enough to expose only study for VizierDBInterface.
Saying: only CreateStudy is enough, we can drop CreateNASStudy and CreateHPStudy in VizierDBInterface and the two can be subfunctions of CreateStudy. The difference between them is that: which field should processed before insert. We can even process in a unified way in CreateStudy to handle both case:
```
var configs, nasConfig string
if in.ParameterConfigs {
  configs, err = (&jsonpb.Marshaler{}).MarshalToString(in.ParameterConfigs)
}
if in.NasConfig {
  nasConfig, err = (&jsonpb.Marshaler{}).MarshalToString(in.NasConfig)
}
```

YujiOshima: I agree with @hougangliu. For example, in  CreateNASStudy and CreateHPStudy, most of codes are duplicate. We can merge them.

cmd/manager/main.go
```diff
@@ -50,13 +68,20 @@ func (s *server) DeleteStudy(ctx context.Context, in *api_pb.DeleteStudyRequest)
...
 	return &api_pb.DeleteStudyReply{StudyId: in.StudyId}, nil
 }
 
 func (s *server) GetStudy(ctx context.Context, in *api_pb.GetStudyRequest) (*api_pb.GetStudyReply, error) {
-	sc, err := dbIf.GetStudyConfig(in.StudyId)
+
+	sc, err := dbIf.GetStudy(in.StudyId)
+
+	if err != nil {
+		log.Printf("Error is %v", err)
```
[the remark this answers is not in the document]
Akado2009: Deleted.

pkg/db/interface.go
Akado2009: Great catch! Didn't notice it firstly

andreyvelich:
@hougangliu @YujiOshima 
Can you finally review this PR, please? 
I think it's ready to merge.

johnugeorge:
Fix in https://github.com/kubeflow/katib/pull/363 is required..  Please rebase and push again.

pkg/db/interface.go
```diff
@@ -132,56 +135,74 @@ func NewWithSQLConn(db *sql.DB) (VizierDBInterface, error) {
...
-	return NewWithSQLConn(db)
+	retMetrics := strings.Split(metrics, ",\n")
+	return retMetrics, nil
 }
 
-func generateRandid() string {
-	// UUID isn't quite handy in the Go world
-	id := make([]byte, 8)
-	_, err := rand.Read(id)
+func (d *dbConn) GetStudyIDsTypesList() ([]string, []string, error) {
```
hougangliu: I did not find where GetStudyIDsTypesList is used. Maybe you can delete it.
BTW,  if you still persist to remain it. I think you can use "map[string]string" to replace "[]string, []string" in return where key is studyid and value is type

hougangliu:
@andreyvelich thanks! some  minor comment.

cmd/manager/main.go
```diff
@@ -50,13 +47,20 @@ func (s *server) DeleteStudy(ctx context.Context, in *api_pb.DeleteStudyRequest)
 	}
 	err := dbIf.DeleteStudy(in.StudyId)
 	if err != nil {
+		log.Printf("Error is %v", err)
```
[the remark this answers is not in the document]
Akado2009: Resolved

cmd/manager/main.go
```diff
@@ -50,13 +47,20 @@ func (s *server) DeleteStudy(ctx context.Context, in *api_pb.DeleteStudyRequest)
...
 	return &api_pb.DeleteStudyReply{StudyId: in.StudyId}, nil
 }
 
 func (s *server) GetStudy(ctx context.Context, in *api_pb.GetStudyRequest) (*api_pb.GetStudyReply, error) {
-	sc, err := dbIf.GetStudyConfig(in.StudyId)
+
+	sc, err := dbIf.GetStudy(in.StudyId)
+
+	if err != nil {
+		log.Printf("Error is %v", err)
```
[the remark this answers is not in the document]
Akado2009: resolved

pkg/db/interface.go
Akado2009: Resolved

pkg/db/interface_test.go
```diff
@@ -119,7 +119,7 @@ func TestGetStudyConfig(t *testing.T) {
 	mock.ExpectQuery("SELECT").WillReturnRows(
 		sqlmock.NewRows(studyColumns).AddRow(
 			"abc", "test", "admin", 1, 0.99, "{}", "", "", "", "test"))
-	study, err := dbInterface.GetStudyConfig(id)
+	study, err := dbInterface.GetStudy(id)
 	if err != nil {
```
andreyvelich: You should also change this things: 
1. Add 2 values in StudyColumns: https://github.com/kubeflow/katib/blob/297324dd7d88c32d81c63b7109f051e558db8a44/pkg/db/interface_test.go#L23
2. Add value "time" in trialColumns:
https://github.com/kubeflow/katib/blob/297324dd7d88c32d81c63b7109f051e558db8a44/pkg/db/interface_test.go#L27
3. Add 2 empty values:
https://github.com/kubeflow/katib/blob/297324dd7d88c32d81c63b7109f051e558db8a44/pkg/db/interface_test.go#L119
4. Add 1 empty value:
https://github.com/kubeflow/katib/blob/297324dd7d88c32d81c63b7109f051e558db8a44/pkg/db/interface_test.go#L216
5. Add 2 Values empty values:
https://github.com/kubeflow/katib/blob/297324dd7d88c32d81c63b7109f051e558db8a44/pkg/db/interface_test.go#L367
6. Add 1 empty value:
https://github.com/kubeflow/katib/blob/297324dd7d88c32d81c63b7109f051e558db8a44/pkg/db/interface_test.go#L233

andreyvelich:
I signed it!

jlewi:
The CLA check is failing because you have multiple commits with different authors. Please rebase and squash the commits so that all your commits are coming from a single author.

andreyvelich:
@YujiOshima @hougangliu @johnugeorge 
We finally fixed all presubmit tests. 
Can you check, please, all the difference that we made in that PR?

pkg/db/interface.go
```diff
@@ -223,33 +247,29 @@ func (d *dbConn) GetStudyList() ([]string, error) {
...
-			&dummyStudy.OptimizationType,
-			&dummyStudy.OptimizationGoal,
-			&dummyConfigs,
-			&dummyTags,
-			&dummyStudy.ObjectiveValueName,
-			&dummyMetrics,
-			&dummyJobID,
-		)
+		var temporaryId string
+		err := d.db.QueryRow("SELECT id FROM studies WHERE job_id = ?", in.JobId).Scan(&temporaryId)
```
YujiOshima: Please check `sql.ErrNoRows` error instead nil.

YujiOshima: Please check `sql.ErrNoRows` instead of nil check.

cmd/manager/main_test.go
```diff
@@ -8,14 +8,14 @@ import (
...
 	mockdb "github.com/kubeflow/katib/pkg/mock/db"
-	mockmodelstore "github.com/kubeflow/katib/pkg/mock/modelstore"
 )
 
 func TestCreateStudy(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 	mockDB := mockdb.NewMockVizierDBInterface(ctrl)
-	mockModelStore := mockmodelstore.NewMockModelStore(ctrl)
+	// Don't need test for Model Store. Right now Katib doesn't use ModelDB.
```
YujiOshima: Removing `mockModelStore` test is reasonable but it is not related to this PR. Could you split it to another PR?

pkg/db/interface.go
```diff
@@ -362,11 +379,13 @@ func (d *dbConn) getTrials(trialID string, studyID string) ([]*api.Trial, error)
 		trial := new(api.Trial)
 
 		var parameters, tags string
+		var timeStamp string
 		err := rows.Scan(&trial.TrialId,
 			&trial.StudyId,
 			&parameters,
 			&trial.ObjectiveValue,
 			&tags,
+			&timeStamp,
```
YujiOshima: In other grpc APIs, `RFC3339Nano` format is used, not MySQL format.
Please convert timeStamp to the format when returning api.Trial.

cmd/manager/main_test.go
andreyvelich: Should I split changing version of the cluster (https://github.com/kubeflow/katib/pull/328/files#diff-3d8f11bfc35c15745c970724619c57c3L36) in another PR, as well? 
Also, should I delete tests for mockModelStore or just comment it?

pkg/db/interface.go
andreyvelich: Right now, in the Trial API we don't have time field.
https://github.com/kubeflow/katib/blob/master/pkg/api/api.proto#L363
We store it in database, but can't get with grpc APIs.
Should we extend this API with time and add parsing in the separate PR?

pkg/db/interface.go
andreyvelich: Done

cmd/manager/main_test.go
YujiOshima: >changing version of the cluster

Yes, please split it.

> `mockModelStore`

You can delete it.

pkg/db/interface.go
YujiOshima: I see. 
Yes, please add a timestamp field to grpc API and write doc about the format.

andreyvelich: Ok, I will do it in a separate PR.

Issue #317, opened 2019-01-07
StudyJob controller shouldn't crash if PyTorch (or other job operators not installed)


jlewi:
See: kubeflow/kubeflow#2212

User reports that the StudyJob controller crashes if pytorch operator isn't installed on the cluster.

Logs for StudyJobController

```
$ kubectl logs -n kubeflow studyjob-controller-7d77f959-pjfzf
2019/01/06 22:43:34 Registering Components.
2019/01/06 22:43:34 controller.AddToManager(mgr)
2019/01/06 22:43:34 no matches for kind "PyTorchJob" in version "kubeflow.org/v1beta1"
```

This seems like a bug. If a particular job controller isn't installed I would still expect katib and StudyJobs to work with other types of job controllers.

johnugeorge:
it should be happening because schemes of job operators(tfjob and pytorch) are added during init.   

https://github.com/kubeflow/katib/blob/master/pkg/api/operators/apis/addtoscheme_tfjob_v1beta1.go#L21

Operator deployments are not needed if corresponding workers are not needed in Katib. However, there is a requirement of having operator crds to be installed. 

https://github.com/kubeflow/katib/blob/master/scripts/deploy.sh#L36

/cc @richardsliu

jlewi:
@johnugeorge Can you elaborate? Why does linking in the spec for TFJob (https://github.com/kubeflow/katib/blob/master/pkg/api/operators/apis/addtoscheme_tfjob_v1beta1.go#L21) create a runtime dependency on the CRD being installed in the cluster? I would have thought that just creates a compile time dependency?

I guess my bigger question is; what's the long term plan here for Katib in terms of how its going to fire of K8s resources to do the actual training? Is this covered in either of the design docs for Katib?

johnugeorge:
@jlewi  Runtime error should be where we do a watch on the jobs during init

https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L114

can you elaborate more on "to fire off K8s resources to do the actual training" ?

jlewi:
@johnugeorge 

> Operator deployments are not needed if corresponding workers are not needed in Katib. However, there is a requirement of having operator crds to be installed.

Why is there a requirement to have operator CRDs to be installed if they aren't being used? This seems like a limitation of the current implementation.


[...]

PR #335, opened 2019-01-22
ignore tfjob/pytorch job if corresponding CRD not created

hougangliu:
As a short-term solution
Fixes: #317

### DIFF
--- a/pkg/controller/studyjob/studyjob_controller.go
+++ b/pkg/controller/studyjob/studyjob_controller.go
@@ -51,6 +51,10 @@ const (
 	cleanDataFinalizer = "clean-studyjob-data"
 )
 
+var (
+	invalidCRDResources [] string
+)
+
 /**
 * USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
 * business logic.  Delete these comments after modifying this file.*
@@ -117,7 +121,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 			OwnerType:    &katibv1alpha1.StudyJob{},
 		})
 	if err != nil {
-		return err
+		invalidCRDResources = append(invalidCRDResources, TFJobWorker)
+		log.Printf("Fail to watch TFJob resource: %v", err)
 	}
 
 	err = c.Watch(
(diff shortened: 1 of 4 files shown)


richardsliu:
@johnugeorge Do we still need this fix based on what we discussed today?

pkg/controller/studyjob/utils.go
```diff
@@ -39,6 +39,58 @@ func createWorkerJobObj(kind string) runtime.Object {
 	return nil
 }
 
+func validateWorkerResource(wkind string) error {
+	for _, crd := range invalidCRDResources {
+		if crd == wkind {
+			return fmt.Errorf("Cannot support %s; If CRD for it installed, please restart studyjob-controller to take effect", wkind)
```
richardsliu: Maybe specifically say "Please install the CRD and restart studyjob-controller"?

hougangliu: Thanks! updated

pkg/controller/studyjob/studyjob_controller.go
```diff
@@ -117,7 +121,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 			OwnerType:    &katibv1alpha1.StudyJob{},
 		})
 	if err != nil {
```
johnugeorge: invalidCRDResources should be updated only when CRD is not installed. We should explicitly check whether CRD exists or not. If watch fails for any other reason(even when CRD is installed), controller should fail by raising error. 

You can use crd clientset  to check. 
 https://github.com/kubeflow/tf-operator/blob/master/cmd/tf-operator.v1beta1/app/server.go#L192

hougangliu: OK, I will handle it

pkg/controller/studyjob/utils.go
```diff
@@ -39,6 +40,70 @@ func createWorkerJobObj(kind string) runtime.Object {
...
+func validateWorkerResource(wkind string) error {
+	for _, crd := range invalidCRDResources {
+		if crd == wkind {
+			return fmt.Errorf("Cannot support %s; Please install the CRD and restart studyjob-controller", wkind)
+		}
+	}
+	return nil
+}
+
+func ignoreWatchError(err error, job string) bool {
```
richardsliu: The name of the function is a little confusing. Maybe change it to "isFatalWatchError" and return the opposite boolean values?

hougangliu: Thanks! Done!

PR #336, opened 2019-01-22
omit empty fields for studyjob status

hougangliu:
EarlyStoppingParameterID should not be shown in studyjob status to mislead user since EarlyStopping feature cannot work now.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/336)


hougangliu:
BTW, will EarlyStopping feature be in katib plan? @YujiOshima @richardsliu @gaocegege

Issue #338, opened 2019-01-23
Suggestion for Neural Architecture Search with Reinforcement Learning


DeeperMind:
Create a suggestion for Neural Architecture Search. 

Related to: https://github.com/kubeflow/katib/issues/257

PR #339, opened 2019-01-23
Suggestion for Neural Architecture Search with Reinforcement Learning

DeeperMind:
Fixes: https://github.com/kubeflow/katib/issues/338.

The first suggestion we are working on is reinforcement learning. The algorithm follows the paper *Neural Architecture Search with Reinforcement Learning* by Zoph & Le (https://arxiv.org/abs/1611.01578), and the implementation is based on the github of *Efﬁcient Neural Architecture Search via Parameter Sharing* (https://github.com/melodyguan/enas).

Related to https://github.com/kubeflow/katib/issues/257

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/339)


YujiOshima:
@DeeperMind Thank you for your hard work!
To test this, we need https://github.com/kubeflow/katib/pull/328 right?
Should we focus on #328 before this?

andreyvelich:
@YujiOshima 
Yes, we should merge PR about database API first. We will remove WIP status from this. PR has been done already, you can review it: https://github.com/kubeflow/katib/pull/328

YujiOshima:
@andreyvelich Okay, Thanks.

pkg/suggestion/NAS_Reinforcement_Learning/README.md
```diff
@@ -0,0 +1,123 @@
+# About the Nerual Architecture Search with Reinforcement Learning Suggestion
+
+The algorithm follows the idea proposed in *Neural Architecture Search with Reinforcement Learning* by Zoph & Le (https://arxiv.org/abs/1611.01578), and the implementation is based on the github of *Efﬁcient Neural Architecture Search via Parameter Sharing* (https://github.com/melodyguan/enas). It uses a recurrent neural network with LSTM cells as controller to generate neural archiecture canddiates. And this controller network is updated by policy gradients. However, it currently does not support parameter sharing. 
+
```
hougangliu: s/canddiates/candidates
s/archiecture/architecture

pkg/suggestion/NAS_Reinforcement_Learning/README.md
```diff
@@ -0,0 +1,123 @@
...
+[5 0 0 1 0 1]
+[1 1 1 0 0 1 0]
+[2 0 0 0 1 1 0 1]
+[0 0 0 1 1 1 1 1 0]
+[2 0 1 0 1 1 1 0 0 0]
+[3 1 1 1 1 1 1 0 0 1 1]
+[0 1 1 1 1 0 0 1 1 1 1 0]
+```
+
+There are n rows, the i<sup>th</sup> row has n elements and describes the i<sup>th</sup> layer. Please notice that layer 0 is the input and is not included in this definition.
```
hougangliu: the i<sup>th</sup> row has **i** elements

pkg/suggestion/NAS_Reinforcement_Learning/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+In each row:
+The first integer ranges from 0 to m-1, indicates the operation in this layer.
+The next (i-1) integers is either 0 or 1. The k<sup>th</sup> (k>=2) integer indicates whether (k-2)<sup>th</sup> layer has a skip connection with this layer. (There will always be a connection from (k-1)<sup>th</sup> layer to k<sup>th</sup> layer)
+
+## Output of `GetSuggestion()`
+The output of `GetSuggestion()` consists of two parts: `architecture` and `nn_config`.
+
+`architecture` is a json string of the deinition of a neural architecture. The format is as stated above. One example is:
+```
```
hougangliu: s/deinition/definition

pkg/suggestion/nasrl_service.py
```diff
@@ -1,22 +1,227 @@
...
+                skip_target=self.suggestion_config['skip-target'],
+                skip_weight=self.suggestion_config['skip-weight'],
+                name="Ctrl_"+self.current_study_id)
+
+            self.controller.build_trainer()
+
+        print("Suggestion for StudyJob {} has been initialized.".format(request.study_id))
 
     def GetSuggestions(self, request, context):
+        if request.study_id != self.current_study_id:
```
hougangliu: A user may submit two studyjob both with nasrl suggestion, and the two studyjob may call GetSuggestions alternately. It seems NasrlService can only serve one studyjob at same time. right?

pkg/suggestion/nasrl_service.py
```diff
@@ -1,22 +1,227 @@
...
+                trials[i].trial_id = ctrep.trial_id
+                self.current_trial_id = ctrep.trial_id
+            print("Trial {} Created\n".format(ctrep.trial_id))
+            self.current_trial_id = ctrep.trial_id
+        
+        self.ctrl_step += 1
         return api_pb2.GetSuggestionsReply(trials=trials)
+
+    def GetEvaluationResult(self, studyID):
+        worker_hist = []
```
hougangliu: s/worker_hist/worker_list

pkg/suggestion/nasrl_service.py
```diff
@@ -1,22 +1,227 @@
...
+        with api_pb2.beta_create_Manager_stub(channel) as client:
+            gwfrep = client.GetWorkerFullInfo(api_pb2.GetWorkerFullInfoRequest(study_id=studyID, trial_id=self.current_trial_id, only_latest_log=True), 10)
+            worker_hist = gwfrep.worker_full_infos
+
+        for w in worker_hist:
+            if w.Worker.status == api_pb2.COMPLETED:
+                for ml in w.metrics_logs:
+                    print("Evaluation result of previous candidate:", ml.values[-1].value)
+
+        return float(ml.values[-1].value)
```
hougangliu: if there is no "w.Worker.status == api_pb2.COMPLETED", GetEvaluationResult will trigger exception

pkg/suggestion/nasrl_service.py
```diff
@@ -1,22 +1,227 @@
...
+        nn_config_str = str(nn_config_json).replace('\"', '\'')
+
+        print("\nNew Neural Network Architecture (internal representation):")
+        print(organized_arc_json)
+        print("\nCorresponding Seach Space Description:")
+        print(nn_config_str)
+        print()
+
         trials = []
+        trials.append(api_pb2.Trial(
```
hougangliu: always only return one trial, ignoring requestNumber?

pkg/suggestion/nasrl_service.py
hougangliu: I suggest keep line 182 same indent with line 179, had better handle "no worker status == api_pb2.COMPLETED" case (although we except at least one worker changing to completed trigger this method).

pkg/suggestion/nasrl_service.py
```diff
@@ -1,22 +1,227 @@
...
+    def GetEvaluationResult(self, studyID):
+        worker_hist = []
+        channel = grpc.beta.implementations.insecure_channel(self.manager_addr, self.manager_port)
+        with api_pb2.beta_create_Manager_stub(channel) as client:
+            gwfrep = client.GetWorkerFullInfo(api_pb2.GetWorkerFullInfoRequest(study_id=studyID, trial_id=self.current_trial_id, only_latest_log=True), 10)
+            worker_hist = gwfrep.worker_full_infos
+
+        for w in worker_hist:
+            if w.Worker.status == api_pb2.COMPLETED:
+                for ml in w.metrics_logs:
```
hougangliu: GetEvaluationResult should only return the last value of objectivevaluename metrics. so I think if `ml.name != objectivevaluename`,  just `continue`

pkg/suggestion/nasrl_service.py
```diff
@@ -1,22 +1,227 @@
...
+                    saver.save(sess, self.ctrl_cache_file)
+
+                self.is_first_run = False
+
+            else:
+                with tf.Session() as sess:
+                    saver.restore(sess, self.ctrl_cache_file)
+
+                    valid_acc = self.controller.reward
+                    result = self.GetEvaluationResult(request.study_id)
```
hougangliu: We can define optimizationtype of a studyjob, which can be `maximize` and `minimize`. saying if objectivevaluename is `loss`, we hope its value change smaller and smaller(optimizationtype is minimize).

GetEvaluationResult should return last value of objectivevaluename metrics. and return of GetEvaluationResult is used to feed controller.reward (which hopes bigger and bigger).

So if a studyjob wants to get suggestion from suggestion-nasrl with `minimize` optimizationtype, GetEvaluationResult should return `-(lastOfobjectivevalue)`?

pkg/suggestion/NAS_Reinforcement_Learning/Operation.py
```diff
@@ -0,0 +1,81 @@
...
+                    double_list = np.arange(spec_min, spec_max+spec_step, spec_step)
+                    if double_list[-1] > spec_max:
+                        del double_list[-1]
+                    avail_space[spec_name] = double_list
+
+                avail_space_len.append(len(avail_space[spec_name]))
+
+            # generate all the combinations of possible operations
+            key_avail_space = list(avail_space.keys())
+            val_avail_space = list(avail_space.values())
```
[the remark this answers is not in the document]
DeeperMind: For the python version inside our container, dict.keys() return a dict_keys object, which does not support indexing. So we have to keep the list() here.

pkg/suggestion/NAS_Reinforcement_Learning/LSTM.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import tensorflow as tf
+
```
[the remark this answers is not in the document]
DeeperMind: tf.contrib module contains volatile or experimental code. The API in tf.contrib can deprecate fast. So it is more advisable not to use this module.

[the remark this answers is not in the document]
YujiOshima: How about  `tf.nn.rnn_cell.LSTMCell` ?

pkg/suggestion/nasrl_service.py
DeeperMind: At the current stage, we will return one trial each time. We will add support for multiple trials in the near future.

pkg/suggestion/NAS_Reinforcement_Learning/Operation.py
```diff
@@ -0,0 +1,81 @@
...
+                    spec_max = int(ispec.feasible.max)
+                    spec_step = int(ispec.feasible.step)
+                    avail_space[spec_name] = range(spec_min, spec_max+1, spec_step)
+                elif ispec.parameter_type == api_pb2.DOUBLE:
+                    spec_min = float(ispec.feasible.min)
+                    spec_max = float(ispec.feasible.max)
+                    spec_step = float(ispec.feasible.step)
+                    if spec_step == 0:
+                        print("Error, NAS Reinforcement Learning algorithm cannot accept continuous search space!")
+                        exit(999)
```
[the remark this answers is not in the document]
DeeperMind: Yes. If the definition of a study is invalid, there is no way we could start a task.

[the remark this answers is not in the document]
hougangliu: so If I submit a invalid studyjob, suggestion service exit; Then I submit another valid studyjob, but nothing can work since I cannot get suggestion from (suggestion service).
I think raising a exception makes sense for a invalid studyjob

pkg/suggestion/NAS_Reinforcement_Learning/Operation.py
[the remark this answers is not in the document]
hougangliu: Ok, I see. thanks

pkg/suggestion/nasrl_service.py
hougangliu: OK, maybe you can create an issue to trace it

pkg/suggestion/NAS_Reinforcement_Learning/Operation.py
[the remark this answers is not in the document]
andreyvelich: I think we should check that StudyJob is valid in initialize part in the StudyJob controller. Just like you notice here: https://github.com/kubeflow/katib/issues/372. 
What do you think about extending ```validateStudy``` function here: https://github.com/kubeflow/katib/blob/f11c13e8c65a9236afe2878e5c915a1c769ad52b/pkg/controller/studyjob/katib_api_util.go#L28 ?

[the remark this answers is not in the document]
hougangliu: agree. so suggestion service can always take each studyjob as valid one.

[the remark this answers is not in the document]
andreyvelich: Ok. I will make this changes tomorrow.

pkg/suggestion/nasrl_service.py
DeeperMind: Could I make the support for multiple studyjobs a separate PR?

pkg/suggestion/NAS_Reinforcement_Learning/LSTM.py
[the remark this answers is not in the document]
DeeperMind: The code is borrowed from https://github.com/melodyguan/enas/blob/master/src/cifar10/image_ops.py, its correctness can be guaranteed. Changing to tf.nn.rnn_cell.LSTMCell may incur many other issues such as incompatible function parameters in my Controller.py. 
If you think it is better to use tensorflow API, I will open another PR after this one is merged

pkg/suggestion/nasrl_service.py
DeeperMind: Sure, I will add a TODO here.

pkg/suggestion/NAS_Reinforcement_Learning/Operation.py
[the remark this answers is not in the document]
DeeperMind: I see. I will remove the checking here after the validateStudy is modified and merged.

pkg/suggestion/nasrl_service.py
hougangliu: sure, please create an new issue to trace it

pkg/suggestion/NAS_Reinforcement_Learning/LSTM.py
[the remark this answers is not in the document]
hougangliu: agree

pkg/suggestion/nasrl_service.py
```diff
@@ -1,22 +1,227 @@
...
+        self.ctrl_cache_file = ""
+        self.ctrl_step = 0
+        self.tf_graph = tf.get_default_graph()
+        self.is_first_run = True
+
+    def reset_controller(self, request):
+        print("-" * 80 + "\nResetting Suggestion for StudyJob {}\n".format(request.study_id) + "-" * 80)
+        self.ctrl_step = 0
+        self.current_study_id = request.study_id
+        self.ctrl_cache_file = "ctrl_cache/{}.ckpt".format(self.current_study_id)
```
hougangliu: please store checkpoint of a controller by pvc in case of suggestion-naslr service pod restarts sowehow.

DeeperMind: Sure, will add it in TODO. I will add the support for multiple jobs, multiple trials, change to use tf package and change to use PVC within one month after this PR is merged.

pkg/suggestion/nasrl_service.py
hougangliu: w.metrics_logs includes all metrics of studyjob worker, including objectivevaluename metrics. Here you should only consider  objectivevaluename metrics.

pkg/suggestion/NAS_Reinforcement_Learning/README.md
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+    }
+}
+```  
+This neural architecture can be visualized as
+![a neural netowrk architecure example](example.png)
+
+## To Do
+1. Add support fot multiple studyjobs
```
hougangliu: s/fot/for

pkg/suggestion/nasrl_service.py
DeeperMind: thank you for your explanation!

pkg/suggestion/nasrl_service.py
```diff
@@ -1,22 +1,271 @@
...
+        channel = grpc.beta.implementations.insecure_channel(self.manager_addr, self.manager_port)
+        with api_pb2.beta_create_Manager_stub(channel) as client:
+            gwfrep = client.GetWorkerFullInfo(api_pb2.GetWorkerFullInfoRequest(study_id=studyID, trial_id=self.prev_trial_id, only_latest_log=True), 10)
+            worker_list = gwfrep.worker_full_infos
+
+        completed_count = 0
+        for w in worker_list:
+            if w.Worker.status == api_pb2.COMPLETED:
+                for ml in w.metrics_logs:
+                    if ml.name != self.objective_name:
```
hougangliu: if w.metrics_logs is like [objectivemetrics, othermetrics], line 214 will return othermetrics value (ml is othermetrics)

how about
```
if ml.name == self.objective_name:
    self.logger.info("Evaluation result of previous candidate: {}".format(ml.values[-1].value))
    return float(ml.values[-1].value)
```
and keep below to keep same indent with line 204 (completed_count can be removed)
```
self.logger.warning("Error. No trial has completed.")
return None
```

DeeperMind: I believe w.metrics_logs will only return the selected metrics defined in https://github.com/kubeflow/katib/blob/master/examples/nasjob-example-RL.yaml#L12. The metrics collector will parse the pattern <metrics_name>: <metrics_value> from the log of training container and give it to suggestion.

Currently, the algorithm only spawns one trial each time and uses the metrics of the very last trials to update. That is why I will go through all the available metrics and index the last one.

hougangliu: w.metrics_logs will return Collection of https://github.com/kubeflow/katib/blob/master/examples/nasjob-example-RL.yaml#L12 and https://github.com/kubeflow/katib/blob/master/examples/nasjob-example-RL.yaml#L15. Maybe you can take a look by debug to verify it

DeeperMind: Yes, you are right! 
The program has been modified as suggested. Thank you for your reminder!

pkg/suggestion/nasrl_service.py
```diff
@@ -125,11 +125,6 @@ def GetSuggestions(self, request, context):
                         time.sleep(20)
                         result = self.GetEvaluationResult(request.study_id)
 
-                    # This lstm cell is designed to maximize the metrics
-                    # However, if the user want to minimize the metrics, we can take the negative of the result
-                    if self.opt_direction == api_pb2.MINIMIZE:
-                        result = -result
-
```
hougangliu: why delete above?

DeeperMind: Sorry, I made a mistake. Should delete the code above that. Already fixed

hougangliu: ok, thanks

Issue #346, opened 2019-01-28
Extending StudyJob API by adding more trials to finished StudyJob


janvdvegt:
For our usecase we would like to be able to extend finished studyjobs by adding more trials to it even after it is finished. If we would submit a new one instead they are seen as a completely new instance which means it's more difficult to compare the results and it cannot leverage the previous results for smarter hyperparameter search.

gaocegege:
Thanks for the issue. Do you mean we need to support creating new trials for the finished studies?

richardsliu:
This is interesting. I think @janvdvegt is asking for a way to resubmit a StudyJob that has already completed, to continue off the results from the previous trials. This brings up some additional questions:

- Do we want to allow resubmitting study jobs with different parameter search spaces?
- What about adding/removing hyperparameters?
- Changing the suggestion algorithm/parameters?

YujiOshima:
@janvdvegt I think my PR #352 is related your use-case. Maybe not enough yet.

janvdvegt:
I'm not 100% sure @YujiOshima how related it is. Unfortunately my Go is not enough to disect your PR.

@gaocegege Yeah I guess. Maybe my proposal is not the best way to do it, but I would like to be able to continue a StudyJob after the initially submitted amount of suggestions.

@richardsliu In my case I would not necessarily need to allow the first two options but it would not hurt either. The third option would not be necessary but would be nice to have. 

For me, most importantly I would just like to be able to start an experiment without having to say up front how many trials I would like to do. This would allow me to just start an experiment run and then later automatically if there are idle resources continue searching, or if the results seem promising continue the job. In that sense your three questions don't apply directly but would not hurt at all.

hougangliu:
#291 is about how we should consider update for `completed` and `failed` studyjob. Maybe if we want to trigger new trial by updating requestCount

gaocegege:
I am thinking if we could implement trial as a CRD. Then we could have two CRDs: StudyJob and Trial. And the relation of the two resources is similar to Job and Pod. In this way, we could fix this issue natively. Besides this, there are some other benefits:

- Make managing resources of katib easier
- Easy to implement the UI (Just CRUD two Kubernetes CRs)
- More K8s native

/cc @YujiOshima @johnugeorge @richardsliu @hougangliu @gyliu513

johnugeorge:
Interesting idea. Can you create a proposal with more details?

gaocegege:
[...]

Issue #122, opened 2018-06-19
Getting optimal parameters after a study is complete


vinaykakade:
Is there a recommended way of getting the optimal parameters when a study is complete? For example, it would be useful to get the parameters that have the maximum (or minimum) value for a the objective metric once the study is complete, so that the model could be further trained with those parameters.

jlewi:
@YujiOshima @gaocegege Do you know the answer?

jlewi:
@YujiOshima ping?

YujiOshima:
@vinaykakade Do you mean retrain or fine-tuning?
If you mean retrain,  the Hyperparameters are needed.
It is store Katib DB and you can find in KatibUI or get Katib API.

For fine-tuning, the trained weight parameters are needed.
It is not managed in Katib now. 
It is exactly the experimental tracking area. (discussed here https://github.com/kubeflow/community/pull/195 )
Katib will integrate Model management(StudioML or MLFlow) and the model data will be managed there.

YujiOshima:
@vinaykakade I agree. Easy to retraining is very useful.
How about introducing a one-shot job API.
We can specify a Trial ID or parameters directly, and it will be evaluated without any suggestion service.

hougangliu:
I think #312 solved this requirement.
After studyJob completed, user can get bestTrialId from studyJob.Status.BestTrialID. Then call grpc api GetTrial to get trial detail which includes optimal parameters. For example:
```
    with grpc.insecure_channel('localhost:6790') as channel:
        stub = api_pb2_grpc.ManagerStub(channel)
        response = stub.GetTrial(api_pb2.GetTrialRequest(trial_id="n51572c70f020c3d"))
```

YujiOshima:
@hougangliu I think it is not enough. We need to add a one-shot job API for specific trial ID. I will work on it.
/assign @YujiOshima

richardsliu:
We can close this one since the design for v1alpha2 API covers this. Implementation will be completed in 0.6.

/close

PR #315, opened 2019-01-04
delete obsolete data in db

hougangliu:
Fixes: #311

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/315)

PR #352, opened 2019-01-30
Manual suggest

YujiOshima:
You can run one-shot suggestion.
Generate trial in two way in manual suggestion.
* Specify all parameters manually.  https://github.com/YujiOshima/hp-tuning/blob/eff191d3f4191e8097f26914f8931aca181fe911/examples/manual-example.yaml
* Use Trial ID that is created in the past. https://github.com/YujiOshima/hp-tuning/blob/eff191d3f4191e8097f26914f8931aca181fe911/examples/manual-example-tid.yaml

By this PR, you can reuse studyID. ref https://github.com/YujiOshima/hp-tuning/blob/eff191d3f4191e8097f26914f8931aca181fe911/examples/manual-example-tid.yaml#L9
It enables
* Add a new trial to completed trial.
* Set a initial value of hyper-parameter serch by manual suggestion. Then run a bayse-opt suggestion.
* Run multiple suggestion paralelly.
* Rerun a best trial.

Related #122 #346 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/352)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -15,7 +15,9 @@ import (
 	"github.com/kubeflow/katib/pkg/manager/modelstore"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/reflection"
+	"google.golang.org/grpc/status"
 )
 
 const (
@@ -32,7 +34,17 @@ func (s *server) CreateStudy(ctx context.Context, in *api_pb.CreateStudyRequest)
 	if in == nil || in.StudyConfig == nil {
 		return &api_pb.CreateStudyReply{}, errors.New("StudyConfig is missing.")
 	}
+	if in.StudyConfig.JobId != "" {
+		get_study_id, _ := dbIf.GetStudyIDfromJob(in.StudyConfig.JobId)
+		if get_study_id != "" {
+			return &api_pb.CreateStudyReply{StudyId: get_study_id}, status.Errorf(codes.AlreadyExists, "Job UUID is already registerd.")
+		}
+	}
 	studyID, err := dbIf.CreateStudy(in.StudyConfig)
+	if err != nil {
+		return &api_pb.CreateStudyReply{}, err
+	}
+	err = dbIf.RegisterStudyJob(in.StudyConfig.JobId, studyID, in.StudyConfig.JobType)
 	if err != nil {
 		return &api_pb.CreateStudyReply{}, err
 	}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -37,7 +37,7 @@ func (s *server) CreateStudy(ctx context.Context, in *api_pb.CreateStudyRequest)
 	if in.StudyConfig.JobId != "" {
 		get_study_id, _ := dbIf.GetStudyIDfromJob(in.StudyConfig.JobId)
 		if get_study_id != "" {
-			return &api_pb.CreateStudyReply{StudyId: get_study_id}, status.Errorf(codes.AlreadyExists, "Job UUID is already registerd.")
+			return &api_pb.CreateStudyReply{StudyId: get_study_id}, status.Errorf(codes.AlreadyExists, "StudyJob with UUID %s registered already", in.StudyConfig.JobId)
 		}
 	}
 	studyID, err := dbIf.CreateStudy(in.StudyConfig)
--- a/examples/manual-example.yaml
+++ b/examples/manual-example.yaml
@@ -11,7 +11,6 @@ spec:
   optimizationtype: maximize
   objectivevaluename: Validation-accuracy
   optimizationgoal: 0.99
-  requestcount: 10
   metricsnames:
     - accuracy
   parameterconfigs:
@@ -70,4 +69,3 @@ spec:
       - 
           name: "--optimizer"
           value: "sgd"
-    requestNumber: 1
--- a/examples/manual-example.yaml
+++ b/examples/manual-example.yaml
@@ -1,0 +1,73 @@
+apiVersion: "kubeflow.org/v1alpha1"
+kind: StudyJob
+metadata:
+  namespace: kubeflow
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: manual-example
+spec:
+  studyName: manual-example
+  owner: crd
+  optimizationtype: maximize
+  objectivevaluename: Validation-accuracy
+  optimizationgoal: 0.99
+  requestcount: 10
+  metricsnames:
+    - accuracy
+  parameterconfigs:
+    - name: --lr
+      parametertype: double
+      feasible:
+        min: "0.01"
+        max: "0.03"
+    - name: --num-layers
+      parametertype: int
+      feasible:
+        min: "1"
+        max: "4"
+    - name: --optimizer
+      parametertype: categorical
+      feasible:
+        list:
+        - sgd
+        - adam
+        - ftrl
+  workerSpec:
+    goTemplate:
+        rawTemplate: |-
+          apiVersion: batch/v1
+          kind: Job
+          metadata:
+            name: {{.WorkerID}}
+            namespace: kubeflow
+          spec:
+            template:
+              spec:
+                imagePullSecrets:
+                  - name: gitlabregcred
+                containers:
+                - name: {{.WorkerID}}
+                  image: katib/mxnet-mnist-example
+                  command:
+                  - "python"
+                  - "/mxnet/example/image-classification/train_mnist.py"
+                  - "--batch-size=32"
+                  {{- with .HyperParameters}}
+                  {{- range .}}
+                  - "{{.Name}}={{.Value}}"
+                  {{- end}}
+                  {{- end}}
+                restartPolicy: Never
+  suggestionSpec:
+    suggestionAlgorithm: "manual"
+    suggestionParameters:
+      - 
+          name: "--lr"
+          value: "0.02"
+      - 
+          name: "--num-layers"
+          value: "2"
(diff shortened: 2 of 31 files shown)


pkg/controller/studyjob/katib_api_util.go
```diff
@@ -22,41 +22,58 @@ import (
...
-		return err
-	}
-	log.Printf("Create Study %s", studyConfig.Name)
-	//CreateStudy
-	studyID, err := createStudy(c, studyConfig)
-	if err != nil {
-		return err
+	if instance.Spec.ReuseStudyID != "" {
+		_, err = getStudyfromDB(c, instance.Spec.ReuseStudyID)
+		instance.Status.Condition = katibv1alpha1.ConditionFailed
```
hougangliu: why set it Failed here

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -249,14 +267,27 @@ func deleteStudy(instance *katibv1alpha1.StudyJob) error {
...
+	}
+	return getStudyrep.StudyConfig, nil
+}
+
 func createStudy(c katibapi.ManagerClient, studyConfig *katibapi.StudyConfig) (string, error) {
 	ctx := context.Background()
 	createStudyreq := &katibapi.CreateStudyRequest{
 		StudyConfig: studyConfig,
 	}
+	log.Printf("StudyJob UUID %s", studyConfig.JobId)
```
hougangliu: Create study for StudyJob UUID %s

cmd/manager/main.go
```diff
@@ -32,10 +34,20 @@ func (s *server) CreateStudy(ctx context.Context, in *api_pb.CreateStudyRequest)
 	if in == nil || in.StudyConfig == nil {
 		return &api_pb.CreateStudyReply{}, errors.New("StudyConfig is missing.")
 	}
+	if in.StudyConfig.JobId != "" {
+		get_study_id, _ := dbIf.GetStudyIDfromJob(in.StudyConfig.JobId)
+		if get_study_id != "" {
+			return &api_pb.CreateStudyReply{StudyId: get_study_id}, status.Errorf(codes.AlreadyExists, "Job UUID is already registerd.")
+		}
```
hougangliu: "StudyJob  with UUID %s registered already"

pkg/api/api.proto
```diff
@@ -26,6 +26,7 @@ service Manager {
     };
     /** 
      * Get a Study Config from DB by ID of Study.
+     * Return AlreadyExists error when the Job UID in GetStudyRequest is already exist in DB.
```
hougangliu: Wrong line

cmd/manager/main.go
```diff
@@ -32,10 +34,20 @@ func (s *server) CreateStudy(ctx context.Context, in *api_pb.CreateStudyRequest)
...
+	}
 	studyID, err := dbIf.CreateStudy(in.StudyConfig)
 	if err != nil {
 		return &api_pb.CreateStudyReply{}, err
 	}
+	err = dbIf.RegisterStudyJob(in.StudyConfig.JobId, studyID, in.StudyConfig.JobType)
+	if err != nil {
```
hougangliu: should we consider redo dbIf.CreateStudy?

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -249,14 +267,27 @@ func deleteStudy(instance *katibv1alpha1.StudyJob) error {
 	return nil
 }
 
+func getStudyfromDB(c katibapi.ManagerClient, studyId string) (*katibapi.StudyConfig, error) {
```
hougangliu: s/getStudyfromDB/getStudy
since createStudy/deleteStudy name style which has no "fromDB"

examples/manual-example.yaml
```diff
@@ -0,0 +1,73 @@
...
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: manual-example
+spec:
+  studyName: manual-example
+  owner: crd
+  optimizationtype: maximize
+  objectivevaluename: Validation-accuracy
+  optimizationgoal: 0.99
+  requestcount: 10
```
hougangliu: return 10 same trial?
maybe we should only consider ignore requestcount for manual suggestion?

hougangliu:
for trials table in db, it has study_id column. if we reused study id and trial id. when old studyjob deleted, all its trials deleted, too. the new studyjob with manual suggestion has deleted bestTrialID?

pkg/controller/studyjob/katib_api_util.go
YujiOshima: Sorry I should move it inside of `if err != nil {`.

pkg/controller/studyjob/katib_api_util.go
YujiOshima: I will fix.

examples/manual-example.yaml
YujiOshima: Yes, the requestcount will be ignored for the manual suggestion.
I will delete it from example and add doc.

cmd/manager/main.go
YujiOshima: No, we shouldn't. The Reconcile() will be called asynchronously. So often a request of CreateStudy is duplicate.
This is a guard for it.

YujiOshima:
@hougangliu 

> when old studyjob deleted, all its trials deleted
Even if we delete studyjob, its trials are not deleted automatically. They are stored in DB until deleted explicitly.
When we delete trials explicitly, we should delete the corresponding bestTrialID.
The new studyjob with a manual suggestion will not delete bestTrialID but it may update it.

cmd/manager/main.go
hougangliu: I see. Thanks!

examples/manual-example.yaml
YujiOshima: rewrite both of requestNumber and requestCount to 1 when suggestion algo is manual.
https://github.com/kubeflow/katib/pull/352/commits/608f3970994cb7b6ac376f39a6d22f6e12986e25#diff-7a381fdaeae1295724b74c0a04f7644cR38

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -22,41 +22,63 @@ import (
...
+	if instance.Spec.ReuseStudyID != "" {
+		_, err = getStudy(c, instance.Spec.ReuseStudyID)
+		if err != nil {
+			instance.Status.Condition = katibv1alpha1.ConditionFailed
+			return true, err
+		}
+		instance.Status.StudyID = instance.Spec.ReuseStudyID
+	} else {
+		studyConfig, err := populateStudyConf(instance)
+		instance.Status.Condition = katibv1alpha1.ConditionFailed
```
hougangliu: move line 62 to line 64

YujiOshima: Thanks, fixed.

hougangliu:
> @hougangliu
> 
> > when old studyjob deleted, all its trials deleted
> 
> Even if we delete studyjob, its trials are not deleted automatically. They are stored in DB until deleted explicitly.
> When we delete trials explicitly, we should delete the corresponding bestTrialID.
> The new studyjob with a manual suggestion will not delete bestTrialID but it may update it.

Since each trial has its own study_id in db, if now a trial may be owned by multiple studyjob, I am afraid it can cause potential confusion.
How about for reused trialid of manual suggest, just create a new trial with same ParameterSet as old trial?

YujiOshima:
My assumption, the studyjob is a temporary resource. The Study, Trial, and Worker(or its metrics) are stored in DB persistently.
Workers belong to a Trial. Trials belong to a Study.
In creating a new trial pattern when a user rerun the best Trial, the new job has a new trialID. It is probably confusing.
WDYT?

hougangliu:
> My assumption, the studyjob is a temporary resource. The Study, Trial, and Worker(or its metrics) are stored in DB persistently.
> Workers belong to a Trial. Trials belong to a Study.
> In creating a new trial pattern when a user rerun the best Trial, the new job has a new trialID. It is probably confusing.
> WDYT?
My understanding is different :)
I think that the Study, Trial, and Worker(or its metrics) stored in DB are affiliated with studyJob.
The information in DB are just hyperparameters of the studyjob's worker. the information in db is data like "--lr 0.01; --epoches 3", which can only serve the studyjob's worker. If I create a new studyjob with different worker, it may cannot know what "--lr" is, maybe it wants "--learningRate".
So If we cannot know a studyjob worker, all Study, Trial, and Worker(or its metrics) stored in DB are non-meaningful.
BTW, db information is deleted when corresponding studyJob deleted #315.

YujiOshima:
> So If we cannot know a studyjob worker, all Study, Trial, and Worker(or its metrics) stored in DB are non-meaningful.

Right. Then, how about making a new table for studyjob info? 
I don't want to manage persistent data in multiple places and I think etcd of k8s is not for persistent data.

> BTW, db information is deleted when corresponding studyJob deleted #315.

Sorry, my mistake.  For the above reason, I hope data will be stored in DB after CR deleted until delete API called explicitly.

hougangliu:
per talk with @YujiOshima, maybe we can store studyjob worker detail in db, too. then another studyjob with same worker can re-use trial/study/worker info in db.
(db information will never be deleted automatically when studyjob deleted)

we can add api to show specified workers detail with related trials and metrics, including objective metrics and value.
Thus, for a studyjob's workerSpec, we can use RawTemplate/TemplatePath, also workerID.
@gaocegege @richardsliu @johnugeorge WDYT?

YujiOshima:
/hold

This depends on #362.

PR #362, opened 2019-02-07
DB: add studyjob table and extend worker table

YujiOshima:
Related to https://github.com/kubeflow/katib/pull/352#issuecomment-459318668
I split changes related to DB scheme from #352 .
In this PR, information of studyjob and worker are stored in DB and users can refer them after the studyjob CR is deleted.
It makes easier to rerun the workers since the manifests of them are stored in DB. 

@hougangliu @gaocegege @richardsliu @johnugeorge

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/362)


johnugeorge:
Needs rebase to get gcloud fix

YujiOshima:
CI passed!
@johnugeorge @hougangliu Could you take a look?

pkg/controller/studyjob/studyjob_controller.go
```diff
@@ -226,11 +227,6 @@ func (r *ReconcileStudyJobController) Reconcile(request reconcile.Request) (reco
 		if !contains(pendingFinalizers, cleanDataFinalizer) {
 			return reconcile.Result{}, nil
 		}
-		err = deleteStudy(instance)
```
hougangliu: I think maybe we can add a field like "Retain" for studyjob to decide if deleteStudy when delete studyjob

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -35,22 +36,33 @@ func initializeStudy(instance *katibv1alpha1.StudyJob) error {
...
 		log.Printf("Connect katib manager error %v", err)
 		instance.Status.Condition = katibv1alpha1.ConditionFailed
-		return nil
+		return true, nil
 	}
 	defer conn.Close()
 	c := katibapi.NewManagerClient(conn)
+	_, err = c.GetStudyJob(context.Background(), &katibapi.GetStudyJobRequest{
+		StudyJobUid: string(instance.UID),
+	})
```
hougangliu: in which case instance.UID can be duplicated?

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -268,6 +277,52 @@ func createStudy(c katibapi.ManagerClient, studyConfig *katibapi.StudyConfig) (s
...
+				return err
+			}
+			mtl, err := sjc.GetMetricsCollectorTemplates()
+			if err != nil {
+				return err
+			}
+			if mt, ok := mtl[instance.Spec.MetricsCollectorSpec.GoTemplate.TemplatePath]; ok {
+				regStudyJobreq.StudyJob.MetricsCollectorTemplate = mt
+			}
+		}
```
hougangliu: please also handle `GoTemplate.TemplatePath == "" && GoTemplate.RawTemplate == "" ` case both for worker and metrics.

hougangliu:
tables studyjobs stores workertemplate/metricscollectortemplate info.
table study stores other info except  workertemplate/metricscollectortemplate.

do you plan to support below configuration?
```
spec:
   reusestudyid: STUDYID
   reusestudyjobid: UUID
```

YujiOshima:
@hougangliu Thank you for your review! 
I will add `reusestudyid`. But can I reuse studyjobid? The studyjobid is a UID of a studyjob CR. 
It is automatically filled by k8s.

pkg/db/interface_test.go
```diff
@@ -341,12 +353,35 @@ func TestGetWorkerList(t *testing.T) {
 }
 
 func TestUpdateWorker(t *testing.T) {
-	mock.ExpectExec(`UPDATE workers SET status = \? WHERE id = \?`).WithArgs(api.State_COMPLETED, defaultWorkerID).WillReturnResult(sqlmock.NewResult(1, 1))
-	err := dbInterface.UpdateWorker(defaultWorkerID, api.State_COMPLETED)
+	wmanufest := "woker manufest"
```
johnugeorge: typo: manufest -> manifest, woker ->worker

pkg/db/interface_test.go
```diff
@@ -341,12 +353,35 @@ func TestGetWorkerList(t *testing.T) {
...
+	mcmanufest := "metricscollector manufest"
```
johnugeorge: Same. Please correct the other lines too

hougangliu:
> @hougangliu Thank you for your review!
> I will add `reusestudyid`. But can I reuse studyjobid? The studyjobid is a UID of a studyjob CR.
> It is automatically filled by k8s.

Do you mean when a new studyjob uses reusestudyid, it will use configuration of row from studys tables and  worker/metrics of row from study jobs tables whose studyid == reusestudyid?
is there one record or multiple records in studyjobs table which studyid == reusestudyid? if there is many, if worker/metrics template keeps same or not?

YujiOshima:
@hougangliu Sorry for confusing.
Yes, I will add 
```
spec:
   reusestudyid: STUDYID
```
But I don't have plan to add `reusestudyjobid: UUID` since studyjobid cannot be reused.

>is there one record or multiple records in studyjobs table which studyid == reusestudyid? if there is many, if worker/metrics template keeps same or not?

Multiple studyjob for a studyid and worker/metrics template do not need to be the same for each studyjobs even they have the same studyid.

hougangliu:
> But I don't have plan to add `reusestudyjobid: UUID` since studyjobid cannot be reused.

studyjob has been stored in db, why we cannot reuse it? (we create a studyjob crd whose uid is xxx, a studyjob with studyjobid xxx also created in db, when we delete studyjob with uid xx, a studyjob with studyjobid xxx still stored in db.)

> Multiple studyjob for a studyid and worker/metrics template do not need to be the same for each studyjobs even they have the same studyid.

I think maybe you plan to use it as below, which can lead to multiple studyjob for a studyid, right?
if yes, we can only reuse a studyjob spec without worker/metrics spec. That is, even we store studyjob in db, we cannot reuse it.
```
spec:
   reusestudyid: STUDYID
   workerSpec:
     ...
   metricsCollectorSpec:
     ...
```

YujiOshima:
@hougangliu 
> studyjob has been stored in db, why we cannot reuse it?
Oh, I misunderstood. We can reuse a config for a studyjob but a UID will change in the new studyjob.

My assumption is below.
* Reuse StudyID: Run another config(algorithm, worker template) for an existing study. Only reuse studyID and studyconfig.
* Reuse StudyJobID: Rerun a studyjob with the same config (same algorithm, same worker template)

pkg/controller/studyjob/katib_api_util.go
YujiOshima: By CR operator implementation, the reconcile function will be called parallelly. So `initializeStudy` is often called two times. I don't know how to avoid it.

pkg/controller/studyjob/katib_api_util.go
YujiOshima: In that case, the `defaultWorkerTemplate.yaml` and `defaultMetricsCollectorTemplate.yaml` will be used. It's not an error.

hougangliu:
@YujiOshima I think the change is a little complex. To avoid misunderstanding, I think a detail example can make it clear.
saying: there are existing studyjob s1 and s2 (what it looks config and worker/metrics, simple one is ok). when create a new studyjob s3 with reusestudy/reusestudyjob, what s3 will look like.
Thanks!

YujiOshima:
@hougangliu 

You have a original studyjob yaml.
https://gist.github.com/YujiOshima/ac143f0d334759e8a812136d607d6cad#original-job-yaml
and you can create original CR `kubectl create -f original-job.yaml`
https://gist.github.com/YujiOshima/ac143f0d334759e8a812136d607d6cad#original-cr

When you reuse StudyID, you can omit StudyConf but need to write other info e.g. WorkerSpec.
https://gist.github.com/YujiOshima/ac143f0d334759e8a812136d607d6cad#reuse-studyid-yaml
You can reused StudyConf like this.
https://gist.github.com/YujiOshima/ac143f0d334759e8a812136d607d6cad#reuse-studyid-cr

When you reuse StudyJobID, you can omit almost all field.
https://gist.github.com/YujiOshima/ac143f0d334759e8a812136d607d6cad#reuse-studyjobid-yaml
You will get job copy of the original.
https://gist.github.com/YujiOshima/ac143f0d334759e8a812136d607d6cad#reuse-studyjobid-cr

hougangliu:
@YujiOshima it makes sense.
Then I think there is only one studyjob for a studyid (one to one). Otherwise, for below case, it's hard to decide selecting one of studyjob to fill workerSpec automatically.

```
apiVersion: "kubeflow.org/v1alpha1"
kind: StudyJob
metadata:
  namespace: kubeflow
  labels:
    controller-tools.k8s.io: "1.0"
  name: random-example-reusejob
spec:
  reuseStudyJobId: f4a4b946-3b33-11e9-adcb-08002750630b
```

YujiOshima:
> Then I think there is only one studyjob for a studyid (one to one).

No, multi studyjobs can use a studyID and the studyjobs share historical data of Trial and Worker belong to the studyID.
The use cases are

* You can get initial values for bayesian opt, running random or manual studyjob. After that, run bayesian opt job reusing the StudyID.
* You can retry or try another algorithm after a job is completed and see all results on the same graph in UI.
* You can add a trial manually during job running since the algorithm doesn't search near the trial.

> Otherwise, for below case, it's hard to decide selecting one of studyjob to fill workerSpec automatically.

A studyjob has only one WorkerSpec. So it can fill with the WorkerSpec.

gaocegege:
/close 

We have dropped v1alpha1. Thus I think we cannot merge it into master.

PR #364, opened 2019-02-08
Use log.Printf instead of log.Print

toshiiw:
In order to supress "Print call has possible formatting directive" warnings.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/364)


johnugeorge:
Thanks @toshiiw.  Merged these changes in #363  as they were required for presubmits.

toshiiw:
@johnugeorge thanks

Issue #349, opened 2019-01-29
support worker/metricsCollector template from specified configmap + path


hougangliu:
For now, when specify worker and  metricsCollector by TemplatePath, this path must exist in kubeflow/configmaps/worker-template for worker and kubeflow/configmaps/metricscollector-template for metrics collector.

we should cover a use case: a user who can only create/update studyjob and configmap in  non-kubeflow namespace , but he wants to persist his worker/metricscollector template in configmap fo re-use (even more, he wants to persist worker/metricscollector of a studyjob in one configmap for better management, saying mnistStudyJobConfigmap, yoloStudyJobConfigmap and so on)

hougangliu:
I plan refactor GoTemplate as below
```
type TemplatePath struct {
	ConfigMapName string `json:"configmapName,omitempty"`
	ConfigMapNamespace string `json:"configmapNamespace,omitempty"`
	TemplatePath string `json:"templatePath,omitempty"`
}

type GoTemplate struct {
	TemplateFromConfigmap TemplatePath `json:"templateFromConfigmap,omitempty"`
	RawTemplate  string `json:"rawTemplate,omitempty"`
}
```
@richardsliu @YujiOshima @gaocegege @johnugeorge  any comments?

richardsliu:
Maybe call the field and type TemplateSpec instead of TemplateFromConfigmap/TemplatePath? Otherwise sounds ok to me.

johnugeorge:
Closing this issue as fields are included in v1alpha2 API
/close

Issue #370, opened 2019-02-12
StudyJob v1alpha2 API version


johnugeorge:
There are couple of changes that are suggested. We can incorporate all these changes in the next API version.

 - [x]  Consistent field naming #351 , #371 
 - [x]  Extending Status fields to contain all conditions #344 
 - [x]  Extending Status fields to have detailed info #356 #122 
 - [x]  Renaming RequestNumber, RequestCount #161 
 - [x] Custom TemplateSpec for workers/metrics #349 
 - [x] StudyName field is unnecessary #380

Please add if I missed out any.

/cc @richardsliu @YujiOshima @hougangliu 

----------------

Implementation (last updated 5/13/2019):
- [x] Controllers implementation #441
- [x] Metrics collection #442
- [x] New UI  #443
- [x] Suggestion algorithms implementation #407 
- [x] DB operations https://github.com/kubeflow/katib/issues/510
- [x] Support custom resources https://github.com/kubeflow/katib/pull/512
- [x] Update kustomize packages #513 
- [x] Add e2e tests #514 
- [x] #551

hougangliu:
@johnugeorge  please also add #349

johnugeorge:
@hougangliu can you create an issue ?

hougangliu:
@johnugeorge Done #371

johnugeorge:
It would be better if  all other current bigger PRs are reviewed and merged before moving to v1alpha2.  Else, PRs need be modified for newer version too.  

PRs which should be reviewed and merged:

@YujiOshima  #362 #352 
@DeeperMind #339 #345 
@Akado2009  #328

richardsliu:
How about #346 ? It might require some API changes too.

#341 is not API related but would require controller side changes, and would be good to add as well.

YujiOshima:
https://github.com/kubeflow/katib/issues/291 is also a big change for the logic of controller.
It would be good to add.

johnugeorge:
@YujiOshima @richardsliu i was listing out issues that need API changes. Except  #346(not finalized yet), other mentioned issues are already tagged with 0.5.0 which can be tracked separately.  Other than issues included in https://github.com/kubeflow/katib/issues/370#issuecomment-462779691,  we will get all new controller changes in v1alpha2.

YujiOshima:
@johnugeorge OK, Thanks.


> It would be better if all other current bigger PRs are reviewed and merged before moving to v1alpha2. Else, PRs need be modified for newer version too.

[...]

PR #375, opened 2019-02-14
WIP: support worker and metricsCollector template spec

hougangliu:
Fixes:  #349

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/375)

### DIFF
--- a/pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go
+++ b/pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go
@@ -114,19 +114,25 @@ const (
 	OptimizationTypeMaximize OptimizationType = "maximize"
 )
 
-type GoTemplate struct {
+type TemplateSpec struct {
+	ConfigMapName string `json:"configMapName,omitempty"`
+	ConfigMapNamespace string `json:"configMapNamespace,omitempty"`
 	TemplatePath string `json:"templatePath,omitempty"`
+}
+
+type GoTemplate struct {
+	TemplateSpec *TemplateSpec `json:"templateSpec,omitempty"`
 	RawTemplate  string `json:"rawTemplate,omitempty"`
 }
 
 type WorkerSpec struct {
 	Retain     bool       `json:"retain,omitempty"`
-	GoTemplate GoTemplate `json:"goTemplate,omitempty"`
+	GoTemplate *GoTemplate `json:"goTemplate,omitempty"`
 }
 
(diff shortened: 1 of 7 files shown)


pkg/controller/studyjob/manifest_parser.go
```diff
@@ -28,30 +27,50 @@ import (
...
-		if err != nil {
-			return "", nil, err
-		}
-		if wt, ok := wtl[wPath]; !ok {
-			return "", nil, fmt.Errorf("No worker template name %s", wPath)
-		} else {
-			wtp, err = template.New("Worker").Parse(wt)
+		tName, tNamespace, tPath := getDefaultTemplateSpec()
+		if goTemplate.TemplateSpec != nil {
+			tName = goTemplate.TemplateSpec.ConfigMapName
```
YujiOshima: When the field of TemplateSpec is emply, we should use the default value.
For example, we can omit the namespace of ConfigMap when we want to use the same namespace as the studyjobcontroller.

pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go
```diff
@@ -114,19 +114,25 @@ const (
 	OptimizationTypeMaximize OptimizationType = "maximize"
 )
 
-type GoTemplate struct {
+type TemplateSpec struct {
```
YujiOshima: This extension of studyjob API would be for v1alpha2? https://github.com/kubeflow/katib/issues/370

YujiOshima:
This issue is listed v1alpha2 API https://github.com/kubeflow/katib/issues/370 .
@johnugeorge Should we make a branch for v1alpha2 and this will be merged to the branch?

johnugeorge:
Yes. I would suggest that only common changes(that won't break the current config) go into v1alpha1 at this point. @richardsliu is working to get v1alpha2 structure. Once it is merged, we can get it in v1alpha2

In general,  I feel that we are moving away from the data scientist point of view, instead thinking from a kubernetes developer perspective.   Currently, a data scientist will find it extremely difficult to write a katib study job config correctly without good kubernetes experience. For eg: Currently in the studyjob config,  there is a namespace for overall studyjob, a namespace for the workerspec, namespace for ConfigMap, namespace for metric collector . Won't this be complicated for the users by exposing much details? WDYT?

hougangliu:
will submit another PR for v1alpha2

Issue #380, opened 2019-02-15
StudyName field is  unnecessary


hougangliu:
When create a studyJob, we need specify its name in metadata.name. But we have studyName filed in studyJob.spec, too. I think .spec.StudyName is unnecessary. We can remove it.

```
type StudyJobSpec struct {
	StudyName            string                `json:"studyName,omitempty"`
       ...
}
```

johnugeorge:
@YujiOshima Was there some discussion about this before?

YujiOshima:
@johnugeorge @hougangliu We have discussed this here https://github.com/kubeflow/katib/pull/152#discussion_r212744338 .
In this PR https://github.com/kubeflow/katib/pull/352 I try to add `reusestudyid` field.
From this PR, we will use `StudyName` only when creating study and use `reusestudyid` in a studyjob for an existing study.
If we merge metadata.name and .spec.StudyName, I worry about a situation below.
We create a studyjob with `reusestudyid`. It will use an existing study.
Then the metadata.name of the studyjob and the name of study corresponding to the studyjob will be different. 
It may be confusing.

hougangliu:
@YujiOshima As you say, when we create a studyjob with `reusestudyid`. It will use an existing study. If we don't remove  spec.StudyName. the metadata.name of the studyjob and the name of study corresponding to the studyjob will be different (this is really confusing).

for now, I have no idea what we can use spec.StudyName for.

johnugeorge:
i agree with @hougangliu. as in the comment, do we really need to support "You can create multiple StudyJob for the same Study."

YujiOshima:
@johnugeorge I think it is needed. 

The use cases are
* To get initial values for bayesian opt, running random or manual studyjob. After that, run bayesian opt job reusing the StudyID.
* Want to retry or try another algorithm after a job is completed. And need to see all results on the same graph in UI
* Add a trail manually during job running since the algorithm doesn't search near the trial.

How about moving `.spec.StudyName` to `.status.StudyName`?
And it will be set by studyjob operator. When creating a study, it will be studyjobName. When reusing a study, it will be the name of the corresponding study.
Thought?

johnugeorge:
> i agree with @hougangliu. as in the comment, do we really need to support "You can create multiple StudyJob for the same Study."

I see.  I didn't realize that you meant reuseStudy by that comment.

hougangliu:
@YujiOshima 
[...]

Issue #356, opened 2019-01-31
Katib status should return optimal parameters values


richardsliu:
Enhance Katib StudyJob status to have more info, e.g.
- Successful/Pending/Failed trials
- best hyper parameters till now
- etc.

Currently, output of “kubectl describe studyjob” gives very limited view of status. Looking at the output status, we can't really tell what is going on with the algorithm right now, how many iterations have passed. It would be interesting to see how far are we  from the "Optimization goal".

jlewi:
Is this on track to land in 0.5 or should move it to 0.6?

richardsliu:
We can close this one since the design for v1alpha2 API covers this. Implementation will be completed in 0.6.

/close

Issue #343, opened 2019-01-24
Katib job and Katib services cannot be in different namespaces due to metrics collector service account


hougangliu:
just copy https://github.com/kubeflow/kubeflow/issues/2311 here to trace

jlewi:
@johnugeorge @YujiOshima @richardsliu Are we planning on fixing this as part of 0.5 or should we move it to 0.6?

Is this P1 or P2?

richardsliu:
Moved out of 0.5.

gaocegege:
I think we still cannot solve it in 0.6

johnugeorge:
Yes. There is no cleaner way to solve it using current metric collector design

gaocegege:
Yeah, then we could put the issue in 0.7

richardsliu:
Moved to 0.7.0.

johnugeorge:
This issue is resolved with new metric collector implementation in v1alpha3

johnugeorge:
Closing this issue as https://github.com/kubeflow/katib/pull/738 is merged.

Proposal: https://github.com/kubeflow/katib/pull/732

Issue #341, opened 2019-01-23
Make Katib generic for operator support


johnugeorge:
Currently in order to support custom K8s to Katib,  explicit support has to be added.  There are explicit checks for each resource in the current code.  Remove such checks so that it is generic to support new operators. 

eg:  In getJobWorkerStatus,  https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L421

Related PRs:
- [x]  StudyJob controller shouldn't crash if operators are not installed #317

richardsliu:
StudyJobController currently implements different logic for each worker type to check if the worker has completed: https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L413

BatchJob also supports the "last condition" format, so in theory we should be able to make this code generic: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/batch/v1/generated.proto#L158

richardsliu:
I looked a bit into this, seems like we are dependent on other operator binaries in a few ways:

1. Watch on resources. Currently we are doing:
```
	err = c.Watch(
		&source.Kind{Type: &tfjobv1beta1.TFJob{}},
		&handler.EnqueueRequestForOwner{
			IsController: true,
			OwnerType:    &katibv1alpha1.StudyJob{},
		})
	if isFatalWatchError(err, TFJobWorker) {
		return err
	}
```
Instead, can we do something like:
```
	u := &unstructured.Unstructured{}
	u.SetGroupVersionKind(schema.GroupVersionKind{
		Kind:    "TFJob",
		Group:   "",
		Version: "v1beta1",
	})
	err = c.Watch(
		&source.Kind{Type: u},
		&handler.EnqueueRequestForOwner{
			IsController: true,
			OwnerType:    &katibv1alpha1.StudyJob{},
		})
	if err != nil {
		return err
	}
```

2. Get/Delete resources at runtime, e.g. https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L313

Seems like these calls just need a pointer to a runtime.Object, and do not need any specific linking to operators.

3. GetStatus, e.g. https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L417

We just need to define an interface with Status.Conditions, and it should just work across operators. The contract is that each resource needs to support Status.Conditions.

[...]

PR #381, opened 2019-02-15
Katib v1alpha2 API for CRDs

richardsliu:
@YujiOshima @gaocegege @johnugeorge @alexandraj777 @hougangliu @xyhuang 

This is an initial proposal for the Katib v1alpha2 API. The changes here reflect the discussion in https://github.com/kubeflow/katib/issues/370. 

Comments and suggestions are welcome. 

Please note that the NAS APIs are not included here since the feature is still in early phase.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/381)


pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
```diff
@@ -0,0 +1,202 @@
...
+	//runtime "k8s.io/apimachinery/pkg/runtime"
+
+	pb "github.com/kubeflow/katib/pkg/api"
+)
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// StudyJobSpec defines the desired state of StudyJob
+type StudyJobSpec struct {
+	StudyName            string                `json:"studyName,omitempty"`
```
hougangliu: I think StudyName is unnecessary #380

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
```diff
@@ -0,0 +1,202 @@
...
+	// be set in happens-before order across separate operations.
+	// It is represented in RFC3339 form and is in UTC.
+	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
+
+	// Represents last time when the StudyJob was reconciled. It is not guaranteed to
+	// be set in happens-before order across separate operations.
+	// It is represented in RFC3339 form and is in UTC.
+	LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`
+
+	Conditions               JobCondition  `json:"conditions,omitempty"`
```
hougangliu: [] JobCondition

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
```diff
@@ -0,0 +1,202 @@
...
+	BestTrialID              string     `json:"bestTrialId,omitempty"`
+	BestWorkerID             string     `json:"bestWorkerId,omitempty"`
+	BestParameterValues      []ParameterValue    `json:"bestParameterValues,omitempty"`
+	CurrentSuggestionCount   int        `json:"currentSuggestionCount,omitempty"`
+}
+
+type WorkerCondition struct {
+	WorkerID       string      `json:"workerId,omitempty"`
+	Kind           string      `json:"kind,omitempty"`
+	Condition      Condition   `json:"condition,omitempty"`
```
hougangliu: s/Condition      Condition    /Condition      JobConditionType

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
```diff
@@ -0,0 +1,202 @@
...
+type OptimizationType string
+
+const (
+	OptimizationTypeUnknown  OptimizationType = ""
+	OptimizationTypeMinimize OptimizationType = "minimize"
+	OptimizationTypeMaximize OptimizationType = "maximize"
+)
+
+type GoTemplate struct {
+	TemplatePath string `json:"templatePath,omitempty"`
```
hougangliu: s/TemplatePath/TemplateSpec
refer to #375

hougangliu:
@richardsliu Thanks! some comment

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
YujiOshima: A condition for a worker does not equal job condition.
https://github.com/kubeflow/katib/blob/master/pkg/api/api.proto#L283
It has `Killed` condition that means the worker is killed by early stopping service.

johnugeorge: Condition not defined?

hougangliu: you're right!
So I think `type Condition string` missed

xyhuang:
@richardsliu are you planning to have a separate PR for nas?

johnugeorge:
Shall we have this PR to have a complete v1alpha2 API so that it will be easier to review/implement

API modifications that are not included:

1. Do we need any more extra status fields wrt to #356  ? 
2. #352 proposes ReuseStudyID  @YujiOshima 
3. Nas specific fields are missing @andreyvelich  
4. #349  proposes configMap related fields @hougangliu

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
gaocegege: Maybe we could name it as WorkerCondition.

jdplatt: What worker conditions are allowed right now? When tracking failed trails I think it is important to distinguish between trails that failed for reasons independent of the suggested parameters (e.g. k8s killed the pod due to resource constraints on a node) and trails that failed due to the suggested parameters (e.g. learning rate was too high and loss blew up). The first type of failures are worth retrying while the second type should be avoided in future.

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+type WorkerCondition struct {
+	WorkerID       string      `json:"workerId,omitempty"`
+	Kind           string      `json:"kind,omitempty"`
+	Condition      Condition   `json:"condition,omitempty"`
+	ObjectiveValue *float64    `json:"objectiveValue,omitempty"`
+	StartTime      metav1.Time `json:"startTime,omitempty"`
+	CompletionTime metav1.Time `json:"completionTime,omitempty"`
+}
```
jdplatt: How do you feel about allowing the option to add metadata to a trial? For example, when a trials fail we'll often add a message containing the error so we can figure out why that parameter combination didn't work (e.g. if it's an GPU OOM error then our model is too large and we might reduce the batch size or the number of neurons, if it's a loss=NaN error we reduce the maximum learning rate).

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
```diff
@@ -0,0 +1,202 @@
...
+
+// StudyJobSpec defines the desired state of StudyJob
+type StudyJobSpec struct {
+	StudyName            string                `json:"studyName,omitempty"`
+	Owner                string                `json:"owner,omitempty"`
+	OptimizationType     OptimizationType      `json:"optimizationType,omitempty"`
+	OptimizationGoal     *float64              `json:"optimizationGoal,omitempty"`
+	ObjectiveValueName   string                `json:"objectiveValueName,omitempty"`
+	MaxSuggestionCount   int                   `json:"maxSuggestionCount,omitempty"`
+	MetricsNames         []string              `json:"metricsNames,omitempty"`
```
jdplatt: Based on our experience using sigopt I think the user should be able to set an ObservationBudget, which sets the number of trails they expect to run, independent of the MaxSuggestionCount, which sets the maximum number of trials the experiment runs before stopping. There are two advantages to this:

1. If the HPO still hasn't found a good model when it hits the ObservationBudget often we will leave a study running until it hits a reasonable value. It would be nice to allow studies to just keep running in this scenario without us having to restart them manually.

2. We often change the ObservationBudget to influence the explore vs. exploit trade-off in the HPO algorithm. If the algorithm thinks it has fewer observations left to find a good solution it will be more aggressive exploring hyperparameter space. The normal use case for this is when we are happy to trade a little long-term accuracy for finding a decent model quickly.

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
richardsliu: Removed

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
richardsliu: The name of this struct is somewhat confusing. A "Condition" should be something like Succeeded, Failed, etc. This is more like "WorkerMetadata".

We should have a separate WorkerCondition type, that can take enum values. We can also support conditions like retryable failures.

richardsliu: I've added a separate WorkerCondition struct with its own condition types, and a WorkerMetadata struct. Let me know if this is confusing or too verbose.

richardsliu:
Thanks for the initial comments. I've made a few changes:

- Incorporated pending changes from #352 and #375 
- Added a WorkerMetadata struct with a list of WorkerConditions. The semantics are similar to JobConditions, but workers can have different status values. 

With regards to NAS (@xyhuang ), I would like to keep the API design separate from this effort. Whenever we finalize this API, I will add the NAS stuff to the schema, with the understanding that NAS is an alpha feature and can change in non-compatible ways.

Please continue to provide feedback. For example:
- What to do with metric collectors? Is this something we should remove and replace with a metadata store API?
- #349, #356 
- General naming conventions. How do people feel about names like "Study", "Trial", "Worker" etc in general? Are there better/more well known terms that we can use here?

Let's aim to have the API structure stabilized by mid-March.

jdplatt:
I'm still a little confused about the relationship between a Trial and a Worker. Am I right to assume that like in the vizier paper a Trial is meant to be an evaluation of a single suggestion and a worker is meant to be the process that runs a Trial? In that case I am unsure:

1. Why there can be multiple workers attached to a single trial?
2. Why the objective value is inside the WorkerMetadata struct not the Trial struct? If the goal of a Trial is to calculate the objective value for a given suggestion then it might be worth putting the resulting value in the Trial rather than the worker than ran it.
3. What does the Kind field in WorkerMetadata mean?

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
```diff
@@ -0,0 +1,240 @@
...
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// StudyJobSpec defines the desired state of StudyJob
+type StudyJobSpec struct {
+	Owner                string                `json:"owner,omitempty"`
+	OptimizationType     OptimizationType      `json:"optimizationType,omitempty"`
+	OptimizationGoal     *float64              `json:"optimizationGoal,omitempty"`
+	ObjectiveValueName   string                `json:"objectiveValueName,omitempty"`
+	MaxSuggestionCount   int                   `json:"maxSuggestionCount,omitempty"`
```
cvenets: Would it make sense to somehow group `OptimizationType`, `OptimizationGoal`, `ObjectiveValueName`, and maybe `MaxSuggestionCount` and or `ObservationBudget` proposed by @jdplatt ? Could we have an Optimization/Objective Spec, so that the generic categories a user needs to define are clear and separated in the `StudyJobSpec`? Same way we have a `SuggestionSpec` for example.

Also, following [this converstation](https://kubeflow.slack.com/archives/CE0BURK1B/p1549135874084100?thread_ts=1548816849.064800&cid=CE0BURK1B) does it make sense to rename:

RequestNumber -> ParallelWorkersNum
RequestCount -> MaxIterations (rather than MaxSuggestionCount)

I think that `MaxSuggestionCount` can be confused with the max number of the aggregated suggestions one wants, rather than the actual iterations we want to do. What do you think?

cvenets:
I agree with @jdplatt's comments above. Also, if a Trial is actually a single run with specific parameters, why don't we can it a 'Run' rather than a 'Trial'? Wouldn't this be easier for users to get what it is?

jdplatt:
Based on the discussion in #386 I think we also need to add the ability to track what order the trials were run in. We could this by adding index field to Trail indicating what order the trials were run in, or by inferring the order later on using the started/completed times. Tracking the order the trials were generated in will make it possible to monitor a running study to see how the metric is improving over time. We look at these sorts of plots a lot to decide when to end a study (for example if a new best model hasn't been found in a while)

pkg/api/operators/apis/studyjob/v1alpha2/studyjob_types.go
richardsliu: This is a great suggestion.

richardsliu:
@jdplatt Your understanding is correct. A "Trial" is a vizier concept mapping to one evaluation of a suggestion; whereas a "Worker" is a k8s resource that runs the Trial.
 
1. According to https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L558, each Trial should correspond to exactly one Worker. @YujiOshima can you confirm this?

2. This is probably because the objective value is calculated by each Worker and fetched by the metrics collector (which maps to workers and not trials). Perhaps we can combine/restructure this.

3. Kind refers to the type of worker - "Job" for generic k8s BatchJob, "TFJob" and "PytorchJob" for kubeflow distributed training jobs. We are also looking into making this more generic, see https://github.com/kubeflow/katib/issues/341.

hougangliu:
> 1. According to https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L558, each Trial should correspond to exactly one Worker. @YujiOshima can you confirm this?
#352 aims to reuse trial so that multiple workers may belongs to one trial.

BTW, now when a worker failed, we mark studyjob failed, too. But for some case, if we re-create another worker, everything may work well (for example, the worker is killed by eviction policy or node error, if the worker is scheduled to other node, everything can work). Maybe for a trial, when a worker failed, we can re-create another one until max-fail-time to mark  studyjob `Failed`. So in this case, multiple workers (max-fail-time at most) can belong to a trial

jlewi:
One alternative to **Study** would be to call it an **Experiment**.

The term **Experiment** appears in other places; I think KF pipelines and MLFlow both use it. I don't know if that is a good thing or bad thing.

jdplatt:
@richardsliu I think some of the confusion over Trials vs Workers is because Katib uses workers differently than Vizier. Below is a figure from the Vizier paper showing the logic inside a worker.

vizierworker

Each worker runs for the length of a Study and ends up running many trials. However, in Katib a worker run only for the life of a single trial. 

1. Is Katib meant to be an open source version of Vizier or do you foresee the project going in it's own direction over time?
2. If we stick with the current approach in Katib where workers and trials have the same lifecycle can we just collapse them into a single object?

jdplatt:
@jlewi Sigopt uses the term Experiment instead of Study as well.

jdplatt:
Another feature I think we should add to the new API is the ability to scale the feasible parameter space (e.g. #224). It is really common to use log scaling on parameters such as learning rate or regularization coefficient

hougangliu:
> Each worker runs for the length of a Study and ends up running many trials. However, in Katib a worker run only for the life of a single trial.

I think katib keeps consistent with vizier, you can take RunTrial(trial) as a worker lifecycle in katib.

richardsliu:
@jdplatt In response to your questions:

1. I see katib as a project inspired by Vizier, but not necessarily the "open source version of Vizier". My long-term vision is for it to evolve into an open source HP-tuning/NAS/AutoML service that integrates well with Kubernetes. @YujiOshima what do you think?

2. I agree with @hougangliu suggestion - we can collapse the vocabulary and just use "Trial/RunTrial".

johnugeorge:
> @jdplatt In response to your questions:
> 
> 1. I see katib as a project inspired by Vizier, but not necessarily the "open source version of Vizier". My long-term vision is for it to evolve into an open source HP-tuning/NAS/AutoML service that integrates well with Kubernetes. @YujiOshima what do you think?
> 2. I agree with @hougangliu suggestion - we can collapse the vocabulary and just use "Trial/RunTrial".

I agree. Though the initial design is inspired from Vizier, evolving it into a k8s native solution with best user experience is more important in the longer run.

YujiOshima:
@jdplatt @richardsliu I'm sorry for late reply.
>1. I see katib as a project inspired by Vizier, but not necessarily the "open source version of Vizier". My long-term vision is for it to evolve into an open source HP-tuning/NAS/AutoML service that integrates well with Kubernetes. @YujiOshima what do you think?

I agree. We do not need to stick to the design of Vizier.
In my first design, the Trial is only a set of parameter.  And the Worker is an evaluating process for a trial. So if you need to evaluate multiple time for one parameter, you can make multiple Worker for a Trial.
Though I think we can use another name instead of Worker, we should need a concept for an evaluating process that independent from Trial.
Because the objective value and metrics may change every time even with the same trial.

For avoiding confusion, how about use consistent names for resources and processes?
Currently:

| Resouce | Process |
----|---- 
| Study | StudyJob |
| Trial | Worker |

Suggestion:

| Resouce | Process |
----|---- 
| Study(Experiment?) | StudyRun(ExperimentRun?) |
| Trial | TrialRun |

alexandraj777:
Re: offline conversation with Richard, I was hoping to comment a bit on the naming topic. I put together this [blog post on how we named our API objects](https://sigopt.com/blog/experiment-suggestion-observation-how-we-named-our-hpo-api-objects/) to give some perspective on the names we chose, and how those fit into our API. I hope this is a useful perspective!

|**Name**|**Definition**|
|--|--|
|Metric|A numeric measure of some trait you care about|
|Experiment|An encapsulation of the process of optimizing one or more Metrics|
|Parameters|A range of possible inputs that you believe effect the Metric|
|Assignments|One specific input|
|Value|One result of evaluating your Metric|
|Observation|The logical encapsulation of Assignments and the resulting Value of each Metric on those assignments|
|Suggestion|The logical encapsulation of a set of Assignments to be evaluated by the user|

cvenets:
@alexandraj777 thank you very much for documenting this.

As commented [here](https://kubeflow.slack.com/archives/CE0BURK1B/p1550703809017300), I'm copying/pasting as a comment to have all the information in one place.

Trying to do a mapping between SigOpt and Katib, my understanding is the following. Please comment/fix:

| SigOpt | Katib |
| --- | --- |
| Metric | Metric |
| Experiment | Study |
| Parameter | Parameter |
| Assignment | Parameter Value |
| Value | Metric Value |
| Observation | Trial |
| Suggestion | Suggestion |
| Optimization Loop | _N/A_ |
| _N/A_ | Objective |

Katib's *Objective*
The desired state/value of a metric(s) that we are trying to reach. It includes `OptimizationType`, `OptimizationGoal`, `ObjectiveValueName`, and maybe `MaxIterations`, `ObservationBudget`.

I think that `Study` should definitely be renamed, since it is Vizier-specific. `Experiment` is a good name, but I think it may be misleading in the context of Kubeflow. In Kubeflow Pipelines terminology an `Experiment` is much broader. It includes multiple runs of even different pipelines, where only one step of the pipeline may be the hyperparameter optimization. This is why I proposed *Optimization* in a previous comment @jlewi @jdplatt.

richardsliu:
@cvenets I prefer the name `Experiment` over `Optimization`. 
1. It is more consistent with existing terminologies as @alexandraj777 mentioned;
2. `Optimization` can be conflated with other katib field names, specifically `OptimzationType` and `OptimizationGoal`;
3. `Experiment` implies that we are testing some hypothesis (in this case a `Suggestion`), and that it is not necessarily conclusive - we can start/resume an experiment or reconfigure it with different parameters. 

Agree with the point about Objective. It should be refactored as its own struct type.

Within the context of katib, the term `Trial` is actually more appropriate since it encapsulates the 3 steps involved in each optimization loop:
1. Get a suggestion;
2. Evaluate the suggestion;
3. Record observations (results).

However as @YujiOshima mentioned, we need to differentiate between a Trial and a running instance of a Trial (currently called 'Worker'). A few options:
- TrialRunner
- TrialEvaluator
- TrialJob
- ???

Hopefully, explicitly prefixing the term with 'Trial' would help identify that the two concepts are closely related.

YujiOshima:
@cvenets  @richardsliu I vote `Experiment`. `Experiment` is already used broadly in similar projects.
Optimization will be also Katib-specific word.

>Hopefully, explicitly prefixing the term with 'Trial' would help identify that the two concepts are closely related.

Agree. I prefer `TrialEvaluator` since it is responsible for an evaluation part of an optimization loop.

cvenets:
@YujiOshima, @richardsliu:

I agree 100% that **Experiment** is the best term if one looks only at Katib, or other similar projects. And I don't have a strong opinion against it either. My only reservation is that, since I'm trying to look at it in a holistic way and put it in the broader Kubeflow context, I think that this will have a clash later on with the Experiment term that other projects that integrate with Kubeflow have.

For example both ModelDB and Kubeflow Pipelines already have the notion of an Experiment. ModelDB's is closer to Katib's and maybe we will be able to do a one-to-one mapping when tracking. In the Pipelines case this is definitely not the case, and we may end up with the following situation:

One creates a Kubeflow Pipelines Experiment. Inside the Experiment they create a Pipelines Run, which executes a few Pipelines Steps. One of the Steps is Tuning, so the Step will create an Experiment again, which will be a Katib Experiment this time. If this is fine, or if we can find a better term to replace the Pipelines Experiment in the future, we can move forward with **Experiment** for Katib. What do you think @jlewi, regarding cross-project use?

Regarding *Trial*, I also agree that it is a good name. I proposed previously to rename it to *Run* (or *TrialRun*) because it wasn't clear to me that somebody could run multiple instances of the same Trial with the same parameter values from a single suggestion, and end up with different results, as @YujiOshima described. If this is the case (how/when does that happen?), then I think it is easier and simpler for the end user to have Trial and TrialRun. I also vote for both to be prefixed with 'Trial' to denote that they are related.

We may also want to consider introducing the observation/result entity, which seems to make a lot of sense, even when defining what a Trial is.

johnugeorge:
Thanks for detailed description. 

From @richardsliu earlier [comment](https://github.com/kubeflow/katib/pull/381#issuecomment-464233002), I feel that we should reconsider metric collector design.  we need a push mechanism(metadata store API) to make design simpler and to report metrics accurately.  @YujiOshima 

Extending from @gaocegege 's [comment](https://github.com/kubeflow/katib/issues/346#issuecomment-463911655) , how about creating a separate 'Trial' crd which contains a `suggestion` and `observation`?  My initial thought would be, StudyJob Controller runs the 'optimization loop' and creates Trials. Trial Controller executes and records observations.

Please post your thoughts

YujiOshima:
@johnugeorge 
I'm not sure about `push mechanism` mean. Do you mean a training container call API and push metrics to Katib?

Since the Trial is one of the core resources of Katib, it should be stored in DB.
Introducing Trial Controller and delegating management of Trial probably make the optimization loop more robust.
But all resources (Trial, observation) should be stored not only CR but DB.

johnugeorge:
@YujiOshima 
 
Yes. Currently we have a pull based mechanism where metric collectors are spawned up at regular intervals,  collects metrics and writes to DB via API. What are your thoughts on push vs pull based metric collector design? 

I have seen a few problems wrt the current metric collector design. 

1. What happens if worker pods are deleted/completed before metric collector pods are created?  Pods can get deleted because of clean pod policies set for the jobs.  Related discussion: https://github.com/kubeflow/tf-operator/issues/722#issuecomment-405669269

2. https://github.com/kubeflow/katib/issues/343 - Issue happens when Katib services and Katib Job are run in different namespaces

3. How can user ensure/debug if all metrics are recorded accurately? 
 
We can discuss this in a separate issue too so as not to move away from the main focus.

jdplatt:
Alongside the issues @johnugeorge raised I think there are two other advantages of having a push mechanism for reporting observations:

1. It is easier to have Workers that evaluate multiple suggestions one after the other (like in vizier or a sigopt optimization loop with multiple iterations), rather than having each worker exit after it has finished evaluating a trial. To do this with the current metrics collector setup we would have to add extra parsing to the logs and figure out when one trial ended and a new one began.

2. It might allow people to run the optimization loop outside of k8s. They would hit the API to generate a suggestion, evaluate it however they like (either in k8s or not), and then report the resulting observation to the API. This would turn katib from a project that tunes HPs only for people running in k8s into a project that runs in k8s and tunes HPs for people running anywhere. I can see this being very useful for organizations that are running a mixture of k8s + non-k8s systems, but want a single HPO API that works for all of their use cases. We would just need a way to control what experiments that run workers inside k8s based on docker images and those that run their own workers and just use the suggestion services in katib.

richardsliu:
Thanks for all the great feedback and contributions so far. For the purpose of this discussion, I believe we should focus on finalizing the terminologies and API structure. Metrics collection is an important aspect of the overall design and will be discussed in a separate thread.

To summarize, here's my current proposal for HP tuning terminologies:

## Core Concepts

These are the terms that we expect data scientists to use when they use Katib, either from notebooks or from UI:

**Experiment**: Encapsulates the process of a HP-tuning optimization. In Katib v1alpha1, an **Experiment** is known as a *Study*.  An **Experiment** [1] is composed of:
- **Parameters**: This is what we are trying to tune: learning rate, number of layers/neurons in a NN, etc. Each **Experiment** will tune one or more **Parameters**.
- **Metric**: This refers to anything that we measure during the course of an **Experiment**. **Metrics** can be collected and stored for analysis during and after the experiment [2].
- **Objective**: This refers to the goal of the experiment - for example, maximizing the accuracy of a model. An **Objective** consists of the target value for some **metric**.   
- **Algorithm**: This refers to the mechanism of finding new values for **parameters**. Examples of this are random search, grid search, Bayesian optimization, etc. In katib v1alpha1, this is known as a *SuggestionSpec*.

**Trial**: This refers to one iteration of the optimization loop that Katib runs during the course of an **Experiment**. A **Trial** is composed of:
- **Suggestion**: A set of specific **parameter** assignments, as determined by the **algorithm**;
- **Evaluation**: The process of actually running the training model with the **suggestion**;
- **Observation**: The results of the **evaluation**, in terms of **metrics**.

## K8s Details

Going deeper, there are some terms that describe how Katib functions internally. These will be part of the Katib API, but ideally we should have some layer of abstraction to separate these from the core concepts above.

**Experiment** - This will be implemented as a CRD. The schema is:
```
type StudyJobSpec struct {
	Owner                     string                `json:"owner,omitempty"`
	ParameterSpecs     []ParameterSpec     `json:"parameterSpecs,omitempty"`
	ObjectiveSpec         ObjectiveSpec  `json:"objectiveSpec,omitempty"`

         // This can be empty if we only care about the objective metric.
         // Note: If we adopt a push instead of pull mechanism, this can be omitted completely.
	AdditionalMetricsNames []string   `json:"additionalMetricsNames,omitempty"`

         // Template for the evaluator - formerly known as a 'Worker'.
	TrialEvaluatorSpec           *TrialEvaluatorSpec           `json:"trialEvaluatorSpec,omitempty"`

        // This is same as a SuggestionSpec today.
	AlgorthmSpec       *AlgorithmSpec       `json:"algorithmSpec,omitempty"`

	ParallelTrialCount   int                   `json:"parallelTrialCount,omitempty"`
        MaxIterations          int                   `json:"maxIterations,omitempty"`

        // Other fields, exact format is TBD
        // - metrics collection
        // - NAS
        // - early stopping
        // - resume experiment
}

type ObjectiveSpec struct {
	OptimizationType     OptimizationType      `json:"optimizationType,omitempty"`
	OptimizationGoal     *float64              `json:"optimizationGoal,omitempty"`
	ObjectiveMetricName   string                `json:"objectiveMetricName,omitempty"`
}

```

The **Experiment** resource is watched and managed by the **ExperimentController**. The **ExperimentStatus** is composed of:
```
type ExperimentStatus struct {
	StartTime *metav1.Time `json:"startTime,omitempty"`
	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
	LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`

	Conditions               []ExperimentCondition  `json:"conditions,omitempty"`
	ExperimentID          string     `json:"experimentId,omitempty"`
	Trials                       []Trial    `json:"trials,omitempty"`

        // Since a Trial encapsulates suggestion + evaluation + observation, the user
        // can access all relevant information here.
	CurrentOptimalTrial Trial      `json:"currentOptimalTrial,omitempty"`
}

```

**Trial** is also implemented as a CRD:
```
type Trial struct {
	TrialID    string            `json:"trialId,omitempty"`

         // Key-value pairs for Parameters and Assignment values
        Suggestion Suggestion  `json:"suggestion,omitempty"`
        
        // Start/completion time, status, error messages, etc - TBD
        Evaluation Evaluation `json:"evaluation,omitempty"`

        // Results - objectives and other metrics values
        Observation Observation `json:"observation,omitempty"`
}
```
A **Trial** is also watched by a **TrialController**, whose responsibility is to spawn **TrialEvaluators**. These map to the existing *WorkerKinds* - BatchJob, TFJob, PytorchJob, etc.


[1] - One question that was raised earlier is whether this conflicts with Pipeline's notion of experiments. This may be acceptable since Katib is designed as an independent service; and if we do integrate together, the two notions can be differentiated by clear documentation ("pipeline experiment" vs "katib experiment").
[2] - Metrics collection and storage mechanism will likely changed from its current implementation. The design for this is in progress, and will be updated soon.

johnugeorge:
Do we need 
1. Trials    []Trial   in Experiment Status?
2. ExperimentID, Trial ID

hougangliu:
We can add `Owner` filed in Trial CRD to refer to its corresponding  Experiment

alexandraj777:
@richardsliu the spec is looking great! In addition to the major naming changes, I also like that you have clear one-word-names and definitions for `Algorithm` and `Objective`.

Comments:
 - `TrialEvaluatorSpec` - have you considered `EvaluatorSpec`?
 - Do the fields need to have the `Spec` suffix? Does this indicate something meaningful to the user, or can it be omitted?
 - re: previous comments on `MaxIterations`, this field looks like it refers to a global experiment maximum, is that the case? 
 - The `Objective` field mixes the terms `Optimization` and `Objective`. Would it work to use the names `ObjectiveType` and `ObjectiveGoal`? Seems like type and goal refer to the specific objective anyways, and not the configuration of the global optimization algorithm (ie, if your objective type is minimization, that applies to one experiment, rather than all experiments run through the global optimization algorithm)
 - I know `NAS` configuration is a TODO, but it would probably be best as a general field to help configuration any algorithm, rather than a NAS algorithm specific field (unless I'm completely misunderstanding, I thought NAS was another Algorithm, just with a specific DL use case?)
 - Does the trial need to hold all objects, or can it just hold IDs? Not quite sure the best design principles for CRDs, so this suggestion comes from my REST API experience

richardsliu:
@alexandraj777 

* I am actually unsure about using the term `Evaluator` - I just realized that it may be confused with a TensorFlow distributed training Evaluator role. My actual preference would be to not introduce another term here - each Trial should be just an association of Suggestions to Observations.

* You are correct, we don't need to add Spec everywhere. I think it makes sense when it specifies what a custom resource should look like, but it is not needed in all the field names.

* `MaxIterations` in this case would mean how many times we should run the Optimization Loop.

* Since we are grouping the objective-related fields in one struct, we can get rid of repeating `optimization` and `objective` in the field names. `Optimization` is redundant anyway; we are trying to run HP optimization experiments so that part is already implied.

* NAS config has more components than algorithm - for example there are configurations for what the DAG looks like. See https://github.com/kubeflow/katib/blob/master/pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go#L188.

* Trial does not need to hold all the objects. As @johnugeorge and @hougangliu suggested, we can simply use labels to denote resource ownership.

richardsliu:
Some revisions based on the above comments:

**Experiment**:
```
type ExperimentJobSpec struct {
	Owner             string              `json:"owner,omitempty"`
	Parameters     []ParameterSpec    `json:"parameters,omitempty"`
	Objective         *ObjectiveSpec       `json:"objective,omitempty"`
        // This is same as a SuggestionSpec today.
	Algorthm       *AlgorithmSpec       `json:"algorithm,omitempty"`

         // Template for each run of the trial - formerly known as a 'Worker'.
         // Need to decide on a name here...
	TrialRunSpec           *TrialRunSpec           `json:"trialRunSpec,omitempty"`

	ParallelTrialCount   int                   `json:"parallelTrialCount,omitempty"`
        MaxIterations        int                   `json:"maxIterations,omitempty"`

        // Other fields, exact format is TBD
        // - metrics collection
        // - NAS
        // - early stopping
        // - resume experiment
}

type ObjectiveSpec struct {
	Type     ObjectiveType      `json:"type,omitempty"`
	Goal     *float64              `json:"goal,omitempty"`
	ObjectiveMetricName   string           `json:"objectiveMetricName,omitempty"`
         // This can be empty if we only care about the objective metric.
         // Note: If we adopt a push instead of pull mechanism, this can be omitted completely.
	AdditionalMetricsNames []string   `json:"additionalMetricsNames,omitempty"`
}

type ExperimentJobStatus struct {
	StartTime *metav1.Time `json:"startTime,omitempty"`
	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
	LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`

	Conditions               []ExperimentCondition  `json:"conditions,omitempty"`

        // Since a Trial encapsulates suggestion + evaluation + observation, the user
        // can access all relevant information here.
	CurrentOptimalTrial *Trial      `json:"currentOptimalTrial,omitempty"`

        // Should we include more metadata about Trials here?
        // E.g. Number of Trials completed/failed/pending
}

```

**Trial**:
```
type TrialSpec struct {
	Owner             string              `json:"owner,omitempty"`
        // Key-value pairs for Parameters and Assignment values
        Suggestion Suggestion  `json:"suggestion,omitempty"`

        // This removes the notion of a worker/evaluator and collapses the raw template into each Trial.
        // The idea is that the ExperimentJobController would take care of template parsing.
        // Does this work? Similar to the relationship between Experiment and Trial, the actual running
        // BatchJob/TfJob/PytorchJobs would reference this Trial as its owner.
	RawTemplate  string `json:"rawTemplate,omitempty"`
}

type TrialStatus struct {
	StartTime *metav1.Time `json:"startTime,omitempty"`
	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
	LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`
	Conditions               []TrialCondition  `json:"conditions,omitempty"`

        // This is included here so that we can see what suggestions/templates were used
        TrialSpec *TrialSpec  `json:"trialSpec,omitempty"`

        // Results - objectives and other metrics values
        Observation Observation `json:"observation,omitempty"`        
}
```

johnugeorge:
Few Questions

1. Do we need a separate Owner field if labels are used? What is the difference?
2. Do we need Rawtemplate or real config in Trial? 
3.  Shall we expand AlgorithmSpec, TrialRunSpec for clarity?

jdplatt:
The new structure looks great!

hougangliu:
make a few update.
```
type ExperimentJob struct {
        metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
        Spec   ExperimentJobSpec   `json:"spec,omitempty"`
	Status ExperimentJobStatus `json:"status,omitempty"`
}

type ExperimentJobSpec struct {
	Parameters     []ParameterSpec    `json:"parameters,omitempty"`
	Objective         *ObjectiveSpec       `json:"objective,omitempty"`
        // This is same as a SuggestionSpec today.
	Algorithm       *AlgorithmSpec       `json:"algorithm,omitempty"`

         // Template for each run of the trial - formerly known as a 'Worker'.
         // Need to decide on a name here...
	TrialRunSpec           *TrialRunSpec           `json:"trialRunSpec,omitempty"`

        MaxIterations        int                   `json:"maxIterations,omitempty"`

        MetricsCollectorType   string `json:"metricsCollectorSpec,omitempty"`
        // Other fields, exact format is TBD
        // - NAS
        // - early stopping
        // - resume experiment
}

type TrialRunSpec struct {
	Retain     bool       `json:"retain,omitempty"`
	GoTemplate *GoTemplate `json:"goTemplate,omitempty"`
}

type TemplateSpec struct {
	ConfigMapName string `json:"configMapName,omitempty"`
	ConfigMapNamespace string `json:"configMapNamespace,omitempty"`
	TemplatePath string `json:"templatePath,omitempty"`
}

type GoTemplate struct {
	TemplateSpec *TemplateSpec `json:"templateSpec,omitempty"`
	RawTemplate  string `json:"rawTemplate,omitempty"`
}

type ObjectiveSpec struct {
	Type     ObjectiveType      `json:"type,omitempty"`
	Goal     float64              `json:"goal,omitempty"`
	ObjectiveMetricName   string           `json:"objectiveMetricName,omitempty"`
         // This can be empty if we only care about the objective metric.
         // Note: If we adopt a push instead of pull mechanism, this can be omitted completely.
	AdditionalMetricsNames []string   `json:"additionalMetricsNames,omitempty"`
}

type AlgorithmSpec struct {
    	Name              string                   `json:"name,omitempty"`
	Parameters     []AlgorithmParameter `json:"parameters,omitempty"`
	ParallelTrialCount   int                   `json:"parallelTrialCount,omitempty"`
}

type ExperimentJobStatus struct {
	StartTime *metav1.Time `json:"startTime,omitempty"`
	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
	LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`

	Conditions               []ExperimentCondition  `json:"conditions,omitempty"`

        // Since a Trial encapsulates suggestion + evaluation + observation, the user
        // can access all relevant information here.
	CurrentOptimalTrial *Trial      `json:"currentOptimalTrial,omitempty"`

        SuggestionRoundCount int `json:" suggestionRoundCount,omitempty"`

        // Should we include more metadata about Trials here?
        // E.g. Number of Trials completed/failed/pending
}

type Trial struct {
        metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
        Spec   TrialSpec   `json:"spec,omitempty"`
	Status TrialStatus `json:"status,omitempty"`
}

type TrialSpec struct {
        // Key-value pairs for Parameters and Assignment values
        Suggestion Suggestion  `json:"suggestion,omitempty"`

        // This removes the notion of a worker/evaluator and collapses the raw template into each Trial.
        // The idea is that the ExperimentJobController would take care of template parsing.
        // Does this work? Similar to the relationship between Experiment and Trial, the actual running
        // BatchJob/TfJob/PytorchJobs would reference this Trial as its owner.
       // ExperimentJobController will fill this field when spawn trial from TrialRunSpec.GoTemplate
	RawTemplate  string `json:"rawTemplate,omitempty"`
}

type TrialStatus struct {
	StartTime *metav1.Time `json:"startTime,omitempty"`
	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
	LastReconcileTime *metav1.Time `json:"lastReconcileTime,omitempty"`
	Conditions               []TrialCondition  `json:"conditions,omitempty"`

        // Results - objectives and other metrics values
        Observation Observation `json:"observation,omitempty"`        
}
```
Remove `Owner` field since metadata.ownerReferences has exist.

richardsliu:
@hougangliu 'ParallelTrialCount' should be removed from AlgorthmSpec since it is part of the ExperimentJobSpec now.

hougangliu:
> @hougangliu 'ParallelTrialCount' should be removed from AlgorthmSpec since it is part of the ExperimentJobSpec now.

I move it from ExperimentJobSpec to AlgorithmSpec since I think AlgorithmSpec is all for suggestion service (suggestion service will decide return how many trials based on ParallelTrialCount)

johnugeorge:
Currently, 'RequestNumber'( which defines the  ParallelTrialCount), is specified per Suggestion algorithm. However, I feel that parallelism is a general property which should be part of the Experiment and has to be supported by all algorithms.

Better name for SuggestionRoundCount?

richardsliu:
Someone also suggested using the word "Budget", which I also like.

YujiOshima:
@richardsliu 
* Why do you delete the TrialID? 
IMO, all info around Trial should need to be stored in DB and reuse it with ID. 

* Early Stopping and NAS are also algorithm.
How about like below?

```
type AlgorithmSpec struct {
	Suggestion{ 
		Name              string                   `json:"name,omitempty"`
		Parameters     []AlgorithmParameter `json:"parameters,omitempty"`
		ParallelTrialCount   int                   `json:"parallelTrialCount,omitempty"`
	}
	//TODO
	EalyStopping{
	}
}
```

richardsliu:
@YujiOshima Do we need TrialID in the CRD? Or just storing to DB? See @johnugeorge comment at https://github.com/kubeflow/katib/pull/381#issuecomment-466504035.

Early stopping spec looks fine.

YujiOshima:
I think both of `Experiment ID` and `Trial ID` are needed in CRD.
Can their operators interact with DB when these IDs are not in CR?

cvenets:
I also agree with @johnugeorge and @richardsliu that `ParallelTrialCount` is generic and should be independent from the Algorithm. It should be part of the Experiment. Maybe we can find an even better name for it to associate it with the Experiment and `MaxIterations`.

What about `TrialsPerIteration`, or `ParallelTrialsPerIteration`?

I also agree with @richardsliu that `Evaluator` will get people confused with TensorFlow's term. 'TrialRun' is good.

richardsliu:
Please take a look at the latest version, thanks!

johnugeorge:
Are we planning to provide asynchronous Trial execution as in https://github.com/kubeflow/katib/issues/223?  What would `MaxIterations` mean in this situation ?

pkg/api/operators/apis/experiment/v1alpha2/types.go
```diff
@@ -0,0 +1,194 @@
...
+	ObjectiveTypeMaximize ObjectiveType = "maximize"
+)
+
+type AlgorithmSpec struct {
+	Suggestion    *SuggestionSpec    `json:"suggestion,omitempty"`
+	EarlyStopping *EarlyStoppingSpec `json:"earlyStopping,omitempty"`
+}
+
+type SuggestionSpec struct {
+	Name       string                   `json:"name,omitempty"`
```
johnugeorge: This refers to algorithm name right?

richardsliu:
Pushed latest changes based on the discussion over the past week.

If this looks ok, let's merge the PR and start working on the database schema and protobuf APIs.

pkg/api/operators/apis/experiment/v1alpha2/types.go
```diff
@@ -0,0 +1,224 @@
...
+type ExperimentStatus struct {
+	// Represents time when the Experiment was acknowledged by the Experiment controller.
+	// It is not guaranteed to be set in happens-before order across separate operations.
+	// It is represented in RFC3339 form and is in UTC.
+	StartTime         *metav1.Time `json:"startTime,omitempty"`
+
+	// Represents time when the Experiment was completed. It is not guaranteed to
+	// be set in happens-before order across separate operations.
+	// It is represented in RFC3339 form and is in UTC.
+	CompletionTime    *metav1.Time `json:"completionTime,omitempty"`
```
alexandraj777: May be more conventional to use `EndTime`?

pkg/api/operators/apis/experiment/v1alpha2/types.go
```diff
@@ -0,0 +1,224 @@
...
+	// How many trials have been killed.
+	TrialsKilled int `json:"trialsKilled,omitempty"`
+
+	// How many trials are currently pending.
+	TrialsPending int `json:"trialsPending,omitempty"`
+}
+
+type OptimalTrial struct {
+	// Key-value pairs for the current optimal hyperparameters.
+	Hyperparameters []trial.Hyperparameter `json:"hyperparameters"`
```
alexandraj777: Is this the same as `Assignments`? Should that go on the `Observation`?

pkg/api/operators/apis/experiment/v1alpha2/types.go
alexandraj777: I'm also unsure what Suggestion represents here

pkg/api/operators/apis/experiment/v1alpha2/types.go
richardsliu: It is the same as Assignments. I think it should be outside of Observation since the two are parallel concepts (input and output basically).

pkg/api/operators/apis/experiment/v1alpha2/types.go
richardsliu: Sorry - this is the suggestion algorithm name. @johnugeorge did we decide to move this to the AlgorithmSpec?

pkg/api/operators/apis/experiment/v1alpha2/types.go
richardsliu: CompletionTime is more consistent with kubernetes native resources, for example BatchJobs: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/types.go#L156

pkg/api/operators/apis/trial/v1alpha2/types.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	pb "github.com/kubeflow/katib/pkg/api"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type TrialSpec struct {
+	// Key-value pairs for Parameters and Assignment values
+	SuggestionParameters []pb.SuggestionParameter `json:"suggestionParameters"`
```
YujiOshima: SuggestionParameter is a parameter for Suggestion service. https://github.com/kubeflow/katib/blob/master/pkg/api/api.proto#L327

This should be `Parameter` https://github.com/kubeflow/katib/blob/master/pkg/api/api.proto#L273
And the name `SuggestionParameters` is confusing. How about only `ParameterSet` or `ParameterList`?

pkg/api/operators/apis/experiment/v1alpha2/types.go
YujiOshima: Do you mean moving SuggestionSpec to AlgorithmSpec and remove EarlyStopping?
IMO, we should keep both of Suggestion and EarlyStopping.
To avoid confusing, we can rename the field `AlgorithmName`.

pkg/api/operators/apis/experiment/v1alpha2/types.go
```diff
@@ -0,0 +1,194 @@
...
+	ExperimentRunning    ExperimentConditionType = "Running"
+	ExperimentRestarting ExperimentConditionType = "Restarting"
+	ExperimentSucceeded  ExperimentConditionType = "Succeeded"
+	ExperimentFailed     ExperimentConditionType = "Failed"
+)
+
+type ParameterSpec struct {
+	Name          string        `json:"name,omitempty"`
+	ParameterType ParameterType `json:"parameterType,omitempty"`
+	FeasibleSpace FeasibleSpace `json:"feasibleSpace,omitempty"`
```
jdplatt: I'd still really like to add the ability to transform parameters here. It's a really common use case (e.g. tuning over the log learning rate form -6 to -3 rather than the actual learning rate from 1e-6 to 1e-3), won't complicate anything else, and easy to implement in the suggestion package.

pkg/api/operators/apis/trial/v1alpha2/types.go
jdplatt: I really liked Alexandra's suggestion of ParameterAssignment. I agree SuggestionParameters is confusing.

pkg/api/operators/apis/experiment/v1alpha2/types.go
richardsliu: That won't require any API or CRD changes right? If I understand this correctly.

pkg/api/operators/apis/experiment/v1alpha2/types.go
richardsliu: I meant moving the field `Name` to AlgorithmSpec, keep the rest the same. I can also rename it as `AlgorithmName` to avoid confusion.

richardsliu:
- Renamed "hyperparameter" and "suggestionparameter" to "ParameterAssignment".
- Restructured AlgorithmSpec

pkg/api/operators/apis/experiment/v1alpha2/types.go
jdplatt: It shouldn't. It will only get used inside the suggestion services to transform the parameter space before generating suggestions

pkg/api/operators/apis/experiment/v1alpha2/types.go
```diff
@@ -0,0 +1,220 @@
...
+const (
+	ObjectiveTypeUnknown  ObjectiveType = ""
+	ObjectiveTypeMinimize ObjectiveType = "minimize"
+	ObjectiveTypeMaximize ObjectiveType = "maximize"
+)
+
+type AlgorithmSpec struct {
+	AlgorithmName string		 `json:"algorithmName,omitempty"`
+	// Key-value pairs for hyperparameters and assignment values.
+	ParameterAssignments []trial.ParameterAssignment `json:"parameterAssignments"`
```
johnugeorge: This is algorithm specific parameters.  eg: https://github.com/kubeflow/katib/blob/master/examples/grid-example.yaml#L41

Will this be confusing with ParameterAssignment of Trial though both are key-value struct type?

YujiOshima: I agree with @johnugeorge . 
In the gRpc API, we distinguish these concepts.
ParameterAssignment of Trial: https://github.com/kubeflow/katib/blob/master/pkg/api/api.proto#L284
Parameter for Suggestion service: https://github.com/kubeflow/katib/blob/master/pkg/api/api.proto#L338

It is also key-value but we should use another name. How about `AlgorithmParameterAssignment` ?

richardsliu: I think it should be something like AlgorithmParameter. "Assignment" is for assigning HP values which should not be confused with internal parameters for suggestion algorithms (sorry if my previous answer was misleading).

richardsliu:
Some small fixes:
- Added "retainHistoricalData" to Experiment spec
- Added a type for "AlgorithmSetting" to avoid confusion with HP assignments

Everyone please take a look and lgtm if you think the API looks ok. We can still make minor edits after merging this PR.

YujiOshima:
@richardsliu Thank you!
/lgtm

johnugeorge:
LGTM
nitpick:   We might not repeat Algorithm term in AlgorithmSpec fields. We can change this later too.

richardsliu:
Thanks everyone!
/approve

Issue #314, opened 2019-01-03
studyJob controller is blocked by bad CR manifests


hougangliu:
refer to [same root cause ](https://github.com/kubernetes/kubernetes/issues/57705) . we can follow the recommendation in [this comment](https://github.com/kubernetes/kubernetes/issues/57705#issuecomment-354552314)

```
ERROR: logging before flag.Parse: E0103 11:58:27.704970       1 streamwatcher.go:109] Unable to decode an event from the watch stream: unable to decode watch event: v1alpha1.StudyJob.Spec: v1alpha1.StudyJobSpec.RequestCount: readUint64: unexpected character: �, error found in #10 byte of ...|stcount":"l","studyN|..., bigger context ...|","parametertype":"categorical"}],"requestcount":"l","studyName":"random-example","suggestionSpec":{|...
ERROR: logging before flag.Parse: E0103 11:58:27.713261       1 streamwatcher.go:109] Unable to decode an event from the watch stream: unable to decode watch event: v1alpha1.StudyJob.Spec: v1alpha1.StudyJobSpec.RequestCount: readUint64: unexpected character: �, error found in #10 byte of ...|stcount":"l","studyN|..., bigger context ...|","parametertype":"categorical"}],"requestcount":"l","studyName":"random-example","suggestionSpec":{|...
ERROR: logging before flag.Parse: W0103 11:58:27.713500       1 reflector.go:341] github.com/kubeflow/katib/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:106: watch of *v1alpha1.StudyJob ended with: very short watch: github.com/kubeflow/katib/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:106: Unexpected watch close - watch lasted less than a second and no items received
ERROR: logging before flag.Parse: E0103 11:58:28.716347       1 reflector.go:205] github.com/kubeflow/katib/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:106: Failed to list *v1alpha1.StudyJob: v1alpha1.StudyJobList.Items: []v1alpha1.StudyJob: v1alpha1.StudyJob.Spec: v1alpha1.StudyJobSpec.RequestCount: readUint64: unexpected character: �, error found in #10 byte of ...|stcount":"l","studyN|..., bigger context ...|","parametertype":"categorical"}],"requestcount":"l","studyName":"random-example","suggestionSpec":{|...
[...]

PR #383, opened 2019-02-15
add validating webhook for studyJob

hougangliu:
If create/update a studyJob with bad CR manifest or invalid configuration, k8s api
server will reject the request.
Fixes: #314

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/383)

### DIFF
--- a/manifests/studyjobcontroller/rbac.yaml
+++ b/manifests/studyjobcontroller/rbac.yaml
@@ -8,6 +8,7 @@ rules:
   resources:
   - configmaps
   - serviceaccounts
+  - services
   verbs:
   - "*"
 - apiGroups:
@@ -32,6 +33,12 @@ rules:
   verbs:
   - create
   - get
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - validatingwebhookconfigurations
+  verbs:
+  - '*'
 - apiGroups:
   - kubeflow.org
   resources:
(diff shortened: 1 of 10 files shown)


richardsliu:
@hougangliu Does this require any specific version of k8s server to work?

hougangliu:
api admission webhook introduced in k8s 1.9

hougangliu:
api admission webhook introduced in k8s 1.9

johnugeorge:
In operators, job goes into failed state when the config is invalid.  What is the recommended way?

hougangliu:
when we submit a k8s resource(saying deployment), API server reject it. so I think we'd better follow k8s way

johnugeorge:
I was thinking about the consistency across kubeflow components.  

There are few pre requisites for API servers.  https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#is-there-a-recommended-set-of-admission-controllers-to-use  I am not sure if admission plugins are enabled by default for api servers across clouds.
https://istio.io/help/ops/setup/webhook/#verify-dynamic-admission-webhook-prerequisites

hougangliu:
In k8s official doc, ValidatingAdmissionWebhook is in recommended set of admission controllers to use and enabled by default.

hougangliu:
in 1.9, ValidatingAdmissionWebhook is RecommendedPlugin. After 1.10, it is enabled by default.

johnugeorge:
Sounds good. Can you add a test too?

hougangliu:
> Sounds good. Can you add a test too?

sure, done! thanks!

johnugeorge:
/lgtm
I will wait for @richardsliu

PR #387, opened 2019-02-17
Removing Operator specific handling during a StudyJob run

johnugeorge:
During StudyJob run, there are no more operator specific code. The unstructured is used to create the resource, get the resource and get the status

The only remaining is the watch changes that has to be separately handled in #317. Current controller runtime code(0.1.7 and less) doesn't support dynamic watch. Refer: https://github.com/kubernetes-sigs/kubebuilder/issues/422  

Related: #341

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/387)

### DIFF
--- a/pkg/controller/studyjob/studyjob_controller.go
+++ b/pkg/controller/studyjob/studyjob_controller.go
@@ -35,7 +35,9 @@ import (
 	batchv1beta "k8s.io/api/batch/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -120,7 +122,6 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if isFatalWatchError(err, TFJobWorker) {
 		return err
 	}
-
 	err = c.Watch(
 		&source.Kind{Type: &pytorchjobv1beta1.PyTorchJob{}},
 		&handler.EnqueueRequestForOwner{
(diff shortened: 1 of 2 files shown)


pkg/controller/studyjob/studyjob_controller.go
```diff
@@ -393,46 +394,55 @@ func (r *ReconcileStudyJobController) updateWorker(c katibapi.ManagerClient, ins
...
+	switch wkind.Kind {
+
 	case DefaultJobWorker:
-		job := runtimejob.(*batchv1.Job)
+		var job batchv1.Job
+		if err := r.Client.Get(context.TODO(), nname, &job); err != nil {
+			log.Printf("Client Get error %v for %v", err, nname)
+			return WorkerStatus{}
+		}
 		if job.Status.Active == 0 && job.Status.Succeeded > 0 {
```
richardsliu: It seems like we should be able to use the similar logic here to check for job completion: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/batch/v1/generated.proto#L158

pkg/controller/studyjob/studyjob_controller.go
```diff
@@ -393,46 +394,55 @@ func (r *ReconcileStudyJobController) updateWorker(c katibapi.ManagerClient, ins
...
-		return WorkerStatus{}
-	}
+func (r *ReconcileStudyJobController) getJobWorkerStatus(ns string, wid string, wkind *schema.GroupVersionKind) WorkerStatus {
+	nname := types.NamespacedName{Namespace: ns, Name: wid}
 	var state katibapi.State = katibapi.State_RUNNING
 	var cpTime *metav1.Time
-	switch w.Kind {
+	switch wkind.Kind {
+
 	case DefaultJobWorker:
```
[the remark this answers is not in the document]
johnugeorge: Yes. But the string values are different currently.  'Complete' in  https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/batch/v1/types.go#L170  vs 'Succeeded' in https://github.com/kubeflow/tf-operator/blob/master/pkg/apis/common/v1beta2/common_types.go#L121 

In general, since 'job' is a built-in K8s  Kind, should we care about it matching with our JobCondition?  I feel that the only necessary condition is that  all custom job operators share the same JobStatus type. 

Or other option is, we can change the common JobConditionType string to match the batch Job's JobConditionType. But new developers might not be aware of this assumption and may break the logic.

pkg/controller/studyjob/studyjob_controller.go
johnugeorge: Please see above comment

pkg/controller/studyjob/studyjob_controller.go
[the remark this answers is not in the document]
richardsliu: I think this is ok for this version. One thing that we have discussed in the past is moving the common types into its own repo. If we go down that route, it will be easier to change then. What do you think?

pkg/controller/studyjob/studyjob_controller.go
```diff
@@ -393,46 +394,55 @@ func (r *ReconcileStudyJobController) updateWorker(c katibapi.ManagerClient, ins
...
-			} else if lc.Type == commonv1beta1.JobFailed {
-				state = katibapi.State_ERROR
+		status, ok, unerr := unstructured.NestedFieldCopy(u.Object, "status")
+
+		if ok {
+			statusMap := status.(map[string]interface{})
+			jobStatus := commonv1beta1.JobStatus{}
+			err := runtime.DefaultUnstructuredConverter.FromUnstructured(statusMap, &jobStatus)
+			if err != nil {
+				log.Printf("Error in converting unstructured to status:%v ", err)
```
hougangliu: s/status:%v/status: %v

hougangliu: return WorkerStatus{}

pkg/controller/studyjob/studyjob_controller.go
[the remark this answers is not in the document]
johnugeorge: Yup. There is a discussion in https://github.com/kubeflow/mpi-operator/issues/92#issuecomment-464777479

PR #392, opened 2019-02-19
 Add separable/depthwise convolution and a new example to reproduce ENAS

DeeperMind:
This pull request added separable convolution and depthwise convolution to the operation library of the training container. It also added a new StudyJob example that completely reproduces the search space in Efficient Neural Architecture Search via Parameter Sharing (https://arxiv.org/abs/1802.03268). Experiments show that it can improve the performance and lower the generalization error.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/392)


DeeperMind:
will close the PR due to some problems in github versions. Will merge from master first and open a new PR.

PR #394, opened 2019-02-20
Database APIs for NAS updated

Akado2009:
This PR introduces a distinct set of API functions to save/get NAS jobs into the mysql databases. For now it creates a separate table, but after new update we're gonna use studies table, just increase the number of columns in that one. This one should be merged after: #327.

New version of https://github.com/kubeflow/katib/pull/328 with squashing

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/394)


andreyvelich:
/assign @YujiOshima @hougangliu @johnugeorge 
We passed all tests, I think we can, finally, merge this PR.

pkg/db/interface.go
```diff
@@ -46,8 +46,10 @@ type VizierDBInterface interface {
 	DBInit()
 	SelectOne() error
 
-	GetStudyConfig(string) (*api.StudyConfig, error)
+	GetStudyMetrics(string) ([]string, error)
```
hougangliu: do GetStudyMetrics really need be exposed in VizierDBInterface? it seems for internal use

andreyvelich: I think, this is just for Select metrics from studies table.
We use it here: https://github.com/kubeflow/katib/blob/869fdf23b70c602b9aa9b30af5bfaff15a9ae8a5/pkg/db/interface.go#L885

hougangliu: if it's only in need in pkg/db/interface.go, you can remove line 49

Issue #372, opened 2019-02-13
need validate nasConfig


hougangliu:
Now for a studyjob, we never check if a nasConfig is valid or not.
#339 is nasrl suggestion service, which always expects graphConfig (numlayers, input size,outputsize) and operations field present. we need validate nasconfig filed before store studyjob info into db.

@andreyvelich WDYT?

andreyvelich:
@hougangliu 
This is a good point! 
Do you think, we should check it in the controller step? 
For example, if something wrong with nasConfig structure, studyJob status should be Failed and controller should stop the job?

johnugeorge:
See validateStudy  in https://github.com/kubeflow/katib/blob/f11c13e8c65a9236afe2878e5c915a1c769ad52b/pkg/controller/studyjob/katib_api_util.go#L28

andreyvelich:
@johnugeorge 
Thats cool! 
Should I add a part there for the NAS job ?

PR #398, opened 2019-02-20
Add validation for NAS job in Katib controller

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/372.
As well, I added validation for HP job.
/assign @hougangliu @YujiOshima

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/398)


pkg/controller/studyjob/utils.go
```diff
@@ -209,3 +210,88 @@ func contains(l []string, s string) bool {
...
+
+	if instance.Spec.NasConfig.GraphConfig.InputSize == nil {
+		return fmt.Errorf("Missing InputSize in NasConfig.GraphConfig: %v", instance.Spec.NasConfig.GraphConfig)
+	}
+
+	if instance.Spec.NasConfig.GraphConfig.OutputSize == nil {
+		return fmt.Errorf("Missing OutputSize in NasConfig.GraphConfig: %v", instance.Spec.NasConfig.GraphConfig)
+	}
+
+	if instance.Spec.NasConfig.GraphConfig.NumLayers == 0 {
```
hougangliu: NumLayers is not always in need for all NAS (ENAS need it). I think we need validate NAS considering suggestion algorithm

andreyvelich: @hougangliu I think we can check these parameters exactly for Reinforcement Learning algorithm for now. For NumLayers and Step here: https://github.com/kubeflow/katib/blob/d407be8252034585bc770ced2de36eb39b72ddd6/pkg/controller/studyjob/utils.go#L287, we should validate only for RL algorithm right now. 
What do you think about it?

pkg/controller/studyjob/utils.go
```diff
@@ -206,6 +207,91 @@ func contains(l []string, s string) bool {
...
+
+	if instance.Spec.NasConfig.GraphConfig.InputSize == nil {
+		return fmt.Errorf("Missing InputSize in NasConfig.GraphConfig: %v", instance.Spec.NasConfig.GraphConfig)
+	}
+
+	if instance.Spec.NasConfig.GraphConfig.OutputSize == nil {
+		return fmt.Errorf("Missing OutputSize in NasConfig.GraphConfig: %v", instance.Spec.NasConfig.GraphConfig)
+	}
+
+	if instance.Spec.NasConfig.GraphConfig.NumLayers == 0 && instance.Spec.SuggestionSpec.SuggestionAlgorithm == suggestionAlgorithmRL {
```
YujiOshima: Is this a specific param for the RL algorithm?

pkg/controller/studyjob/utils.go
```diff
@@ -206,6 +207,91 @@ func contains(l []string, s string) bool {
...
+		if pc.Name == "" {
+			return fmt.Errorf("Missing Name in ParameterConfig: %v", pc)
+		}
+		if pc.ParameterType == "" {
+			return fmt.Errorf("Missing ParameterType in ParameterConfig: %v", pc)
+		}
+		if reflect.DeepEqual(pc.Feasible, katibv1alpha1.FeasibleSpace{}) {
+			return fmt.Errorf("Missing Feasible in ParameterConfig: %v", pc)
+		}
+		if pc.ParameterType == katibv1alpha1.ParameterTypeCategorical {
```
YujiOshima: Not only categorical but discrete.

andreyvelich: In discrete we also use only List structure or something else?

YujiOshima: Currently discrete only need List. But we may need more type for discretes, int or double.
Please treated in the same way as Categorical for now.

andreyvelich: Done

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -25,6 +25,25 @@ import (
...
+			return validJobErr
+		}
+	} else {
+		//Validate NAS job
+		if validJobErr := validateNASJob(instance); validJobErr != nil {
+			instance.Status.Condition = katibv1alpha1.ConditionFailed
+			return validJobErr
+		}
+	}
+
```
hougangliu: move validateHPJob and validateNASJob into validateStudy, and move validateStudy away from initializeStudy

pkg/controller/studyjob/utils.go
```diff
@@ -206,6 +207,91 @@ func contains(l []string, s string) bool {
...
+	log.Printf("Validation for the HP job")
+	if instance.Spec.ParameterConfigs == nil {
+		return fmt.Errorf("No Spec.ParameterConfigs specified")
+	}
+	err := validateParameterConfigs(instance.Spec.ParameterConfigs, jobTypeHP, instance.Spec.SuggestionSpec.SuggestionAlgorithm)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
hougangliu: `return validateParameterConfigs` to replace line 215-219

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -25,6 +25,25 @@ import (
 )
 
 func initializeStudy(instance *katibv1alpha1.StudyJob) error {
+	if validErr := validateStudy(instance); validErr != nil {
+		instance.Status.Condition = katibv1alpha1.ConditionFailed
+		return validErr
+	}
+
+	if getJobType(instance) != jobTypeNAS {
```
hougangliu: please use  suggestionSpec.suggestionAlgorithm to return jobTypeNAS or jobTypeHP in getJobType to handle the invalid studyjob case: .Spec.NasConfig is not nil and suggestionSpec.suggestionAlgorithm is not nas-like

pkg/controller/studyjob/const.go
```diff
@@ -12,11 +12,12 @@ limitations under the License.
...
-	TFJobWorker      = "TFJob"
-	PyTorchJobWorker = "PyTorchJob"
-	jobTypeNAS       = "NAS"
-	jobTypeHP        = "HP"
+	DefaultJobWorker      = "Job"
+	TFJobWorker           = "TFJob"
+	PyTorchJobWorker      = "PyTorchJob"
+	jobTypeNAS            = "NAS"
+	jobTypeHP             = "HP"
+	suggestionAlgorithmRL = "nasrl"
```
hougangliu: s/suggestionAlgorithmRL/suggestionNASRL

pkg/controller/studyjob/utils.go
```diff
@@ -206,6 +207,91 @@ func contains(l []string, s string) bool {
...
+		}
+		if pc.ParameterType == katibv1alpha1.ParameterTypeDouble || pc.ParameterType == katibv1alpha1.ParameterTypeInt {
+			if pc.Feasible.Min == "" {
+				return fmt.Errorf("Missing Min in ParameterConfig.Feasible: %v", pc.Feasible)
+			}
+			if pc.Feasible.Max == "" {
+				return fmt.Errorf("Missing Max in ParameterConfig.Feasible: %v", pc.Feasible)
+			}
+			if jobType == jobTypeNAS && pc.Feasible.Step == "" && pc.ParameterType == katibv1alpha1.ParameterTypeDouble && suggestionAlgorithm == suggestionAlgorithmRL {
+				return fmt.Errorf("Missing Step in ParameterConfig.Feasible for NAS job: %v", pc.Feasible)
```
hougangliu: define a suggestionNASList (now only nasrl is in)
s/suggestionAlgorithm == suggestionAlgorithmRL/contains(suggestionNASList, suggestionAlgorithm)
so that when we add other nas suggestion, we only need update suggestionNASList

pkg/controller/studyjob/katib_api_util.go
andreyvelich: You want to put validateStudy function before initializeStudy?
Right here: https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L250 ?

hougangliu: no, validateStudy has been moved to https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/validating_webhook.go#L44

andreyvelich:
@hougangliu Thank you! I changed it.

pkg/controller/studyjob/utils.go
```diff
@@ -237,7 +251,7 @@ func validateNASJob(instance *katibv1alpha1.StudyJob) error {
 		return fmt.Errorf("Missing OutputSize in NasConfig.GraphConfig: %v", instance.Spec.NasConfig.GraphConfig)
 	}
 
-	if instance.Spec.NasConfig.GraphConfig.NumLayers == 0 && instance.Spec.SuggestionSpec.SuggestionAlgorithm == suggestionAlgorithmRL {
+	if instance.Spec.NasConfig.GraphConfig.NumLayers == 0 && contains(suggestionNASList, instance.Spec.SuggestionSpec.SuggestionAlgorithm) {
```
hougangliu: I think line 298 can use contains(suggestionNASList,..), but here I don't think so.
If we add another nas suggestion, it may not need NumLayers.

andreyvelich: You think we should specify this block only for ```suggestionNASRL``` ?

johnugeorge: Is there a better place to put algorithm specific checks other than the controller?  This list can grow over time with more algorithms and these checks are very specific to the algorithm implementation. @andreyvelich @hougangliu  Thoughts?

andreyvelich: From my point of view, it is important to check it in the controller. We don't need to run getSuggestion if something wrong with yaml file. We can't set failed condition for the StudyJob inside Suggestion. Controller do that.

pkg/controller/studyjob/const.go
```diff
@@ -12,12 +12,14 @@ limitations under the License.
...
+	TFJobWorker      = "TFJob"
+	PyTorchJobWorker = "PyTorchJob"
+	jobTypeNAS       = "NAS"
+	jobTypeHP        = "HP"
+	suggestionNASRL  = "nasrl"
 )
 
 var ValidWorkerKindList = [...]string{DefaultJobWorker, TFJobWorker, PyTorchJobWorker}
+
+var suggestionNASList = []string{suggestionNASRL}
```
johnugeorge: Our goal is to keep controller independent of worker type and suggestion algos.  Currently there are no worker specific code in the code(except for watch. Hope, it can be fixed soon and ValidWorkerKindList won't not needed any more).  Similarly i feel that there should not be any suggestion algorithm specific changes in the controller code.

andreyvelich: In that case, how we can make a validation for the NAS job ?

pkg/controller/studyjob/utils.go
johnugeorge: Same here. Better way to handle this?

pkg/controller/studyjob/utils.go
YujiOshima: We should do algorithm-specific check in suggestion service.

* Add ValidateSuggestionParameter API to Suggestion service 
or
* GetSuggesiton returns a specific error code if the config is invalid then the controller handle the error code

@andreyvelich @johnugeorge @hougangliu Thoughts?

hougangliu: > Add ValidateSuggestionParameter API to Suggestion service
Good idea

andreyvelich: @YujiOshima How it will work, if we add ValidateSuggestionParameter API in Suggestion Service? Can you explain, please? Right now, suggestion return just trials that need to be created by controller.

YujiOshima: @andreyvelich At initialization, the controller calls ValidateSuggestionParameter and send the parameters of a suggestion algorithm to a corresponding suggestion service.
Then the SuggestionService validates the parameters and if it returns an error, the controller won't start the optimization loop.

andreyvelich: Do you think we should do it for each Suggestion or make it just for parameters specific for current suggestion?
For example for RL Suggestion parameter Number of Layers here: 
https://github.com/kubeflow/katib/blob/76f8397980979ae2c97932330196933e23d4cfd4/pkg/controller/studyjob/utils.go#L34
and Step here:
https://github.com/kubeflow/katib/blob/76f8397980979ae2c97932330196933e23d4cfd4/pkg/controller/studyjob/utils.go#L298 ?
Other parameters specific to each NAS or HP algorithm, we can validate inside validation function in controller.

YujiOshima: We should do it for every Suggestion algos.
For example in BO, [this function](https://github.com/kubeflow/katib/blob/76f8397980979ae2c97932330196933e23d4cfd4/pkg/suggestion/bayesian_service.py#L138) should be done in ValidateSuggestionParameter.

And same as GetSuggestion, it can automatically select algorithm which user requested. https://github.com/kubeflow/katib/blob/master/cmd/manager/main.go#L104

andreyvelich: Should we also put inside Suggestion validation of these parameters: https://github.com/kubeflow/katib/blob/76f8397980979ae2c97932330196933e23d4cfd4/pkg/controller/studyjob/utils.go#L277 ?

johnugeorge: @YujiOshima In such a design, what should be the validation response  when the suggestion service doesn't exist in the cluster(not yet deployed)? 
Sample scenario: User applies a valid Studyjob config. However, the corresponding Suggestion service doesn't exist in the cluster  (may not be deployed at that instant)

YujiOshima: @andreyvelich  Both is OK that calling SetSuggestionParameters before validation or put parameters in Validate args.

@johnugeorge 
Same as GetSuggestion, the manager will receive the request like a proxy. If the Suggestion service does not exist or not responsible, the manager can handle and return Unavailable  error code https://github.com/grpc/grpc-go/blob/master/codes/codes.go#L139

andreyvelich: @YujiOshima In that case, should we send StudyConfig to ValidateSuggestionParameters, because we don't save StudyConfig in db, if it is not valid. 
We can't GetStudy like here: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/nasrl_service.py#L219 ?
As well, should we put this function after this line: https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/katib_api_util.go#L43 ?
We generate StudyConfig and can send it to Suggestion to check, whether it is valid or not.

YujiOshima: @andreyvelich You're right. StudyConf and SuggestionParams should be included in an args of ValidateSuggestionParameters.

andreyvelich: @YujiOshima  Ok.  In this PR, should I make a validation only for NAS RL service, or we should include BO here, as well?

YujiOshima: @andreyvelich Only for NAS RL is enough here. If you could add validation for BO in another PR, it would be fantastic.

andreyvelich: @YujiOshima @hougangliu What do you think about name of this API: ```ValidateStudyParameters``` ? 
It looks confused with ```ValidateSuggestionParameters``` since we validate StudyConfig and SuggestionParameters in this function

andreyvelich:
Validation in NAS RL service is not ready
/hold

andreyvelich:
I created a new API function called ValidateSuggestionParameters and added validation inside NAS RL service.
Right now, we can add this function inside Suggestion service and validate your StudyJob.
Logic works like this:

```
StudyJobController -> vizier-core-manager -> Suggestion service

Case1: Suggestion can't reply (Suggestion doesn't have ValidateSuggestionParameters function)

StudyJobController (Request)-> vizier-core-manager  (Request)->x Failed Suggestion service
StudyJobController <-(Reply{}, error code Unknown) vizier-core-manager - Suggestion service
instance.Status.Condition = katibv1alpha1.ConditionRunning

Case2: StudyJob is Valid

StudyJobController (Request)-> vizier-core-manager  (Request)-> Suggestion service
StudyJobController <-(Reply{}, nil) vizier-core-manager <-(Reply{}, nil) Suggestion service
instance.Status.Condition = katibv1alpha1.ConditionRunning

Case3: StudyJob is Invalid

StudyJobController (Request)-> vizier-core-manager  (Request)-> Suggestion service
StudyJobController <-(Reply{}, error code InvalidArgument + error message from Suggestion) vizier-core-manager <-(Reply{}, error code InvalidArgument + error message from Suggestion) Suggestion service
instance.Status.Condition = katibv1alpha1.ConditionFailed
```

@YujiOshima @hougangliu @johnugeorge @richardsliu Please, check it.

/hold cancel

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -42,8 +48,20 @@ func initializeStudy(instance *katibv1alpha1.StudyJob) error {
 
 	studyConfig, err := getStudyConf(instance)
 	if err != nil {
+		instance.Status.Condition = katibv1alpha1.ConditionFailed
 		return err
 	}
+
+	log.Println("Validate Parameters inside Suggestion Start")
```
hougangliu: Start to validate Suggestion  Parameters

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -42,8 +48,20 @@ func initializeStudy(instance *katibv1alpha1.StudyJob) error {
...
+	isValidSuggestionParameters := validateSuggestionParameters(c, studyConfig, instance.Spec.SuggestionSpec.SuggestionParameters, instance.Spec.SuggestionSpec.SuggestionAlgorithm)
+
+	if isValidSuggestionParameters {
+		log.Println("Suggestion Parameters is valid")
+	} else {
+		instance.Status.Condition = katibv1alpha1.ConditionFailed
+		return errors.New("Suggestion Parameters is not valid")
+	}
```
hougangliu: /s/is/are
or `Valid suggestion Parameters` and `Invalid  suggestion Parameters`

pkg/api/api.proto
```diff
@@ -784,3 +794,15 @@ message EarlyStoppingParameterSet {
...
+    StudyConfig study_config = 1;
+    string suggestion_algorithm = 2;
+    repeated SuggestionParameter suggestion_parameters = 3;    
+}
+
+/**
+ * Return 1 if Suggestion Parameters is Valid, 0 if not
+ */
+message ValidateSuggestionParametersReply {
+}
```
hougangliu: bool is_valid 1;
string msg 2;

andreyvelich:
@jlewi @YujiOshima @hougangliu @richardsliu 
Should we switch test server to us-east1-c since us-east1-d has VM shortage?
Like here: https://github.com/kubeflow/kubeflow/pull/2583

pkg/api/api.proto
```diff
@@ -212,10 +212,20 @@ service Manager {
             get: "/api/Manager/GetSavedModels"
         };
     };
+    /** 
+     * Validate Suggestion Parameters from Study Job.
```
YujiOshima: Please explain a suggestion service should return `INVALID_ARGUMENT` error when the parameter is invalid.

cmd/manager/main.go
```diff
@@ -115,6 +115,24 @@ func (s *server) GetSuggestions(ctx context.Context, in *api_pb.GetSuggestionsRe
 	return r, nil
 }
 
+func (s *server) ValidateSuggestionParameters(ctx context.Context, in *api_pb.ValidateSuggestionParametersRequest) (*api_pb.ValidateSuggestionParametersReply, error) {
+	if in.SuggestionAlgorithm == "" {
+		return &api_pb.ValidateSuggestionParametersReply{}, errors.New("No suggest algorithm specified")
+	}
+
+	conn, err := grpc.Dial("vizier-suggestion-"+in.SuggestionAlgorithm+":6789", grpc.WithInsecure())
+	if err != nil {
```
YujiOshima: It should return specific error code here (e.g. `Unavailable` ) and set message.

pkg/controller/studyjob/katib_api_util.go
```diff
@@ -328,3 +355,38 @@ func getSuggestion(c katibapi.ManagerClient, studyID string, suggestionSpec *kat
...
+
+	_, err := c.ValidateSuggestionParameters(ctx, validateSuggestionParametersReq)
+
+	statusCode, _ := status.FromError(err)
+
+	if statusCode.Code() == codes.Unknown {
+		log.Printf("Method ValidateSuggestionParameters not found inside Suggestion service: %s", suggestionAlgorithm)
+		return true
+	}
```
YujiOshima: Please handle `InvalidArgument` and `Unavailable` error and print the reason for the error (the parameter is invalid or service is unavailable).

cmd/manager/main.go
andreyvelich: I think we should delete this part: 
https://github.com/kubeflow/katib/pull/398/files/ea896ae1a4c4690ef8873ce416a019e782710f9e#diff-9e63f7b0791edcb4a53c944552b88970R119.
Since you add "random" algorithm, if it is not specified: 
https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/katib_api_util.go#L28

andreyvelich: @YujiOshima By default this Error https://github.com/kubeflow/katib/pull/398/files#diff-9e63f7b0791edcb4a53c944552b88970R120 will return Unavailable code. I think, we don't need to specify this error code there. 
Right now, I handle this error here, as you said:  https://github.com/kubeflow/katib/pull/398/commits/b88030b9e9a149b335189598d70ef881f38a6bf8#diff-7a381fdaeae1295724b74c0a04f7644cR386.
Please check, if it is ok.

pkg/api/api.proto
```diff
@@ -212,10 +212,21 @@ service Manager {
...
+     * Suggestion service should return INVALID_ARGUMENT Error when the parameter is invalid
+     */
+     rpc ValidateSuggestionParameters(ValidateSuggestionParametersRequest) returns (ValidateSuggestionParametersReply){
+        option (google.api.http) = {
```
hougangliu: I find that ValidateSuggestionParametersReply is used as `return  &api.ValidateSuggestionParametersReply{}` all the time. can we drop ValidateSuggestionParametersReply, alway return nothing?
or you can use below struct to replace current one 
```
message ValidateSuggestionParametersReply {
    bool is_valid 1;
    string msg 2;
 }
```

andreyvelich: @hougangliu  But we don't need to return anything, like here: 
https://github.com/kubeflow/katib/blob/master/cmd/manager/main.go#L213. 
If parameter is not valid, we return ```INVALID_ARGUMENT``` Error message.

andreyvelich:
@YujiOshima @hougangliu Do you have any other suggestions on this PR?

hougangliu:
/lgtm
/approve
Thanks!

Issue #382, opened 2019-02-15
Support parallel StudyJobs in the NAS job


andreyvelich:
Currently RL Suggestion for the NAS tasks doesn't support more than one StudyJob in the same time. We should change this suggestion to make it possible.
/assign @DeeperMind 
/area nas

PR #404, opened 2019-02-21
Add support for parallel studyjobs

DeeperMind:
Now the suggestion of neural architecture search with reinforcement learning supports multiple studyjobs  running in parallel. 

Fixes https://github.com/kubeflow/katib/issues/382

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/404)


pkg/suggestion/nasrl_service.py
```diff
@@ -18,9 +17,21 @@ def __init__(self, logger=None):
...
+        self.is_first_run = dict()
+        self.suggestion_configs = dict()
+        self.controllers = dict()
+        self.num_layers = dict()
+        self.input_size = dict()
+        self.output_size = dict()
+        self.num_operations = dict()
+        self.search_space = dict()
+        self.opt_direction = dict()
+        self.objective_name = dict()
```
hougangliu: how about self.studyinfo = {"studyid": {name: "", "tf_graph": xx, "prev_trial_id": xx, "ctrl_cache_file": xx, ..., "objective_name": xx}} to replace above dicts since they has same key?

pkg/suggestion/nasrl_service.py
```diff
@@ -63,24 +79,27 @@ def setup_controller(self, request):
...
+        self.logger.info("Suggestion for StudyJob {} (ID: {}) has been initialized.\n".format(self.study_names[studyID], studyID))
 
     def GetSuggestions(self, request, context):
+        studyID = request.study_id
+
         if request.study_id not in self.registered_studies:
             self.setup_controller(request)
-            self.is_first_run = True
-            self.registered_studies.append(request.study_id)
+            self.is_first_run[studyID] = True
```
hougangliu: all self.dict will always increase, and nasrl-suggestion service will consume all memory (I think before that OOM killer will kill the pod)

andreyvelich: @hougangliu Is it possible to get this value: https://github.com/kubeflow/katib/blob/master/pkg/api/operators/apis/studyjob/v1alpha1/studyjob_types.go#L34 
inside Suggestion, right now?

DeeperMind: RequestCount is the only information to determine whether a job has ended or not. If we cannot get that, we could not delete the StudyJob information.

hougangliu: In fact, a studyjob may stop calling GetSuggestion from suggestion service when: 1. changed to `failed`; 2. changed to `completed` (exceed RequestCount or exceed goal).
I think we had to add a new api like `StopSuggestion`  in above two conditions to make suggestionService to clear something. @YujiOshima @richardsliu @johnugeorge WDYT?

pkg/suggestion/nasrl_service.py
DeeperMind: I have reorganized the program a little bit. Now, the information of every StudyJob is stored in a `NAS_RL_StudyJob` class. And the initialization of a StudyJob and retrieving parameters are all handled by this class. In this way, GetSuggestion gets much cleaner.

pkg/suggestion/nasrl_service.py
andreyvelich: @hougangliu @YujiOshima Should we ignore this parameter at this time? 
It will work, but we can't clean dictionary with studyIDs and information about it.
After we add RequestCount parameter to the API, we can modify this functionality.

hougangliu: Please trace cleaning unused controller by an issue. thanks!

PR #405, opened 2019-02-22
Metric collector must fail on error

johnugeorge:
Instead of returning, metric collector must fail if unable to collect logs.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/405)


YujiOshima:
@johnugeorge Please set `backofflimit` 0 in the MetricsCollectorTemplate.
MetricsCollector doesn't need to restart since it will run every minute.

johnugeorge:
@YujiOshima updated

Issue #230, opened 2018-10-29
Test for Bayesian Optimization Algo


garganubhav:
I am working on this

jlewi:
@garganubhav What is the current status of this? Can you provide more information about what you are planning to deliver and when? 

For example, will the test just verify that the bayesian algorithm generates suggestions without errors, or will it also check correctness?

garganubhav:
Currently, I am not working on this. I started but then the BO algo was not working in Katib.

YujiOshima:
@garganubhav What is the current status of this? BO can work now.

garganubhav:
@YujiOshima I don't plan to work on this in near future.

jdplatt:
Can I get a few more details on what is required here? I'm not sure if you are looking for a set of unit tests to make sure the BO code runs correctly, or more of a benchmark on a standard ML tasks to compare the performance of BO with something simpler like random search. Either way I am happy to help out.

johnugeorge:
Thanks @jdplatt   This issue is to track unit tests for BO algo. See unit tests for other algos https://github.com/kubeflow/katib/tree/master/pkg/suggestion

jdplatt:
I took a look at the code and have a few questions:

1. Do you have any existing unit tests for python code in the repo? I couldn’t find any but wanted to double check. If there are I’ll make sure to follow the existing workflow.
2. I have a few questions about the structure (e.g. why do each of the other algorithms get their own service while the BO service contains multiple algorithms) and what different parameters mean. Where is the best place to get answers to these sort of questions? I’ll try to add some docs to the code along side the tests.

YujiOshima:
@jdplatt Thank you!
Unfortunately, we don't have a test with python in the repo.  We want a function level unit test for BO. Please use some libraries like  `unittest`.
The algorithm level test will be defined [here](https://github.com/kubeflow/katib/blob/master/test/e2e/test-client.go). This code will call API and test end to end algorithms. It is an advanced task.

>  why do each of the other algorithms get their own service while the BO service contains multiple algorithms

You are right. It would be better to split them. 
> I’ll try to add some docs to the code along side the tests.

Great! Please make a new issue for your questions.

jlewi:
@YujiOshima any update on this?

jdplatt:
@jlewi This got merged in #406

YujiOshima:
@jdplatt Thanks!
/close

PR #406, opened 2019-02-22
Test for Bayesian Optimization Algo

jdplatt:
#230 

Here is a PR containing tests for the code in pkg/suggestion/bayesianoptimization. You can run them using test/scripts/python_tests.sh. I have close to full coverage but didn't cover the parts of the code that deal with the beginning of tuning when there is no data. I think there is a cleaner way to handle this using a different suggestion algorithm such as random search. I fixed a few bugs in the way the models were trained alongside writing the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/406)


YujiOshima:
@jdplatt Thank you for PR! Could you sign the CLA?

jdplatt:
I signed it!

jdplatt:
@YujiOshima 

Had a few issues getting corporate CLA to work but should be all good now.

YujiOshima:
Thanks, I will take a look!

test/scripts/python_tests.sh
```diff
@@ -0,0 +1,5 @@
+pip install -r cmd/suggestion/bayesianoptimization/requirements.txt
```
YujiOshima: You should add this script to https://github.com/kubeflow/katib/blob/master/test/scripts/unit-test.sh or add a new workflow for python test to argo.

pkg/suggestion/bayesianoptimization/src/algorithm_manager.py
```diff
@@ -1,39 +1,31 @@
...
     diff = np.absolute(diff)
     return feasible_values[np.argmin(diff)]
 
+
 def deal_with_categorical(feasible_values, one_hot_values):
     """ function to do the one hot encoding of the categorical values """
-    #index = np.argmax(one_hot_values)
-    index = one_hot_values.argmax()
+    index = np.argmax(one_hot_values)
+    #index = one_hot_values.argmax()
```
YujiOshima: Why did you change this?

jdplatt: The two lines perform the same task. However, the np.argmax function works on any array_like object (e.g. list, tuple, or ndarray), while the ndarray.argmax() method only works on ndarrays. I switched for two reasons, i) the old version raised an exception because it was being passed a list at some point, ii) the function approach is more flexible and requires less type casts to force array_like variables into ndarrays.

test/scripts/python_tests.sh
jdplatt: Thanks for catching this! I added it at the end of unit-test.sh

jdplatt:
Thanks for looking at my PR. I fixed the two issues you raised. Do you need me to do anything else?

YujiOshima:
@jdplatt Thanks!
/ok-to-test

pkg/suggestion/tests/test_boa.py
```diff
@@ -0,0 +1,23 @@
+import numpy as np
+
+from ..bayesianoptimization.src.bayesian_optimization_algorithm import BOAlgorithm
+
+
+def test_boa(dim, request_num, lower_bounds, upper_bounds, X_train, y_train):
```
richardsliu: General comment (for all test files): 
For all the tests, add a brief comment section that explains what this test does, what are the expectations, etc. This makes it easier for other contributors to follow.

Issue #397, opened 2019-02-20
Add time to Trial API


andreyvelich:
After this PR https://github.com/kubeflow/katib/pull/394 we will store create time in the Trial table.
We should modify Katib API to be able to get this time through gRPC server. 
See comment: https://github.com/kubeflow/katib/pull/328#discussion_r257864354
/assign
/area nas

PR #410, opened 2019-02-26
Add create time to Trial API

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/397.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/410)


pkg/api/api.proto
```diff
@@ -366,6 +366,7 @@ message Trial {
     repeated Parameter parameter_set = 3; /// Hyperparameter set
     string objective_value = 4; /// Objective Value
     repeated Tag tags = 5; /// Tags of Trial.
+    string create_time = 6; /// Trial create timestamp RFC3339 format.
```
YujiOshima: Please explain that the create_time will be filled in the server side even user set the value.

andreyvelich: Done

Issue #166, opened 2018-09-13
grid-example failed


AnberLu:
i canot  see metrics in UI; and the log is like ： 
```
2018/09/13 06:44:08 Fail to check status Suggestion count mismatched. May be duplicate suggestion request
2018/09/13 06:44:09 Fail to Update StudyJob c1e236530496ef99 : Operation cannot be fulfilled on studyjobs.kubeflow.org "grid-example": the object has been modified; please apply your changes to the latest version and try again
ERROR: 2018/09/13 06:44:10 pickfirstBalancer: failed to NewSubConn: rpc error: code = Canceled desc = grpc: the client connection is closing
ERROR: 2018/09/13 07:04:00 pickfirstBalancer: failed to NewSubConn: rpc error: code = Canceled desc = grpc: the client connection is closing
```

jlewi:
@YujiOshima do we have any tests or other verification to ensure examples are working?

/cc @inc0 @gaocegege

YujiOshima:
I think this problem is fixed in latest version.
But we need test for algorithms ( #42 )
/close

PR #411, opened 2019-02-26
Retry studyjobcontroller

toshiiw:
This hopefully fixes Update race condition from Reconcile.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/411)

### DIFF
--- a/pkg/controller/studyjob/studyjob_controller.go
+++ b/pkg/controller/studyjob/studyjob_controller.go
@@ -212,7 +212,7 @@ func (r *ReconcileStudyJobController) Reconcile(request reconcile.Request) (reco
 		}
 		log.Printf("Fail to read Object %v", err)
 		// Error reading the object - requeue the request.
-		return reconcile.Result{}, err
+		return reconcile.Result{Requeue: true}, err
 	}
 	var update bool = false
 	deleted := instance.GetDeletionTimestamp() != nil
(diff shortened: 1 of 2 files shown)


pkg/controller/studyjob/studyjob_controller.go
```diff
@@ -240,44 +240,66 @@ func (r *ReconcileStudyJobController) Reconcile(request reconcile.Request) (reco
...
-		log.Printf("Fail to check status %v", err)
-		return reconcile.Result{}, err
-	}
-	if update {
-		err = r.Update(context.TODO(), instance)
-		if err != nil {
+	ret := reconcile.Result{}
+	var updateErr error
+	if update || err != nil {
+		updateErr := r.Update(context.TODO(), instance)
```
johnugeorge: What is the root cause of the race condition?

toshiiw: I heard that some jobs can finish meanwhile and the StudyJob get updated.
I've never seen the actual failure though.

YujiOshima: Like this https://github.com/kubeflow/katib/issues/166#issue-359780752
`2018/09/13 06:44:09 Fail to Update StudyJob c1e236530496ef99 : Operation cannot be fulfilled on studyjobs.kubeflow.org "grid-example": the object has been modified; please apply your changes to the latest version and try again`

When any updates worker or metrics-collector happen during reconcile is running, the update for studyjob will be failed.

johnugeorge:
@YujiOshima I was asking if we know the reason behind the error. Else, this error will be hidden with this fix.

YujiOshima:
I think we can't know the exactly reason.
But because of mutex I believe the reasons of this error are only completion of Worker, creation of job from MetricsCollector cronjob, and Fail of Worker.
Since these updates are not controled by studyjob operator.
And they can handle at the next reconcile.

YujiOshima:
The  reason why we need this is avoiding endless retry.
When we use a heavy suggestion service which take longer than one minute for creating a suggestion, the metrics collector must create a new job during  studyjob operator wait for the suggestion.
Then the studyjob will fail to update and retry.
The studyjob won't update any more.

johnugeorge:
> I think we can't know the exactly reason.
> But because of mutex I believe the reasons of this error are only completion of Worker, creation of job from MetricsCollector cronjob, and Fail of Worker.

@YujiOshima Can you please explain more on this situation?  Why will  this error happen during these conditions?  Error says that the object got modified before the update call.

> Since these updates are not controled by studyjob operator.
> And they can handle at the next reconcile.

YujiOshima:
@johnugeorge When any updates on a resource that is corresponding to a studyjob, the operator can't update the studyjob.

A studyjob named "A" has a worker "A-w1" and a metrics collector cron-job "A-mc1".
First, The operator get `instance` by `r.Get` [here](https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L202)
And the operator waiting for some process (e.g. GetSuggestion).
During the operator waiting, "A-mc1" create a job. Since "A-mc1" is a cron-job, it will create a job every minute.
At this time, the studyjob CR is modified from the operator getted.
So the operator get error when call `r.Update` [here](https://github.com/kubeflow/katib/blob/master/pkg/controller/studyjob/studyjob_controller.go#L267)

johnugeorge:
`At this time, the studyjob CR is modified from the operator getted.`
@YujiOshima  In the above example, i didn't understand how studyJobCR is getting modified after mc job is created.  Can you point out to the code?  AFAIK, error happens when update is called on a object that already got modified  from the initial Get

toshiiw:
/hold cancel
Seems to work in my local env.

johnugeorge:
I am doubting this https://github.com/kubernetes-sigs/controller-runtime/issues/245.

Need to confirm if this is the issue in our case

johnugeorge:
Related: https://github.com/kubernetes-sigs/controller-runtime/issues/343

 Since we are watching on job pods, there are higher chances of Reconcile functions called continuously. Immediate Get after Update can have stale entries.  This will cause the next Update to fail. However, this should not be a problem since we can handle this during the next Reconcile.

YujiOshima:
@johnugeorge Thanks. So we should not believe values in CR, believe values in DB.
I think it would be a problem for currently Katib. For example, when fail to update after getSuggestion, the suggestionCount will be inconsistent between DB and CR.
Should we fix that bug and other related bugs now or think about this after v2API? WDYT?

toshiiw:
> For example, when fail to update after getSuggestion, the suggestionCount will be inconsistent between DB and CR.

This patch treats the suggestionCount in the DB as the single source of the truth. Inconsistency with this should be resolved in the next reconcile run.

YujiOshima:
@toshiiw When the operator get  stale version of CR, operator won't set false to nextSuggestionSchedule  [here](https://github.com/kubeflow/katib/blob/51948ea4339c364a8e173704e0af9cd82956a82e/pkg/controller/studyjob/studyjob_controller.go#L547) even the GetSuggestion is alrealy called in previous Reconcile.
Then GetSuggestion is called duplicatedly and create workers but update will be fail. So the workers will be orphan.

johnugeorge:
It is not accurate to say that values in CR should not be believed.  From the k8s philosophy, We have to make sure that design is level triggered based than edge triggered based.  If it is a level triggered design, reconcile function will make sure that actual state moves towards the desired state.  

i think, we should think after v2API because there will be significant code changes. We have to ensure this.

toshiiw:
@johnugeorge Agreed. The current Reconcile() is written somewhat in a edge-triggered way. We might need to put some worker related info in CR or DB to ensure a level-triggered design works.

gaocegege:
/close

Thanks for the PR. But we already dropped v1alpha1. We can't merge it since v1alpha1 is deprecated.

Issue #396, opened 2019-02-20
Add support for multiple trials


DeeperMind:
Currently the reinforcement learning suggestion only spawns one trails each time. It is necessary to integrate the support for multiple trials so that the training can be parallelized. 

/assign @DeeperMind
/area nas

PR #416, opened 2019-03-02
Multiple Trials for Reinforcement Learning Suggestion

DeeperMind:
Now the reinforcement learning suggestion supports spawning multiple trials in each iteration. Users can activate this function by specifying the `num_trials` suggestion parameter in StudyJob yaml file. 

Please notice that this multiple trials support is for extending exploration only, not for asynchronous update. The suggestion will use the average metrics of all the trials to calculate the policy gradient, which provides a more justified evaluation of the LSTM cell's internal state compared to using single trial. And the suggestion will not generate new candidates until all the previous ones are finished.

Fixes https://github.com/kubeflow/katib/issues/396

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/416)


pkg/suggestion/nasrl_service.py
```diff
@@ -33,11 +35,13 @@ def __init__(self, request, logger):
         self.search_space = None
         self.opt_direction = None
         self.objective_name = None
+        self.num_trials = 1
```
hougangliu: api.GetSuggestionsRequest.request_number has defined num_trials, you need not add a new one in  suggestionParameters
Thanks!

andreyvelich: Thank you for notice! We will change it.

DeeperMind: Fixed. I removed num_trials from the suggestion parameters and used requestNumber instead.

DeeperMind:
retest

PR #417, opened 2019-03-04
Print verbose error when studyjob updates fail

toshiiw:
This is to test the scenario behind kubeflow/katib/pull/411

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/417)

### DIFF
--- a/pkg/controller/studyjob/studyjob_controller.go
+++ b/pkg/controller/studyjob/studyjob_controller.go
@@ -17,6 +17,7 @@ package studyjob
 
 import (
 	"context"
+	"encoding/json"
 	"log"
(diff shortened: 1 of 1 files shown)


toshiiw:
I realized studyjobcontroller logs should be collected, if we are going to test in the CI.

gaocegege:
/close

Thanks for the PR. But we already dropped v1alpha1. We can't merge it since v1alpha1 is deprecated.

gaocegege:
/close

Thanks for the PR. But we already dropped v1alpha1. We can't merge it since v1alpha1 is deprecated.

Issue #423, opened 2019-03-08
gRpc API v2


YujiOshima:
Now we merged v2 CRD API https://github.com/kubeflow/katib/pull/381
We need a new version of gRpc API which comples with the CRD.

richardsliu:
Also need to create a new version of database schema.

jlewi:
@richardsliu @YujiOshima @johnugeorge It looks like the PR is still in WIP. Since we are planning on cutting tomorrow I'm going to say this isn't going to make 0.5 so I'm moving it to 0.6.

johnugeorge:
Closing this PR as GRPC PR is merged
/close

PR #427, opened 2019-03-12
add v1alpha2 grpc api

YujiOshima:
#423 
I added v1alpha2 gRPC API complies with the v1alpha2 CRD.
https://github.com/kubeflow/katib/pull/427/files#diff-c0f6760a929f1056527071edf1f57db8

I will add a new DB schema soon.

@richardsliu @johnugeorge @hougangliu @gaocegege 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/427)


pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,461 @@
...
+ */
+ message Operation {
+     string operationType = 1; /// Type of operation in DAG
+      /** 
+     * List of ParameterAssignment
+     */
+     message ParameterAssignments {
+        repeated ParameterAssignment assignments = 1;
+    }
+    ParameterAssignments parameter_assignments = 2; /// List of ParameterAssignments
```
andreyvelich: I think, here we should define `ParameterSpec` instead of `ParameterAssignments`

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,461 @@
...
+     */
+     rpc ValidateAlgorithmSettings(ValidateAlgorithmSettingsRequest) returns (ValidateAlgorithmSettingsReply){
+        option (google.api.http) = {
+            post: "/api/Manager/ValidateAlgorithmSettings"
+            body: "*"
+        };
+    };
+}
+
+service Suggestion {
```
andreyvelich: Should we still name this service `Suggestion` or move to `Algorithm` ?

pkg/api/v1alpha2/api.proto
YujiOshima: Is `ParameterSpec` you said the same as this https://github.com/kubeflow/katib/pull/427/files/2ed4f1f3eedaa0048441676d849ac44cc3426ea8#diff-c0f6760a929f1056527071edf1f57db8R184 ?

pkg/api/v1alpha2/api.proto
YujiOshima: I think we should remain the name `Suggestion Service`. 
The `Algorithm` means the algorithm for generating suggestions.
WDYT @richardsliu?

YujiOshima:
I added a new version of DB IF
https://github.com/kubeflow/katib/pull/427/commits/1a6d4ca4f9e39123dcf5280909aae2a5603721a2

pkg/api/v1alpha2/api.proto
richardsliu: Suggestion makes more sense. Algorithm is how suggestions are generated.

pkg/api/v1alpha2/api.proto
andreyvelich: @YujiOshima Yes, that's right.

richardsliu: Assignment seems correct here, ParameterSpec is for the feasible space.

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,479 @@
...
+ * Spec of a Experiment. Experiment represents a single optimization run over a feasible space. 
+ * Each Experiment contains a configuration describing the feasible space, as well as a set of Trials.
+ * It is assumed that objective function f(x) does not change in the course of a Experiment.
+ */
+message ExperimentSpec {
+    repeated ParameterSpec parameters = 1; 
+    ObjectiveSpec objective = 2; 
+    AlgorithmSpec algorithm = 3;
+    string trial_template = 4;
+    int32 pallarel_trial_count = 5;
```
richardsliu: Typo - parallel

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,479 @@
...
+    string name = 1;
+    string value = 2;
+}
+
+message Metric {
+    string name = 1;
+    string value = 2;
+}
+
+message MetricLog {
```
richardsliu: What is this used for?

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,479 @@
...
+    }
+    string name = 1;
+    repeated ValueLog valuelog = 2;
+}
+
+message Observation {
+    repeated Metric metrics = 1;
+}
+
+message ObservationLog {
```
richardsliu: Document what this is used for.

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,479 @@
...
+message ExperimentSummary {
+    string experiment_name = 1;
+    ExperimentStatus.ExperimentConditionType condition = 4;
+}
+
+message GetExperimentListRequest {
+}
+
+message GetExperimentListReply {
+    repeated ExperimentSummary experiment_summarys = 1;
```
richardsliu: summaries

pkg/db/v1alpha2/db_init.go
```diff
@@ -0,0 +1,65 @@
...
+	db := d.db
+
+	_, err := db.Exec(`CREATE TABLE IF NOT EXISTS experiments
+		(id INT AUTO_INCREMENT PRIMARY KEY,
+		name VARCHAR(255),
+		parameters TEXT,
+		objective TEXT,
+		algorithm TEXT,
+		trial_template TEXT,
+		pallarel_trial_count INT,
```
richardsliu: Parallel

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,479 @@
...
+    }
+    Operations operations = 2; /// List of Operation
+}
+
+/**
+ * GraphConfig contains a config of DAG
+ */
+ message GraphConfig {
+    int32 num_layers = 1; /// Number of layers
+    repeated int32 input_size = 2; ///  Dimenstions of input size
```
andreyvelich: Dimenstions -> Dimensions

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,479 @@
...
+message GetExperimentReply {
+    Experiment experiment = 1;
+}
+
+message GetStudyListRequest {
+}
+
+message ExperimentSummary {
+    string experiment_name = 1;
+    ExperimentStatus.ExperimentConditionType condition = 4;
```
andreyvelich: Need to change 4 to 2 ?

pkg/db/v1alpha2/db_init.go
```diff
@@ -0,0 +1,65 @@
...
+		max_trial_count INT,
+		condition TINYINT,
+		start_time DATETIME(6),
+		metrics_collector_type TEXT,
+		completion_time DATETIME(6),
+		last_reconcile_time DATETIME(6))`)
```
andreyvelich: Will we add NasConfig in `experiments` table in the future?

YujiOshima:
@richardsliu @andreyvelich Thanks! I apply your comments.

pkg/api/v1alpha2/api.proto
YujiOshima: I added doc here https://github.com/kubeflow/katib/pull/427/commits/2e3444d2fda2b69c546538ece0f36393a8018ae4#diff-c0f6760a929f1056527071edf1f57db8R117
Katib will store every log of metrics. Users can look at the history of training. And the logs are used by suggestion algorithm and early stopping.

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+    UNKNOWN = 0; /// Undefined type and not used.
+    MINIMIZE = 1; /// Minimize
+    MAXIMIZE = 2; /// Maximize
+}
+
+message ObjectiveSpec{
+    ObjectiveType type = 1;
+    float goal = 2;
+    string objective_metric_name = 3;
+    repeated string additional_metrics_name = 4;
```
johnugeorge: additional_metrics_names?

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+    string value = 2;
+}
+
+message MetricLog {
+    message ValueLog {
+        string time_stamp = 1; /// RFC3339 format
+        string value = 2;
+    }
+    string name = 1;
+    repeated ValueLog valuelog = 2;
```
johnugeorge: why would be it list of [timestamp-value] pair for a particular key?   Isn't it, list of [key-value] pair for a particular timestamp?

johnugeorge: does the name correspond to trail name?

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+
+message GetExperimentRequest {
+    string experiment_name = 1;
+}
+
+message GetExperimentReply {
+    Experiment experiment = 1;
+}
+
+message GetStudyListRequest {
```
johnugeorge: what is this used for?

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+message DeleteTrialRequest {
+    string trial_id = 1;
+}
+
+message DeleteTrialReply {
+}
+
+message GetTrialListRequest {
+    string experiment_name = 1;
+    string start_time = 2; ///The start of the time range. RFC3339 format
```
johnugeorge: are these filter options?

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+}
+
+message GetObservationLogReply {
+    ObservationLog observation_log = 1;
+}
+
+message GetSuggestionsRequest {
+    string experiment_name = 1;
+    string algorithm_name = 2;
+    int32 request_number = 3;
```
johnugeorge: what is this field used for?

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+message UpdateExperimentStatusReply {
+    string experiment_name = 1;
+}
+
+message RegisterTrialRequest {
+    Trial trial = 1;
+}
+
+message RegisterTrialReply {
+    string trial_id = 1;
```
johnugeorge: DB Interface can have multiple storage backends. Can we safely assume that every storage backend(eg: ModelDB) will return unique triail_id?   Which is better? Controller creating name and call RegisterTrial   or Controller calling RegisterTrial and DB returning new trial_id?

@achalshant @richardsliu

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+message GetExperimentReply {
+    Experiment experiment = 1;
+}
+
+message GetStudyListRequest {
+}
+
+message ExperimentSummary {
+    string experiment_name = 1;
+    ExperimentStatus.ExperimentConditionType condition = 2;
```
johnugeorge: Should we add more fields as in ExperimentStatus https://github.com/kubeflow/katib/blob/master/pkg/api/operators/apis/experiment/v1alpha2/types.go#L75 
 OptimalTrial ,TrialsCompleted,TrialsFailed ,TrialsKilled ,TrialsPending

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+    enum TrialConditionType {
+        PENDING = 0;
+        RUNNING = 1;
+        COMPLETED = 2;
+        KILLED = 3;
+        FAILED = 4;
+    }
+    string start_time = 1; /// RFC3339 format
+    string completion_time = 2; /// RFC3339 format
+    string last_reconcile_time = 3; /// RFC3339 format
```
johnugeorge: Do we need last_reconcile_time in DB?

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+    enum ExperimentConditionType {
+        CREATED = 0;
+        RUNNING = 1;
+        RESTARTING = 2;
+        SUCCEEDED = 3;
+        FAILED = 4;
+    }
+    string start_time = 1; /// RFC3339 format
+    string completion_time = 2; /// RFC3339 format
+    string last_reconcile_time = 3; /// RFC3339 format
```
johnugeorge: Do we need last_reconcile_time in DB?

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+    TrialConditionType condition = 4;
+    Observation observation = 5; /// The best observation in logs.
+}
+
+message Trial {
+    string id = 1;
```
johnugeorge: Is this ID or name(like in experiment)?

pkg/api/v1alpha2/api.proto
johnugeorge: DB interface can have multiple storage backend(Mysql, ModelDb etc). Is the controller creating Id?  What is the return trial_id

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+    rpc GetExperimentList(GetExperimentListRequest) returns (GetExperimentListReply){
+        option (google.api.http) = {
+            get: "/api/Manager/GetExperimentList"
+        };
+    };
+
+    /** 
+     * Update Status of a experiment.
+     */
+    rpc UpdateExperimentStatus(UpdateExperimentStatusRequest) returns (UpdateExperimentStatusReply){
```
johnugeorge: Currently, Suggestion algorithms can modify the algorithm parameters. https://github.com/kubeflow/katib/blob/master/pkg/suggestion/grid_service.go#L195 

How is this handled?

pkg/api/v1alpha2/api.proto
YujiOshima: Since the trials are generated by the controller, users won't name them.
I think `ID` is more suitable here.

pkg/api/v1alpha2/api.proto
YujiOshima: My assumption, it is created by DB backend. 
But it is OK the controller create the ID. It will probably be more consistent behavior.

pkg/api/v1alpha2/api.proto
YujiOshima: I will create UpdateAlgorithmParameter API.

pkg/api/v1alpha2/api.proto
johnugeorge: I am not sure if every storage backend supports id creation.

pkg/api/v1alpha2/api.proto
johnugeorge: It is true. But, I feel that it might be better to keep the same terminology as in the Kubernetes config(matadata.name)

YujiOshima: make sense. "name" would be more user friendly.

pkg/api/v1alpha2/api.proto
YujiOshima: OK. let's make the name of trial is generated by controller.

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+    int32 num_layers = 1; /// Number of layers
+    repeated int32 input_size = 2; /// Dimensions of input size
+    repeated int32 output_size = 3; /// Dimensions of output size
+}
+
+/**
+ * Config for operations in DAG
+ */
+ message Operation {
+     string operationType = 1; /// Type of operation in DAG
```
hougangliu: **s/operationType/operation_type** to keep same name style as others

pkg/api/v1alpha2/api.proto
hougangliu: s/valuelog/value_log

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,481 @@
...
+    string name = 1;
+    repeated ValueLog valuelog = 2;
+}
+
+message Observation {
+    repeated Metric metrics = 1;
+}
+
+message ObservationLog {
+    repeated MetricLog metriclogs = 1;
```
hougangliu: s/metriclogs/metric_logs

pkg/api/v1alpha2/api.proto
andreyvelich: @richardsliu @YujiOshima We need feasible space inside Operation as well.

pkg/api/v1alpha2/api.proto
YujiOshima: When you set 3 to `request_number`, you can get three Suggestions at one time.

YujiOshima:
@johnugeorge @hougangliu Thank you for your comments. I updated.
 PTAL.

pkg/db/v1alpha2/db_init.go
```diff
@@ -43,10 +44,10 @@ func (d *dbConn) DBInit() {
 	}
 
 	_, err = db.Exec(`CREATE TABLE IF NOT EXISTS observation_logs
-		(trial_id INT NOT NULL,
+		(trial_name INT NOT NULL,
 		id INT AUTO_INCREMENT PRIMARY KEY,
 		time DATETIME(6),
-		name VARCHAR(255),
+		metric_name VARCHAR(255),
 		value TEXT,
```
johnugeorge: is this metric_value field?

johnugeorge:
@YujiOshima  I see that new API UpdateAlgorithmSpec is added.  How is experiment Spec update handled in general?  Say If user updates fields like MaxTrials, ParallelTrials  etc  in K8s config , how is DB updated?

YujiOshima:
@johnugeorge You have a point.
The problem is how to sync the AlgorithmSpec between k8s CR and DB.
One idea, the AlgorithmSpec would be an initial config of an algorithm for the experiment and it will not change.
And adding a new table `AlgorithmStatus` to DB. SuggestionService will save the status of the algorithm with  UpdateAlgorithmStatus to DB.
WDYT?

johnugeorge:
@YujiOshima How would `AlgorithmStatus` structure look ? Is it a key-value pair(similar to AlgorithmSetting)  that would be only configured by Algorithms internally?

YujiOshima:
@johnugeorge Yes. Since the parameter that each algorithm need is different, the AlgorithmStatus should have a general struct like key-value pair.

> that would be only configured by Algorithms internally?

Yes.

pkg/api/v1alpha2/api.proto
johnugeorge: Isn't this same as the `parallel_trial_count` which is already in the experiment Spec ?

johnugeorge:
@YujiOshima 
Few thoughts

1.  What does `status` mean in `AlgorithmStatus' call? I think, term is slightly confusing.  

2. From what i understood,  this rpc call can be used to set extra key-value pairs (Algorithm specific parameters, in addition to the ones set by user in ExperimentSpec). Is it right?

3 . Do we need UpdateAlgorithmSpec call in such a case?

YujiOshima:
@johnugeorge Sorry for confusing you.

> From what i understood, this rpc call can be used to set extra key-value pairs (Algorithm specific parameters, in addition to the ones set by user in ExperimentSpec). Is it right?

Right.
We need to store the extra key-value pairs to DB.
As you pointed, we should not use `UpdateAlgorithmSpec` in this case (I will delete this API since I think we don't need it now).
I called The extra parameters `AlgorithmStatus` in the previous comment.
My idea is the extra parameters are handled independently from AlgorithmSpec of ExperimentSpec.
When a Suggestion service calls `UpdateAlgorithmStatus` gRpc API to save the parameters, the Katib-manager save the parameter to DB and it doesn't any effect on Experiments CR. 

make sense?

pkg/api/v1alpha2/api.proto
YujiOshima: a little bit different.
When parallel_trial_count is 4 and 2 trials are completed at the same time, the operator needs to set 2 to request_number in GetSuggesitonRequest.

johnugeorge: Makes sense

johnugeorge:
> @johnugeorge Sorry for confusing you.
> 
> > From what i understood, this rpc call can be used to set extra key-value pairs (Algorithm specific parameters, in addition to the ones set by user in ExperimentSpec). Is it right?
> 
> Right.
> We need to store the extra key-value pairs to DB.
> As you pointed, we should not use `UpdateAlgorithmSpec` in this case (I will delete this API since I think we don't need it now).
> I called The extra parameters `AlgorithmStatus` in the previous comment.
> My idea is the extra parameters are handled independently from AlgorithmSpec of ExperimentSpec.
> When a Suggestion service calls `UpdateAlgorithmStatus` gRpc API to save the parameters, the Katib-manager save the parameter to DB and it doesn't any effect on Experiments CR.
> 
> make sense?

Yes. Should we rename the api `UpdateAlgorithmStatus` to a better one?  Status looks misleading since we are just updating algorithm specific parameters

pkg/api/v1alpha2/api.proto
```diff
@@ -0,0 +1,487 @@
...
+
+/**
+ * Config for operations in DAG
+ */
+ message Operation {
+     string operation_type = 1; /// Type of operation in DAG
+      /** 
+     * List of ParameterAssignment
+     */
+     message ParameterAssignments {
```
andreyvelich: @YujiOshima @richardsliu I think we need `ParameterSpec` instead of `ParameterAssignments`, that I told you. We need Feasible space inside ParameterConfigs in Operations.

YujiOshima: OK, I will update.

andreyvelich: Thank you!

YujiOshima:
@johnugeorge How about
* AlgorithmVariables
* AdditionalAlgorithmSetting

YujiOshima:
@johnugeorge @andreyvelich  @richardsliu @hougangliu I think I applied all comments. PTAL.

Issue #407, opened 2019-02-23
Move all suggestion algorithms to python


jdplatt:
What do you think of moving all of the code in pkg/suggestion used to generate suggestions to python? Right now three of the algorithms (grid, random, hyperband) are in go while two (rf + gp) are in python. I think the advantages of a single language would be:

1. It would open up the option of using combinations of the algorithms. For example, with the the gp algorithm it is common to use something like random search for the burn-in phase at the start of an experiment when there is not enough data to train an accurate model. GPs also don't scale well to a high number of observations, which is why the vizier paper transitioned to a local search algorithm for high N. Finally, while the original hyperband paper uses random search to generate suggestions for each bracket, they highlight the algorithm can be combined with other suggestion algorithms such as a latin hypercube or ones such as GP that adapt as they see more examples.

2. We only need to manage a single service rather than one for each of the go algorithms and a final one for all of the python algorithms. Having one service per algorithm will make it hard to scale to a wide range of algorithms. It will also probably be a little easier to develop + test if it's all the in the same language.

I'm happy to take the lead on this if you guys think it makes sense. Grid search and random search shouldn't take too long though hyperband may be a bigger lift.

@YujiOshima @richardsliu

YujiOshima:
Thanks @jdplatt. 

I don’t want to merge suggestion services because of robustoness for whole system.
In currently archtecture, a failure of a service won't affect other services.
And user can add thier custom algorithm service without restarting any services.

But we have big motivation to move python. We can reuse code in several algorithm and enrich algorithm with many liblaries in python.

jdplatt:
That makes sense. I'll aim to migrate the suggestion code to python but make sure it's possible to run each of the suggestion algorithms in a separate service.

richardsliu:
Makes sense to me.
@gaocegege @johnugeorge

johnugeorge:
@YujiOshima What failures do you anticipate? 

With the current design, I see the advantage of adding custom algorithm service without restarting the service. However, I agree with @jdplatt Point 2.  I feel that It is not a scalable solution to add wide range of algorithms. Eg: In https://github.com/tobegit3hub/advisor, 17 algorithms are listed which would need 17 pods in our case.  What do you think?

jdplatt:
[...]

PR #435, opened 2019-03-16
Python migration

jdplatt:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/435)


jdplatt:
Now there are multiple algorithms I decided to refactor the code in places. The main goals were:

1. Move the bayesian_service.py file inside the main package folder so it’s easy to test.
2. Break all the parsing routines in algorithm manager into functions so they can be used independently (e.g. not all algorithms require us to parse the past observations).
3. Push all algorithm specific code (e.g. parsing or setting default parameters) inside the relevant algorithms classes.
4. Move from a long list of function arguments (e.g. when creating an algorithm) to a few objects that mirror similar concepts in the api (e.g. a parameter search space, the algorithm spec).
5. Run both grid, random, and BO services from a single dockerfile.

I was a little confused about the format of the SuggestionSpec for grid search. I looked at the file in the examples folder but was wondering if there is another example somewhere? If not can someone clarify what the structure of the SuggestionSpec for grid search is?

jdplatt:
This addresses #407 .

YujiOshima:
@jdplatt Thank you for your hard work!
We need to rename the suggestion image. Currently, every suggestion is `bayesianoptimization`.
And we are discussing new API https://github.com/kubeflow/katib/pull/427
It affects on suggestion services.
You don't need to change the logic of algorithm but need to modify API.
I think it would be better to pend this until new API is merged.
@jdplatt @johnugeorge WDTY?

jdplatt:
Thanks @YujiOshima! I agree with your plan. How about I rename bayesianoptimization to katib-suggestion? I'll rename the python package at the same time. Do we have a timeline for the new API? It shouldn't take me too long to get update the suggestion code because all contact with the APIs is limited to a few functions.

johnugeorge:
i feel that we should wait for v1alpha2 as manifests have changed.  We might need to create separate directories for v1alpha2 in order to support both versions. 

@jdplatt we expect to merge GRPC API by end of this week. Once GRPC API is merged, I will try to get initial version of v1alpha2 version in 1-2 weeks.   Can you please hold this PR till then and then make changes for v1alpha2 directly? It will be much easier
   Btwn, I see that `grid_search` and `random_search` python files were moved into `pkg/suggestion/bayesianoptimization` Should we keep them in separate directories?

scripts/deploy.sh
```diff
@@ -31,7 +31,7 @@ kubectl apply -f manifests/vizier/ui
 kubectl apply -f manifests/vizier/suggestion/random
 kubectl apply -f manifests/vizier/suggestion/grid
 kubectl apply -f manifests/vizier/suggestion/hyperband
-kubectl apply -f manifests/vizier/suggestion/bayesianoptimization
+kubectl apply -f manifests/vizier/suggestion/katib_suggestion
```
andreyvelich: Why do we rename folder only for BayesianOptimization?

test/e2e/suggestion-config-grid.yml
```diff
@@ -1,9 +1,6 @@
...
-        name: "DefaultGrid"
-        value: "1"
-    - 
-        name: "learning-rate"
-        value: "4"
-
+  - name: "DefaultGrid"
+    value: "1"
+  - name: "learning-rate"
+    value: "4"
```
andreyvelich: Please, add newline.

andreyvelich:
@jdplatt Thank you for your work! 
I left few comments.

pkg/suggestion/katib_suggestion/algorithm_register.py
```diff
@@ -0,0 +1,8 @@
+from .random_search import RandomSearch
+from .grid_search import GridSearch
+from .bayesian_optimization_algorithm import BOAlgorithm
+
+
+ALGORITHM_REGISTER = {"random_search": RandomSearch,
+                      "grid_search": GridSearch,
+                      "bayesian_optimization": BOAlgorithm}
```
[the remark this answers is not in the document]
jdplatt: I am, but was worried this PR had already got really large. I plan to do it soon after this is merged.

scripts/deploy.sh
jdplatt: Now that multiple algorithms (some of which aren't BO) sit inside a single image we start renaming bayesianoptimization to katib_suggestion.

gaocegege:
test/e2e/v1alpha2/test-client.go:9:2: no Go files in /mnt/test-data-volume/kubeflow-katib-presubmit-e2e-435-ac68cb5-9088-fe5c/go/src/github.com/kubeflow/katib/pkg/api

johnugeorge:
Can you update the description with the details of restructuring done in this PR ?

manifests/v1alpha2/studyjobcontroller/crd.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
johnugeorge: studyjobcontroller is not needed in v1alpha2. Why are these files added?

jdplatt:
Sure! Along with my changes to the python code to generate suggestions I added v1alpha2 folders for several components. For the following folders I copied my version into a new folder called v1alpha2 and the copied the v1alpha1 folder over from master:

1. manifests
2. scripts
3. pkg/suggestion
4. test/e2e
5. test/scripts

For the cmd/suggestion and test/workflow folders I left the current folders in the same place to avoid creating merge conflicts for other people and added a v1alpha2 folder containing my version. I figured we can clean up the cmd/suggestion folder when we merge the suggestion service into a single image, and the test/workflow folder when we do the integrations tests for v1alpha2. 

Because my v1alpha2 is just the current state of the manifests, there will be need to be more changes as we add the v1alpha2 versions of the different components. For example, if the studyjobcontroller is not needed in v1alpha2 you can delete it in the PR that implements the replacement. This may not be the ideal approach, but this PR was already getting huge and I wanted to get it merged and not have to roll changes for even more components into it.

johnugeorge:
Do you really need to copy all  the files to v1alpha2 folder in this PR? Is there any reason?  I feel that it will be better to add v1alpha2 only for relevant components related to the changes in this PR . eg: manifests for suggestion  

For other components, we can add whenever corresponding v1alpha2 code is added.

jdplatt:
The code for v1alpha1 and v1alpha2 is very similar so it seemed natural to take a copy of v1alpha1 as the starting point of v1alpha2 and make the required alterations, rather than build up v1alpha2 from scratch. I guess I could have created the v1alpha2 folders and cherry-picked only the files I changed to move over, but it seemed easier to move everything at once to make sure nothing was missed.

richardsliu:
@jdplatt We've changed the API implementation quite a bit from v1alpha1 to v1alpha2, so taking a copy of all the files would make it harder for others to determine which files were copied from v1alpha1 and which ones are under development. It also unfortunately makes this PR harder to review. Can we include only the relevant changed files? Or alternatively mark the copied files with a note/comment so we know which files should be removed later?

jlewi:
@jdplatt @richardsliu How's this PR coming? Do we have a goal in mind for getting it merged?

richardsliu:
@jdplatt We should try to get this PR merged soon, otherwise we will be dealing with more conflicts later on. Let me know if you need help with this.

manifests/v1alpha2/vizier/suggestion/katib_suggestion/deployment.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: extensions/v1beta1
```
richardsliu: Name of the directory should be bayestian-optimization?

pkg/suggestion/v1alpha2/katib_suggestion/algorithm_register.py
```diff
@@ -0,0 +1,8 @@
+from .random_search import RandomSearch
+from .grid_search import GridSearch
+from .bayesian_optimization_algorithm import BOAlgorithm
+
+
+ALGORITHM_REGISTER = {"random_search": RandomSearch,
```
richardsliu: How would someone distinguish between an algorithm that is still in development and one that is stable? What if users want to enable or disable a specific algorithm?

Also if a user wants to add a new algorithm, how would they achieve this without changing the shared Docker image?

test/scripts/v1alpha2/run-tests.sh
```diff
@@ -66,6 +66,11 @@ echo "REPO_NAME ${REPO_NAME}"
 echo "VERSION ${VERSION}"
 
 sed -i -e "s@image: katib\/katib-controller@image: ${REGISTRY}\/${REPO_NAME}\/katib-controller:${VERSION}@" manifests/v1alpha2/katib-controller/katib-controller.yaml
+sed -i -e "s@image: katib\/suggestion@image: ${REGISTRY}\/${REPO_NAME}\/suggestion:${VERSION}@" manifests/vizier/suggestion/random/deployment.yaml
```
richardsliu: These seem to be missing the v1alpha2 path.

manifests/v1alpha2/vizier/suggestion/grid/deployment.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: extensions/v1beta1
```
richardsliu: Is "vizier" necessary here? I think we can just use "suggestion" here. Also going forward we should avoid using the name vizier since it can confuse users.

cmd/suggestion/v1alpha2/hyperband/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:alpine AS build-env
+# The GOPATH in the image is /go.
```
hougangliu: we had `cmd/suggestion/hyperband/v1alpha1` already. can you move it in `cmd/suggestion/hyperband/v1alpha2`? do dose others

cmd/suggestion/v1alpha2/katib-suggestion/main.py
```diff
@@ -0,0 +1,26 @@
+import time
+from concurrent import futures
+
+import grpc
+
+from pkg.api.python import api_pb2_grpc
+from pkg.suggestion.katib_suggestion.rpc_service import SuggestionService
```
hougangliu: wrong python path

cmd/suggestion/v1alpha2/hyperband/main.go
```diff
@@ -0,0 +1,27 @@
+package main
```
hougangliu: do you plan to migrate hyperband to python one in another PR? if yes, please move it in this PR

test/e2e/v1alpha2/invalid-studyjob.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: "kubeflow.org/v1alpha1"
+kind: StudyJob
```
hougangliu: I think you need sync up most part in test/ for v1alpha2

pkg/suggestion/v1alpha2/katib_suggestion/algorithm_register.py
hougangliu: can we make it configurable instead of hard-code

pkg/suggestion/v1alpha2/katib_suggestion/rpc_service.py
```diff
@@ -0,0 +1,102 @@
...
+REQUIRES_PAST_OBSERVATIONS = ["bayesian_optimization", "grid_search"]
+
+
+class SuggestionService(api_pb2_grpc.SuggestionServicer):
+
+    def __init__(self, search_algorithm=SEARCH_ALGORITHM, logger=None):
+        self.manager_addr = MANAGER_ADDRESS
+        self.manager_port = MANAGER_PORT
+        self.logger = logger if (logger is not None) else get_logger()
+        self.search_algorithm = search_algorithm
```
hougangliu: if search_algorithm is not in ALGORITHM_REGISTER, we should exit with right error message in __init__ instead of exception occurring in line 30 when GetSuggestions.
That is, the corresponding deployment should not be `Running` if search_algorithm is not supported

pkg/suggestion/v1alpha2/katib_suggestion/random_search.py
```diff
@@ -0,0 +1,25 @@
...
+"""
+from .parsing_utils import parse_x_next_vector
+
+
+class RandomSearch:
+
+    def __init__(self, parameter_config, suggestion_config, logger=None):
+        self.parameter_config = parameter_config
+        self.suggestion_config = suggestion_config
+        self.logger = logger
```
hougangliu: self.logger unused

pkg/suggestion/v1alpha2/katib_suggestion/rpc_service.py
```diff
@@ -0,0 +1,102 @@
...
+        self.requires_previous = search_algorithm in REQUIRES_PAST_OBSERVATIONS
+
+    def GetSuggestions(self, request, context):
+        suggestion_config = self._parse_suggestion_parameters(request.param_id)
+        suggestion_config = {param.name: param.value for param in suggestion_config}
+        study_conf = self._get_study_config(request.study_id)
```
hougangliu: study_id field is dropped in v1alpha2

hougangliu:
@jdplatt GetSuggestionsRequest of v1alpha2 changes a lot, I think you need sync up suggestion services of v1alpha2 with it.
Thanks!

richardsliu:
Discussed offline with @jdplatt and @johnugeorge - we (myself and Johnu) will patch this PR and make the necessary fixes based on recent changes to the v1alpha2 API. Thanks John for your contributions.

Issue #273, opened 2018-12-04
[cleanup] delete CLI code


jlewi:
According to: https://github.com/kubeflow/katib/issues/78#issuecomment-440470532

We no longer use the CLI and we could delete the code for it.

jdplatt:
I think we can close this now that #436 is merged.

johnugeorge:
Thanks @jdplatt . Closing as PR #436  is merged.
/close

PR #436, opened 2019-03-16
Remove katibcli

jdplatt:
Addresses #273. I'm pretty sure this is everything related to the CLI. Did I miss anything, or worse delete something we still need?

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/436)


jdplatt:
Ok I fixed the failing test. There was some code to copy the go dependencies to the mounted volume in the build-cli.sh test script that we still needed. I copied it over to the unit-test.sh script.

PR #453, opened 2019-04-10
fix py client import error

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/453)

### DIFF
--- a/py/study_job_client.py
+++ b/py/study_job_client.py
@@ -14,6 +14,7 @@
 import datetime
 import json
 import logging
+import multiprocessing
 import time
 
 from kubernetes import client as k8s_client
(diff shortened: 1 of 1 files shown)


johnugeorge:
@hougangliu  Can you please explain why you need this change?

hougangliu:
> @hougangliu Can you please explain why you need this change?

pipeline katib-laucher uses studyjob python client. and this patch fixes 2 import error:
1. multiprocessing is used but not import
2. util.JobTimeoutError is not defined.

When timeout occurs, the two issue triggers

johnugeorge:
Ok. It was not used earlier 
/lgtm
/approve

PR #455, opened 2019-04-15
fix(readme): Merge image directory

gaocegege:
Currently, we have two directories, `docs/images` and `img`. We could merge the two since they serve the same purposes.

/assign @johnugeorge @YujiOshima 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/455)


gaocegege:
New README preview: https://github.com/kubeflow/katib/blob/0addb334b48e48d2dfe32864042f72a82b971eaf/README.md

PR #456, opened 2019-04-15
v1alpha2 api server implementation

YujiOshima:

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/456)


YujiOshima:
I will add some tests.

pkg/db/v1alpha2/interface.go
```diff
@@ -120,40 +124,526 @@ func New() (KatibDBInterface, error) {
...
+		}
+		start_time = s_time.UTC().Format(mysqlTimeFmt)
+		c_time, err := time.Parse(time.RFC3339Nano, experiment.ExperimentStatus.CompletionTime)
+		if err != nil {
+			log.Printf("Error parsing completion time %s: %v", experiment.ExperimentStatus.CompletionTime, err)
+		}
+		completion_time = c_time.UTC().Format(mysqlTimeFmt)
+	}
+	_, err = d.db.Exec(
+		`INSERT INTO experiments (
```
richardsliu: What about the proposal for an abstract DB interface? https://docs.google.com/document/d/1xAi4V9SCAVZvrwiVc6a3VpK7BXPx-liQbr6YPdqBThs/edit#heading=h.1hri22e2jta

cmd/manager/v1alpha2/main.go
```diff
@@ -24,93 +25,142 @@ type server struct {
...
 }
 
 // Get Suggestions from a Suggestion service.
-func (s *server) GetSuggestions(context.Context, *api_pb.GetSuggestionsRequest) (*api_pb.GetSuggestionsReply, error) {
-	return nil, nil
+func (s *server) GetSuggestions(ctx context.Context, in *api_pb.GetSuggestionsRequest) (*api_pb.GetSuggestionsReply, error) {
+	if in.AlgorithmName == "" {
+		return &api_pb.GetSuggestionsReply{Trials: []*api_pb.Trial{}}, errors.New("No algorithm name is specified")
+	}
+	conn, err := grpc.Dial("vizier-suggestion-"+in.AlgorithmName+":6789", grpc.WithInsecure())
```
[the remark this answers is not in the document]
johnugeorge: i think, we should start renaming to katib.  i had renamed studyjobcontroller to katib-controller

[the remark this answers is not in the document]
YujiOshima: agree. I will change them.

pkg/db/v1alpha2/interface.go
YujiOshima: @richardsliu I think it is an important proposal. Should I include it to this PR? or make another PR after this?
If I include it, this PR may become too big.

gaocegege: Personally, I think it is better to place in another PR.

richardsliu: Let's change it in another PR.

pkg/api/v1alpha2/api.proto
```diff
@@ -439,9 +451,7 @@ message DeleteTrialReply {
 
 message GetTrialListRequest {
     string experiment_name = 1;
-    string start_time = 2; ///The start of the time range. RFC3339 format
-    string end_time = 3; ///The end of the time range. RFC3339 format
-    string filter = 4;
+    string filter = 2;
```
hougangliu: what does `filter` look like? I don't find it is used in GetTrialList()

cmd/manager/v1alpha2/main.go
```diff
@@ -24,93 +25,142 @@ type server struct {
...
 
 // Validate AlgorithmSettings in an Experiment.
 // Suggestion service should return INVALID_ARGUMENT Error when the parameter is invalid
-func (s *server) ValidateAlgorithmSettings(context.Context, *api_pb.ValidateAlgorithmSettingsRequest) (*api_pb.ValidateAlgorithmSettingsReply, error) {
-	return nil, nil
+func (s *server) ValidateAlgorithmSettings(ctx context.Context, in *api_pb.ValidateAlgorithmSettingsRequest) (*api_pb.ValidateAlgorithmSettingsReply, error) {
+	if in.AlgorithmName == "" {
+		return &api_pb.ValidateAlgorithmSettingsReply{}, errors.New("No algorithm name is specified")
+	}
+	conn, err := grpc.Dial("vizier-suggestion-"+in.AlgorithmName+":6789", grpc.WithInsecure())
```
hougangliu: we can add a function like `getSuggestionServiceUrl(name String) string` or even a function like `getSuggestionServiceConnection(name String)`

pkg/api/v1alpha2/api.proto
YujiOshima: Sorry, I forget it. I will add it to GetTrialList.

cmd/manager/v1alpha2/main.go
YujiOshima: done

YujiOshima:
I applied comments and added tests.
PTAL @richardsliu @johnugeorge @hougangliu @gaocegege

gaocegege:
cc @anchovYu PTAL, too

pkg/api/v1alpha2/api.proto
hougangliu: Since the filter is used to filter trial name, can we rename it to filter_by_name? Also in db function, please handle `filter == ""` case.

YujiOshima: Thanks! updated.

pkg/api/v1alpha2/api.proto
```diff
@@ -439,9 +451,7 @@ message DeleteTrialReply {
 
 message GetTrialListRequest {
     string experiment_name = 1;
-    string start_time = 2; ///The start of the time range. RFC3339 format
-    string end_time = 3; ///The end of the time range. RFC3339 format
-    string filter = 4;
+    string filter_by_name = 2;
```
richardsliu: I actually prefer "filter", because semantically it allows us to do things like:
- filter:name=foo
- filter:start_time>x
- filter:end_time<=y

We probably will only support filtering by name for now, but I can see a use case for the others. What do you think?

scripts/mockgen.sh
```diff
@@ -29,7 +29,6 @@ echo "Generating v1alpha1 VizierDBInterface..."
 mockgen -package mock -destination pkg/mock/v1alpha1/db/db.go github.com/kubeflow/katib/pkg/db/v1alpha1 VizierDBInterface
```
richardsliu: Vizier -> Katib

pkg/api/v1alpha2/api.proto
hougangliu: Make sense. A general filter will really help. So  we need use `d.db.Query("SELECT * FROM trials WHERE experiment_name = ? AND ?, experimentName, filter)` instead, not limited in name filed filter. So when update, please comment what `filter` looks like.

YujiOshima: @richardsliu @hougangliu Thanks. I revert to `filter` and add comments.

cmd/manager/v1alpha2/main.go
```diff
@@ -24,94 +26,160 @@ type server struct {
...
+		Status: health_pb.HealthCheckResponse_SERVING,
+	}
+
+	// We only accept optional service name only if it's set to suggested format.
+	if in != nil && in.Service != "" && in.Service != "grpc.health.v1.Health" {
+		resp.Status = health_pb.HealthCheckResponse_UNKNOWN
+		return &resp, fmt.Errorf("grpc.health.v1.Health can only be accepted if you specify service name.")
+	}
+
+	// Check if connection to vizier-db is okay since otherwise manager could not serve most of its methods.
```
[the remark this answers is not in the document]
YujiOshima: done

PR #465, opened 2019-04-24
enable test for v1alpha2

hougangliu:
This is initial version to enable test for v1alpha2.
Now, for v1alpha2 test, only build katib-controller and test if the deployment Pod is running or not.
After other components done, add corresponding tests one by one.

@richardsliu @johnugeorge @YujiOshima @gaocegege Please help review it. thanks!

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/465)


gaocegege:
Thanks for your contribution! Let's wait for the CI result.

cmd/katib-controller/v1alpha1/Dockerfile
```diff
@@ -7,10 +7,8 @@ ADD . /go/src/github.com/kubeflow/katib
 WORKDIR /go/src/github.com/kubeflow/katib/cmd/katib-controller
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o studyjobcontroller  ./v1alpha1
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o katib-controller.v1alpha2  ./v1alpha2
 # Copy the controller-manager into a thin image
 FROM alpine:3.7
 WORKDIR /app
 COPY --from=build-env /go/src/github.com/kubeflow/katib/cmd/katib-controller/studyjobcontroller .
-COPY --from=build-env /go/src/github.com/kubeflow/katib/cmd/katib-controller/katib-controller.v1alpha2 .
```
johnugeorge: If we want to support both v1alpha1 and v1alpha2 with same image, don't we need to build both of the versions together?
@richardsliu

test/scripts/v1alpha2/build-ui.sh
```diff
@@ -0,0 +1,43 @@
...
+exit 0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+export PATH=${GOPATH}/bin:/usr/local/go/bin:${PATH}
+REGISTRY="${GCP_REGISTRY}"
+PROJECT="${GCP_PROJECT}"
+GO_DIR=${GOPATH}/src/github.com/${REPO_OWNER}/${REPO_NAME}-modeldb
```
andreyvelich: Should it be `modeldb` or `katib-ui` here?

examples/v1alpha2/random-example.yaml
```diff
@@ -0,0 +1,59 @@
...
+                  - "--batch-size=64"
+                  {{- with .HyperParameters}}
+                  {{- range .}}
+                  - "{{.Name}}={{.Value}}"
+                  {{- end}}
+                  {{- end}}
+                restartPolicy: Never
+  parameters:
+    - name: --lr
+      parametertype: double
```
andreyvelich: `parametertype` -> `parameterType`

examples/v1alpha2/random-example.yaml
```diff
@@ -0,0 +1,59 @@
...
+      feasibleSpace:
+        min: "0.01"
+        max: "0.03"
+    - name: --num-layers
+      parametertype: int
```
andreyvelich: `parametertype` -> `parameterType`

examples/v1alpha2/random-example.yaml
```diff
@@ -0,0 +1,59 @@
...
+      feasibleSpace:
+        min: "2"
+        max: "5"
+    - name: --optimizer
+      parametertype: categorical
```
andreyvelich: `parametertype` -> `parameterType`

manifests/v1alpha2/katib-controller/crds.yaml
```diff
@@ -0,0 +1,25 @@
...
+metadata:
+  name: experiments.kubeflow.org
+spec:
+  group: kubeflow.org
+  version: v1alpha2
+  scope: Namespaced
+  names:
+    kind: Experiment
+    singular: experiment
+    plural: experiments
```
andreyvelich: Should we add condition in Experiment crd, like here: https://github.com/kubeflow/katib/blob/master/manifests/v1alpha1/studyjobcontroller/crd.yaml#L13 ?

test/scripts/v1alpha2/build-ui.sh
hougangliu: ui will be refactor, update this shell then

manifests/v1alpha2/katib-controller/crds.yaml
hougangliu: we can add it when experiment conditions done. now, we cannot know which one condition we should show

cmd/katib-controller/v1alpha1/Dockerfile
hougangliu: v1alpha1 and v1alpha2 have big difference, I think maybe we should separate them in two images.

richardsliu: But this is the docker file for v1alpha1, based on the file path. Why is it building v1alpha2?

richardsliu: For tf-operator and pytorch-operator, we've always built multiple versions with the same image. In the pod template we override the entrypoint for different versions.

hougangliu: OK, I can follow operators.

prow_config.yaml
```diff
@@ -24,3 +24,25 @@ workflows:
     - test/*
     params:
       registry: "gcr.io/kubeflow-images-public"
+  - app_dir: kubeflow/katib/test/workflows
+    component: workflows-v1alpha2
+    name: e2e-v1alpha2
+    job_types:
+      - presubmit
+    include_dirs:
+    - pkg/*
```
richardsliu: Can we configure the conditions such that only v1alpha1 tests are run if v1alpha1 code is changed, and similarly for v1alpha2?

hougangliu: that will list two many files, especially we plan to share common files between the two versions, such as `Dockerfile`

hougangliu: it is a good idea to trigger corresponding version workflow when related code changed. maybe we can implement it when we refactor katib dir.

hougangliu: #472 trace it

prow_config.yaml
```diff
@@ -24,3 +24,25 @@ workflows:
...
+    - cmd/*
+    - test/*
+    params:
+      registry: "gcr.io/kubeflow-ci"
+  - app_dir: kubeflow/katib/test/workflows
+    component: workflows-v1alpha2
+    name: e2e-v1alpha2-release
+    job_types:
+      - postsubmit
```
johnugeorge: do we need postsubmit for v1alpha2 if the same image is built ?

manifests/v1alpha2/katib-controller/trialTemplateConfigmap.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: ConfigMap
+metadata:
+  name: trial-template
+  namespace: kubeflow
+data:
+  defaultTrialTemplate.yaml : |-
+    apiVersion: batch/v1
+    kind: Job
+    metadata:
+      name: {{.WorkerID}}
```
johnugeorge: Change to Trial?

manifests/v1alpha2/katib-controller/trialTemplateConfigmap.yaml
```diff
@@ -0,0 +1,20 @@
...
+      namespace: {{.NameSpace}}
+    spec:
+      template:
+        spec:
+          containers:
+          - name: {{.WorkerID}}
```
johnugeorge: Same comment

scripts/v1alpha2/build.sh
```diff
@@ -0,0 +1,28 @@
...
+set -o pipefail
+
+PREFIX="katib"
+CMD_PREFIX="cmd"
+
+SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/../..
+
+cd ${SCRIPT_ROOT}
+
+docker build -t ${PREFIX}/katib-controller -f ${CMD_PREFIX}/katib-controller/Dockerfile .
```
johnugeorge: build script might be same for both versions if same image contains both versions. We can change it later when all components are built

prow_config.yaml
hougangliu: In future, I wonder if all the images containing both versions (at least, metricsCollectors may be dropped).  keep it as it is now, update if in need

scripts/v1alpha2/build.sh
hougangliu: ok

andreyvelich:
Thanks, lgtm!

hougangliu:
Thanks for review
/approve

Issue #374, opened 2019-02-13
Don't run full testing suite for docs/ changes


alexandraj777:
PR #369 added a few links to [docs/developer-guide.md](https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md), and was not expected to have any effects on katib code. However, a full test suite was required to pass before merging the PR. Due to repeated test failures, the suite needed to be run three times before passing.

Blocking docs changes on full test failures may discourage frequent and casual documentation updates. Is there a smaller / lighter test suite that can run for either `*.md` files, or the `README.md` + `docs/*.md` files?

jlewi:
You can use include_dirs in the prow_config to only trigger an E2E test if files in certain directories are changed.
https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml#L24

alexandraj777:
Excuse me for not knowing, but what is an E2E test? End-to-end?

gaocegege:
@alexandraj777 Yeah, End to end.

PR #467, opened 2019-04-24
chore: Skip test when code is not changed

gaocegege:
Fix https://github.com/kubeflow/katib/issues/374

/cc @alexandraj777

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/467)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -6,6 +6,10 @@ workflows:
     name: e2e
     job_types:
       - presubmit
+    include_dirs:
+    - pkg/*
+    - cmd/*
+    - test/*
     params:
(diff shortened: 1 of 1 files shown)


alexandraj777:
Thanks @gaocegege! I don't know enough about configuring prow and the structure of go projects to know if this is what we want, but it seems good. The only suggestion I'd make is to include the prow config itself...is there any way to re-run tests when the prow config file changes?

alexandraj777:
Oh, I assigned @jlewi because it looks like he was the other person to touch these files, not sure if you're allowed to "lgtm" on your own PRs, so lmk if I've done something non-standard 😄

PR #473, opened 2019-04-25
validation and mutating webhook for experiment

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/473)


pkg/api/operators/apis/experiment/v1alpha2/experiment_defaults.go
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+// Code generated by main. DO NOT EDIT.
+
+package v1alpha2
+
+func (e *Experiment) Default() {
```
johnugeorge: Many be SetDefaults?

pkg/controller/v1alpha2/experiment/util/webhook_util.go
```diff
@@ -0,0 +1,88 @@
...
+	if len(instance.Spec.Parameters) == 0 && instance.Spec.NasConfig == nil {
+		return fmt.Errorf("spec.Parameters or spec.NasConfig must be specified.")
+	}
+
+	if len(instance.Spec.Parameters) > 0 && instance.Spec.NasConfig != nil {
+		return fmt.Errorf("Only one of spec.Parameters and spec.NasConfig can be specified.")
+	}
+
+	if len(instance.Spec.Parameters) > 0 {
+		if err := validateParameters(instance.Spec.Parameters); err != nil {
```
andreyvelich: Will we use `ValidateAlgorithmSetting` function to validate parameters inside Suggestion ? 
https://github.com/kubeflow/katib/blob/master/pkg/api/v1alpha2/api.proto#L170

hougangliu: sure :)

andreyvelich: Good, thank you!

pkg/api/operators/apis/experiment/v1alpha2/experiment_defaults.go
hougangliu: I can update it (in fact, I see many kube project use Default(), too).

johnugeorge:
@hougangliu Can you have a look at my previous comment?  Is it possible to avoid client-go dependency directly?  The problem is that we will then have to keep it as a direct dependency in Godep and then have to worry about the changes in client-go library. However, if we use client of controller-runtime(iwhich indirectly uses client-go), we have to just upgrade controller-runtime whenever required

hougangliu:
> @hougangliu Can you have a look at my previous comment? Is it possible to avoid client-go dependency directly? The problem is that we will then have to keep it as a direct dependency in Godep and then have to worry about the changes in client-go library. However, if we use client of controller-runtime(iwhich indirectly uses client-go), we have to just upgrade controller-runtime whenever required

OK, I can try. Thanks!

johnugeorge:
Sorry. Forgot to publish

pkg/controller/v1alpha2/experiment/util/template_util.go
```diff
@@ -0,0 +1,109 @@
...
+	"bytes"
+	"errors"
+	"os"
+	"text/template"
+
+	apiv1alpha2 "github.com/kubeflow/katib/pkg/api/v1alpha2"
+	experimentsv1alpha2 "github.com/kubeflow/katib/pkg/api/operators/apis/experiment/v1alpha2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
```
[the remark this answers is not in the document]
hougangliu: done! thanks

pkg/api/operators/apis/experiment/v1alpha2/constants.go
```diff
@@ -27,4 +27,7 @@ const (
 
 	// Default value of Spec.TemplatePath
 	DefaultTrialTemplatePath = "defaultTrialTemplate.yaml"
+
+	OptimizationTypeMinimize = "minimize"
```
johnugeorge: Do we need this? Isn't it already defined in types.go?

hougangliu: I will remove it

pkg/controller/v1alpha2/experiment/util/template_util.go
```diff
@@ -0,0 +1,107 @@
...
+
+	trialTemplate := instance.Spec.TrialTemplate
+	if trialTemplate != nil && trialTemplate.GoTemplate.RawTemplate != "" {
+		tpl, err = template.New("Trial").Parse(trialTemplate.GoTemplate.RawTemplate)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		configMapNS := os.Getenv(experimentsv1alpha2.DefaultKatibNamespaceEnvName)
+		configMapName := experimentsv1alpha2.DefaultTrialConfigMapName
```
johnugeorge: Should we keep these too in the defaults?

hougangliu: make sense

pkg/controller/v1alpha2/experiment/util/webhook_util.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"bytes"
+	"fmt"
+
+	ep_v1alpha2 "github.com/kubeflow/katib/pkg/api/operators/apis/experiment/v1alpha2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func ValidateExperiment(instance *ep_v1alpha2.Experiment) error {
```
YujiOshima: Could you add validation for Experiment name that the name is already registered in DB or not?
If users set `RetainHistoricalData` true, the previous data will remain even the CR was deleted.

hougangliu: done! thanks!

PR #474, opened 2019-04-26
Get experiment config from the instance

andreyvelich:
In this PR I have added few thing in the v1alpha2 experiment controller:
1. Change structure for `ParameterSpec` in `Operation` inside api.proto. I think, it should be the same as `ParameterSpecs` in `ExperimentSpec`
2. Add parsing from the instance to experiment config. I think, we should parse it before save in DB.

We need to think, what status of Experiment config should be on the step of getting `Experiment config` from instance.
As well, do we need a JobType inside `Experiment config` like [here](https://github.com/kubeflow/katib/blob/master/pkg/api/v1alpha1/api.proto#L456) or we can handle it with checking nasConfig ?
/assign @johnugeorge @hougangliu @YujiOshima 
/cc @richardsliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/474)


YujiOshima:
My opinion, we don't need to distinguish NASJobs from HPJobs.
We can populate both of HPconf and NASconf (if it is not nil) and send to specified suggestion service. And then, the suggestion service can handle the Experiment configs.

andreyvelich:
@YujiOshima I can make it in the same function, like you said. 
What do you think @johnugeorge @hougangliu ?

hougangliu:
> @YujiOshima I can make it in the same function, like you said.
> What do you think @johnugeorge @hougangliu ?

agree with @YujiOshima

johnugeorge:
SGTM

andreyvelich:
@hougangliu @YujiOshima @johnugeorge 
I made these changes.

pkg/controller/v1alpha2/experiment/experiment_controller.go
```diff
@@ -184,8 +184,10 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 
 	}
 	if !instance.IsCreated() {
+		//Get Experiment config from instance
+		experimentConfig := util.GetExperimentConf(instance)
 		//Experiment not created in DB
-		err = util.CreateExperimentInDB(instance)
+		err = util.CreateExperimentInDB(experimentConfig)
```
johnugeorge: As we would like to remove dependency of api manager from controller, I would recommend to keep grpc api related config out of the controller code.  Let util take care of getting the config and calling the API

andreyvelich: I put it inside `CreateExperimentInDB` function in `util`. Is it fine?

johnugeorge:
@andreyvelich can you rebase

andreyvelich:
@johnugeorge Done

pkg/controller/v1alpha2/experiment/util/api_util.go
```diff
@@ -37,7 +38,142 @@ func GetExperimentFromDB(instance *experimentsv1alpha2.Experiment) (*api_pb.GetE
...
+				AdditionalMetricsNames: []string{},
+			},
+			Algorithm: &api_pb.AlgorithmSpec{
+				AlgorithmSetting: []*api_pb.AlgorithmSetting{},
+			},
+		},
+	}
+
+	experiment.Name = instance.ObjectMeta.Name
+
```
hougangliu: s/instance.ObjectMeta.Name/instance.Name

andreyvelich: Done

PR #476, opened 2019-04-28
katib manager db error

hougangliu:
`condition` is keyword of mysql, we need escape it in sql, otherwise an error in your SQL syntax occurs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/476)


johnugeorge:
instead of this, is it better to rename thee field? may be `status`?  WDYT?

YujiOshima:
I agree with @johnugeorge.

hougangliu:
@YujiOshima @johnugeorge agree with you, I will update it to `status`

PR #478, opened 2019-04-29
enable test for katib-manager

hougangliu:
enable test for katib-manager and fix RegisterTrial error

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/478)

### DIFF
--- a/manifests/v1alpha2/katib/db/deployment.yaml
+++ b/manifests/v1alpha2/katib/db/deployment.yaml
@@ -1,0 +1,52 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: katib-db
+  namespace: kubeflow
+  labels:
+    app: katib
+    component: db
+spec:
+  replicas: 1
+  template:
+    metadata:
+      name: katib-db
+      labels:
+        app: katib
+        component: db
+    spec:
+      containers:
+      - name: katib-db
+        image: mysql:8.0.3
+        args:
+        - --datadir
+        - /var/lib/mysql/datadir
+        env:
+          - name: MYSQL_ROOT_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: katib-db-secrets
+                key: MYSQL_ROOT_PASSWORD
+          - name: MYSQL_ALLOW_EMPTY_PASSWORD
+            value: "true"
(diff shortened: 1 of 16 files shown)


manifests/v1alpha2/katib/db/deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+          initialDelaySeconds: 5
+          periodSeconds: 2
+          timeoutSeconds: 1
+        volumeMounts:
+        - name: katib-mysql
+          mountPath: /var/lib/mysql
+      volumes:
+      - name: katib-mysql
+        persistentVolumeClaim:
+          claimName: katib-mysql
```
gaocegege: Will we provide the PV config for users?

andreyvelich: I think, we have to provide it, like in v1alpha1. 
Some users don't have storage classes in their k8s clusters to bound PVC.

gaocegege: If some users do not have storage classes, then how can we create PV for the users? Do you mean that we use the default storage class to do it?

andreyvelich: If users don't have StorageClass they need to follow this instruction: https://github.com/kubeflow/katib#katib.
They need to create PV manually from the yaml file.

hougangliu: Ah! I missed pv/pvc.

gaocegege: Yeah, I agree. Thus we could add a default config for the PV and PVC.

hougangliu: in fact, files in manifest are used to test. In kubeflow product, only PV will not be created in ksonnet/kustimize

andreyvelich: We can create README file under kustomize Katib manifest, how to create PV, like we have here: https://github.com/kubeflow/kubeflow/tree/master/kubeflow/katib#katib-1

hougangliu:
@YujiOshima @johnugeorge @richardsliu @gaocegege @andreyvelich 
test passed.

Issue #442, opened 2019-03-27
[v1alpha2] Metrics collection


richardsliu:

richardsliu:
@YujiOshima @johnugeorge @gaocegege 
Per our discussion last week, until the metadata store is ready, the least costly solution would be to keep metrics collection as-is (i.e. using cron jobs). We will migrate to the new push model when the metadata APIs are ready.

jlewi:
@johnugeorge  @richardsliu Could you provide an update for this issue please?

johnugeorge:
@jlewi This is done. 
/close

PR #481, opened 2019-04-30
Add metrics collector spec to v1alpha2 API

richardsliu:
Part of #442

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/481)


pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
```diff
@@ -50,6 +50,9 @@ type ExperimentSpec struct {
 	// TODO - figure out what to do with metric collectors
 	MetricsCollectorType string `json:"metricsCollectorSpec,omitempty"`
```
johnugeorge: Is this needed?

hougangliu: I think we can keep it here, since MetricsCollectorSpec is a workaround. In long term, we will use MetricsCollectorType

richardsliu: If we need to change it in the future, should we deprecate metricsCollectorSpec directly or create a new API version?

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
```diff
@@ -50,6 +50,9 @@ type ExperimentSpec struct {
...
 
+	// For v1alpha2 we will keep the metrics collector implementation same as v1alpha1. 
+	MetricsCollectorSpec *MetricsCollectorSpec `json:"metricsCollectorSpec,omitempty"`
```
hougangliu: how about use TrialTemplate to replace MetricsCollectorSpec? or we can rename TrialTemplate to a general name so that MetricsCollectorSpec and TrialTemplate can be same struct.
after that, we can handle both MetricsCollectorSpec and TrialTemplate in same method

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
hougangliu: I suggest just  deprecate metricsCollectorSpec if there is no big change

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
richardsliu: This is a temporary fix so I think we should keep the definitions separate and more readable.

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
johnugeorge: I feel that we needn't include any fields that are not supported in current release.  We will need a new API version anyways.

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
hougangliu: ok

richardsliu:
Removed metricsCollectorType. If we need to change the metrics collector behavior in the future, we will add this back in a new API version.

PR #484, opened 2019-05-03
V1alpha2 Metrics collector (part 1)

richardsliu:
Part of #442 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/484)


pkg/util/v1alpha2/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,116 @@
...
+	return &MetricsCollector{
+		clientset: clientset,
+	}, nil
+
+}
+
+func (d *MetricsCollector) CollectObservationLog(tId string, jobKind string, objectiveValueName string, metrics []string, namespace string) (*v1alpha2.ObservationLog, error) {
+	labelMap := make(map[string]string)
+
+	// TODO: Create labelMap based on jobKind
```
johnugeorge: For now, we can add the labels for the default "Job"

cmd/metricscollector/v1alpha2/main.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	ctx := context.Background()
+	getreq := &api.GetExperimentRequest{
+		ExperimentName: *experimentName,
+	}
+	getrep, err := c.GetExperiment(ctx, getreq)
+	if err != nil {
+		log.Fatalf("Failed to GetExperiment: %v", err)
+	}
+	olog, err := mc.CollectObservationLog(*trialName, *jobKind, getrep.Experiment.ExperimentSpec.Objective.ObjectiveMetricName, getrep.Experiment.ExperimentSpec.Objective.AdditionalMetricsNames, *namespace)
```
johnugeorge: Why do we need to distinguish between ObjectiveMetricName and additional metrics?  All metrics have to be collected right?  Just wondering. I see the same in v1alpha1 also.

cmd/metricscollector/v1alpha2/main.go
```diff
@@ -0,0 +1,92 @@
...
+	c := api.NewManagerClient(conn)
+	mc, err := metricscollector.NewMetricsCollector()
+	if err != nil {
+		log.Fatalf("Failed to create MetricsCollector: %v", err)
+	}
+	ctx := context.Background()
+	getreq := &api.GetExperimentRequest{
+		ExperimentName: *experimentName,
+	}
+	getrep, err := c.GetExperiment(ctx, getreq)
```
johnugeorge: Should we consider populating  metric names similar to other flags? We can then avoid this API call, reducing the dependency

manifests/v1alpha2/katib-controller/metricsControllerConfigMap.yaml
```diff
@@ -0,0 +1,38 @@
...
+kind: ConfigMap
+metadata:
+  name: metrics-collector-template
+  namespace: kubeflow
+data:
+  defaultMetricsCollectorTemplate.yaml : |-
+    apiVersion: batch/v1beta1
+    kind: CronJob
+    metadata:
+      name: {{.TrialName}}
```
johnugeorge: Template names are different from the trial Template.

pkg/util/v1alpha2/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,116 @@
...
+		for _, kv := range kvpairs {
+			v := strings.Split(kv, "=")
+			if len(v) > 2 {
+				log.Printf("Ignoring trailing garbage: %s", kv)
+			}
+			if len(v) == 1 {
+				continue
+			}
+			metricName := ""
+			if v[0] == objectiveValueName {
```
johnugeorge: Earlier comment: Why do we need to distinguish between objectiveMetric and other additional  Metrics?

pkg/util/v1alpha2/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
```
johnugeorge: Can we use controller runtime client itself instead of directly using client-go?

pkg/util/v1alpha2/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (d *MetricsCollector) CollectObservationLog(tId string, jobKind string, objectiveValueName string, metrics []string, namespace string) (*v1alpha2.ObservationLog, error) {
+	labelMap := make(map[string]string)
+
+	// TODO: Create labelMap based on jobKind
+
+	pl, _ := d.clientset.CoreV1().Pods(namespace).List(metav1.ListOptions{LabelSelector: labels.Set(labelMap).String(), IncludeUninitialized: true})
+	if len(pl.Items) == 0 {
+		return nil, errors.New(fmt.Sprintf("No Pods are found in Trial %v", tId))
```
gaocegege: fmt.Errorf is better than `errors.New(fmt.Sprintf` here.

pkg/util/v1alpha2/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,116 @@
...
+	}, nil
+
+}
+
+func (d *MetricsCollector) CollectObservationLog(tId string, jobKind string, objectiveValueName string, metrics []string, namespace string) (*v1alpha2.ObservationLog, error) {
+	labelMap := make(map[string]string)
+
+	// TODO: Create labelMap based on jobKind
+
+	pl, _ := d.clientset.CoreV1().Pods(namespace).List(metav1.ListOptions{LabelSelector: labels.Set(labelMap).String(), IncludeUninitialized: true})
```
gaocegege: We should check the error here.

pkg/util/v1alpha2/metricscollector/metricscollector.go
richardsliu: This is only used to get the in-cluster config, which needs the REST client. The controller runtime client also calls rest client under the hood: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/config/config.go#L78

cmd/metricscollector/v1alpha2/main.go
richardsliu: Manager service is still needed to make the Report call

pkg/util/v1alpha2/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,116 @@
...
+	"errors"
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
```
johnugeorge: use Katib-client instead?

richardsliu: According to https://stackoverflow.com/questions/53852530/how-to-get-logs-from-kubernetes-using-golang, we cannot get logs without using client-go for now.

pkg/util/v1alpha2/metricscollector/metricscollector.go
richardsliu: Changed to controller-runtime

PR #485, opened 2019-05-03
Update util for experiment in v1alpha2

andreyvelich:
I added new functions in Experiment util. After that, we can get new 2 conditions of Experiment and get current condition type of Experiment.
I will use it in the UI.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/485)


pkg/api/operators/apis/experiment/v1alpha2/util.go
```diff
@@ -73,10 +73,34 @@ func (exp *Experiment) IsFailed() bool {
...
+func (exp *Experiment) IsRestarting() bool {
+	return hasCondition(exp, ExperimentRestarting)
+}
+
 func (exp *Experiment) IsCompleted() bool {
 	return exp.IsSucceeded() || exp.IsFailed()
 }
 
+func (exp *Experiment) GetCurrentConditionType() ExperimentConditionType {
+	if exp.IsCreated() {
```
johnugeorge: What would this function used for?  If you want the latest condition, it would be better to directly lookup the Type of the last status.conditions?

andreyvelich: Is that true that the latest `Condition` in this array: https://github.com/kubeflow/katib/blob/master/pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go#L77, will be current? 
I need to get Current Condition of Experiment.

pkg/api/operators/apis/experiment/v1alpha2/util.go
```diff
@@ -73,10 +73,34 @@ func (exp *Experiment) IsFailed() bool {
...
+
+func (exp *Experiment) IsRestarting() bool {
+	return hasCondition(exp, ExperimentRestarting)
+}
+
 func (exp *Experiment) IsCompleted() bool {
 	return exp.IsSucceeded() || exp.IsFailed()
 }
 
+func (exp *Experiment) GetCurrentConditionType() ExperimentConditionType {
```
hougangliu: it will always return `ExperimentCreated`, I don't think you except it

andreyvelich: @hougangliu We can leave here only `Succeeded` or `Failed` or `Running` or `Restarting` condition. I think, for 1 experiment it can be one of the 4 conditions at the same time, isn't it?

hougangliu: I mean `exp.IsCreated()` will always `true`

andreyvelich: What will happen in that situation:
Experiment ran and after that failed
Experiment restarted
Experiment is running again.
Will `exp.IsRunning()` and `exp.IsFailed()` return true ?

YujiOshima: My understanding, the `exp.Status.Conditions` is a history of conditions of the experiment. So the exp.IsCreated() will be always `true`.
The last value of `exp.Status.Conditions` is the latest condition.

andreyvelich: @YujiOshima @hougangliu Ok, thank you! I changed this function to get the latest condition, please , take a look. In the UI, where we monitor experiments, I think, we should show the latest condition.

pkg/api/operators/apis/experiment/v1alpha2/util.go
```diff
@@ -73,10 +73,22 @@ func (exp *Experiment) IsFailed() bool {
...
+
+func (exp *Experiment) IsRestarting() bool {
+	return hasCondition(exp, ExperimentRestarting)
+}
+
 func (exp *Experiment) IsCompleted() bool {
 	return exp.IsSucceeded() || exp.IsFailed()
 }
 
+func (exp *Experiment) GetLastConditionType() ExperimentConditionType {
```
hougangliu: just for safe, we'd better guard `len(exp.Status.Conditions) > 0` before below

andreyvelich: @hougangliu Added, with an additional error, if Experiment doesn't have any Condition.

hougangliu:
/lgtm
thanks!

hougangliu:
@andreyvelich I think you can try to rebase upstream to fix test error

andreyvelich:
@hougangliu Thanks, done.

andreyvelich:
@richardsliu @hougangliu Please, lgtm this PR again.

PR #489, opened 2019-05-08
Add metrics collector spec and objective spec to Trial

richardsliu:
Part of #442

Adding some fields to Trial CRD:

- ObjectiveSpec: This is needed because the trial needs to store the best objective metric value in trial status. So the trial has to know the objective metric name and type.
- MetricsCollectorSpec: Raw text representation of the metrics collector. Needed to spawn the corresponding cron job.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/489)


pkg/api/operators/apis/trial/v1alpha2/trial_types.go
```diff
@@ -29,6 +29,9 @@ type TrialSpec struct {
 	// and let the corresponding resource controller (e.g. tf-operator) handle
 	// the rest.
 	RunSpec string `json:"runSpec,omitempty"`
+
+	// Raw text for the metrics collector spec. This must be a CronJob object.
+	MetricsCollectorSpec string `json:"metricsCollectorSpec,omitempty"`
```
johnugeorge: add Objective name and type  to spec to figure out the best values?

richardsliu: Added ObjectiveSpec to Trial.

pkg/api/operators/apis/trial/v1alpha2/trial_types.go
```diff
@@ -54,6 +60,23 @@ type TrialStatus struct {
...
+}
+
+type ObjectiveType string
+
+const (
+	ObjectiveTypeUnknown  ObjectiveType = ""
+	ObjectiveTypeMinimize ObjectiveType = "minimize"
+	ObjectiveTypeMaximize ObjectiveType = "maximize"
+)
+
```
hougangliu: maybe move above to a common_types.go since they are shared by by trial and experiment

richardsliu: Done

pkg/db/v1alpha2/db_init.go
```diff
@@ -18,7 +18,6 @@ func (d *dbConn) DBInit() {
 		parallel_trial_count INT,
 		max_trial_count INT,
 		status TINYINT,
-		metrics_collector_type TEXT,
```
johnugeorge: Are we not keeping the metric collector template in experiments table?

pkg/api/operators/apis/trial/v1alpha2/trial_types.go
```diff
@@ -51,22 +58,7 @@ type TrialStatus struct {
...
-
-type Metric struct {
-	Name  string  `json:"name,omitempty"`
-	Value float64 `json:"value,omitempty"`
-}
-
-type Observation struct {
-	// Key-value pairs for metric names and values
-	Metrics []Metric `json:"metrics"`
+	Observation common.Observation `json:"observation,omitempty"`
```
gaocegege: I think it is better to use `*common.Observation` since it is omitempty

pkg/db/v1alpha2/db_init.go
richardsliu: We should. Added

pkg/api/operators/apis/trial/v1alpha2/trial_types.go
richardsliu: Changed.

PR #491, opened 2019-05-09
add common package

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/491)

### DIFF
--- a/pkg/common/v1alpha1/common.go
+++ b/pkg/common/v1alpha1/common.go
@@ -1,4 +1,4 @@
-package pkg
+package v1alpha1
 
 import (
 	"os"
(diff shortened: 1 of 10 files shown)


pkg/common/v1alpha1/common.go
```diff
@@ -1,4 +1,4 @@
-package pkg
+package v1alpha1
```
andreyvelich: Maybe name this package `common` ?

hougangliu: based on go best practices, package name should keep same as its directory name. all Katit package follows this style.

gaocegege: Yeah, I think we should keep v1alpha1

PR #493, opened 2019-05-09
chore: Delete useless files

gaocegege:
I am not sure why we have setup.py. But it seems that we can delete it safely.

/assign @YujiOshima 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/493)

### DIFF
--- a/py/study_job_client.py
+++ b/py/study_job_client.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# Notice: The file is used in https://github.com/kubeflow/pipelines/blob/master/components/kubeflow/katib-launcher/Dockerfile, please do not delete the file.
 import datetime
 import json
 import logging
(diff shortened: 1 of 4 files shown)


py/study_job_client.py
```diff
@@ -1,160 +0,0 @@
-# Copyright 2019 Google LLC
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#      http://www.apache.org/licenses/LICENSE-2.0
-#
```
hougangliu: this file is used in https://github.com/kubeflow/pipelines/blob/master/components/kubeflow/katib-launcher/Dockerfile

gaocegege: OK, then we need to keep it. I will add a comment about it.

gaocegege:
@hougangliu PTAL again.

setup.py
```diff
@@ -1,4 +0,0 @@
-from setuptools import setup
```
hougangliu: this file used in python_tests.py, but not sure if it is useful now

gaocegege: I am trying if it works after the deletion.

gaocegege:
It seems that we cannot delete it...

PR #496, opened 2019-05-09
v1alpha2 metrics collector - controller

richardsliu:
Controller-side changes for #442

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/496)

### DIFF
--- a/pkg/controller/v1alpha2/experiment/util/manifest_util.go
+++ b/pkg/controller/v1alpha2/experiment/util/manifest_util.go
@@ -27,7 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/controller/v1alpha2/experiment/util/manifest_util.go
```diff
@@ -27,7 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func getMetricsCollectorManifest(experimentName string, trialName string, jobKind string, namespace string, metricNames []string, mcs *katibv1alpha2.MetricsCollectorSpec) (*bytes.Buffer, error) {
+func GetMetricsCollectorManifest(experimentName string, trialName string, jobKind string, namespace string, metricNames []string, mcs *katibv1alpha2.MetricsCollectorSpec) (*bytes.Buffer, error) {
```
hougangliu: can we use `podSelector []string` to replace `jobKind` here since MetricsCollector use it just for getting corresponding pod?
we can add a function like `getRunPodSelector(jobKind string)`

this is just a tip, it's ok for me for this PR, too

Issue #463, opened 2019-04-20
GetSuggestion API change


johnugeorge:
Currently  GetSuggestionsRequest and GetSuggestionsReply structure are as follows
```
message GetSuggestionsRequest {
    string experiment_name = 1;
    string algorithm_name = 2;
    int32 request_number = 3; ///The number of Suggestion you request at one time. When you set 3 to request_number, you can get three Suggestions at one time.
}

message GetSuggestionsReply {
    repeated Trial trials = 1;
}


```
1. Currently, GetSuggestions and ValidateAlgorithmSettings API calls are called to corresponding suggestion service via Manager Service. (https://github.com/kubeflow/katib/blob/master/cmd/manager/v1alpha2/main.go#L138) Is it actually needed? Why can't the controller directly connect to the suggestion service?

2.Can GetSuggestionsReply have just ParameterAssignments(parameter name-value pair) instead of entire Trial? Suggestion algorithm need not construct entire Trial object which is a overhead. eg: Trial has TrialSpec and TrialStatus.    SuggestionAlgorithm should not call CreateTrial. Instead Creating Trial object from ParameterAssignments  should be left to the controller. 

/cc @richardsliu 
/cc @YujiOshima 
/cc @jdplatt 
/cc @andreyvelich 

WDYT?

richardsliu:
I agree, returning ParameterAssignments is more consist with the semantics of a Get API. Also since we are defining Trial as its own resource type, its lifecycle should be handled by a controller.

andreyvelich:
1. Is it possible to call this functions from controller directly?
2. As well, we send StudyID in Trial structure. You can see it here: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1alpha1/nasrl_service.py#L356.
We can define this structure like that:

```
message GetSuggestionsReply {
   string study_id = 1;
   message ParameterAssignments{
        repeated ParameterAssignment assignments = 1;
    }
    ParameterAssignments parameter_assignments = 2;
}
```
What do you think @johnugeorge ?

richardsliu:
@andreyvelich For NAS wouldn't the GetSuggestionsReply need to contain the architecture and operations?

andreyvelich:
@richardsliu Yes, but we store it in two different categorical parameters, architecture and nn_config: 
https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1alpha1/nasrl_service.py#L357

gaocegege:
[...]

Issue #471, opened 2019-04-25
[discussion][feature request] Experiment Specific Suggestion Support


gaocegege:
There are some algorithms that cannot be running as a long-running service. It needs to be trained first and serve for the specific NAS tasks.

Currently, we create the suggestion services before we run the system, the approach may not work for the case. We may create some suggestions dynamically during the running.

/cc @andreyvelich @DeeperMind

andreyvelich:
It is a good idea. For current RL suggestion https://github.com/kubeflow/katib/tree/master/pkg/suggestion/v1alpha1/NAS_Reinforcement_Learning it will work. This Suggestion can be running not as a long-running service. We can run this suggestion during live of one Experiment.

gaocegege:
/close

We can support it with v1alpha3

PR #500, opened 2019-05-13
feat(suggestion): Define new API for suggestion services.

gaocegege:
Related to #463


gaocegege:
/cc @johnugeorge 

https://github.com/kubeflow/katib/pull/500/commits/6c9f7405a5ea78cbac9e59bd14fb6f0d78422e54

I am working on the investigation of the new API to make sure if it could support all cases.

pkg/api/suggestion/v1alpha3/suggestion.proto
```diff
@@ -0,0 +1,110 @@
...
+message GetSuggestionsReply {
+    repeated Trial trials = 1; // trials should be created in the next run.
+}
+
+message Experiment {
+    string name = 1;
+    ExperimentSpec experiment_spec = 2;
+}
+
+message ExperimentSpec {
```
andreyvelich: You don't want to add `NasConfig` in `ExperimentSpec` on that step of implementation?

gaocegege: This PR will add it, I think. It is still WIP.

andreyvelich: Cool, thank you!

johnugeorge:
@gaocegege 
When I thought a little more, Implementing Suggestion CR might be a better long term solution instead of this change.  Experiment controller requests suggestions using Suggestion CR. Suggestion Controller talks to algorithm service and populates the suggestion results.

With Suggestion CR and controller,
1. This design can take care of long running experiments. Suggestion CR is created per experiment  by the experiment controller. 
2. State or algorithm settings(key-value pair) will be stored in the CR instead of external DB. This will make it k8s native. Suggestion controller can thus send just the required information to each algorithm service instead of sending whole experiment and trails(in the current PR). 
  
Please share your thoughts.

gaocegege:
@johnugeorge 

I want to introduce the suggestion CRD to solve https://github.com/kubeflow/katib/issues/471

As for the algorithm settings, I think your idea is awesome. If we could save the state in suggestion CR, it will be painless. Do you have a further design about it? In my initial design, the suggestion controller is only responsible for creating suggestions for the experiments. When the suggestions are created, the experiment controller will communicate with suggestion CR directly, thus I did not think about storing the state of suggestions in CR.

But inspired by your thoughts, I think it is better. But I did not have the idea about how to do it. If we store the state in Suggestion CR, how to send it to the Suggestion itself?

PR #501, opened 2019-05-13
feat(api): Add total number of trials

### LINKED ISSUE #499: [feature] Add Trials in Experiment Spec
https://github.com/kubeflow/katib/blob/master/pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go

We have TrialsSucceeded, TrialsFailed and other fields, but we do not have a field for the total number of the trials. We need it.

gaocegege:
Fix #499 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/501)

### DIFF
--- a/pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
+++ b/pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
@@ -47,7 +47,7 @@ type ExperimentSpec struct {
 	// Whether to retain historical data in DB after deletion.
 	RetainHistoricalData bool `json:"retainHistoricalData,omitempty"`
 
-	// For v1alpha2 we will keep the metrics collector implementation same as v1alpha1. 
+	// For v1alpha2 we will keep the metrics collector implementation same as v1alpha1.
 	MetricsCollectorSpec *MetricsCollectorSpec `json:"metricsCollectorSpec,omitempty"`
 
 	NasConfig *NasConfig `json:"nasConfig,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
```diff
@@ -79,6 +79,9 @@ type ExperimentStatus struct {
 	// Current optimal trial parameters and observations.
 	CurrentOptimalTrial OptimalTrial `json:"currentOptimalTrial,omitempty"`
 
+	// Trials is the total number of trials owned by the experiment.
+	Trials int `json:"trials,omitempty"`
```
andreyvelich: What do you think about `TotalTrials` name?

gaocegege: I think in k8s community, we prefer trials. See also https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L185 and https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L383

andreyvelich: Ok

gaocegege:
Weird, https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_katib/501/kubeflow-katib-presubmit/1128140356542533632/ shows succeeded, but the bot told me failed.

gaocegege:
Rebased, PTAL

Issue #495, opened 2019-05-09
[RFC] Directory Refactor


gaocegege:
Now the directory layout is a little confusing:

```
./pkg
├── api
│   ├── health
│   ├── operators
│   │   └── apis
│   │       ├── common
│   │       │   └── v1alpha2
│   │       ├── experiment
│   │           └── v1alpha2
│   │       ├── studyjob
│   │       │   └── v1alpha1
│   │       └── trial
│   │           └── v1alpha2
│   ├── v1alpha1
│   └── v1alpha2
├── controller
│   ├── v1alpha1
│   │   └── studyjob
│   └── v1alpha2
│       ├── experiment
│       └── trial
├── db
│   ├── v1alpha1
│   └── v1alpha2
├── earlystopping
│   └── v1alpha1
├── manager
│   ├── modelstore
│   └── v1alpha1
│       ├── file-metricscollector
│       ├── metricscollector
│       ├── studyjobclient
│       └── visualise
│           └── tensorboard
├── mock
│   ├── modelstore
│   ├── v1alpha1
│   └── v1alpha2
├── suggestion
│   ├── v1alpha1
├── ui
└── util
    └── v1alpha2
```

I propose some changes:

- Move `pkg/mock/modelstore` to `pkg/mock/v1alpha1/db`. Maybe we need to change the generate script to do it automatically.
- Move `pkg/manager/modelstore` to `pkg/manager/v1alpha1`, since it is version specific implementation. And it will be changed in v1alpha2.
- Change `pkg/controller/v1alpha2/experiment` to `pkg/controller/experiment/v1alpha2/`. Becuase the API layout is `pkg/api/operators/apis/experiment/v1alpha2`
- Change `pkg/controller/v1alpha2/trial` to `pkg/controller/trial/v1alpha2/`
- Change `pkg/controller/v1alpha1/studyjob` to `pkg/controller/studyjob/v1alpha2/`
- Rename `pkg/api/operators` to `pkg/apis/controller/`. The new name follows the same convention with `pkg/controller`
- Put `pkg/api/v1alpha1` and `pkg/api/v1alpha2` in `pkg/apis/manager`. At the same time, make API proto about suggestions in `pkg/apis/suggestion/v1alpha1` and `pkg/apis/suggestion/v1alpha2`. Becuase it will be called directly in v1alpha2.

The new layout will look like:

```diff
./pkg
+── apis
-── api
│   ├── health
-   ├── operators
+   ├── controller
│   │   ├── common
[...]

PR #502, opened 2019-05-13
refactor(suggestion): Use interface

gaocegege:
There are some benefits to define an interface instead of using util.GetSuggestions function:

- Easy to test and maintain. We can write unit test cases for the interface.
- Easy to integrate. We can mock the suggestion when testing the experiment controller, or use a fake controller to run the experiment without affecting the real logic.

This PR:

- added a CLI argument to control if we use fake implementation in the controller. If not, now we will always return an error becuase we do not implement suggestion logic now.

```
  -fake-experiment-suggestion
    	Use the fake suggestion in experiment controller
````

- added a interface about suggestion.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/502)

### DIFF
--- a/cmd/katib-controller/v1alpha2/main.go
+++ b/cmd/katib-controller/v1alpha2/main.go
@@ -39,17 +39,17 @@ func main() {
 	logf.SetLogger(logf.ZapLogger(false))
 	log := logf.Log.WithName("entrypoint")
 
-	var fakeExperimentSuggestion string
+	var experimentSuggestionName string
 
-	flag.StringVar(&fakeExperimentSuggestion, "experiment-suggestion-implementation",
+	flag.StringVar(&experimentSuggestionName, "experiment-suggestion-name",
 		"default", "The implementation of suggestion interface in experiment controller (default|fake)")
 
 	flag.Parse()
 
-	viper.Set(consts.ConfigFakeExperimentSuggestion, fakeExperimentSuggestion)
+	viper.Set(consts.ConfigExperimentSuggestionName, experimentSuggestionName)
 	log.Info("Config:",
-		consts.ConfigFakeExperimentSuggestion,
-		viper.GetString(consts.ConfigFakeExperimentSuggestion))
+		consts.ConfigExperimentSuggestionName,
+		viper.GetString(consts.ConfigExperimentSuggestionName))
 
 	// Get a config to talk to the apiserver
 	cfg, err := config.GetConfig()
--- a/cmd/katib-controller/v1alpha2/main.go
+++ b/cmd/katib-controller/v1alpha2/main.go
@@ -39,10 +39,10 @@ func main() {
 	logf.SetLogger(logf.ZapLogger(false))
 	log := logf.Log.WithName("entrypoint")
 
-	var fakeExperimentSuggestion bool
+	var fakeExperimentSuggestion string
 
-	flag.BoolVar(&fakeExperimentSuggestion, "fake-experiment-suggestion",
-		false, "Use the fake suggestion in experiment controller")
+	flag.StringVar(&fakeExperimentSuggestion, "experiment-suggestion-implementation",
+		"default", "The implementation of suggestion interface in experiment controller (default|fake)")
 
 	flag.Parse()
 
--- a/cmd/katib-controller/v1alpha2/main.go
+++ b/cmd/katib-controller/v1alpha2/main.go
@@ -20,49 +20,70 @@ limitations under the License.
 package main
 
 import (
-	"log"
+	"flag"
+	"os"
 
-	"github.com/kubeflow/katib/pkg/api/operators/apis"
-	controller "github.com/kubeflow/katib/pkg/controller/v1alpha2"
+	"github.com/spf13/viper"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by gaocegege** [comment]
/cc @johnugeorge @richardsliu @YujiOshima 

PTAL

File: pkg/controller/v1alpha2/experiment/experiment_controller.go [hunk 1]
```diff
@@ -36,12 +37,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/builder"
 
-	experimentsv1alpha2 "github.com/kubeflow/katib/pkg/api/operators/apis/experiment/v1alpha2"
-	trialsv1alpha2 "github.com/kubeflow/katib/pkg/api/operators/apis/trial/v1alpha2"
+	experimentv1alpha2 "github.com/kubeflow/katib/pkg/api/operators/apis/experiment/v1alpha2"
```
johnugeorge: I feel, Kubebuilder recommends plural form of CR.  Ideally, it should have been `apis/experiments`, `apis/trials` as per Kubebuilder notation. I was planning to take this up post 0.6 release

gaocegege: @johnugeorge SGTM. I reverted the change. Maybe we could rename the package in #495

johnugeorge: Sounds good

cmd/katib-controller/v1alpha2/main.go
```diff
@@ -20,49 +20,70 @@ limitations under the License.
...
 )
 
 func main() {
 	logf.SetLogger(logf.ZapLogger(false))
+	log := logf.Log.WithName("entrypoint")
+
+	var fakeExperimentSuggestion bool
+
+	flag.BoolVar(&fakeExperimentSuggestion, "fake-experiment-suggestion",
+		false, "Use the fake suggestion in experiment controller")
```
hougangliu: how about use args like `experiment-suggestion-name`, and now its value can be "fake|default", `default` by default. in future, we can may support more.

gaocegege: Good point. I will update it.

gaocegege: Addressed

gaocegege:
@hougangliu PTAL

pkg/controller/v1alpha2/experiment/experiment_controller.go
```diff
@@ -58,7 +62,21 @@ func Add(mgr manager.Manager) error {
...
+		Client: mgr.GetClient(),
+		scheme: mgr.GetScheme(),
+	}
+	// If the flag is set in CLI, use the fake implementation.
+	imp := viper.GetString(consts.ConfigFakeExperimentSuggestion)
+	switch imp {
+	case "fake":
+		log.Info("Using the fake suggestion implementation")
+		r.Suggestion = suggestionfake.New()
+	default:
```
hougangliu: if user sets it as "invalid-one", we can still use default, but a warning log is necessary

gaocegege: SGTM

hougangliu:
> @hougangliu PTAL
@gaocegege Thanks. some mini comment

gaocegege:
> > @hougangliu PTAL
> > @gaocegege Thanks. some mini comment

Addressed, Thanks for your review. PTAL

hougangliu:
/lgtm
thanks

pkg/controller/v1alpha2/experiment/experiment_controller.go
```diff
@@ -58,7 +62,30 @@ func Add(mgr manager.Manager) error {
...
+		scheme: mgr.GetScheme(),
+	}
+	imp := viper.GetString(consts.ConfigExperimentSuggestionName)
+	r.Suggestion = newSuggestion(imp)
+	return r
+}
+
+// newSuggestion returns the new Suggestion for the given config.
+func newSuggestion(implementation string) suggestion.Suggestion {
+	// If the flag is set in CLI, use the fake implementation.
```
YujiOshima: The comment is outdated.

gaocegege: Thanks for your review, addressed, PTAL again https://github.com/kubeflow/katib/pull/502/commits/04b2032fc379203cca0a5f93a33ef7eda154d35d

gaocegege:
@hougangliu PTAL again, fixed a bug.

pkg/controller/v1alpha2/experiment/experiment_controller.go
```diff
@@ -58,7 +62,30 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileExperiment{Client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	r := &ReconcileExperiment{
+		Client: mgr.GetClient(),
+		scheme: mgr.GetScheme(),
+	}
+	imp := viper.GetString(consts.ConfigExperimentSuggestionName)
```
johnugeorge: Just to check: Is there a better way to pass flags to the manager than using viper?

gaocegege: I have no idea. Kubebuilder does not support a built-in way, thus I use viper here.

gaocegege: Does it work for you?

PR #509, opened 2019-05-13
create experiment in db

hougangliu:
Part of #510 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/509)

### DIFF
--- a/pkg/common/v1alpha2/common.go
+++ b/pkg/common/v1alpha2/common.go
@@ -2,8 +2,11 @@ package v1alpha2
 
 import (
 	"os"
+	"context"
 
 	experimentsv1alpha2 "github.com/kubeflow/katib/pkg/api/operators/apis/experiment/v1alpha2"
+	api_pb "github.com/kubeflow/katib/pkg/api/v1alpha2"
+	"google.golang.org/grpc"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/common/v1alpha2/common.go
```diff
@@ -29,3 +37,27 @@ func GetManagerAddr() string {
...
+		return nil, err
+	}
+	kc := &katibClientAndConnection {
+		Conn: conn,
+		KatibClient: api_pb.NewManagerClient(conn),
+	}
+	return kc, nil
+}
+
+func RegisterExperiment(request *api_pb.RegisterExperimentRequest) (*api_pb.RegisterExperimentReply, error) {
```
johnugeorge: Does it have to be put in api_util itself? I feel that We should keep grpc api in one place and It doesn't fit to be in common.

hougangliu: I think we should put it into common part since both experiment and trial, or even suggestion service will call it to access DB by katib-manager grpc

johnugeorge: When would `trial` need to call `RegisterExperiment`?

hougangliu: I mean I will also add `RegisterTrial` into this file, both `RegisterTrial`  and `RegisterExperiment` need katib-client to call katib-manager to visit DB.

pkg/common/v1alpha2/common.go
```diff
@@ -15,6 +18,11 @@ const (
 	ManagerAddr                   = KatibManagerService + ":" + KatibManagerPort
 )
 
+type katibClientAndConnection struct {
+	Conn *grpc.ClientConn
+	KatibClient api_pb.ManagerClient
```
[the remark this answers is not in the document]
hougangliu: Ok, I will rename it to `KatibManagerClient`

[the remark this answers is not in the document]
hougangliu: done, thanks

PR #519, opened 2019-05-15
Delete experiment in DB if in need

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/519)

### DIFF
--- a/pkg/api/operators/apis/experiment/v1alpha2/util.go
+++ b/pkg/api/operators/apis/experiment/v1alpha2/util.go
@@ -22,6 +22,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	cleanDataFinalizer = "clean-data-in-db"
+)
+
 func getCondition(exp *Experiment, condType ExperimentConditionType) *ExperimentCondition {
 	for _, condition := range exp.Status.Conditions {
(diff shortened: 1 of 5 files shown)


pkg/api/operators/apis/experiment/v1alpha2/util.go
```diff
@@ -137,3 +141,33 @@ func (exp *Experiment) MarkExperimentStatusFailed(reason, message string) {
...
+			contained = true
+			break
+		}
+	}
+
+	if !deleted && !contained {
+		if exp.Spec.RetainHistoricalData {
+			return false, []string{}
+		}
+		finalizers := append(pendingFinalizers, cleanDataFinalizer)
```
YujiOshima: Why do we need `cleanDataFinalizer` and exclude it at line 166?

hougangliu: by k8s [finalizer feature](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers), we can make sure when "kubectl delete experiment xx", the related data in DB deleted before experiment resource deleted in k8s.

YujiOshima: I see. Thanks.

PR #520, opened 2019-05-15
Fix prow to trigger corresponding workflow

hougangliu:
files in `manifests`, `scripts` and `test/e2e` are used for test, they are useful during presubmit, but for postsubmit, these files should not trigger corresponding workflow

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/520)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -7,7 +7,8 @@ workflows:
     job_types:
       - presubmit
     include_dirs:
-    - pkg/api/operators/*
+    - pkg/api/operators/apis/studyjob
+    - pkg/api/operators/apis/*.go
     - pkg/api/health/*
     - pkg/api/v1alpha1/*
     - pkg/common/v1alpha1/*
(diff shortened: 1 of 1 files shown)


prow_config.yaml
```diff
@@ -7,7 +7,8 @@ workflows:
     job_types:
       - presubmit
     include_dirs:
-    - pkg/api/operators/*
+    - pkg/api/operators/apis/studyjob
```
gaocegege: Maybe pkg/api/operators/apis/studyjob/* is better

hougangliu: done! thanks

richardsliu: Yeah, it does a regex match, so this needs to have the /*

richardsliu: See https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/run_e2e_workflow.py#L187

PR #524, opened 2019-05-15
Fix ut test and enable ut-test of v1alpha2

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/524)


hougangliu:
> dep ensure

In test workflow, `dep-ensure` run it

Issue #510, opened 2019-05-13
DB operations for v1alpha2


hougangliu:
- [x] RegisterExperiment when Experiment CR created #509
- [x] DeleteExperiment when Experiment CR deleted and `RetainHistoricalData` is false #519
- [x] UpdateExperimentStatus when Experiment CR status changed #534
- [x] RegisterTrial when TrialCR created #530 
- [x] DeleteTrial when Trial CR deleted and `RetainHistoricalData` is false (I think when delete experiment, its trials will be deleted in db automatically)
- [x] UpdateTrialStatus when Trial CR status changed #537 
- [x] ReportObservationLog when metricCollector get metrics and insert it into DB. #539 484

johnugeorge:
Can you group these issues? It will be easier to track i feel

andreyvelich:
@hougangliu I will take `RegisterTrial when TrialCR created` from this list.
/cc @johnugeorge

richardsliu:
@hougangliu This is done now, right?

hougangliu:
right, maybe we can close it now

PR #530, opened 2019-05-16
Register Trial in DB

andreyvelich:
I added part to register Trial in DB.
As well, I rename Status and Spec to TrialStatus and TrailSpec here: https://github.com/andreyvelich/katib/blob/issue-510-register-trial-db/pkg/api/v1alpha2/api.proto#L381,
like we have with Experiment.

/cc @hougangliu @johnugeorge @richardsliu @gaocegege 

Part of https://github.com/kubeflow/katib/issues/510.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/530)


pkg/api/v1alpha2/api.proto
```diff
@@ -378,8 +378,8 @@ message TrialStatus {
 
 message Trial {
     string name = 1;
-    TrialSpec spec = 2;
-    TrialStatus status = 3;
+    TrialSpec trial_spec = 2;
```
hougangliu: In fact, I like spec/status rather than trial_spec/trial_status. Maybe we should refactor experiment_spec/experiment_status instead of here

andreyvelich: I can rename experiment_spec to spec, if we want to. 
What do you think? @johnugeorge @gaocegege @richardsliu

gaocegege: Prefer spec/status but it may require lots of work to update the code base.

pkg/common/v1alpha2/katib_manager_util.go
```diff
@@ -87,3 +87,14 @@ func DeleteExperiment(request *api_pb.DeleteExperimentRequest) (*api_pb.DeleteEx
...
 	return kc.DeleteExperiment(ctx, request)
 }
+
+func RegisterTrial(request *api_pb.RegisterTrialRequest) (*api_pb.RegisterTrialReply, error) {
+	ctx := context.Background()
+	kcc, err := getKatibManagerClientAndConn()
+	if err != nil {
+		return nil, err
+	}
+	defer kcc.Conn.Close()
```
gaocegege: It is a little hack here. Could we define a func in katibManagerClientAndConn to close the conn?

pkg/controller/v1alpha2/trial/util/api_util.go
```diff
@@ -18,11 +18,20 @@ package util
 import (
 	//v1 "k8s.io/api/core/v1"
```
gaocegege: Please remove the never used package here

pkg/api/v1alpha2/api.proto
andreyvelich: I will update it in this PR, no problem.

gaocegege: Thanks

andreyvelich:
@gaocegege @hougangliu Done.

andreyvelich:
@gaocegege @hougangliu @johnugeorge I fixed all tests, I think, we can merge this PR.

pkg/controller/v1alpha2/trial/util/api_util.go
```diff
@@ -35,3 +42,51 @@ func GetTrialObservation(instance *trialsv1alpha2.Trial) error {
...
+			&api_pb.ParameterAssignment{
+				Name:  p.Name,
+				Value: p.Value,
+			})
+	}
+
+	trial.Spec.RunSpec = instance.Spec.RunSpec
+
+	trial.Spec.MetricsCollectorSpec = instance.Spec.MetricsCollectorSpec
+
```
hougangliu: miss trial.Status?

andreyvelich:
@hougangliu I have added trial status and change status names in `api.proto` like in trial_types.go.

PR #528, opened 2019-05-15
feat(trial): Refactor trial controller and add basic test cases

gaocegege:
**What this PR does / why we need it**:

I make DB an interface, and mock it during the unit test.

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/528)

PR #532, opened 2019-05-16
fix: Avoid side effect

### LINKED ISSUE #531: [bug] Side effect in Trial controller spawnMetricsCollector
/kind bug

Run trial controller and create a trial, the metrics collector will be created multiple times. We do not check if the resource is created in the cluster. Thus it will return an error and run the endless reconcile loop

/assign

**Environment:**

- Kubeflow version:
- Minikube version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

gaocegege:

Fix #531 

We will meet an error during the creation and retry it again and again. I strongly suggest using state compare pattern instead of current state machine based implementation. But it is another topic.

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/532)

### DIFF
--- a/pkg/controller/v1alpha2/trial/trial_controller.go
+++ b/pkg/controller/v1alpha2/trial/trial_controller.go
@@ -189,10 +189,20 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1alpha2.Trial) error {
 		logger.Error(err, "Job Spec Get error")
 		return err
 	}
+	desiredMetricsCollector, err := r.getDesiredMetricsCollectorSpec(instance)
+	if err != nil {
+		logger.Error(err, "Metrics Collector Get error")
+		return err
+	}
 
 	deployedJob, err := r.reconcileJob(instance, desiredJob)
 	if err != nil {
 		logger.Error(err, "Reconcile job error")
+		return err
+	}
+	_, err = r.reconcileMetricsCollector(instance, desiredMetricsCollector)
+	if err != nil {
+		logger.Error(err, "Reconcile Metrics Collector error")
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/v1alpha2/trial/trial_controller.go
```diff
@@ -224,24 +234,20 @@ func (r *ReconcileTrial) reconcileJob(instance *trialsv1alpha2.Trial, desiredJob
...
 		if errors.IsNotFound(err) {
-			logger.Info("Creating Job", "kind", kind)
+			logger.Info("Creating Job", "kind", kind,
+				"name", desiredJob.GetName())
 			err = r.Create(context.TODO(), desiredJob)
 			if err != nil {
 				logger.Error(err, "Create job error")
 				return nil, err
 			}
+			deployedJob = desiredJob
```
johnugeorge: Is this needed?  deployed job is returned if job is already in running state. In reconcile function(https://github.com/kubeflow/katib/pull/532/files#diff-63bd3a953ae0c850443cefebf8365018R21), update status if deployed job is not null

johnugeorge:
Thanks. This is a required change. I was about to give the  comment in https://github.com/kubeflow/katib/pull/496 before it got merged :)

pkg/controller/v1alpha2/trial/trial_controller.go
gaocegege: If we do not set it, deployedJob is an empty job without status. And we have to update it in the next reconcile loop. I think we can do it in this loop.

gaocegege:
@johnugeorge Yeah my mistake. I did not find this potential problem. And when I am running unit test in #528 I find the logic has bugs.

pkg/controller/v1alpha2/trial/trial_controller.go
gaocegege: WDYT, actually we keep an eventually consistent model, thus it is not a critical change.

gaocegege: Let's remove it since I think it is not a necessary change.

gaocegege:
@johnugeorge PTAL

pkg/controller/v1alpha2/trial/trial_controller.go
johnugeorge: OK. agree

gaocegege:
I am not why why bot does not merge the PR automatically.

johnugeorge:
Tests failed @gaocegege

gaocegege:
So the CI will be run again after we approve one PR.

PR #537, opened 2019-05-16
Update trial status DB operation

hougangliu:
Part of #510

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/537)


pkg/controller/v1alpha2/experiment/experiment_controller.go
```diff
@@ -233,7 +232,6 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 			logger.Error(err, "Create experiment in DB error")
 			return reconcile.Result{}, err
 		}
-		requeue = true
```
johnugeorge: why is this removed?

pkg/controller/v1alpha2/experiment/experiment_controller.go
```diff
@@ -257,7 +255,7 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
-	return reconcile.Result{Requeue: requeue}, nil
```
johnugeorge: Same comment?

pkg/controller/v1alpha2/experiment/experiment_controller.go
hougangliu: in this loop, instance status changed or fails(line 234). for status changed case, there is no need to requeue since changes of obj will make it into work queue again; for db write fail case, the obj key will not be pop out of the queue, similar as `Requeue: true`

pkg/controller/v1alpha2/experiment/experiment_controller.go
hougangliu: after my change, `requeue` will always false, and reconcile.Result.Requeue is 'false' by default

pkg/controller/v1alpha2/experiment/experiment_controller.go
johnugeorge: Can you check if reconcile function will be called immediately or wait for the sync period?

hougangliu: if `Requeue` is true even update done, it will be called immediately. Then we calls Get operation, the return nearly always is obsolete data (That's why I find I need do this change).

gaocegege:
Please rebase the master, Thanks

PR #540, opened 2019-05-17
Adding random suggestion algorithm implementation and manifests

johnugeorge:
This PR implements the random suggestion algorithm.  I have tested it independently using API manager.  Controller changes will be added separately. 

Related: #407 

Fixes are made into the base changes from #435

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/540)


pkg/suggestion/v1alpha2/test.py
```diff
@@ -0,0 +1,54 @@
+#from . import parsing_utils
```
gaocegege: Could we name it random_service_test?

johnugeorge: I will remove the file for now.  It is not complete

johnugeorge: We can add tests separately

gaocegege: SGTM

pkg/suggestion/v1alpha2/random_service.py
```diff
@@ -0,0 +1,44 @@
...
+from pkg.api.v1alpha2.python import api_pb2_grpc
+import grpc
+from . import parsing_util
+
+class RandomService(api_pb2_grpc.SuggestionServicer):
+    def __init__(self):
+        self.manager_addr = "katib-manager"
+        self.manager_port = 6789
+
+    def _get_experiment(self, name):
```
richardsliu: Should this be part of a common lib?

pkg/suggestion/v1alpha2/parsing_util.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    Parse the metric to the dictionary
+    """
+    y_array = np.array(y_train, dtype=np.float64)
+    if goal == api_pb2.MINIMIZE:
+        y_array *= -1
+    return y_array
+
+
+def parse_x_next_vector(x_next, param_types, param_names, discrete_info, categorical_info):
```
richardsliu: Nit - The names of these methods (and the class) feel somewhat weird. "Parse*" seems to suggest taking in a string and extracting the contents from it.

pkg/suggestion/v1alpha2/parameter.py
```diff
@@ -0,0 +1,52 @@
...
+        self.lower_bounds = np.array(lower_bounds).reshape((1, dim))
+        self.upper_bounds = np.array(upper_bounds).reshape((1, dim))
+        self.parameter_types = parameter_types
+        self.names = names
+        self.discrete_info = discrete_info
+        self.categorical_info = categorical_info
+        if len(self.names) != len(set(self.names)):
+            raise Exception("Parameter names are not unique.")
+
+    def create_scaler(self):
```
richardsliu: Do these methods need to be in this class? They seem to be specific to Bayestian optimization and random search respectively, so maybe they should just be in the specific algorithms.

Also the file name should be parameter_config.py.

pkg/suggestion/v1alpha2/parsing_util.py
```diff
@@ -0,0 +1,144 @@
...
+                    if ci["name"] == p.name:
+                        value_num = ci["values"].index(p.value)
+                        parsed_X[row_idx, offset + value_num] = 1
+                        offset += ci["number"]
+    return parsed_X
+
+
+def parse_metric(y_train, goal):
+    """
+    Parse the metric to the dictionary
```
richardsliu: This doesn't return a dictionary?

pkg/suggestion/v1alpha2/parsing_util.py
```diff
@@ -0,0 +1,144 @@
...
+    return feasible_values[np.argmin(diff)]
+
+
+def _deal_with_categorical(feasible_values, one_hot_values):
+    """ function to do the one hot encoding of the categorical values """
+    index = np.argmax(one_hot_values)
+    return feasible_values[int(index)]
+
+
+def parse_parameter_configs(parameter_configs):
```
richardsliu: Seems like this would belong in the ParameterConfig class.

PR #486, opened 2019-05-04
Katib UI for v1alpha2

andreyvelich:
This PR about new UI for v1alpha2 version of Katib built with React. 
The UI supports HP and NAS jobs.
You can find all functionality of this UI in the Youtube link: 
https://www.youtube.com/watch?v=WAK37UW7spo

Fixes: https://github.com/kubeflow/katib/issues/373,  https://github.com/kubeflow/katib/issues/443.

Backend is done, but need to be tested and fixed small issues.
Frontend needs to be changed according to new Backend. I marked in Backend what needs to be done. I have started to work on Frontend.
As well, I refactored folder structure to support both UI for v1alpha1 and v1alpha2 version.

Please, start to review it.
/assign @richardsliu @hougangliu @johnugeorge @YujiOshima 
/cc @gaocegege @MattSich @jlewi 

Developed by: @Akado2009 @andreyvelich

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/486)

PR #544, opened 2019-05-17
Earlystopping changes trigger CI based on version

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/544)


johnugeorge:
do we still need to keep include dirs in Katib prow?  I feel that it is error prone.

@richardsliu @gaocegege

hougangliu:
Basically we will seldom change prow soon since after Katib components finished one by one. But including files in prow can reduce workload of CI in long term.
So I suggest to keep include dirs in Katib prow

hougangliu:
#486 seems to have fix ui part, maybe this PR can fix the last part: earlystop
/hold cancel

PR #558, opened 2019-05-20
Implement GetExperimentInDB

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/558)


pkg/common/v1alpha2/katib_manager_util.go
```diff
@@ -134,4 +134,15 @@ func GetSuggestions(request *api_pb.GetSuggestionsRequest) (*api_pb.GetSuggestio
...
+func GetExperiment(request *api_pb.GetExperimentRequest) (*api_pb.GetExperimentReply, error) {
+	ctx := context.Background()
+	kcc, err := getKatibManagerClientAndConn()
+	if err != nil {
+		return nil, err
+	}
+	defer closeKatibManagerConnection(kcc)
+	kc := kcc.KatibManagerClient
+	return kc.GetExperiment(ctx, request)
 }
```
gaocegege: New line here please.

pkg/controller/v1alpha2/experiment/validator/validator.go
```diff
@@ -140,8 +141,13 @@ func (g *DefaultValidator) validateSupportedJob(job *unstructured.Unstructured)
 }
 
 func (g *DefaultValidator) validateForCreate(inst *experimentsv1alpha2.Experiment) error {
+	isErrNoRows := func (e error) bool {
```
gaocegege: Why should we use such a func instead of sql.ErrNoRows, is there any problem that we want to avoid?

hougangliu: sql.ErrNoRows is raised by db interface in grpc server, in grpc client side,  we cannot get sql.ErrNoRows, which has been wrapped.

johnugeorge: It looks hacky. Just wondering if there is a better way to handle this

johnugeorge: I feel that DB errors should not be exposed to controller

gaocegege:
@richardsliu Could we allow the decrease of coveralls? We can set a floor for it. Maybe 50%

pkg/controller/v1alpha2/experiment/validator/validator.go
gaocegege: Could we catch the error in GetExperimentFromDB? If the error is  sql.ErrNoRows we return nil, or we return error to the controller.

richardsliu:
@gaocegege We can set a failure threshold at 50%, and a coverage decrease threshold at maybe 5%. Does that sound ok?

gaocegege:
SGTM, thanks.

richardsliu:
@gaocegege Done.

gaocegege:
@hougangliu Please rebase, then I think we could merge it.

pkg/controller/v1alpha2/experiment/validator/validator.go
hougangliu: In fact, I don't like handling like this. I will try to investigate what general solution of gprc error is
thanks for your comments

hougangliu: Per investigation, we have to parse the Error message of GetExperimentFromDB (in grpc client side) to decide if grpc server GetExperiment error is sql.ErrNoRows or not owning to grpc implementation.
Considering that in future release, if we support more db backend, GetExperiment may not return sql.ErrNoRows, it can be modeldb.NoRecords and so on. So I prefer to define a general new api as below (since here call GetExperiment to check if a same name experiment exists in DB).
WDYT?
```
rpc PreCheckRegisterExperiment(RegisterExperimentRequest) returns (PreCheckRegisterExperimentReply, error)

message PreCheckRegisterExperimentReply {
    bool can_register = 1;
}
```

johnugeorge: how will you handle similar errors for other api calls?  eg: Alreadyexists for RegisterExperiment etc

hougangliu: For others APIs, our use cases don't need care what the error type is now when error is not nil.

hougangliu:
PTAL @gaocegege @johnugeorge @richardsliu

pkg/controller/v1alpha2/experiment/validator/validator.go
gaocegege: Can we update managerclient.go as follows:

```go
func (d *DefaultClient) GetExperimentFromDB(instance *experimentsv1alpha2.Experiment) (*api_pb.GetExperimentReply, error) {
	request := &api_pb.GetExperimentRequest{
		ExperimentName: instance.Name,
	}
	if exp, err := commonv1alpha2.GetExperiment(request); err != nil {
		if !isErrNoRows(err) {
			return nil, fmt.Errorf("Fail to check record for the experiment in DB: %v", err)
		}
	}
	return exp, nil
}
```

Then the error will be handled in manager client and not be exposed to the controller.

hougangliu: I had used "PreCheckRegisterExperiment" to replace "GetExperimentFromDB" in latest commit. I think we should avoid to parse Error message string to judge what its original error in grpc server side is, especially when we support more DB backends, we must refactor `isErrNoRows` to parse the error message string to match other backends ErrNoRows counterparts.

gaocegege: SGTM

johnugeorge:
LGTM

In general, i feel that the right solution is to have a translation layer between DB and GRPC to translate the DB errors to common set of errors defined for GRPC that can be used by any caller(eg: controller, api user, UI etc).  I am sure that there will be more errors to be differentiated at the caller level( controller here) and this solution might not scale. 
Also, returning DB error directly to the caller is also not good IMO as we are exposing backend details to the user.

hougangliu:
> LGTM
> 
> In general, i feel that the right solution is to have a translation layer between DB and GRPC to translate the DB errors to common set of errors defined for GRPC that can be used by any caller(eg: controller, api user, UI etc). I am sure that there will be more errors to be differentiated at the caller level( controller here) and this solution might not scale.
> Also, returning DB error directly to the caller is also not good IMO as we are exposing backend details to the user.

For now, the reason why we didn't do the translation is that we don't care what the error exactly is except this validation case. So I don't think it's worth spending much time to add  a translation layer  just to hide backend details, at least, its priority is not high. Anyway, we can create a issue to trace discuss for it so that we can do it in coming release

PR #559, opened 2019-05-20
Added metric name to GetObservationLogRequest

johnugeorge:
Added optional field 'metric_name' to GetObservationLogRequest in order to filter logs based on metric name if set

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/559)


pkg/api/v1alpha2/api.proto
```diff
@@ -491,8 +492,9 @@ message ReportObservationLogReply {
 
 message GetObservationLogRequest {
     string trial_name = 1;
-    string start_time = 2; ///The start of the time range. RFC3339 format
-    string end_time = 3; ///The end of the time range. RFC3339 format
+    string metric_name = 2;
```
hougangliu: how about `s/string metric_name/repeat string metric_name_list` to make it more scalable

johnugeorge: Can we target this enhancement in a separate PR? Currently, controller will need to pass only one metric name which would be the objective metric name.

hougangliu: sure

Issue #551, opened 2019-05-17
implement UpdateTrialStatusObservation


hougangliu:
Now UpdateTrialStatusObservation does nothing, we need it to fill-back trial.Status.Observation

PR #564, opened 2019-05-20
Implement trial observation metrics

johnugeorge:
Fixes: #551 

With this PR , e2e experiment run is completed with metric collection.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/564)

### DIFF
--- a/pkg/controller/v1alpha2/trial/util/status_util.go
+++ b/pkg/controller/v1alpha2/trial/util/status_util.go
@@ -131,10 +131,6 @@ func getBestObjectiveMetricValue(metricLogs []*api_pb.MetricLog, objectiveType c
 	metricLogSize := len(metricLogs)
 	if metricLogSize == 0 {
 		return nil
-	} else if metricLogSize == 1 {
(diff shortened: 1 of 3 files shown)


pkg/controller/v1alpha2/trial/util/status_util.go
```diff
@@ -96,5 +99,58 @@ func UpdateTrialStatusCondition(instance *trialsv1alpha2.Trial, deployedJob *uns
...
+		}
+	}
+	return nil
+}
+
+func getBestObjectiveMetricValue(metricLogs []*api_pb.MetricLog, objectiveType commonv1alpha2.ObjectiveType) *float64 {
+	metricLogSize := len(metricLogs)
+	if metricLogSize == 0 {
+		return nil
+	} else if metricLogSize == 1 {
```
richardsliu: Is this case needed? I think the else block would take care of it.

johnugeorge: done

PR #567, opened 2019-05-21
fix: Check if the deletion time is nil or zero

gaocegege:

This PR is to avoid the potential bug of finalizers. The object cannot be deleted without the PR.

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/567)


gaocegege:
/assign @hougangliu 

IsZero returns true if the value is nil or time is zero.

PR #569, opened 2019-05-22
refactor: Use manager client to get log for test

gaocegege:

Now the code to communicate with katib manager to get the log is hard coded in the main workflow. Thus we cannot run unit test because in unit test we do not have such a manager instance. This PR moves the function call to managerclient to mock it in test.

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/569)

### DIFF
--- a/pkg/controller/v1alpha2/trial/trial_controller_util.go
+++ b/pkg/controller/v1alpha2/trial/trial_controller_util.go
@@ -87,6 +87,7 @@ func (r *ReconcileTrial) UpdateTrialStatusObservation(instance *trialsv1alpha2.T
 	objectiveMetricName := instance.Spec.Objective.ObjectiveMetricName
 	reply, err := r.GetTrialObservationLog(instance)
 	if err != nil {
+		log.Error(err, "Get trial observation log error")
 		return err
 	}
 	if reply.ObservationLog != nil {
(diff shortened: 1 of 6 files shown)


pkg/controller/v1alpha2/trial/trial_controller_util.go
```diff
@@ -96,29 +83,23 @@ func UpdateTrialStatusCondition(instance *trialsv1alpha2.Trial, deployedJob *uns
...
-	// read GetObservationLog call and update observation field
+func (r *ReconcileTrial) UpdateTrialStatusObservation(instance *trialsv1alpha2.Trial, deployedJob *unstructured.Unstructured) error {
 	objectiveMetricName := instance.Spec.Objective.ObjectiveMetricName
-	request := &api_pb.GetObservationLogRequest{
-		TrialName:  instance.Name,
-		MetricName: objectiveMetricName,
-	}
-	if reply, err := common.GetObservationLog(request); err != nil {
+	reply, err := r.GetTrialObservationLog(instance)
+	if err != nil {
```
johnugeorge: Can you add a error og here? Missed earlier

gaocegege: SGTM

gaocegege:
@johnugeorge Addressed, PTAL

PR #571, opened 2019-05-22
Fix finalizer cannot work

hougangliu:
#567 make logic opposite

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/571)


hougangliu:
> /lgtm
> Sorry for the bug

That's OK. Everyone creates bug :)

Issue #565, opened 2019-05-20
ignore Retain field


hougangliu:
in experiment.spec.trialTemplate and experiment.spec.metricsCollectorSpec, `Retain` filed is defined, which is used to delete worker job and metricCollecoter job when trial completed, but now we ignore the field when create new trial( Trial struct has no counterpart field of  the two kind of `Retain`)

should we drop `Retain` in experiment and always delete worker job and metricCollecoter job when trial completed? or add two new field in trial.spec?

/cc @richardsliu @gaocegege @johnugeorge @YujiOshima which one do you prefer?

gaocegege:
I have a question may not be related to the issue. How to support the case that the users want to create new trials for the completed experiment? Do we plan to support it? Ref https://github.com/kubeflow/katib/issues/346

johnugeorge:
I think, `Retain` is not needed now.  `RetainHistoricalData` serves the same purpose from the experiment level. Right?

hougangliu:
> I have a question may not be related to the issue. How to support the case that the users want to create new trials for the completed experiment? Do we plan to support it? Ref #346

We need discuss it more in next meeting or in #346. Anyway, I think the feature may be not included in release 0.6

johnugeorge:
> I have a question may not be related to the issue. How to support the case that the users want to create new trials for the completed experiment? Do we plan to support it? Ref #346

Other k8s objects follow the same convention right?  eg: K8s Job

PR #572, opened 2019-05-22
Retain for job and metricsCollector

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/572)

### DIFF
--- a/pkg/controller/v1alpha2/trial/trial_controller.go
+++ b/pkg/controller/v1alpha2/trial/trial_controller.go
@@ -144,11 +144,6 @@ func (r *ReconcileTrial) Reconcile(request reconcile.Request) (reconcile.Result,
 
 	instance := original.DeepCopy()
 
-	if instance.IsCompleted() {
-
-		return reconcile.Result{}, nil
-
-	}
 	if !instance.IsCreated() {
 		//Trial not created in DB
 		if instance.Status.StartTime == nil {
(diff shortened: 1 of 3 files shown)


hougangliu:
Fixes: #565

pkg/controller/v1alpha2/trial/trial_controller.go
```diff
@@ -245,6 +240,9 @@ func (r *ReconcileTrial) reconcileJob(instance *trialsv1alpha2.Trial, desiredJob
 	err = r.Get(context.TODO(), types.NamespacedName{Name: desiredJob.GetName(), Namespace: desiredJob.GetNamespace()}, deployedJob)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			if instance.IsCompleted()  {
```
johnugeorge: Which condition is being targeted here?

hougangliu:
```
func (trial *Trial) IsCompleted() bool {
	return trial.IsSucceeded() || trial.IsFailed() || trial.IsKilled()
}
```

johnugeorge: Is it the case that Get returns a stale entry after trial is completed?

johnugeorge: https://github.com/kubeflow/katib/blob/master/pkg/controller/v1alpha2/trial/trial_controller.go#L149 returns if instance is completed.

hougangliu: do you mean add a comment like `IsCompleted() targets "IsSucceeded() || IsFailed() || IsKilled()"` everywhere IsCompleted() is called?

hougangliu: Line 147~151 deleted in this PR

johnugeorge: Sorry. I missed that change.  Can you explain  the difference? And what I meant, is to add a comment why this check is added.  Readers will find it easier to follow the code.

PR #586, opened 2019-05-24
WIP Do not use /tmp/katib as host path.

gyliu513:

Fixes #

**Release note**:

```release-note
NONE
```

/tmp/katib will disappear after host reboot, it is better use a stable dir.

/cc @hougangliu @gaocegege

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/586)

### DIFF
--- a/manifests/v1alpha2/katib/pv/pv.yaml
+++ b/manifests/v1alpha2/katib/pv/pv.yaml
@@ -12,4 +12,4 @@ spec:
   accessModes:
     - ReadWriteOnce
   hostPath:
-    path: /tmp/katib
+    path: /data/katib


manifests/v1alpha2/katib/pv/pv.yaml
```diff
@@ -12,4 +12,4 @@ spec:
   accessModes:
     - ReadWriteOnce
   hostPath:
-    path: /tmp/katib
+    path: /data/katib
```
gaocegege: /cc @hougangliu 

/lgtm

hougangliu: In the initial version, I use "/data/katib" here. But test cannot pass with error saying "there is no dir /data in host path"-like. I have no idea why it occurs since in v1alpha1, it is "/data/katib" and everything goes well.
LGTM if the test says "LGTM", too :)

gaocegege: Maybe we should create the dir in the host?

gaocegege: No idea. Let's wait the result

andreyvelich: Since we are using hostPath, we need to create this folder in the test cluster for v1alpha2. @gaocegege was right.

hougangliu: I didn't find where we create "/data" in v1alpha1 test. that's weird

PR #587, opened 2019-05-24
Change undeploy script

andreyvelich:
I think, we should move deletion for PV under Katib-db.
Persistent Volume can't be deleted before db.
It will stuck in deletion step, because it used by Katib-db.

/assign @gyliu513 @hougangliu 
/cc @gaocegege @johnugeorge

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/587)


gyliu513:
Good catch, thanks @andreyvelich 

/lgtm

Issue #340, opened 2019-01-23
vizier-db issue with readiness check


jax79sg:
I am currently experiencing an issue with vizier-db pod whereby the readiness probe failed. 

Is the database vizier supposed to be created when the pod is created or am I supposed to create it myself?

The following steps are my troubleshooting process:

Output from `kubectl describe pod -n kubeflow vizier-db-5b59cc744-nh8tp`:
```
Events:
  Type     Reason            Age                 From               Message
  ----     ------            ----                ----               -------
  Warning  FailedScheduling  18m (x25 over 18m)  default-scheduler  persistentvolumeclaim "katib-mysql" is being deleted
  Normal   Pulled            18m                 kubelet, dh-ud     Container image "mysql:8.0.3" already present on machine
  Normal   Created           18m                 kubelet, dh-ud     Created container
  Normal   Started           18m                 kubelet, dh-ud     Started container
  Warning  Unhealthy         17m (x21 over 17m)  kubelet, dh-ud     Readiness probe failed: mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
  Warning  Unhealthy  17m                 kubelet, dh-ud  Readiness probe errored: rpc error: code = Unknown desc = container not running (93f6c700f66ed68bbdaefb4be0a9843eebdcfbc66cf8ad832cc4cf002e8c279d)
  Warning  Unhealthy  3m (x425 over 17m)  kubelet, dh-ud  Readiness probe failed: mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)
```
I have noticed that the readiness check for vizier-db pod is by using: 
`Readiness:      exec [/bin/bash -c mysql -D $$MYSQL_DATABASE -p$$MYSQL_ROOT_PASSWORD -e 'SELECT 1'] delay=5s timeout=1s period=2s #success=1 #failure=3`

So I have tried to run a bash shell in the pod in order to execute the command: 
`mysql -D $$MYSQL_DATABASE -p$$MYSQL_ROOT_PASSWORD -e 'SELECT 1'
`
The output is:
```
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'dh'@'localhost' (using password: YES)
```
Then I tried to execute the command without password:
[...]

PR #588, opened 2019-05-25
Check error in OpenSQLConnection

andreyvelich:
I faced with the same error as here: https://github.com/kubeflow/katib/issues/340.
Problem was with my Coredns in Kubernetes cluster. After I changed pod network from `Flannel` to `Calico`, it was fixed.
I didn't see any logs from the katib-manager and it was hard to found this problem. 
I think, we should print these error messages to the user, then it will be easier to find the problem.

/cc @gaocegege @johnugeorge @hougangliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/588)

### DIFF
--- a/pkg/db/v1alpha1/interface.go
+++ b/pkg/db/v1alpha1/interface.go
@@ -111,9 +111,9 @@ func openSQLConn(driverName string, dataSourceName string, interval time.Duratio
 				if err = db.Ping(); err == nil {
 					return db, nil
 				}
-				klog.Infof("Ping to Katib db failed: %v", err)
+				klog.Errorf("Ping to Katib db failed: %v", err)
 			} else {
-				klog.Infof("Open sql connection failed: %v", err)
+				klog.Errorf("Open sql connection failed: %v", err)
 			}
 		case <-timeoutC:
 			return nil, fmt.Errorf("Timeout waiting for DB conn successfully opened.")
--- a/pkg/db/v1alpha1/interface.go
+++ b/pkg/db/v1alpha1/interface.go
@@ -111,6 +111,9 @@ func openSQLConn(driverName string, dataSourceName string, interval time.Duratio
 				if err = db.Ping(); err == nil {
 					return db, nil
 				}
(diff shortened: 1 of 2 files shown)


pkg/db/v1alpha1/interface.go
```diff
@@ -111,6 +111,9 @@ func openSQLConn(driverName string, dataSourceName string, interval time.Duratio
 				if err = db.Ping(); err == nil {
 					return db, nil
 				}
+				klog.Infof("Ping to Katib db failed: %v", err)
```
gyliu513: We should use `klog.Errorf` here.

pkg/db/v1alpha2/interface.go
```diff
@@ -95,6 +95,9 @@ func openSQLConn(driverName string, dataSourceName string, interval time.Duratio
 				if err = db.Ping(); err == nil {
 					return db, nil
 				}
+				klog.Infof("Ping to Katib db failed: %v", err)
```
gaocegege: Should we return nil, err here?

johnugeorge: @gaocegege This is in a for-select loop. Shouldn't it wait till timeout?

johnugeorge: @andreyvelich Just wondering. Have you seen this log https://github.com/kubeflow/katib/blob/master/cmd/manager/v1alpha2/main.go#L199 after the timeout?

gaocegege: Oh gotcha, yeah we should wait instead of returning an error.

pkg/db/v1alpha1/interface.go
andreyvelich: Thanks, fixed.

pkg/db/v1alpha2/interface.go
andreyvelich: @johnugeorge I didn't check it, because after the timeout my manager container was automatically restarting.

Issue #557, opened 2019-05-20
Hyperband example doesn't terminate


sdgs72:
/kind bug

**What steps did you take and what happened:**
I tried running the hyperband example(katib/examples/v1alpha1/hypb-example.yaml) by running
```sh
kubectl apply -f hypb-example.yaml
```
in my local minikube cluster. (I tried running other algorithm’s such as random and it terminates)

Currently the hyperband study keeps running non stop and doesn’t seem to increase the resource size(num-epochs) at all (i waited for 2 hours and the epoch is still at 1 when it spawn new trials)

**What did you expect to happen:**

The algorithm should increase the resource size given to new trials based on the previous results

**Anything else you would like to add:**

It seems that the hyperband service isn’t able to get the results from previously created trials needed to create child brackets and so it continuously creates master brackets.

Sample logs from hyperband service:
```sh
2019/05/20 04:21:38 HyperBand Suggestion Service
2019/05/20 04:45:58 Unknown Suggestion Parameter SuggestionCount
2019/05/20 04:45:58 Hyb Param sMax 2
2019/05/20 04:45:58 Hyb Param B 27
2019/05/20 04:45:58 Hyb Param n 9
2019/05/20 04:45:58 Hyb Param currentS 2
2019/05/20 04:45:58 Hyb Param r 1
2019/05/20 04:45:58 Hyb Param evaluatingTrials []
2019/05/20 04:45:58 SH loop i = 0
2019/05/20 04:45:58 Make MasterBracket 9 Trials
2019/05/20 04:48:31 Unknown Suggestion Parameter SuggestionCount
2019/05/20 04:48:31 Hyb Param sMax 2
2019/05/20 04:48:31 Hyb Param B 27
2019/05/20 04:48:31 Hyb Param n 9
2019/05/20 04:48:31 Hyb Param currentS 2
2019/05/20 04:48:31 Hyb Param r 1
2019/05/20 04:48:31 Hyb Param evaluatingTrials []
2019/05/20 04:48:31 SH loop i = 0
2019/05/20 04:48:31 Make MasterBracket 9 Trials
```

**Environment:**

- Kubeflow version: v 0.4.1

- Minikube version: minikube version: v0.35.0

- Kubernetes version: (use `kubectl version`):
```sh
[...]

PR #596, opened 2019-05-28
Fix issue of hyperband suggestion service cannot move on

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/596)

### DIFF
--- a/pkg/controller/v1alpha1/studyjob/studyjob_controller.go
+++ b/pkg/controller/v1alpha1/studyjob/studyjob_controller.go
@@ -542,21 +542,6 @@ func (r *ReconcileStudyJobController) checkStatus(instance *katibv1alpha1.StudyJ
 }
 
 func (r *ReconcileStudyJobController) getAndRunSuggestion(instance *katibv1alpha1.StudyJob, c katibapi.ManagerClient, ns string) (bool, error) {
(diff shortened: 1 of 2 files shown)


hougangliu:
Fixes: #557

pkg/controller/v1alpha1/studyjob/studyjob_controller.go
```diff
@@ -542,21 +542,6 @@ func (r *ReconcileStudyJobController) checkStatus(instance *katibv1alpha1.StudyJ
...
-	for i := range sps {
-		if sps[i].Name == "SuggestionCount" {
-			count, _ := strconv.Atoi(sps[i].Value)
-			if count >= instance.Status.SuggestionCount+1 {
-				//Suggestion count mismatched. May be duplicate suggestion request
-				return false, nil
-			}
-			sps[i].Value = strconv.Itoa(instance.Status.SuggestionCount + 1)
-		}
-	}
```
hougangliu: in getSuggestion(line 561), hyperband service will update suggestionParam and store it into DB. If we use `sps` cache from line 546 to update suggestionParam line 614 in DB, it will overwrite values hyperband service updated. That's why hyperband service cannot move on

Issue #607, opened 2019-05-29
Add make check to travis


gyliu513:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

There are often some errors for `make check`, it is better to put the checking to CI.

Related issues for `make check` at https://github.com/kubeflow/katib/pull/606

/cc @hougangliu @gaocegege @richardsliu 

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version:
- Minikube version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

gyliu513:
@hougangliu not clear how to enable `make check` in katib prow at https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubeflow/kubeflow-presubmits.yaml#L82-L92 , do you have any comments for this?

```yaml
kubeflow/katib:
  - name: kubeflow-katib-presubmit
    always_run: true         # Run for every PR, or only when requested.
    branches:
    - master
    labels:
      preset-service-account: "true"
    spec:
      containers:
      - image: gcr.io/kubeflow-ci/test-worker:latest
        imagePullPolicy: Always
```

hougangliu:
you can follow how "run-tests" enabled

johnugeorge:
What do you mean by enabling in prow? You mean, in  presubmit workflow? Alternate solution is to add gometalinter in travis

gyliu513:
@johnugeorge  I mean adding this to CI, so that the CI will run `make check` to help us detect some coding issues.

gaocegege:
@gyliu513 Maybe we could add make check in Travis.

@johnugeorge gometalinter is not maintained any more.

gyliu513:
> @gyliu513 Maybe we could add make check in Travis.

@gaocegege seems adding to travis is more simple, let me fix this in travis instead.

johnugeorge:
LGTM

PR #608, opened 2019-05-29
Enabled make check in travis.

gyliu513:

Fixes https://github.com/kubeflow/katib/issues/607

**Release note**:

```release-note
NONE
```

/cc @gaocegege @hougangliu @johnugeorge

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/608)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: go
 
 go:
-  - "1.10"
-  - "1.11"
-  - "1.12"
+  - "1.12.5"
 
 go_import_path: github.com/kubeflow/katib
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,6 @@ install:
   - go get github.com/mattn/goveralls
 
 script:
+  - make check
   - make test
   - goveralls -coverprofile=coverage.out


.travis.yml
```diff
@@ -1,9 +1,7 @@
 language: go
 
 go:
-  - "1.10"
-  - "1.11"
-  - "1.12"
+  - "1.12.5"
```
gaocegege: Can we use 1.12 here?

gyliu513: I followed kubefed here https://github.com/kubernetes-sigs/kubefed/pull/902 , it is always good to use latest stable version.

gaocegege: Personally, I do not think so. I think we should support multiple go versions.

gyliu513: Different go version sometimes have different standard. In my previous PR, I found that `gofmt` always failed in go1.10 but succeed in go 1.12 and go 1.11, this caused the `make check` failed. Using the latest go can leverage the latest features in Go.

gaocegege: Yeah, go 1.10 can be removed. Format check may be failed in 1.10. I also noticed it.

Issue #615, opened 2019-05-31
Check Metrics Collector before mark Trial succeeded


andreyvelich:
Here https://github.com/kubeflow/katib/blob/master/pkg/controller/v1alpha2/trial/trial_controller_util.go#L49 we mark Trial succeeded right after corresponding Trial job finished. 
After that, we run GetSuggestion and inside Suggestion we, probably, have empty results, if Metrics Collector didn't collect logs.

I think, we should check that Metrics from the Trial were collected and saved in DB before mark Trial as succeeded. 
Metrics collector is the part of the Trial.
What do you think @johnugeorge @richardsliu @gaocegege @hougangliu ?

gaocegege:
Yeah, I think we should have a condition in Trial for metrics collector, for example, metrics-collected. We should check the condition before we set the trial succeeded.

johnugeorge:
`trial.status.Observation` field check should be good enough to verify if metric collection is completed

PR #616, opened 2019-05-31
Set trial completion status only after metric collection

johnugeorge:
Fixes: #615 

Trial status is updated only if trial observation field is updated. (This will be set if metrics are collected)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/616)


johnugeorge:
Will wait for review
/hold

johnugeorge:
@gaocegege Shall we clean up requests in tests in a separate PR?

gaocegege:
Ref https://github.com/kubeflow/katib/issues/619

Yeah

andreyvelich:
Yes, that should work for me. 
Thank you @johnugeorge! 
/lgtm

Issue #610, opened 2019-05-30
Name of the Trial from Suggestion


andreyvelich:
Right now, in v1alpha2, we generate name of the Trial here: https://github.com/kubeflow/katib/blob/master/pkg/controller/v1alpha2/experiment/experiment_util.go#L27 with experiment name + random string of 8 symbols.

In `GetSuggestionsReply` we can return the name of the Trial: https://github.com/kubeflow/katib/blob/master/pkg/api/v1alpha2/api.proto#L525.

I think, we should provide functionality, when user can specify special name to create Trial. 
In that case, from the Controller, we can assign random name for the Trial, if Trial doesn't have a name or name from the `GetSuggestionReply`. 

What do you think @gaocegege @hougangliu @johnugeorge @richardsliu ?

/kind feature

gaocegege:
SGTM

johnugeorge:
This is no more needed as suggestion service gets the list of trials in each GetSuggestions call

johnugeorge:
Closing the issue

PR #624, opened 2019-06-03
Set TrialName from the GetSuggestionReply

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/610.

After this PR, we can set Name of the Trial from the Suggestion.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/624)


johnugeorge:
In longer run, we need to move to a design where Trial is created in controller itself while suggestions give just parameter assignments.  Is there a better way to handle this?

andreyvelich:
@johnugeorge What if Suggestion wants to know which trial was created on the previous run? 
How we can handle, if Suggestion sends only  parameter assignments?
/cc @gaocegege @hougangliu @richardsliu

johnugeorge:
can we get it from the trial list?

andreyvelich:
@johnugeorge If we use TrialList, how Suggestion understands on what steps what trial was created ? 
TrialList returns all Trial names.

johnugeorge:
Holding this PR till we decide the design
/hold

PR #633, opened 2019-06-05
Fix sql syntax for UpdateAlgorithmExtraSettings

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/633)


gaocegege:
https://travis-ci.com/kubeflow/katib/builds/114374171#L564

Test failed

johnugeorge:
Just wondering why this bug was not caught in tests earlier.

hougangliu:
> Just wondering why this bug was not caught in tests earlier.

We never execute “update” operation but "insert"

Issue #625, opened 2019-06-04
Dashboard should not plot a trial if it is not finished


richardsliu:
Currently the UI will plot the graph even if the trials have not finished, with the objective metric assumed to be 0.

This can be misleading - say if we are trying to minimize the error, it would appear as if we've found a set of hyperparameters that leads to 0 error.

PR #646, opened 2019-06-12
UI: Show only succeeded Trials

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/625
/assign @richardsliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/646)


pkg/ui/v1alpha2/backend.go
```diff
@@ -231,33 +231,33 @@ func (k *KatibUIHandler) FetchHPJobInfo(w http.ResponseWriter, r *http.Request)
...
+				log.Printf("GetObservationLog from HP job failed: %v", err)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
 
-		trialResText := make([]string, len(metricsList)+len(paramList))
-		for _, m := range obsLogResp.ObservationLog.MetricLogs {
-			trialResText[metricsList[m.Metric.Name]] = m.Metric.Value
+			trialResText := make([]string, len(metricsList)+len(paramList))
+			for _, m := range obsLogResp.ObservationLog.MetricLogs {
```
hougangliu: same metricName may exist in obsLogResp.ObservationLog.MetricLogs with different value, do you plan to make trialResText stores last value for each metricName?

andreyvelich: This loop will return only the latest value from `obsLogResp.ObservationLog.MetricLogs` in `trialResText`, because they have the same metric name.

johnugeorge:
@hougangliu

PR #647, opened 2019-06-12
feat(experiment): Delete dup trials

gaocegege:

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/647)

### DIFF
--- a/pkg/controller/v1alpha2/experiment/experiment_controller.go
+++ b/pkg/controller/v1alpha2/experiment/experiment_controller.go
@@ -273,7 +273,7 @@ func (r *ReconcileExperiment) ReconcileExperiment(instance *experimentsv1alpha2.
 
 	logger := log.WithValues("Experiment", types.NamespacedName{Name: instance.GetName(), Namespace: instance.GetNamespace()})
 	trials := &trialsv1alpha2.TrialList{}
-	labels := map[string]string{"experiment": instance.Name}
+	labels := map[string]string{consts.LabelExperimentName: instance.Name}
 	lo := &client.ListOptions{}
 	lo.MatchingLabels(labels).InNamespace(instance.Namespace)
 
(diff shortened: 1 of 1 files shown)


pkg/controller/v1alpha2/experiment/experiment_controller.go
```diff
@@ -368,6 +368,26 @@ func (r *ReconcileExperiment) createTrials(instance *experimentsv1alpha2.Experim
...
+	labels := map[string]string{consts.LabelExperimentName: instance.Name}
+	lo := &client.ListOptions{}
+	lo.MatchingLabels(labels).InNamespace(instance.Namespace)
 
+	if err := r.List(context.TODO(), lo, trials); err != nil {
+		logger.Error(err, "Trial List error")
+		return err
+	}
+	actualDeleteCount := int(deleteCount)
+	if int32(len(trials.Items)) < deleteCount {
```
johnugeorge: when do this condition happen?

pkg/controller/v1alpha2/experiment/experiment_controller.go
```diff
@@ -368,6 +368,26 @@ func (r *ReconcileExperiment) createTrials(instance *experimentsv1alpha2.Experim
...
+		actualDeleteCount = len(trials.Items)
+	}
+
+	for i := 0; i < actualDeleteCount; i++ {
+		if err := r.Delete(context.TODO(), &trials.Items[len(trials.Items)-1-i]); err != nil {
```
johnugeorge: We have to delete the most recently created trial. Is the last item of the list guaranteed to be the latest trial?

pkg/controller/v1alpha2/experiment/experiment_controller.go
gaocegege: No idea, just to make sure the program won't crash.

pkg/controller/v1alpha2/experiment/experiment_controller.go
gaocegege: During my test, it is. But I think I should make sure of it. Please wait a while

pkg/controller/v1alpha2/experiment/experiment_controller.go
johnugeorge: I am not sure if it is needed. deleteCount is updated on the results of List operation. https://github.com/kubeflow/katib/blob/master/pkg/controller/v1alpha2/experiment/experiment_controller.go#L280 

One question is if List operation returns stale entries. But this is not solved here too

gaocegege: Because we do not use the trials from https://github.com/kubeflow/katib/blob/master/pkg/controller/v1alpha2/experiment/experiment_controller.go#L280 . Thus the result may be not consistent, IMO.

gaocegege: This just to avoid IndexOutOfBoundsException

gaocegege:
Comments are addressed, please take a look.
Thanks for your time :smile:

pkg/controller/v1alpha2/experiment/experiment_controller.go
johnugeorge: What if you use the earlier `trial.Items` results  than a new List operation?

gaocegege: It works for me. I am working on it now.

Issue #604, opened 2019-05-28
Fix flaky unit tests


johnugeorge:
/kind bug

Currently, unit tests are flaky with frequent errors during its run.

--- FAIL: TestReconcileExperiment (6.29s)
   experiment_controller_suite_test.go:76:
       Expected error:
           <*errors.errorString | 0xc00053a850>: {
               s: "an empty namespace may not be set during creation",
           }
           an empty namespace may not be set during creation
       not to have occurred
   experiment_controller_test.go:268:
       Timed out after 5.000s.
       Expected
           <bool>: false
       to be true

gaocegege:
Can you reproduce it? I think we have CI to run tests for every PR. Thus unit test should be passed.

I tried locally and does not meet any error.

johnugeorge:
This happens much more frequently now

`2019-06-16T11:32:15.530+0530	ERROR	kubebuilder.controller	controller/controller.go:144	Could not wait for Cache to sync	{"Controller": "experiment-controller", "error": "failed to wait for experiment-controller caches to sync"}
github.com/kubeflow/katib/vendor/github.com/go-logr/zapr.(*zapLogger).Error
	/Users/johnugeo/GO-ROOT/src/github.com/kubeflow/katib/vendor/github.com/go-logr/zapr/zapr.go:128
github.com/kubeflow/katib/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start
	/Users/johnugeo/GO-ROOT/src/github.com/kubeflow/katib/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:144
github.com/kubeflow/katib/vendor/sigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).start.func2
	/Users/johnugeo/GO-ROOT/src/github.com/kubeflow/katib/vendor/sigs.k8s.io/controller-runtime/pkg/manager/internal.go:206
2019-06-16T11:32:15.543+0530	INFO	kubebuilder.controller	controller/controller.go:120	Starting EventSource	{"Controller": "experiment-controller", "Source": "kind source: /, Kind="}
2019-06-16T11:32:15.543+0530	INFO	kubebuilder.controller	controller/controller.go:120	Starting EventSource	{"Controller": "experiment-controller", "Source": "kind source: /, Kind="}
2019-06-16T11:32:15.544+0530	INFO	experiment-controller	experiment/experiment_controller.go:131	Experiment controller created
[...]

PR #655, opened 2019-06-16
Adding extra check for Katib core namespace env

johnugeorge:
If DefaultKatibNamespaceEnvName is not set, it will use default namespace.  This will be used in unit tests where env is not set.
/hold
Fixes: #604 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/655)


johnugeorge:
Closing this as web hooks will be disabled in unit tests after https://github.com/kubeflow/katib/pull/657
/close

PR #660, opened 2019-06-18
MetricController: Run only a single job per task

### LINKED ISSUE #659: Metric-collector cronjob spawns unlimited jobs
/kind bug

Run a "high" amount of paralell jobs relative to your cluster size.

Things to work, but slowly.

**What happened:**
The metric-collector cron jobs created by katib keeps spawning new jobs, which don't complete before the new ones are created (since the cluster is under pressure).  

**Proposed solution:**
I know that there is a issue to change to a push-based #577 metric collector, but a short-term fix for this is, I think, to change the [concurrency-policy](https://kubernetes.io/docs/tasks/job/#concurrency-policy) of the cron-jobs to have `Forbid` instead of the default `Allow`. Then at least only a single instance of the metric-collector jobs is initiated at a time. 

**Environment:**

- Katib version: v0.1.2-alpha-156-g4ab3dbd

epa095:
This changes the `spec.concurrencyPolicy` of the metric collector 
cron-job from "Allow" (default) to "Forbid". The cronjob used to
create a new job even if the previous job had not succeeded. On
high-load clusters this could lead to a high number of jobs which
never finished. 

**Which issue(s) this PR fixes** *:
This fixes #659 

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/660)

### DIFF
--- a/manifests/v1alpha1/studyjobcontroller/metricsControllerConfigMap.yaml
+++ b/manifests/v1alpha1/studyjobcontroller/metricsControllerConfigMap.yaml
@@ -14,6 +14,7 @@ data:
       schedule: "*/1 * * * *"
       successfulJobsHistoryLimit: 0
       failedJobsHistoryLimit: 1
+      concurrencyPolicy: Forbid
       jobTemplate:
         spec:
           backoffLimit: 0
--- a/manifests/v1alpha2/katib-controller/metricsControllerConfigMap.yaml
+++ b/manifests/v1alpha2/katib-controller/metricsControllerConfigMap.yaml
@@ -14,6 +14,7 @@ data:
       schedule: "*/1 * * * *"
       successfulJobsHistoryLimit: 0
       failedJobsHistoryLimit: 1
+      concurrencyPolicy: Forbid
       jobTemplate:
         spec:
           backoffLimit: 0


epa095:
I signed it!

johnugeorge:
Wondering why previous jobs have not succeeded? 

Can you add changes to v1alpha2 also?

epa095:
> Wondering why previous jobs have not succeeded?
> 
> Can you add changes to v1alpha2 also?

Added it to v1alpha2 as well. 

In our case we actually saw two problems, and we think both are related to a flaky cluster (on AKS):
1. Pods started, but slowly. This would cause the metric-collectors to pile up, but finish eventually. In this case katib just put unnecessary stress on the system.
2. Some pods never started, because... who knows, but probably AKS/network issues. In this case katib/cronjobs kept spawning jobs which never finished, and it brought everything down (which is maybe as expected with a broken cluster). We have a limit of 110 pods on each node, and I *think* the everlasting-starting pods counted towards this limit. 

It should be noted that in the second case above this patch has the effect that we dont spawn infinite non-starting pods, but it also causes some trials to never get their results collected, while in the previous setup one could be lucky and have some of the collectors for a trial start up and manage collecting the result. I think its better to not start up arbitrary many instances, but its a bit of a tradeof.

johnugeorge:
@epa095 i am little confused about 2.  If pods were not started, cronjobs should fail(https://github.com/kubeflow/katib/blob/master/pkg/util/v1alpha2/metricscollector/metricscollector.go#L49) and  `failedJobsHistoryLimit ` is set which determines the max number of failed jobs that should be kept.

epa095:
That is failed jobs, not jobs stuck in 'starting'.

gaocegege:
@epa095 

PR #671, opened 2019-06-28
[WIP] E2E test for NAS RL Algorithm in v1alpha2

andreyvelich:
I have added NAS RL e2e test to v1alpha2.

/cc @johnugeorge @richardsliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/671)


andreyvelich:
I think we can close this PR since we have https://github.com/kubeflow/katib/blob/master/test/suggestion/v1alpha3/test_nasrl_service.py 
and 
https://github.com/kubeflow/katib/blob/master/test/scripts/v1alpha3/run-suggestion-nasrl.sh 
What do you think @johnugeorge @gaocegege ?

PR #675, opened 2019-07-03
add release scripts

kunmingg:
Add release script for releasing katib images.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/675)

### DIFF
--- a/submit_release_job.sh
+++ b/submit_release_job.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+#
+# A simple script to submit the Argo workflow to build the release.
+#
+# Usage submit_release_job.sh ${COMMIT}
+#
+# COMMIT=commit to build at
+# release workflow will submit to kubeflow-ci until release cluster updated
+set -ex
+
+COMMIT=$1
+
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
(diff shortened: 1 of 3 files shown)


submit_release_job.sh
```diff
@@ -0,0 +1,39 @@
...
+
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+JOB_NAME="katib-release"
+JOB_TYPE=katib-release
+BUILD_NUMBER=$(uuidgen)
+BUILD_NUMBER=${BUILD_NUMBER:0:4}
+REPO_OWNER=kubeflow
+REPO_NAME=katib
+ENV=test
```
hougangliu: I think we'd better use `release` ENV instead

hougangliu:
@kunmingg please also rebase your PR to make kubeflow-katib-presubmit passed

submit_release_job.sh
kunmingg: Ideally we should use `release` ENV, but it's been out of date for a while...
We need to rebuild release infra, and before that `test` ENV also work for release workflows.

hougangliu: sgtm, so please rebase your PR to make kubeflow-katib-presubmit passed (a fix to the error has been merged into master branch)

PR #682, opened 2019-07-11
Use latest katib images to deploy katib

hougangliu:
After this patch, user can install katib individually (not install kubeflow) with latest images by `deploy.sh`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/682)

### DIFF
--- a/manifests/v1alpha1/studyjobcontroller/metricsControllerConfigMap.yaml
+++ b/manifests/v1alpha1/studyjobcontroller/metricsControllerConfigMap.yaml
@@ -23,7 +23,7 @@ data:
               serviceAccountName: metrics-collector
               containers:
               - name: {{.WorkerID}}
-                image: katib/metrics-collector
+                image: gcr.io/kubeflow-images-public/katib/v1alpha1/metrics-collector
                 command: ["./metricscollector"]
                 args:
                 - "-s"
--- a/manifests/v1alpha1/studyjobcontroller/studyjobcontroller.yaml
+++ b/manifests/v1alpha1/studyjobcontroller/studyjobcontroller.yaml
@@ -19,7 +19,7 @@ spec:
       serviceAccountName: studyjob-controller
       containers:
       - name: studyjob-controller
-        image: katib/studyjob-controller
+        image: gcr.io/kubeflow-images-public/katib/v1alpha1/studyjob-controller
         imagePullPolicy: Always
         ports:
         - containerPort: 443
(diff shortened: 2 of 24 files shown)


manifests/v1alpha1/studyjobcontroller/metricsControllerConfigMap.yaml
```diff
@@ -23,7 +23,7 @@ data:
               serviceAccountName: metrics-collector
               containers:
               - name: {{.WorkerID}}
-                image: katib/metrics-collector
+                image: gcr.io/kubeflow-images-public/katib/v1alpha1/metrics-collector
```
gaocegege: @johnugeorge Do we have the repository in dockerhub? If not I think the PR LGTM

hougangliu: AFAIK, we hasn't maintained katib images in dockerhub for a long time

johnugeorge: I think, it is better to keep single source of truth. Since our images are built in gcr, this PR  looks good to me than maintaining in two different repos.  
One related question is, should we point to release tagged image than picking up the latest image from master? Release image can be updated in the manifests after every release. eg:  0.6 Else, users might see new untested features and different behavior from the docs which will be confusing to them.

johnugeorge:
test env is broken in all repos now.

manifests/v1alpha1/studyjobcontroller/metricsControllerConfigMap.yaml
hougangliu: @johnugeorge, Agree with you. But I think for master branch, we can just keep latest  images which keep consistent with master source code. For a released branch, we should modify the images with released version tag.

johnugeorge: @hougangliu But, won't users deploy Katib from master branch always?

hougangliu: we can doc about what images the installer will use. In fact, for end user, I think he should install katib by kubeflow kfctl, which always installs released images. For `deploy.sh` here, it is prepared for developers, who is likely to test how current code runs.

johnugeorge: SGTM. If developers need to use our scripts to deploy(without Katib), we can provide link to the manifests of last deployed branch.  We need to add this to README

PR #687, opened 2019-07-15
Fix error of undeploy.sh

hougangliu:
`kubectl delete resourceType namespace/resourceName` is invalid.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/687)

### DIFF
--- a/examples/v1alpha2/hyperband-example.yaml
+++ b/examples/v1alpha2/hyperband-example.yaml
@@ -4,6 +4,7 @@ metadata:
   namespace: kubeflow
   name: hyperband-example
 spec:
+  parallelTrialCount: 9
   objective:
     type: maximize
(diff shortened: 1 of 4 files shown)


examples/v1alpha2/hyperband-example.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   namespace: kubeflow
   name: hyperband-example
 spec:
+  parallelTrialCount: 9
```
johnugeorge: may be added maxTrials too?

hougangliu: hyperband algorithm can terminate itself when loop exceeds. we doesn't need set maxTrials for it.

Issue #685, opened 2019-07-12
katib metrics collector solution


hougangliu:
- [X] metricsCollector API #697
- [x] Pod level webhook for sidecar
- [ ] Job level webhook for sidecar (traced by #928)
- [x] StdOutCollector metricsCollector executor
- [x]	FileCollector metricsCollector executor
- [x] TfEventCollector metricsCollector executor
- [ ] PrometheusMetricCollector metricsCollector executor  **P2**  (traced by #929)

hougangliu:
## Overview
This solution key ideas are:
1. Inject collector container as a sidecar into `Job`/`Tfjob`/`PytorchJob` Pod. Collector container collects metrics from worker container by parsing file or http based on metrics source, and persist metrics into persistent server, such as `katib-db` by `Katib-manager` or `metadata server`;
2. Introduce `TrialMetrics` CRD to trace metrics of trial by Kubernetes native way;
3. Introduce `Katib-Metrics-Manager` server to poll metrics of running Trial from metric persistent server and update related `TrialMetrics` CR. `Trial` controller updates trial's status based on coresponding `TrialMetrics` CR;

As for Early-Stop service (Maybe discuss it in other issue)
1. Early-Stop service will watch `Trial` metrics observation, once meeting goal, marks corresponding experiments as `Completed`.
1. Experiment controller will mark all running `Trial` as `Killed` or else if experiment `Completed`
1. Trial controller stops Job/TFJob/PytorchJob if `Trial` status is `Killed` or else 

See the architecture as below:

<img width="772" alt="image" src="https://user-images.githubusercontent.com/10126721/61432863-4c185e80-a964-11e9-90da-f111316c0751.png">

## Data Structure
```
type ExperimentSpec struct {
	...
	MetricsCollectorSpec *MetricsCollectorSpec `json:"metricsCollectorSpec,omitempty"`
    ...
}

type MetricsCollectorSpec struct {
	Source    *SourceSpec    `json:"source,omitempty"`
	Collector *CollectorSpec `json:"collector,omitempty"`
}

type SourceSpec struct {
    // maybe model-train source code exposes metrics by http, such as HTTP endpoint in prometheus metric format
    HttpGet        *v1.HTTPGetAction `json:"httpGet,omitempty"`
    // during training model, metrics maybe be persistent into local file in source code, such as tfevent
    FileSystemPath *FileSystemPath   `json:"fileSystemPath,omitempty"`
}

type FileSystemKind string
const (
[...]

PR #697, opened 2019-07-26
API for metricCollector

hougangliu:
For solution #685 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/697)


pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
```diff
@@ -236,12 +236,64 @@ type Operation struct {
...
+
+type CollectorKind string
+
+const (
+	StdOutCollector           CollectorKind = "stdOutCollector"
+	FileCollector             CollectorKind = "fileCollector"
+	TfEventCollector          CollectorKind = "tfEventCollector"
+	PrometheusMetricCollector CollectorKind = "prometheusMetricCollector"
+	CustomCollector           CollectorKind = "customCollector"
+	// When model train source code persists metrics into persistent layer
```
gaocegege:
```suggestion
	// When model training source code persists metrics into persistent layer
```

hougangliu:
> Can you split the commit into 2, one for changes and other for code generated changes

only *_types.go are code change. others are nearly all auto-generated changes

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
```diff
@@ -236,12 +236,64 @@ type Operation struct {
...
+	// Model-train source code can expose metrics by http, such as HTTP endpoint in
+	// prometheus metric format
+	HttpGet *v1.HTTPGetAction `json:"httpGet,omitempty"`
+	// During training model, metrics may be persisted into local file in source
+	// code, such as tfEvent use case
+	FileSystemPath *FileSystemPath `json:"fileSystemPath,omitempty"`
+	// Default metric output format is {"metric": "<metric_name>",
+	// "value": <int_or_float>, "epoch": <int>, "step": <int>}, but if the output doesn't
+	// follow default format, please extend it here
+	Filter *FilterSpec `json:"filter,omitempty"`
```
johnugeorge: Are we doing a type check here when parsing?

pkg/api/operators/apis/trialmetrics/v1alpha2/trialmetrics_types.go
```diff
@@ -0,0 +1,52 @@
+/*
```
johnugeorge: is this meant to store final metrics of Trials?

hougangliu: I had thought to use it persist best metrics by interval from DB. but per discuss, maybe we can drop it and trial controller will get final metrics to fill trial status when worker done

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
hougangliu: metricsCollector will try to convert the value to int or float and take it as an invalid entry with warning message if fail to convert. WDYT?

johnugeorge: Will this be a bigger constraint for the user?  There can be errors in Type cast of int vs float ?

johnugeorge: Currently, we are not doing any check when we write into DB https://github.com/kubeflow/katib/blob/master/pkg/api/v1alpha2/api.proto#L346

May be accept string and convert to float ?

johnugeorge:
Added one final comment.

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
```diff
@@ -236,12 +236,64 @@ type Operation struct {
...
+}
+
+type FilterSpec struct {
+	// When the metrics output follows format as this field specified, metricsCollector
+	// collects it and reports to metrics server, it can be "<metric_name>: <float>" or else
+	MetricsFormat []string `json:"metricsFormat,omitempty"`
```
johnugeorge: MetricsFormat is not clear to me. Does user need to specify objective metric name again here?

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
hougangliu: I just think that we can use `Filter` filed in term of extensibility.  If user source code metrics output style follows katib default suggested format, user doesn't need this field indeed. But if user source code doesn't follow our default style, we have no way to make it run by current katib except that user writes its new metricsCollector from scratch. in this case with this field, we at least can make it possible that user just point us which its format look like so that our metricsCollector can recognize a new pattern.
as for value type, I think maybe we can just guide users that we will always take it as float (for a db column, its value type can be uniquely defined)
In sum, `Filter` no harm but good IMO.

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
hougangliu: for example, we can make MetricsFormat  like `["{{metricName}} := {{metricValue}}", "{{metricName}} is {{metricValue}}"]`, and the experiment objectivemetrics is `loss`, and additional metrics `accuracy`, metrics will run logic like
```
for line of outputLines:
  for metric in ["loss", "accuracy"]:
    for format in MetricsFormat:
      mformat = convert_metrics(metric, format) // use metric replace {{metricName}}
      value = parse(line, mformat) // try to match the output like "loss  := $something" or "loss is $something", and return $something
      if !value:
         continue
      report_metrics(metric, value)
```

johnugeorge: So, you mean to say that `metricName`, `metricValue`  are keywords to be used in Spec for specifying metricFormat.  But the comment says - "<metric_name>: <float>"

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
johnugeorge: Also, are we looking at multiple metrics in single line? (as in comment?)

pkg/api/operators/apis/experiment/v1alpha2/experiment_types.go
hougangliu: the comment is about default format with json format, which keeps same as floydhub.
in `MetricsFormat`, we can use {{metricName}} and {{metricValue}} as a placeholder.
anyway, maybe we can talk more about what the placeholder should looks like in another thread.

johnugeorge:
We can track specific issues in separate PRs
/approve

PR #738, opened 2019-09-03
MetricsCollector pod injection for katib job

hougangliu:
1. support kubernetes`Job` pod metrics collector injection, not just Pytorch and tfjob worker kind
2. Tfjob/pytorch job labels backward compatibility

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/738)

### DIFF
--- a/pkg/webhook/v1alpha2/pod/const.go
+++ b/pkg/webhook/v1alpha2/pod/const.go
@@ -27,7 +27,7 @@ const (
 )
 
 var JobRoleMap = map[string][]string{
-	"TFJob":      []string{JobRoleLabel, TFJobRoleLabel},
-	"PyTorchJob": []string{JobRoleLabel, PyTorchJobRoleLabel},
-	"Job":        []string{},
+	"TFJob":      {JobRoleLabel, TFJobRoleLabel},
+	"PyTorchJob": {JobRoleLabel, PyTorchJobRoleLabel},
+	"Job":        {},
 }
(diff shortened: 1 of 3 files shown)


hougangliu:
> Can we keep both pod and job level webhooks?

sorry that my message misleads you, here I mean before this PR, we only support inject pod of pytorch and tfjob created by Katib, not including k8s Job worker kind. My PR fixes this gap

pkg/webhook/v1alpha2/pod/const.go
```diff
@@ -0,0 +1,33 @@
...
+	JobNameLabel = "job-name"
+	// JobRoleLabel represents the label key for the job role, e.g. the value is master
+	JobRoleLabel        = "job-role"
+	TFJobRoleLabel      = "tf-job-role"
+	PyTorchJobRoleLabel = "pytorch-job-role"
+	MasterRole          = "master"
+)
+
+var JobRoleMap = map[string][]string{
+	"TFJob":      {JobRoleLabel, TFJobRoleLabel},
```
gaocegege: Maybe we could use Plural in tf-operator instead of a string. But here LGTM, too

hougangliu: sure, we can do it. maybe another PR to cover all such replacement in katib

gaocegege: SGTM

Issue #507, opened 2019-05-13
Redesign Suggestion service design


johnugeorge:
Proposal is to make new Suggestion CR  to serve suggestion requests from the experiment controller. Currently, experiment controller directly talks to suggestion service and gets suggestions. 

In the new design,
Experiment controller creates Suggestion CR for each experiment to get suggestion results.Suggestion controller picks it up and talks to right algorithm service(as currently) and updates the Suggestion CR status with suggestion results and state. Experiment controller reads the suggestion results from CR and creates trials.

With new Suggestion CR and controller implementation,
1. It is more K8s native as DB dependency is avoided
2. It can take care of long running experiments. Suggestion CR is created per experiment by the experiment controller.
3. State or algorithm settings(key-value pair) will be stored in the CR instead of external DB.  Suggestion controller can thus send just the required information to each algorithm service.

/assign
/cc @gaocegege @richardsliu @hougangliu  @YujiOshima 

/kind feature

Related: #463 #471 #500

gaocegege:
According to the experiment, I found that we can support grid search with the new design. We can get all grids each time and calculate which grids will be used according to the numbers of succeeded trials. The idea is inspired by advisor here: https://github.com/tobegit3hub/advisor/blob/master/advisor_server/suggestion/algorithm/grid_search.py#L65

johnugeorge:
Draft of the Suggestion CRD: 
  The main idea is that we keep the state of the suggestions(per experiment) as CR without any external dependency.  

```
type SuggestionSpec struct {

//Name of the algorithm 
AlgorithmName string

// Number of suggestions requested
 RequestCount int  

//Algorithm settings set by the user in the experiment config
AlgorithmSettings []AlgorithmSetting

}

type AlgorithmSetting struct {
	Name  string 
	Value string
}

type SuggestionStatus struct {
   // Suggestion results
  Assignments [] ParameterAssignment
   
  Conditions []SuggestionCondition
  // include all common fields
   
}

type ParameterAssignment struct {
	Name  string 
	Value string
}

type Suggestion struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   SuggestionSpec   `json:"spec,omitempty"`
	Status SuggestionStatus `json:"status,omitempty"`
}

```
[...]

PR #735, opened 2019-09-02
WIP: Suggestion Refactor PoC

gaocegege:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/735)

PR #742, opened 2019-09-04
feat: Add suggestion controller

gaocegege:

Add the basic controller for suggestion CRD. It does not contain:

- Logic to create suggestions according to algorithms (only supports random now).
- Logic to communicate with suggestions (Use a fake implementation instead).

Ref #507 

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/742)

### DIFF
--- a/pkg/controller.v1alpha3/suggestion/composer/composer.go
+++ b/pkg/controller.v1alpha3/suggestion/composer/composer.go
@@ -58,7 +58,7 @@ func (g *General) DesiredDeployment(s *suggestionsv1alpha3.Suggestion) (*appsv1.
 
 func (g *General) DesiredService(s *suggestionsv1alpha3.Suggestion) (*corev1.Service, error) {
 	ports := []corev1.ServicePort{
-		corev1.ServicePort{
+		{
 			Name: "katib-api",
 			Port: consts.DefaultSuggestionPort,
 		},
--- a/pkg/controller.v1alpha3/suggestion/composer/composer.go
+++ b/pkg/controller.v1alpha3/suggestion/composer/composer.go
@@ -1,0 +1,94 @@
+package composer
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	suggestionsv1alpha3 "github.com/kubeflow/katib/pkg/apis/controller/suggestions/v1alpha3"
+	"github.com/kubeflow/katib/pkg/controller.v1alpha3/consts"
+	"github.com/kubeflow/katib/pkg/controller.v1alpha3/util"
+)
+
+type Composer interface {
+	DesiredDeployment(s *suggestionsv1alpha3.Suggestion) (*appsv1.Deployment, error)
+	DesiredService(s *suggestionsv1alpha3.Suggestion) (*corev1.Service, error)
+}
+
+type General struct {
+	scheme *runtime.Scheme
+}
+
+func New(scheme *runtime.Scheme) Composer {
+	return &General{
+		scheme: scheme,
+	}
+}
+
+func (g *General) DesiredDeployment(s *suggestionsv1alpha3.Suggestion) (*appsv1.Deployment, error) {
+	d := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      s.Name,
+			Namespace: s.Namespace,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: util.SuggestionLabels(s),
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: util.SuggestionLabels(s),
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						*g.desiredContainer(s),
+					},
+				},
+			},
+		},
+	}
+
+	if err := controllerutil.SetControllerReference(s, d, g.scheme); err != nil {
+		return nil, err
+	}
+	return d, nil
+}
+
+func (g *General) DesiredService(s *suggestionsv1alpha3.Suggestion) (*corev1.Service, error) {
+	ports := []corev1.ServicePort{
+		corev1.ServicePort{
+			Name: "katib-api",
+			Port: consts.DefaultSuggestionPort,
+		},
+	}
+
+	service := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      s.Name,
+			Namespace: s.Namespace,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: util.SuggestionLabels(s),
+			Ports:    ports,
+			Type:     corev1.ServiceTypeClusterIP,
+		},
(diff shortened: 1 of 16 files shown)


pkg/controller.v1alpha3/suggestion/suggestion_controller.go
```diff
@@ -0,0 +1,176 @@
...
+		if errors.IsNotFound(err) {
+			// For additional cleanup logic use finalizers.
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+	oldS := instance.DeepCopy()
+
+	deploy, err := r.DesiredDeployment(instance)
```
johnugeorge: I am confused here. Why do we need a deployment and service for a suggestion?

pkg/controller.v1alpha3/suggestion/composer/composer.go
```diff
@@ -0,0 +1,94 @@
...
+	// Add owner reference to the service so that it could be GC after the suggestion is deleted
+	if err := controllerutil.SetControllerReference(s, service, g.scheme); err != nil {
+		return nil, err
+	}
+
+	return service, nil
+}
+
+// TODO(gaocegege): Implement switch logic.
+func (g *General) desiredContainer(s *suggestionsv1alpha3.Suggestion) *corev1.Container {
```
johnugeorge: why do  we need separate container per process?  Am I missing something?

pkg/controller.v1alpha3/suggestion/suggestion_controller.go
xieydd: I guess, suggestion controller will depend on the deployment controller and service controller, and binding the deployment and the service will expose a service port in cluster.
@gaocegege Can you correct my guess.

pkg/controller.v1alpha3/suggestion/composer/composer.go
gaocegege: We need to create the algorithm container for the given algorithm name and settings.

pkg/controller.v1alpha3/suggestion/suggestion_controller.go
gaocegege: @johnugeorge I think we should create the suggestion first, then we can communicate with the suggestion.

For example, if an experiment uses random, we need to create a random suggestion for the experiment.

pkg/controller.v1alpha3/suggestion/suggestion_controller_util.go
```diff
@@ -0,0 +1,36 @@
...
+		log.Info("Creating Deployment", "namespace", deploy.Namespace, "name", deploy.Name)
+		err = r.Create(context.TODO(), deploy)
+		return nil, err
+	} else if err != nil {
+		return nil, err
+	}
+	return foundDeploy, nil
+}
+
+func (r *ReconcileSuggestion) createOrUpdateService(service *corev1.Service) (*corev1.Service, error) {
```
johnugeorge: Aren't we updating the service here?

pkg/controller.v1alpha3/suggestion/suggestion_controller_util.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (r *ReconcileSuggestion) createOrUpdateDeployment(deploy *appsv1.Deployment) (*appsv1.Deployment, error) {
```
johnugeorge: Do we need to update the deployment if already present?

pkg/controller.v1alpha3/suggestion/suggestion_controller_util.go
gaocegege: No idea about if we have such a use case now.

pkg/controller.v1alpha3/suggestion/suggestion_controller_util.go
gaocegege: No idea about if we have such a use case now.

pkg/controller.v1alpha3/suggestion/composer/composer.go
```diff
@@ -0,0 +1,94 @@
...
+	c := &corev1.Container{
+		Name: consts.ContainerSuggestion,
+	}
+	c.Image = "katib/v1alpha2/suggestion-random:1"
```
johnugeorge: Do you plan to add the mapping of algorithms in this PR?

gaocegege: No, I think we can do it in another PR.

pkg/controller.v1alpha3/suggestion/suggestion_controller.go
```diff
@@ -0,0 +1,176 @@
...
+		return reconcile.Result{}, err
+	}
+
+	if err := r.List(context.TODO(),
+		client.MatchingLabels(util.TrialLabels(experiment)), trials); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	logger.Info("Sync assignments", "suggestions", instance.Spec.Suggestions)
+	if err = r.SyncAssignments(instance, experiment, trials.Items); err != nil {
```
johnugeorge: Where is this function defined?

pkg/controller.v1alpha3/suggestion/suggestion_controller_util.go
johnugeorge: Ok. Function name is confusing. Either we can name to ReconcileDeployment?

pkg/controller.v1alpha3/suggestion/suggestion_controller_util.go
johnugeorge: Ok.  we can rename to ReconcileService?

pkg/controller.v1alpha3/suggestion/suggestion_controller.go
gaocegege: https://github.com/kubeflow/katib/pull/742/files#diff-1cdb066fdd3258bfce6a7218681d43b6R14

Now we only have a fake implementation. The real implementation should be added after GRPC API is updated. Ref https://github.com/kubeflow/katib/pull/735/files#diff-5ad38cda167bce34ad354bdef2f05d51R36

pkg/controller.v1alpha3/suggestion/composer/composer.go
hougangliu: I introduced katib-config as https://github.com/kubeflow/katib/pull/745/files#diff-916572ac262b489ac2debd624cb44116, you can add `suggestion` configuration in the configmap,  for each algorithm, except `image` field, maybe you can also add "is-long-running" field

pkg/controller.v1alpha3/suggestion/suggestion_controller_util.go
gaocegege: Sure

pkg/controller.v1alpha3/suggestion/suggestion_controller_util.go
gaocegege: Sure

pkg/controller.v1alpha3/suggestion/composer/composer.go
johnugeorge: @hougangliu   Why not `image` field ? what is the structure you propose?

johnugeorge:
Merging this PR 
/approve

PR #743, opened 2019-09-05
feat(GRPC): Update API for Suggestion

### LINKED ISSUE #507: Redesign Suggestion service design

gaocegege:

This PR is to update the GRPC API, to avoid DB in the main workflow and remove manager requests from the suggestion side.

Ref #507 

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/743)


pkg/controller.v1alpha3/experiment/suggestion/suggestion.go
```diff
@@ -18,14 +17,15 @@ func New() Suggestion {
...
+	// 	ExperimentName: instance.Name,
+	// 	AlgorithmName:  instance.Spec.Algorithm.AlgorithmName,
+	// 	RequestNumber:  addCount,
+	// }
+	// if reply, err := common.GetSuggestions(request); err != nil {
+	// 	return nil, err
+	// } else {
+	// 	return reply.Trials, nil
+	// }
+	return nil, nil
```
gaocegege: This is to make the compilation succeeded.

pkg/apis/manager/v1alpha3/api.proto
```diff
@@ -516,13 +516,14 @@ message GetObservationLogReply {
...
-    string algorithm_name = 2;
+    Experiment experiment = 1;
+    repeated Trial trials = 2; // all completed trials owned by the experiment.
     int32 request_number = 3; ///The number of Suggestion you request at one time. When you set 3 to request_number, you can get three Suggestions at one time.
 }
 
 message GetSuggestionsReply {
-    repeated Trial trials = 1;
+    repeated Trial trials = 1; // trials should be created in the next run.
+    AlgorithmSpec algorithm = 2;
```
johnugeorge: How do we intend to use AlgorithmSpec in the reply?

gaocegege: It is used to support hyperband. If the algorithm wants to maintain some states using algorithm settings, I think we should support it in the current version. In the future, we should remove it.

gaocegege:
@johnugeorge I think we cannot only update the GRPC API. CI will fail.

Can you update API, manager implementation and controllers in one PR?

pkg/apis/manager/v1alpha3/api.proto
```diff
@@ -516,13 +516,14 @@ message GetObservationLogReply {
...
-    string experiment_name = 1;
-    string algorithm_name = 2;
+    Experiment experiment = 1;
+    repeated Trial trials = 2; // all completed trials owned by the experiment.
     int32 request_number = 3; ///The number of Suggestion you request at one time. When you set 3 to request_number, you can get three Suggestions at one time.
 }
 
 message GetSuggestionsReply {
-    repeated Trial trials = 1;
+    repeated Trial trials = 1; // trials should be created in the next run.
```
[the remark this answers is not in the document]
gaocegege: I think we can fix it in another PR.

PR #745, opened 2019-09-05
Inject sidecar container based on metrics collector kind

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/745)


pkg/apis/controller/trials/v1alpha3/trial_types.go
```diff
@@ -42,6 +42,13 @@ type TrialSpec struct {
 	// Whether to retain the trial metrics collector CronJob object after completed.
 	// Deprecated
 	RetainMetricsCollector bool `json:"retainMetricsCollector,omitempty"`
+	// Describes how metrics will be collected
+	MetricsCollector MetricsCollectorSpec `json:"metricsCollector,omitempty"`
```
gaocegege: Should we move https://github.com/kubeflow/katib/pull/745/files#diff-43075120b08ebfab109947ed650154cbR41 into the spec?

hougangliu: In fact, moving "image" into trial cannot replace Collector, we need  Collector.Kind to decide how to create empty-dir volume to share files between worker container and metricsCollector container. So I think we can get image and inject it into worker Pod whiling create share volume in inject webhook logic.

pkg/webhook/v1alpha3/pod/const.go
```diff
@@ -20,10 +20,12 @@ const (
 	// JobNameLabel represents the label key for the job name, the value is job name
 	JobNameLabel = "job-name"
 	// JobRoleLabel represents the label key for the job role, e.g. the value is master
-	JobRoleLabel        = "job-role"
-	TFJobRoleLabel      = "tf-job-role"
-	PyTorchJobRoleLabel = "pytorch-job-role"
-	MasterRole          = "master"
+	JobRoleLabel                 = "job-role"
```
johnugeorge: Can we create an issue to track these labels so that these can be removed in the next release?

hougangliu: sure, I can create it right now

hougangliu: #746

pkg/apis/controller/common/v1alpha3/common_types.go
```diff
@@ -65,3 +69,57 @@ type Observation struct {
...
+type FilterSpec struct {
+	// When the metrics output follows format as this field specified, metricsCollector
+	// collects it and reports to metrics server, it can be "<metric_name>: <float>" or else
+	MetricsFormat []string `json:"metricsFormat,omitempty"`
+}
+
+type FileSystemKind string
+
+const (
+	DirectoryKind FileSystemKind = "diretory"
```
johnugeorge: Typo: directory

hougangliu: get it, I will include it in a coming PR soon.

pkg/apis/controller/common/v1alpha3/common_types.go
```diff
@@ -65,3 +69,57 @@ type Observation struct {
...
+// +k8s:deepcopy-gen=true
+type FileSystemPath struct {
+	Path string         `json:"path,omitempty"`
+	Kind FileSystemKind `json:"kind,omitempty"`
+}
+
+type CollectorKind string
+
+const (
+	StdOutCollector           CollectorKind = "stdOutCollector"
```
johnugeorge: Since `collector kind` is user provided string, should we keep `kind` string simple  to `stdout`, `file` etc?  What do you think?

hougangliu: SGTM, I can include it in coming PR

PR #747, opened 2019-09-05
Ignore injecting sidecar for none collector kind

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/747)

### DIFF
--- a/pkg/apis/controller/common/v1alpha3/common_types.go
+++ b/pkg/apis/controller/common/v1alpha3/common_types.go
@@ -94,7 +94,7 @@ type FilterSpec struct {
 type FileSystemKind string
 
 const (
-	DirectoryKind FileSystemKind = "diretory"
+	DirectoryKind FileSystemKind = "directory"
 	FileKind      FileSystemKind = "file"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/apis/controller/common/v1alpha3/common_types.go
```diff
@@ -107,14 +107,14 @@ type FileSystemPath struct {
...
 
 const (
-	StdOutCollector           CollectorKind = "stdOutCollector"
-	FileCollector             CollectorKind = "fileCollector"
-	TfEventCollector          CollectorKind = "tfEventCollector"
-	PrometheusMetricCollector CollectorKind = "prometheusMetricCollector"
-	CustomCollector           CollectorKind = "customCollector"
+	StdOutCollector           CollectorKind = "stdout"
+	FileCollector             CollectorKind = "file"
+	TfEventCollector          CollectorKind = "tensorflow-event"
```
johnugeorge: Won't hyphen create issues in yaml?

gaocegege: I think it is not a key, it is a value.

/cc @hougangliu

hougangliu: hyphen can work well in yaml
in fact, we often name a Kubernetes resource with hyphen in by yaml format

hougangliu: yes, it's value @gaocegege

johnugeorge: Oh, right. It is a value here.  In general, Do we have a convention to follow camel case or hyphen separated ? eg: https://github.com/kubeflow/tf-operator/blob/master/pkg/apis/common/v1/types.go#L150

hougangliu: ok for me to use upper camel case style

PR #751, opened 2019-09-10
Populate Parameter assignments in trials

johnugeorge:
Best Parameter assignment is not populated in Experiment as trial spec was not updated.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/751)


johnugeorge:
/cc @gaocegege 
rebased

PR #753, opened 2019-09-11
feat: Support HyperOpt

gaocegege:

This is to support hyperopt:

- random
- TPE

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/753)


pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -0,0 +1,26 @@
...
+        """
+        Main function to provide suggestion.
+        """
+        base_serice = BaseHyperoptService(
+            algorithm_name=request.experiment.spec.algorithm.algorithm_name)
+        search_space = HyperParameterSearchSpace.convert(request.experiment)
+        trials = Trial.convert(request.trials)
+        new_trials = base_serice.getSuggestions(
+            search_space, trials, request.request_number)
+        return api_pb2.GetSuggestionsReply(
```
johnugeorge: @gaocegege  Are we planning to change to assignments after all algorithms?

gaocegege: No, after this PR, I will change to assignments, I think.

gaocegege: WDYT, I think the PR is huge, thus do not want to add a new feature or refactor in it.

cmd/suggestion/hyperopt/v1alpha3/requirements.txt
```diff
@@ -0,0 +1,10 @@
+grpcio
+duecredit
+cloudpickle==0.5.6
+numpy>=1.13.3
+scikit-learn>=0.19.0
+scipy>=0.19.1
+forestci
+protobuf
+googleapis-common-protos
+hyperopt
```
johnugeorge: add versions?

gaocegege: SGTM

pkg/suggestion/v1alpha3/hyperopt_service.py
johnugeorge: Sounds good

pkg/webhook/v1alpha3/experiment/validator/validator.go
```diff
@@ -70,33 +63,13 @@ func (g *DefaultValidator) ValidateExperiment(instance *experimentsv1alpha3.Expe
 		return fmt.Errorf("Only one of spec.parameters and spec.nasConfig can be specified.")
 	}
 
-	if err := g.validateAlgorithmSettings(instance); err != nil {
```
hougangliu: so do you think if we have some any method to validate AlgorithmSettings? how about implementing it in validateAlgorithm?

pkg/controller.v1alpha3/suggestion/suggestionclient/suggestionclient.go
```diff
@@ -16,3 +26,172 @@ type SuggestionClient interface {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	request := &suggestionapi.GetSuggestionsRequest{
+		Experiment:    g.ConvertExperiment(e),
+		Trials:        g.ConvertTrials(ts),
+		RequestNumber: int32(requestNum),
+	}
+	response, err := client.GetSuggestions(ctx, request)
+	logger.V(0).Info("Getting suggestions", "endpoint", endpoint, "response", response, "request", request)
```
hougangliu: I have no idea what happened here if err != nil since I don't know what response looks like in such case
for safe, I suggest we can move line 64 in line 68

hougangliu: I have no idea what may happen here if err != nil since I don't know what response looks like in such case
for safe, I suggest we can move line 64 in line 68

pkg/suggestion/v1alpha3/hyperopt/base_hyperopt_service.py
```diff
@@ -0,0 +1,150 @@
...
+            self.hyperopt_algorithm = hyperopt.tpe.suggest
+        elif algorithm_name == 'hyperopt-random':
+            self.hyperopt_algorithm = hyperopt.rand.suggest
+        # elif algorithm_name == 'hyperopt-anneal':
+        #     self.hyperopt_algorithm = hyperopt.anneal.suggest_batch
+        # elif algorithm_name == 'hyperopt-mix':
+        #     self.hyperopt_algorithm = hyperopt.mix.suggest
+        else:
+            logger.error("Failed to create the algortihm: %s", algorithm_name)
+            self.hyperopt_algorithm = None
```
hougangliu: I think we should raise exception here. since in getSuggestions we always take self.hyperopt_algorithm as valid one

pkg/webhook/v1alpha3/experiment/validator/validator.go
johnugeorge: Since we have suggestion deployment per experiment, we cannot do validate in web hook anymore. We can do a validate in controller itself and update status based on it. what do you think?

hougangliu: sgtm

gaocegege: Yes, we can do it within another PR, I think.

pkg/suggestion/v1alpha3/hyperopt/base_hyperopt_service.py
gaocegege: SGTM

pkg/webhook/v1alpha3/experiment/validator/validator.go
johnugeorge: Sounds good

gaocegege: Yeah, today I will work on it and https://github.com/kubeflow/katib/pull/753#discussion_r324583651

PR #754, opened 2019-09-12
Adding suggestion images to configmap

johnugeorge:
Suggestion image mappings are populated in ConfigMap so that right suggestion algorithm services are spawned for each experiment.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/754)

### DIFF
--- a/manifests/v1alpha3/katib-controller/katib-config.yaml
+++ b/manifests/v1alpha3/katib-controller/katib-config.yaml
@@ -15,7 +15,7 @@ data:
       "TensorFlowEvent": {
         "image": "gcr.io/kubeflow-images-public/katib/v1alpha3/tfevent-metricscollector"
       }
-    },
+    }
   suggestion: |-
     {
       "random": {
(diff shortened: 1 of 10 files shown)


manifests/v1alpha3/katib-controller/katib-config.yaml
```diff
@@ -15,4 +15,19 @@ data:
       "TensorFlowEvent": {
         "image": "gcr.io/kubeflow-images-public/katib/v1alpha3/tfevent-metricscollector"
       }
+    },
```
gaocegege: Should we have a `,` here?

hougangliu: I don't think we should add ","

johnugeorge: Removed

johnugeorge:
@hougangliu @gaocegege 
New var `DefaultKatibNamespace` is added which will be env value of `DefaultKatibNamespaceEnvName`. If not set, default value of `kubeflow` is chosen.

PR #755, opened 2019-09-12
WIP: Update Suggestion deployment 

johnugeorge:
Deployments are updated when there is a new change in deployment spec

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/755)


johnugeorge:
This feature is not currently required. If deployment spec is changed after experiment creation, it might lead to inconsistent state.

johnugeorge:
Closing it. Will reopen when required.
/close

PR #498, opened 2019-05-11
Decoupling DB interface from katib manager

achalshant:
Goal: To decouple the DB interface from katib manager in order to make the backend store pluggable. Please refer to the proposal doc for more details - https://docs.google.com/document/d/1xAi4V9SCAVZvrwiVc6a3VpK7BXPx-liQbr6YPdqBThs/edit#heading=h.1hri22e2jta

Major changes:

pkg/db/interface.go was moved to run as a service in cmd/dbif/mysql/main.go
cmd/manager/main.go is now a thin client that connects to the DBIF service using GRPC and operates on the DB.
Some services and messages from api.proto have moved to dbif.proto.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/498)

PR #760, opened 2019-09-16
Refactor DB interface

johnugeorge:
DB interface is refactored so that backend DB is pluggable

Related: #498

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/760)


johnugeorge:
For new DB support,  add corresponding implementation in `pkg/db/v1alpha3` and use `DB_NAME` to point at it.

cmd/manager/v1alpha3/main.go
```diff
@@ -194,8 +196,11 @@ func (s *server) Check(ctx context.Context, in *health_pb.HealthCheckRequest) (*
 func main() {
 	flag.Parse()
 	var err error
-
-	dbIf, err = kdb.New()
+	dbName := os.Getenv("DB_NAME")
```
hougangliu: maybe we can add pkg/db/v1alpha3/common/const.go to define all costants including ENV name, db backend name and so on.

johnugeorge: Done

PR #761, opened 2019-09-17
Remove old version metrics collector

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/761)


pkg/util/v1alpha3/katibclient/katib_client.go
```diff
@@ -41,7 +41,6 @@ type Client interface {
 	GetTrialList(name string, namespace ...string) (*trialsv1alpha3.TrialList, error)
 	GetTrialTemplates(namespace ...string) (map[string]string, error)
 	UpdateTrialTemplates(newTrialTemplates map[string]string, namespace ...string) error
-	GetMetricsCollectorTemplates(namespace ...string) (map[string]string, error)
 	UpdateMetricsCollectorTemplates(newMCTemplates map[string]string, namespace ...string) error
```
johnugeorge: Is UpdateMetricsCollectorTemplates  needed?

examples/v1alpha3/random-example.yaml
```diff
@@ -26,13 +26,15 @@ spec:
           spec:
             template:
               spec:
+                serviceAccountName: metrics-collector
```
gaocegege: I am not sure why do we need to set service account for the trial

pkg/util/v1alpha3/katibclient/katib_client.go
hougangliu: delete it, too
Thanks

examples/v1alpha3/random-example.yaml
hougangliu: for now, metrics-collector get logs from k8s api, it need some role. I will delete it after implement file reading

gaocegege: SGTM

PR #772, opened 2019-09-18
Removing suggestions from manager interface

johnugeorge:
GetSuggestions and ValidateAlgorithmSettings are removed from the manager interface.  Suggestion interface is directly invoked to call these functions instead of going through the manager interface. 

Related: #463

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/772)


hougangliu:
do we need also drop old suggestion algorithm service, such as random, grid and so on? since they cannot work now which depends on Katib-manager API

gaocegege:
I think so. Should we drop them in this PR? @johnugeorge

johnugeorge:
Sure. I will remove them.  But I am not sure why e2e tests were not getting completed in this case. When I am trying in my setup, it works.

PR #777, opened 2019-09-19
feat: Add bayesian

gaocegege:
Signed-off-by: Ce Gao <gaoce@caicloud.io>

**What this PR does / why we need it**:

Support bayesian with the help of skopt.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/777)

PR #779, opened 2019-09-20
feat: Add GRPC health check in suggestions

gaocegege:

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/779)

### DIFF
--- a/cmd/suggestion/hyperopt/v1alpha3/Dockerfile
+++ b/cmd/suggestion/hyperopt/v1alpha3/Dockerfile
@@ -1,13 +1,18 @@
 FROM python:3
 
-ADD . /usr/src/app/github.com/kubeflow/katib
-WORKDIR /usr/src/app/github.com/kubeflow/katib/cmd/suggestion/hyperopt/v1alpha3
 RUN if [ "$(uname -m)" = "ppc64le" ]; then \
         apt-get -y update && \
         apt-get -y install gfortran libopenblas-dev liblapack-dev && \
         pip install cython; \
     fi
+RUN GRPC_HEALTH_PROBE_VERSION=v0.3.0 && \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
+    chmod +x /bin/grpc_health_probe
+
+ADD . /usr/src/app/github.com/kubeflow/katib
+WORKDIR /usr/src/app/github.com/kubeflow/katib/cmd/suggestion/hyperopt/v1alpha3
 RUN pip install --no-cache-dir -r requirements.txt
-ENV PYTHONPATH /usr/src/app/github.com/kubeflow/katib:/usr/src/app/github.com/kubeflow/katib/pkg/apis/manager/v1alpha3/python
+
+ENV PYTHONPATH /usr/src/app/github.com/kubeflow/katib:/usr/src/app/github.com/kubeflow/katib/pkg/apis/manager/v1alpha3/python:/usr/src/app/github.com/kubeflow/katib/pkg/apis/manager/health/python
 
 ENTRYPOINT ["python", "main.py"]
--- a/pkg/controller.v1alpha3/suggestion/composer/composer.go
+++ b/pkg/controller.v1alpha3/suggestion/composer/composer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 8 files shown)


pkg/controller.v1alpha3/suggestion/composer/composer.go
```diff
@@ -109,6 +116,36 @@ func (g *General) desiredContainer(s *suggestionsv1alpha3.Suggestion) (*corev1.C
...
+				Command: []string{
+					defaultGRPCHealthCheckProbe,
+					fmt.Sprintf("-addr=:%d", consts.DefaultSuggestionPort),
+					fmt.Sprintf("-service=%s", consts.DefaultGRPCService),
+				},
+			},
+		},
+		InitialDelaySeconds: defaultInitialDelaySeconds,
+	}
+	c.LivenessProbe = &corev1.Probe{
```
johnugeorge: Will liveness probe has any hidden consequences?  Say if service is not accessible for some time

pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -2,15 +2,18 @@
...
 from pkg.suggestion.v1alpha3.internal.trial import Trial, Assignment
 from pkg.suggestion.v1alpha3.hyperopt.base_hyperopt_service import BaseHyperoptService
+from pkg.suggestion.v1alpha3.base_health_service import HealthServicer
 
 logger = logging.getLogger("HyperoptRandomService")
 
 
 class HyperoptService(
-        api_pb2_grpc.SuggestionServicer):
+        api_pb2_grpc.SuggestionServicer, HealthServicer):
```
johnugeorge: With this change, it is mandatory for all suggestion service to use HealthServices. Right?

pkg/controller.v1alpha3/suggestion/composer/composer.go
gaocegege: Yeah, it will. But I think we should have it. Since we can mark the deployment unavailable then avoid timeout requests

pkg/suggestion/v1alpha3/hyperopt_service.py
gaocegege: Yes, we need to add health servicer to all algorithms

johnugeorge: Sounds good. We should add them in the suggestion doc separately. Also, Dockerfile instructions to have graph-health-check binary also has to be documented.

pkg/controller.v1alpha3/suggestion/composer/composer.go
johnugeorge: sound good

pkg/controller.v1alpha3/suggestion/composer/composer.go
```diff
@@ -19,6 +20,12 @@ import (
 	"github.com/kubeflow/katib/pkg/controller.v1alpha3/util"
 )
 
+const (
+	defaultInitialDelaySeconds = 10
```
johnugeorge: Do we need higher value for defaultInitialDelaySeconds?  If defaultInitialDelaySeconds needed is just 10 sec, why did we see timeouts when we had set 60 sec to context timeouts earlier?

johnugeorge: Did we miss something earlier?

gaocegege: SGTM, I will update it in #777

cmd/suggestion/hyperopt/v1alpha3/Dockerfile
```diff
@@ -1,13 +1,18 @@
...
 
-ADD . /usr/src/app/github.com/kubeflow/katib
-WORKDIR /usr/src/app/github.com/kubeflow/katib/cmd/suggestion/hyperopt/v1alpha3
 RUN if [ "$(uname -m)" = "ppc64le" ]; then \
         apt-get -y update && \
         apt-get -y install gfortran libopenblas-dev liblapack-dev && \
         pip install cython; \
     fi
+RUN GRPC_HEALTH_PROBE_VERSION=v0.3.0 && \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
```
MrXinWang: Hi @gaocegege , I am wondering if you have tested the multi-arch binary file of the GRPC health probe, as your patch only wget the amd64 architecture.

gaocegege: No, I have not tried to run on other architecture.

But we could support it since the probe can be built other platforms.

https://github.com/grpc-ecosystem/grpc-health-probe/releases

gaocegege: Do you have some problem when you use it?

MrXinWang: Since I am working on the arm64 support of images, therefore I will propose a patch to fix this, thanks for noticing.

MrXinWang: well I didn't test but I am working on the arm64 image support, by going through the Dockerfile I noticed this. I will then try to take care of it :)

gaocegege: Sure, thanks. It seems that we can support it with an arch check.

PR #783, opened 2019-09-20
Implement file metrics collector

hougangliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/783)


pkg/webhook/v1alpha3/pod/inject_webhook.go
```diff
@@ -199,3 +201,110 @@ func (s *sidecarInjector) getMetricsCollectorImage(cKind common.CollectorKind) (
...
+		return common.DefaultFilePath
+	} else if mc.Collector.Kind == common.FileCollector {
+		return mc.Source.FileSystemPath.Path
+	} else {
+		return ""
+	}
+}
+
+func wrapWorkerContainer(pod *v1.Pod, jobKind, metricsFile string, mc trialsv1alpha3.MetricsCollectorSpec) {
+	if mc.Collector.Kind != common.StdOutCollector {
```
gaocegege: What does the function do

hougangliu: this function aims to redirect stdout to a file so that both StdoutCollector and FileCollector can share same image to parse file to get metrics.
I will refactor this function to handle container.command is nil case, which will parse image entrypoint just like how Tekton project handle the case

PR #800, opened 2019-09-24
Removing metric collector templates from UI

johnugeorge:
Metric collector templates are removed from UI.  Also "kubeflow" hardcoding is removed as deployed namespace is chosen by default

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/800)


pkg/ui/v1alpha3/backend.go
```diff
@@ -199,7 +200,7 @@ func (k *KatibUIHandler) FetchHPJobInfo(w http.ResponseWriter, r *http.Request)
 	defer conn.Close()
 
 	resultText := "trialName"
-	experiment, err := k.katibClient.GetExperiment(experimentName, "kubeflow")
+	experiment, err := k.katibClient.GetExperiment(experimentName)
```
gaocegege: Does it mean that we will get experiments in all namespaces?

johnugeorge: No. This has to be done in a separate PR. By default(if we do not pass namespace), it will use own deployed namespace. https://github.com/kubeflow/katib/blob/master/pkg/util/v1alpha3/katibclient/katib_client.go#L175

PR #801, opened 2019-09-24
Using example as the default trial 

johnugeorge:
Current default doesn't emit any metrics. Because of this, experiment won't get succeeded if default trial template is used.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/801)


johnugeorge:
For a beginner, he can use default trial to see Katib working

PR #812, opened 2019-09-25
feat: Add validation for grid

gaocegege:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/suggestion/v1alpha3/chocolate_service.py
+++ b/pkg/suggestion/v1alpha3/chocolate_service.py
@@ -4,7 +4,7 @@ from pkg.apis.manager.v1alpha3.python import api_pb2
 from pkg.apis.manager.v1alpha3.python import api_pb2_grpc
 from pkg.apis.manager.health.python import health_pb2
 
-from pkg.suggestion.v1alpha3.internal.search_space import HyperParameter, HyperParameterSearchSpace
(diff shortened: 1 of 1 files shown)


pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py
```diff
@@ -32,12 +32,8 @@ def getSuggestions(self, search_space, trials, request_number):
                 chocolate_search_space[key] = choco.quantized_uniform(
                     int(param.min), int(param.max), 1)
             elif param.type == DOUBLE:
-                if param.step != None:
-                    chocolate_search_space[key] = choco.quantized_uniform(
-                        float(param.min), float(param.max), float(param.step))
-                else:
-                    chocolate_search_space[key] = choco.uniform(
```
johnugeorge: what was the default value here?

gaocegege: Validation will fail without step.

johnugeorge: I meant, what does the algorithm assume if step is not set?

gaocegege: The algorithm will return validation error if step is not set

Issue #810, opened 2019-09-25
[enhancement] Show logs of trials using kubectl logs


gaocegege:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

/assign @hougangliu 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

PR #819, opened 2019-09-26
Fix stdout of worker container show nothing

hougangliu:
Fixes: #810

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/819)


johnugeorge:
Will tee command exist in all user images?/

gaocegege:
tee is a Linux command, I think it should be in all Linux distributions.

Issue #774, opened 2019-09-18
Can not find experiments in frontend if namespace is not kubeflow


Jeffwan:
/kind bug

**What steps did you take and what happened:**
I am new to katib and not sure if this is an issue.

Based on my finding, I notice if I submit experiments in different namespaces other than kubeflow. There're two problems

1. Can not find experiments on UI
2. Sometimes experiment never get done. Seems pod are done but `experiment` status are not updated. Checked status using command `kubectl get experiments`

**What did you expect to happen:**
We want every individual user create experiment in their own namespace. 

**Environment:**

- Kubeflow version: 0.6.2
- Kubernetes version: (use `kubectl version`): 1.13.8

SatyaNarayan1:
facing the same issue kubeflow version 0.6.2. Cant see any thing in experiments UI also experiments/trials in running state.

johnugeorge:
Yes. This is a known issue. This issue will be solved in upcoming 0.7 release

SatyaNarayan1:
@johnugeorge is there any earlier version in which experiment UI is working?

johnugeorge:
I was referring to issue where Katib job and services are in different namespaces: 

Related: https://github.com/kubeflow/katib/issues/343

johnugeorge:
@SatyaNarayan1 Do you have them in different namespaces?

SatyaNarayan1:
@johnugeorge I tried both approach 
1) Keeping in same namespace "kubeflow"
2) Keeping in two different namespace in both cases experiment UI is blank

<img width="1378" alt="Screen Shot 2019-09-23 at 12 25 38 PM" src="https://user-images.githubusercontent.com/6360289/65406498-67d53080-ddfd-11e9-8cf4-70997681d3d6.png">
<img width="1375" alt="Screen Shot 2019-09-23 at 12 25 24 PM" src="https://user-images.githubusercontent.com/6360289/65406516-6e63a800-ddfd-11e9-94dc-25bb3c1e63f6.png">

gaocegege:
So, current implementation is such that we do not provide namespace, it is uses the namespace of the deployment (which is KF)

From @johnugeorge

Jeffwan:
@gaocegege @johnugeorge 

[...]

PR #827, opened 2019-09-27
feat: support namespace for trial template

gaocegege:

Add a namespace field for the template.

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/827)

### DIFF
--- a/pkg/ui/v1alpha3/frontend/src/components/HP/Create/HPParameters.jsx
+++ b/pkg/ui/v1alpha3/frontend/src/components/HP/Create/HPParameters.jsx
@@ -128,7 +128,7 @@ const HPParameters = (props) => {
             goTemplate: {
                 templateSpec: {
                     configMapName: "trial-template",
-                    configMapNamespace: data.metadata.namespace,
+                    configMapNamespace: props.trialNamespace,
                     templatePath: props.trial,
                 }
             }
--- a/pkg/util/v1alpha3/katibclient/katib_client.go
+++ b/pkg/util/v1alpha3/katibclient/katib_client.go
@@ -19,6 +19,7 @@ import (
 	"context"
 
 	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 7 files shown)


pkg/ui/v1alpha3/frontend/src/components/HP/Create/HPParameters.jsx
```diff
@@ -128,7 +130,7 @@ const HPParameters = (props) => {
             goTemplate: {
                 templateSpec: {
                     configMapName: "trial-template",
-                    configMapNamespace: data.metadata.namespace,
+                    configMapNamespace: deploymentNamespace,
```
johnugeorge: Is this correct? Can't the template be outside kubeflow namespace?

gaocegege: I think the config map should be kept in kubeflow namespace. We have some built-in templates. If we use data.metadata.namespace, we cannot get them if users use other namespaces.

gaocegege:
Need to add an `onChange` to fetch trial templates.

pkg/util/v1alpha3/katibclient/katib_client.go
```diff
@@ -150,7 +151,10 @@ func (k *KatibClient) GetTrialTemplates(namespace ...string) (map[string]string,
 	ns := getNamespace(namespace...)
 	trialTemplates := &apiv1.ConfigMap{}
 
-	if err := k.client.Get(context.Background(), types.NamespacedName{Name: experimentsv1alpha3.DefaultTrialConfigMapName, Namespace: ns}, trialTemplates); err != nil {
+	err := k.client.Get(context.TODO(), types.NamespacedName{Name: experimentsv1alpha3.DefaultTrialConfigMapName, Namespace: ns}, trialTemplates)
```
hougangliu: `return k.GetConfigMap(experimentsv1alpha3.DefaultTrialConfigMapName, namespace)`

BTW, now trialTemplate can be stored in any configmap, should we always get/update it in default ConfigMap?

gaocegege:
Peek 2019-09-28 09-49

pkg/util/v1alpha3/katibclient/katib_client.go
gaocegege: I can refactor it in another PR. Now #774 is in higher priority.

gaocegege:
/cc @hougangliu 

Addressed the comments, PTAL

pkg/util/v1alpha3/katibclient/katib_client.go
hougangliu: sure

PR #861, opened 2019-10-08
Finish reconcile only after running trials are complete 

johnugeorge:
Experiment reconcile is set to complete only after running trials are complete

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/861)

### DIFF
--- a/pkg/controller.v1alpha3/experiment/experiment_controller.go
+++ b/pkg/controller.v1alpha3/experiment/experiment_controller.go
@@ -176,7 +176,7 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 	instance := original.DeepCopy()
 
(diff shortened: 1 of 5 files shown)


pkg/controller.v1alpha3/experiment/experiment_controller.go
```diff
@@ -176,7 +176,7 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 	instance := original.DeepCopy()
 
-	if instance.IsCompleted() {
+	if instance.IsCompleted() && !instance.HasRunningTrials() {
```
hougangliu: willn't this line make new trial never be created?

hougangliu: ignore my comment, get it now

Issue #334, opened 2019-01-21
vizier-db needs enhanced liveness/readiness checks


pdmack:
vizier-db pod is shown as "Running" after deployment but it actually is not fully configured if there was a problem with the katib-mysql PV not being writable.

OpenShift 3.11

```
$ oc logs vizier-db-7b4475c695-7vwt9
2019-01-21T14:33:28.470292Z 27562 [Note] Access denied for user 'UNKNOWN_USER'@'localhost' (using password: YES)
2019-01-21T14:33:30.469526Z 27563 [Note] Access denied for user 'UNKNOWN_USER'@'localhost' (using password: YES)
2019-01-21T14:33:32.471171Z 27564 [Note] Access denied for user 'UNKNOWN_USER'@'localhost' (using password: YES)
```

jlewi:
@johnugeorge @YujiOshima @richardsliu Any plans to fix this in time for 0.5? Or should we move it into 0.6?

richardsliu:
Moving this out of 0.5.

gaocegege:
MySQL liveness probe requires mysqladmin. 

https://github.com/helm/charts/blob/master/stable/mysql/templates/deployment.yaml#L122

In our MySQL image, we missed mysqladmin

Ref https://github.com/docker-library/mysql/issues/410

Thus it is hard to add a liveness probe. Going to replace the image with a new one with mysqladmin.

PR #871, opened 2019-10-11
feat: Add liveness probe for DB

gaocegege:

Fixes https://github.com/kubeflow/katib/issues/334

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/871)

### DIFF
--- a/manifests/v1alpha3/db/deployment.yaml
+++ b/manifests/v1alpha3/db/deployment.yaml
@@ -39,10 +39,19 @@ spec:
             command:
             - "/bin/bash"
             - "-c"
-            - "mysql -D $$MYSQL_DATABASE -p$$MYSQL_ROOT_PASSWORD -e 'SELECT 1'"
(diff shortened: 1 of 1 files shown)


gaocegege:
/assign @johnugeorge 

Not sure if we should merge it before 0.7

It requires some change to kf/manifests, I think.

johnugeorge:
This would help users right?. In that case, shall we add to kf/manifests too?

gaocegege:
It will be helpful. But there may be a blocker that we cannot upgrade from 8.0.3 to 8(8.0.17) painlessly. I need to delete /tmp/katib to clean the directory first.

gaocegege:
But I think it works for us since it solves #334

johnugeorge:
@gaocegege are you talking backward compatibility ? It is not compatible now anyways

gaocegege:
@johnugeorge Yeah, then I think we can merge it into kf/manifests.

johnugeorge:
I will wait for @hougangliu

PR #876, opened 2019-10-11
Remove unused katib-manager-rest

hougangliu:

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/876)


johnugeorge:
I think, this will be useful to use rest calls for our DB interface.   Mostly for debugging I guess, WHYT?

hougangliu:
for now, only metrics related operations (create/delete/get) are available. I think the rest service is of little value now.

gaocegege:
SGTM

/lgtm

hougangliu:
@johnugeorge 
katib-manager-rest is introduced in v1alpha1 when katib-manager works as a core role by which controller persists studyjob info into db and call it to get suggestion result, 3rd-party can call it to check StudyJob information (katib UI), suggestion service to call it to check trial result or else. by katib-manager-rest, 3rd-party can get studyjob info in REST way.
now, katib-manager just plays an internal DB interface role to persist only metrics and 3rd-party doesn't need it. so katib-manager-rest can say goodbye now.
WDYT?

johnugeorge:
Sounds good.  Just thinking if it is late to add to 0.7 release?

hougangliu:
> Sounds good. Just thinking if it is late to add to 0.7 release?

The risk is little since the change doesn't change katib-manger, katib-controller and suggestion algorithm code.

PR #900, opened 2019-10-28
Fix katib-manager crash in kubeflow cluster

hougangliu:
In kubeflow cluster, there is a service named "mysql" from pipeline. so in each kubeflow POD, there is will ENV for mysql service like below, which will make `getDbName()` return in katib mysql DB invalid (owning to env `MYSQL_PORT`)
```
# env|grep MYSQL
MYSQL_PORT_3306_TCP_ADDR=10.0.251.111
MYSQL_PORT_3306_TCP_PORT=3306
MYSQL_SERVICE_HOST=10.0.251.111
MYSQL_PORT_3306_TCP_PROTO=tcp
MYSQL_HOST=katib-db.kubeflow
MYSQL_PORT=tcp://10.0.251.111:3306
MYSQL_SERVICE_PORT=3306
MYSQL_PORT_3306_TCP=tcp://10.0.251.111:3306
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/900)

### DIFF
--- a/pkg/db/v1alpha3/common/const.go
+++ b/pkg/db/v1alpha3/common/const.go
@@ -7,8 +7,8 @@ const (
 
 	DBPasswordEnvName = "DB_PASSWORD"
 
-	MySQLDBHostEnvName = "MYSQL_HOST"
-	MySQLDBPortEnvName = "MYSQL_PORT"
+	MySQLDBHostEnvName = "KATIB_MYSQL_HOST"
+	MySQLDBPortEnvName = "KATIB_MYSQL_PORT"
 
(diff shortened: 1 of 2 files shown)


johnugeorge:
Great! what we suspected about env var.

Still, how does env variable get propagated into Katib pod?

hougangliu:
> propagated

it is Kubernetes feature, all services in the same namespace will be imported into the Pods as env

gaocegege:
Ref https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables

Issue #896, opened 2019-10-24
cannot support training job restart


hougangliu:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
When a training container exit abnormally, metrics collector will go on parsing metrics file and reporting the metrics with exit 0. And even `Restart policy` make the the training container restart, the  metrics collector container will never restart

**What did you expect to happen:**
metrics collector container should only start to parse metrics file when training job exit with 0, otherwise it should exit 1.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version:
- Minikube version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

PR #901, opened 2019-10-30
Support restarting training job

hougangliu:
Fixes: #896

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/901)

### DIFF
--- a/pkg/metricscollector/v1alpha3/common/pns.go
+++ b/pkg/metricscollector/v1alpha3/common/pns.go
@@ -17,8 +17,11 @@ package common
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"runtime"
+	"strings"
 	"time"
 
 	gops "github.com/mitchellh/go-ps"
@@ -27,9 +30,10 @@ import (
 var ErrWaitPidTimeout = fmt.Errorf("Timed out waiting for PID to complete")
 
 type WaitPidsOpts struct {
-	PollInterval time.Duration
-	Timeout      time.Duration
-	WaitAll      bool
+	PollInterval           time.Duration
+	Timeout                time.Duration
+	WaitAll                bool
+	CompletedMarkedDirPath string
 }
 
 func Wait(opts WaitPidsOpts) error {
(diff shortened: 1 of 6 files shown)


pkg/metricscollector/v1alpha3/common/pns.go
```diff
@@ -95,6 +99,16 @@ func WaitPIDS(pids []int, opts ...WaitPidsOpts) error {
 				_, err := os.Stat(path)
 				if err != nil {
 					if os.IsNotExist(err) {
+						if opts[0].CompletedMarkedDirPath != "" {
```
gaocegege: Cannot figure out how this helps solve the problem, could you please explain more about it?

hougangliu: This PR adds `echo completed > $mountPath/$$$$.pid` after training command as below, if the training container succeed, it will touch a file named `$mountPath/$processID.pid` with "completed"
MetricsCollector container watches the training process, once find the process exit, it will check file `$mountPath/$processID.pid` to judge if the training process succeeds or not; if succeed, parse the metrics file; otherwise raise exception to exit.
Before this PR, metrics collector once find training process exit, it starts to parse metrics file, ignoring if training process exit status.

I tries to use "strace" to implement it, but it need more linux capability.
```
  - args:
    - python /mxnet/example/image-classification/train_mnist.py --batch-size=64 --lr=0.02273874688380991
      --num-layers=3 --optimizer=sgd 1>/var/log/katib/metrics.log 2>&1 && echo completed
      > /var/log/katib/$$$$.pid
    command:
    - sh
    - -c
```

gaocegege: Will this cause the similar pipe exit code problem as tee does?

hougangliu: no
1. if training process fails, it will return training process exit code ( `&& echo` will not be executed)
2. if training process succeeds (exit code is 0), `&& echo` will return 0, too. so the container exit code is 0, too

gaocegege: Oh yeah, misunderstand the logic here, SGTM

PR #914, opened 2019-11-07
Handle metricscollector case worker container have no command

hougangliu:

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/914)


pkg/webhook/v1alpha3/pod/utils.go
```diff
@@ -74,3 +80,64 @@ func getLabel(pod *v1.Pod, targetLabel string) (string, error) {
...
+		Namespace:          namespace,
+		ServiceAccountName: pod.Spec.ServiceAccountName,
+		ImagePullSecrets:   imagePullSecrets,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create k8schain: %v", err)
+	}
+
+	mkc := authn.NewMultiKeychain(kc)
+	img, err := remote.Image(ref, remote.WithAuthFromKeychain(mkc))
```
gaocegege: What will be created when we call it? Will we get the image from the remote registry in a blocking manner?

hougangliu: it only fetches manifest instead of image file

Gopkg.toml
```diff
@@ -48,28 +53,32 @@ required = [
...
+
+[[override]]
+  name = "k8s.io/apiextensions-apiserver"
+  version = "kubernetes-1.12.9"
 
 [[override]]
-  name="k8s.io/apiextensions-apiserver"
-  version="kubernetes-1.12.3"
+  name = "k8s.io/kubernetes"
+  version = "v1.13.3"
```
gaocegege: Not sure if we can upgrade now. /cc @johnugeorge

pkg/webhook/v1alpha3/pod/utils.go
```diff
@@ -74,3 +80,64 @@ func getLabel(pod *v1.Pod, targetLabel string) (string, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get container image %q info from registry: %v", c.Image, err)
+	}
+
+	return img, nil
+}
+
+func getImageCommand(pod *v1.Pod, namespace string, containerIndex int) ([]string, error) {
```
gaocegege: The name is confusing. We may not get the command from the image if the container has command. How about getContainerCommand?

Gopkg.toml
hougangliu: I tried, it works well.

pkg/webhook/v1alpha3/pod/utils.go
hougangliu: agree with you

gaocegege:
/lgtm
/approve

@johnugeorge If you have problems, we can revert it. Prefer not to block @hougangliu 's work.

PR #925, opened 2019-11-21
Support multiple metric logs in one line

hougangliu:
for now, in one line, only $metricsName=$value (without blank) can be parsed.
By this PR, blank can be tolerant before and after "="; Also in the same line, multiple metrics couple can be parsed

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/925)

### DIFF
--- a/examples/v1alpha3/file-metrics-collector/mnist.py
+++ b/examples/v1alpha3/file-metrics-collector/mnist.py
@@ -63,7 +63,7 @@ def test(args, model, device, test_loader, epoch):
             correct += pred.eq(target.view_as(pred)).sum().item()
 
     test_loss /= len(test_loader.dataset)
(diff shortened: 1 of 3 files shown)


hougangliu:
> Can we add some unit test cases for the change?

Sure, I can update it soon

gaocegege:
Thanks. I cannot confirm the effect of the code by code review, thus prefer having such test cases to illustrate it.

hougangliu:
> Thanks. I cannot confirm the effect of the code by code review, thus prefer having such test cases to illustrate it.

I had updated `image` in `examples/v1alpha3/file-metricscollector-example.yaml`  to test the change

johnugeorge:
Can we document this now?  There will be more similar things in the future

hougangliu:
> Can we document this now? There will be more similar things in the future

I will implement `filter` filed in coming days and then will document how filter should be defined and what default filter is.

PR #935, opened 2019-11-27
Remove katib webhook when undeploy

hougangliu:

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/935)

### DIFF
--- a/scripts/v1alpha3/deploy.sh
+++ b/scripts/v1alpha3/deploy.sh
@@ -22,6 +22,8 @@ set -o xtrace
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/../..
 
 cd ${SCRIPT_ROOT}
+kubectl get validatingwebhookconfigurations katib-validating-webhook-config && kubectl delete validatingwebhookconfigurations katib-validating-webhook-config
+kubectl get mutatingwebhookconfigurations katib-mutating-webhook-config && kubectl delete mutatingwebhookconfigurations katib-mutating-webhook-config
 kubectl apply -f manifests/v1alpha3
 kubectl apply -f manifests/v1alpha3/katib-controller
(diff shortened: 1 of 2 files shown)


scripts/v1alpha3/deploy.sh
```diff
@@ -22,6 +22,8 @@ set -o xtrace
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/../..
 
 cd ${SCRIPT_ROOT}
+kubectl get validatingwebhookconfigurations katib-validating-webhook-config && kubectl delete validatingwebhookconfigurations katib-validating-webhook-config
+kubectl get mutatingwebhookconfigurations katib-mutating-webhook-config && kubectl delete mutatingwebhookconfigurations katib-mutating-webhook-config
```
johnugeorge: Adding `kubectl delete` in deploy script is confusing.  Why do we need this here?

gaocegege:
/retest

/hold

for @johnugeorge 's question.

I think it is used to ensure that we will recreate the webhook.

scripts/v1alpha3/deploy.sh
hougangliu: if there is same name webhook, controller won't recreate a new one. And the old one has configured with obsolete secret token, which will lead webhook call fail

johnugeorge: @hougangliu But I see that you have added the delete logic(webhook deletion) already in undeploy script. Am i mistaken here?

hougangliu: yes, here is just a guard to make sure the obsolete webhook removed in case that there is katib webhooks existing before run deploy.sh (image that in a cluster 0.7.0 kubeflow installed, a user wants to run deploy.sh to update katib to latest)

PR #948, opened 2019-12-04
upgrade deployments API for k8s 1.16+

ChenjunZou:
As k8s [deprecated](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/) APIs In 1.16,  
the API of deployment "extensions/v1beta1" is deprecated,  so the deploy scripts are breaking now.

This pr aims to fix the break by using "apps/v1" deployment API instead. 
Also, the selector sections are added, otherwise a warning would throw
It works at my computer using k8s version 1.16.1。

But backward compatibility is not tested due to lack of machine resources.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/948)

### DIFF
--- a/manifests/v1alpha2/db/deployment.yaml
+++ b/manifests/v1alpha2/db/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: katib-db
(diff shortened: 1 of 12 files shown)


ChenjunZou:
> @googlebot I signed it!

@googlebot I signed it!

ChenjunZou:
> @googlebot I signed it!

ChenjunZou:
already have .

Issue #891, opened 2019-10-23
[feature] Add new trials to a succeeded Experiment


gaocegege:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Now we cannot achieve it because of:

```go
	reconcileRequired := !instance.IsCompleted()
	if reconcileRequired {
		r.ReconcileTrials(instance, trials.Items)
	}
```

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

johnugeorge:
Yes. We have to add extra check when we support resuming trials

johnugeorge:
#952

Issue #768, opened 2019-09-17
[enhancement] Validate Experiment editable fields


johnugeorge:
Add web hook validation for fields that are editable.

Only editable fields should be ParallelTrialCount, MaxTrialCount, MaxFailedTrialCount

johnugeorge:
2. Validate that maxTrials if set is not set to zero

PR #952, opened 2019-12-06
Resume experiment with extra trials from last checkpoint

johnugeorge:
Fixes: #891 
Resuming experiment feature is added.  If user wants to try the same experiment with more number of trials, max trials can be reconfigured to restart the experiment from the last checkpoint. 

Experiment is restarted only when experiment is in succeeded state with max trials reached previously.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/952)

### DIFF
--- a/pkg/controller.v1alpha3/experiment/experiment_controller.go
+++ b/pkg/controller.v1alpha3/experiment/experiment_controller.go
@@ -186,8 +186,22 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 		return r.updateFinalizers(instance, finalizers)
 	}
 
-	if instance.IsCompleted() && !instance.HasRunningTrials() {
-		return reconcile.Result{}, nil
+	if instance.IsCompleted() {
+		// Check if completed instance is restartable
+		// Experiment is restartable only if it is in succeeded state by reaching max trials
+		if util.IsCompletedExperimentRestartable(instance) {
+			// Check if max trials is reconfigured
+			if (instance.Spec.MaxTrialCount != nil) &&
+				(*instance.Spec.MaxTrialCount != instance.Status.Trials) {
+				msg := "Experiment is restarted"
+				instance.MarkExperimentStatusRestarting(util.ExperimentRestartingReason, msg)
+			}
+		} else {
+			// If experiment is completed with no running trials, stop reconcile
+			if !instance.HasRunningTrials() {
+				return reconcile.Result{}, nil
+			}
+		}
 	}
 	if !instance.IsCreated() {
 		if instance.Status.StartTime == nil {
--- a/pkg/controller.v1alpha3/experiment/experiment_controller.go
+++ b/pkg/controller.v1alpha3/experiment/experiment_controller.go
@@ -191,8 +191,9 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 		// Experiment is restartable only if it is in succeeded state by reaching max trials
 		if util.IsCompletedExperimentRestartable(instance) {
 			// Check if max trials is reconfigured
-			if (instance.Spec.MaxTrialCount != nil) &&
-				(*instance.Spec.MaxTrialCount != instance.Status.Trials) {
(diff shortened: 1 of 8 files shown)


johnugeorge:
/hold 
Holding till reviews are completed. 
/cc @gaocegege 
/cc @hougangliu

johnugeorge:
@gaocegege  added a test to reconfigure max trails and parallel trial count

pkg/util/v1alpha3/katibclient/katib_client.go
```diff
@@ -123,6 +124,14 @@ func (k *KatibClient) CreateExperiment(experiment *experimentsv1alpha3.Experimen
 	return nil
 }
 
+func (k *KatibClient) UpdateExperiment(experiment *experimentsv1alpha3.Experiment, namespace ...string) error {
```
hougangliu: is this method used for UI? and it seems `namespace` param isn't used

pkg/controller.v1alpha3/experiment/experiment_controller.go
```diff
@@ -186,8 +186,22 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
...
 	}
 
-	if instance.IsCompleted() && !instance.HasRunningTrials() {
-		return reconcile.Result{}, nil
+	if instance.IsCompleted() {
+		// Check if completed instance is restartable
+		// Experiment is restartable only if it is in succeeded state by reaching max trials
+		if util.IsCompletedExperimentRestartable(instance) {
+			// Check if max trials is reconfigured
+			if (instance.Spec.MaxTrialCount != nil) &&
```
hougangliu: I think if a user changes `MaxTrialCount` to nil (infinity), here we should also allow to restart it

pkg/util/v1alpha3/katibclient/katib_client.go
johnugeorge: I am not sure about it. Shall we do this separately?

pkg/controller.v1alpha3/experiment/experiment_controller.go
johnugeorge: Updated.

pkg/util/v1alpha3/katibclient/katib_client.go
hougangliu: If this method is not useful in your PR topic, I think we'd better remove it.

pkg/controller.v1alpha3/experiment/experiment_controller.go
```diff
@@ -191,8 +191,9 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 		// Experiment is restartable only if it is in succeeded state by reaching max trials
 		if util.IsCompletedExperimentRestartable(instance) {
 			// Check if max trials is reconfigured
-			if (instance.Spec.MaxTrialCount != nil) &&
-				(*instance.Spec.MaxTrialCount != instance.Status.Trials) {
+			if (instance.Spec.MaxTrialCount != nil &&
+				*instance.Spec.MaxTrialCount != instance.Status.Trials) ||
+				(instance.Spec.MaxTrialCount == nil && instance.Status.Trials != 0) {
```
hougangliu: should we consider the case: instance.Spec.MaxTrialCount == 0 and the experiment will be marked completed with instance.Status.Trials == 0; then update instance.Spec.MaxTrialCount to nil?

pkg/util/v1alpha3/katibclient/katib_client.go
johnugeorge: No. this method is currently used in resume_e2e_experiment.go script in this PR to update the experiment.

pkg/controller.v1alpha3/experiment/experiment_controller.go
johnugeorge: i think, this is a invalid case. It doesn't make sense to set MaxTrialCount to be zero if set. We should add a validation for this separately. 
Related: https://github.com/kubeflow/katib/issues/768

johnugeorge:
/hold cancel
as reviews are completed

PR #953, opened 2019-12-06
Delete v1alpha2 api

johnugeorge:
v1alpha2 is deprecated. Hence, corresponding files are removed from the repository

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/953)


johnugeorge:
What can we do about coverage failure? @richardsliu

gaocegege:
I think we could lower the limit about test coverage in coveralls. But I have no permission to do it.

PR #958, opened 2019-12-09
Increase Suggestion memory limit

andreyvelich:
For NAS RL Suggestion we need at least 200Mi memory limit. 
Suggestion fails with OOM error if we have 100Mi there.

/cc @johnugeorge @gaocegege @hougangliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/958)


pkg/controller.v1alpha3/suggestion/composer/composer.go
```diff
@@ -28,7 +28,7 @@ const (
 
 	cpuLimit   = "500m"
 	cpuRequest = "50m"
-	memLimit   = "100Mi"
+	memLimit   = "200Mi"
```
hougangliu: I'd like that we can add resource field in `katib-config` suggestion data to make suggestion resource configurable for each one

johnugeorge: Agree.  If resource field is added in katib-config for any suggestion algoroithm, it will override the default settings.

andreyvelich: Ok, I will add these values to `katib-config` in this PR.

andreyvelich:
I changed `GetSuggestionContainerImage` function to `GetSuggestionConfigData` which return `map`, where key is the name of the specific parameter and value is the corresponding value from the configMap.

Please take a look @hougangliu @johnugeorge

andreyvelich:
By this way, we can add more suggestion parameters to the configMap and then parse it in `GetSuggestionConfigData` function.

manifests/v1alpha3/katib-controller/katib-config.yaml
```diff
@@ -34,6 +34,7 @@ data:
         "image": "gcr.io/kubeflow-images-public/katib/v1alpha3/suggestion-hyperopt"
       },
       "nasrl": {
-        "image": "gcr.io/kubeflow-images-public/katib/v1alpha3/suggestion-nasrl"
+        "image": "gcr.io/kubeflow-images-public/katib/v1alpha3/suggestion-nasrl",
+        "memLimit": "200Mi"
```
hougangliu: I prefer that here keep same style as k8s resources structure like below
```
   image: gcr.io/kubeflow-images-public/katib/v1alpha3/suggestion-nasrl
   resources:
      requests:
        memory: "64Mi"
        cpu: "250m"
      limits:
        memory: "128Mi"
        cpu: "500m"
```

gaocegege: Agree.

andreyvelich: I changed structure for configMap.

hougangliu:
/lgtm
thanks

PR #960, opened 2019-12-11
Use port higher than 1024 to be able to run as a non-root user

### LINKED ISSUE #959: Katib Controller fails to run as non-root user in OpenShift
/kind bug

* Deploy Katib with `kfctl` on OpenShift 4.x cluster
  * You can follow this WIP PR: https://github.com/opendatahub-io/manifests/pull/4
* Katib Controller fails with error about not being able to bind to port `:443`

Katib Controller starts and runs successfully.

**Environment:**

- OpenShift 4.x

vpavlin:
This PR allows Katib Controller to run as a non-root user. This is important from security point of view and especially to enable Katib on enterprise distributions of Kubernetes like OpenShift.

This PR adds a parameter `--webhook-port` to allow configuring port > 1024 for admission webhooks and also changes the default port (`443`) to `8443`.

It also changes the `USER` in `Dockerfile` as it is not necessary to run as root with this change even on the vanilla Kubernetes.

Fixes #959

No image changes

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/960)

### DIFF
--- a/cmd/katib-controller/v1alpha3/main.go
+++ b/cmd/katib-controller/v1alpha3/main.go
@@ -39,10 +39,12 @@ func main() {
 
 	var experimentSuggestionName string
 	var metricsAddr string
+	var webhookPort int
 
 	flag.StringVar(&experimentSuggestionName, "experiment-suggestion-name",
 		"default", "The implementation of suggestion interface in experiment controller (default|fake)")
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.IntVar(&webhookPort, "webhook-port", 8443, "The port number to be used for admission webhook server.")
 
 	flag.Parse()
 
@@ -83,7 +85,7 @@ func main() {
 	}
 
 	log.Info("Setting up webhooks")
-	if err := webhook.AddToManager(mgr); err != nil {
+	if err := webhook.AddToManager(mgr, int32(webhookPort)); err != nil {
 		log.Error(err, "unable to register webhooks to the manager")
 		os.Exit(1)
 	}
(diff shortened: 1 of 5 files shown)


cmd/katib-controller/v1alpha3/main.go
```diff
@@ -83,7 +85,7 @@ func main() {
 	}
 
 	log.Info("Setting up webhooks")
-	if err := webhook.AddToManager(mgr); err != nil {
+	if err := webhook.AddToManager(mgr, int32(webhookPort)); err != nil {
```
hougangliu: why not `var webhookPort int32` in line 42?

cmd/katib-controller/v1alpha3/main.go
```diff
@@ -39,10 +39,12 @@ func main() {
 
 	var experimentSuggestionName string
 	var metricsAddr string
+	var webhookPort int
 
 	flag.StringVar(&experimentSuggestionName, "experiment-suggestion-name",
 		"default", "The implementation of suggestion interface in experiment controller (default|fake)")
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.IntVar(&webhookPort, "webhook-port", 443, "The port number to be used for admission webhook server.")
```
hougangliu: keep 8443 as default value

cmd/katib-controller/v1alpha3/main.go
vpavlin: Because defining the variable as `int32` fails in `flag.IntVar` with 

```
v1alpha3/main.go:47:14: cannot use &webhookPort (type *int32) as type *int in argument to flag.IntVar
```

So I tried `flag.Var`, but `int32` does not have `flag.Set` method implemented

```
*int32 does not implement flag.Value (missing Set method)
```

So the simplest solution seemed to be to just load `int` and change the type. Do you have different suggestion for solution?

cmd/katib-controller/v1alpha3/main.go
vpavlin: Done

Issue #928, opened 2019-11-24
Job level webhook for metrics collector sidecar container


hougangliu:
/kind feature
/P2
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
for now, we inject metrics collector sidecar container in Pod webhook, we can consider metrics collector sidecar container in Job webhook level for performance.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

johnugeorge:
This is no more required in k8s 1.15 as we could use more efficient object selector in web hooks

vpavlin:
This could actually solve https://github.com/kubeflow/katib/issues/963 as it would reorder actions - first add the container in the Job and then let a job controller generate the pod, so it makes sense form my point of view:)

hougangliu:
> This could actually solve #963 as it would reorder actions - first add the container in the Job and then let a job controller generate the pod, so it makes sense form my point of view:)

@vpavlin yes, it can solve #963, but it is low priority as plan and will be included in Kubeflow 1.0 release

PR #964, opened 2019-12-11
Initialize securityContext in injected metrics container

### LINKED ISSUE #963: Experiments execution fails due to non-initialized securityContext in injected container
/kind bug

* Deploy Katib using Kubeflow on OpenShift 4.x
* Submit https://github.com/kubeflow/katib/blob/master/examples/v1alpha3/random-example.yaml
* In `kubectl describe jobs`:

```
  Warning  FailedCreate  2s    job-controller  Error creating: pods "random-example-vvtcwr95-fvf9k" is forbidden: unable to validate against any security context constraint: []
```

The core issue is that the `mutating.pod.katib.kubeflow.org` injects a container after some other webhook had populated `securityContext` in `container` spec. This "bare" container (https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1alpha3/pod/inject_webhook.go#L181) does not have `securityContext` initialized and OpenShift will prevent it from start unless the `serviceAccount` is `privileged` (which is, on the other hand, a terrible practice:) ).

The ideal solution would be to use `reinvocationPolicy` (https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy), but that is not available in OpenShift 4.2 and the webhook builder used by Katib Controller does not support it either.

My proposed solution then is to use the `securityContext` from the container that is already present in the pod.

Pods for Trial jobs are successfully started

**Environment:**

OpenShift  4.x

vpavlin:

The issue is that the injected container does not have `securityContext` initialized (as that is done by some earlier mutating webhook). The simplest solution for now is to simply copy the `securityContext` from the container that already exists (and is fully mutated) in the pod.

Fixes #963

No image changes

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/964)

### DIFF
--- a/pkg/webhook/v1alpha3/pod/inject_webhook.go
+++ b/pkg/webhook/v1alpha3/pod/inject_webhook.go
@@ -137,7 +137,7 @@ func (s *sidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 		return nil, err
 	}
 
-	injectContainer, err := s.getMetricsCollectorContainer(trial)
+	injectContainer, err := s.getMetricsCollectorContainer(trial, pod)
 	if err != nil {
 		return nil, err
 	}
@@ -162,7 +162,7 @@ func (s *sidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 	return mutatedPod, nil
 }
 
-func (s *sidecarInjector) getMetricsCollectorContainer(trial *trialsv1alpha3.Trial) (*v1.Container, error) {
+func (s *sidecarInjector) getMetricsCollectorContainer(trial *trialsv1alpha3.Trial, originalPod *v1.Pod) (*v1.Container, error) {
 	mc := trial.Spec.MetricsCollector
 	if mc.Collector.Kind == common.CustomCollector {
 		return mc.Collector.CustomCollector, nil
@@ -178,11 +178,13 @@ func (s *sidecarInjector) getMetricsCollectorContainer(trial *trialsv1alpha3.Tri
 	}
 	args := getMetricsCollectorArgs(trial.Name, metricName, mc)
 	sidecarContainerName := getSidecarContainerName(trial.Spec.MetricsCollector.Collector.Kind)
+	securityContext := originalPod.Spec.Containers[0].SecurityContext.DeepCopy()
 	injectContainer := v1.Container{
 		Name:            sidecarContainerName,
 		Image:           image,
 		Args:            args,
 		ImagePullPolicy: v1.PullIfNotPresent,
+		SecurityContext: securityContext,
 	}
 	return &injectContainer, nil
 }


pkg/webhook/v1alpha3/pod/inject_webhook.go
```diff
@@ -178,11 +178,13 @@ func (s *sidecarInjector) getMetricsCollectorContainer(trial *trialsv1alpha3.Tri
 	}
 	args := getMetricsCollectorArgs(trial.Name, metricName, mc)
 	sidecarContainerName := getSidecarContainerName(trial.Spec.MetricsCollector.Collector.Kind)
+	securityContext := originalPod.Spec.Containers[0].SecurityContext.DeepCopy()
```
hougangliu: how can it make sure that originalPod.Spec.Containers[0].SecurityContext here can fit the metricsCollector and the  originalPod.Spec.Containers[0].SecurityContext may not be generated by other mutating Webhook but defined explicitly?

johnugeorge:
@vpavlin Is this `some earlier mutating webhook` specific to your deployment?

vpavlin:
@johnugeorge It is a problem on OpenShift in general - OpenShift is very security focused and enforces quite limited SCC by default. This is happening as part of Pod Security Policies controller and the securityContext is first generated on pod level and then from the pod into containers. (See https://github.com/openshift/apiserver-library-go/blob/master/pkg/securitycontextconstraints/sccmatching/provider.go#L152)

This apparently runs before the mutating webhooks, so when Katib then injects new container without securityContext initialized, OpenShfit is not happy about it.

So the solution I am proposing for now is to replicate the securityContext from a container that already passed through the PSP mutation into the injected container.

Does it make sense?

pkg/webhook/v1alpha3/pod/inject_webhook.go
vpavlin: I am not sure I fully understand the questions. Please look at my comment below which explains how the securityContext is generated and see if it helps.

For how I understand your question - the metricsCollector should have the same securityContext as the `[0]` container as that is how it would be generated by PSP by default - based on the ServiceAccount privileges it would generate securityContext for Pod and from there for all the containers. So as the `[0]` container already has the right securityContext it is safe to simply replicate it into the metricsCollector

johnugeorge:
> @johnugeorge It is a problem on OpenShift in general - OpenShift is very security focused and enforces quite limited SCC by default. This is happening as part of Pod Security Policies controller and the securityContext is first generated on pod level and then from the pod into containers. (See https://github.com/openshift/apiserver-library-go/blob/master/pkg/securitycontextconstraints/sccmatching/provider.go#L152)
> 
> This apparently runs before the mutating webhooks, so when Katib then injects new container without securityContext initialized, OpenShfit is not happy about it.
> 
> So the solution I am proposing for now is to replicate the securityContext from a container that already passed through the PSP mutation into the injected container.
> 
> Does it make sense?

Got it.
/lgtm
will wait for @hougangliu 's review.

pkg/webhook/v1alpha3/pod/inject_webhook.go
hougangliu: The container securityContext can be generated by PSP, but also can be specified explicitly, right?
For example, a user adds a training job in a container with `securityContext: runAsUser: 1000`, if user 1000 can work well for metricsCollector container, too?

vpavlin: Ah, I see what you mean now. That is a fair point. I would say it would work fine - IIUC both containers are working with the same files, so if the user container works with UID `1000`, the metrics container should work too, right?

I am trying to figure out if there is a common pattern about how to solve these cases with the OpenShift team.

vpavlin: As I mentioned somewhere else the best solution would be to use `reinvocationPolicy` which should take care of generating the SC in the new container after it is added by webhook. That has 2 issues at the moment - 

1. it is coming in next version of OpenShift, so I cannot even test it:) 
2. the code katib uses to register webhooks does not know about the field, so I'll have to figure out how to actually do it when it is available

It seems (after discussing with OpenShift team) that using the existing SC from the other container might be our best shot here. I will try to git into it a bit more, but as I mentioned - as long as the sidecar container does not do anything special, it should work fine with the SC from existing container. The only way I see this could potentially fail if if the user configures the container very restrictively and metrics container cannot get to what it needs to read.

The only other way I could think of is to configure the SC explicitly to the state that we know will work for most configurations. Question is what happens if the user then configures his container to run as `root` and metrics container does not get the same SC and thus will probably not be able to access files produced by the main container.

That said, I still think copying the SC is the best solution at the moment. I will try to play around with it a bit more tomorrow to see if I can actually break this approach:)

hougangliu: metrics collector container now reads metrics file and then reports metrics to DB server, I wonder if two operations fail owning to training job SC configuration.

hougangliu: maybe we can try another solution: now k8s runs Webhooks in string order by webhook name, we can try to run katib mutating before other hooks by rename its name.
now its name is `katib-mutating-webhook-config`, we can rename it to `admission-katib-mutating-webhook-config`, not sure what openshift related Webhooks name is.

vpavlin: I asked RH istio team and their issue seems to be different because they set the `SC` explicitly as they need their container to run as `privileged`.

I am not sure about reordering by name because AFAIK the component generating `SC` is PSP (Pod Security Policies) which does not run as a webhook.

Adding the metrics container on `Job` level might help though - I added comment here https://github.com/kubeflow/katib/issues/928#issuecomment-566475922

vpavlin: > metrics collector container now reads metrics file and then reports metrics to DB server, I wonder if two operations fail owning to training job SC configuration.

I tested this before sending the PR and the operations surely did not fail.

hougangliu: I don't know why RH istio team think their issue is different.
the root cause is: when a Pod generated and before persisted, PSP adds security context to all containers; then mutatingwebhook injecting new container into a Pod which leads that the new container has no consistent SC which conflicts with PSP, right?
so I think it is a common use case, many projects would inject new container into a Pod by mutatingwebhook, such as istio, kfserving.
we had better handle this case in a best practice

gaocegege: @hougangliu @vpavlin 

Can we have a command line flag to control the behavior whether we use other containers's securityContext if we cannot come to an agreement now?

gaocegege:
As we discussed, we decide to merge it and add an additional flag to control it. I will add it after the PR is merged.

Issue #944, opened 2019-12-02
MetricsCollector does not parse time of the metric


andreyvelich:
/kind bug

I can't see time in `observation_logs` table after experiment finish.

**What steps did you take and what happened:**
I ran [random-example](https://github.com/kubeflow/katib/blob/master/examples/v1alpha3/random-example.yaml) on my GCP cluster. In the Trial's containers I got warning like this:
```
W1202 17:26:56.160885      15 file-metricscollector.go:50] Metrics will not have timestamp since error parsing time INFO:root:Epoch[0]: parsing time "INFO:root:Epoch[0]" as "2006-01-02T15:04:05.999999999Z07:00": cannot parse "INFO:root:Epoch[0]" as "2006"
```
After that, I checked information in `observation_logs` table and I saw this time: `001-01-01T00:00:00` in each recorded metric.
From my understanding, if we use default metrics collector, it parses `/var/log/katib/metrics.log` file. This file doesn't contain timestamp of the metric and we can't get information about metrics changing through time.
Maybe we can save also timestamp for all lines in the log file?
Or we can find better solution to indicate timestamp of the metrics somehow.
/cc @hougangliu @gaocegege @johnugeorge

hougangliu:
default metricsCollector always try to parse each line of user-application log, and timestamp is expected at start of the line, if there is no timestamp, ignore timestamp (use NULL in time field).
So maybe we can fix the example image and make its log follow this format

andreyvelich:
@hougangliu Yes, I agree we need to create example with printing timestamp in training container, like we did with `file-metrics-collector`.
Will take it.
/assign

andreyvelich:
@hougangliu Should I use docker.io/kubeflowkatib repo to push my image or I should use my own docker hub?

hougangliu:
we should use `docker.io/kubeflowkatib`, please leave your docker hub account so that I can add you to the org

andreyvelich:
@hougangliu Sure, my id is: andreyvelichkevich

PR #970, opened 2019-12-12
Example with collecting timestamp of the metrics

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/944.
Hold until I change docker hub repository.
/hold

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/970)


examples/v1alpha3/metrics-with-timestamp-example.yaml
```diff
@@ -0,0 +1,53 @@
...
+          kind: Job
+          metadata:
+            name: {{.Trial}}
+            namespace: {{.NameSpace}}
+          spec:
+            template:
+              spec:
+                containers:
+                - name: {{.Trial}}
+                  image: docker.io/andreyvelichkevich/timestamp-metric
```
hougangliu: I had added you into kubeflowkatib, you can now push kubeflowkatib/timestamp-metric now
BTW, I think you can rename the image name to `kubeflowkatib/mxnet-mnist-example` to make all example has timestamp.

johnugeorge: I suggest to rename folder and file name also to `metrics-with-timestamp` to `mxnet-mnist-example`. There is no need  to highlight the term`with-timestamp`

andreyvelich: This is not a `mxnet-mnist-example`. 
I took code of pytorch-mnist from here: https://github.com/kubeflow/katib/blob/master/examples/v1alpha3/file-metrics-collector/mnist.py. 

Should I change example to Mxnet in that case? Mxnet example is that one: https://github.com/apache/incubator-mxnet/blob/master/example/image-classification/train_mnist.py.
Or we can name this example `kubeflowkatib/pytorch-mnist-example`?

johnugeorge: I missed it. IIRC @richardsliu created this image to show mxnet support.

andreyvelich: @johnugeorge So what can we do with this example?

examples/v1alpha3/metrics-with-timestamp/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM pytorch/pytorch:1.0-cuda10.0-cudnn7-runtime
+
+WORKDIR /var
```
vpavlin: I don't believe /var is appropriate location for the script. If you check FSH (https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) `/var` should be used for changing files.

I would suggest using `/opt` and ideally a subdir in there - e.g. `/opt/mnist/`

WDYT?

examples/v1alpha3/metrics-with-timestamp-example.yaml
johnugeorge: @richardsliu Can you point out at the Dockerfile used?

examples/v1alpha3/metrics-with-timestamp/Dockerfile
andreyvelich: Good catch. What do you think about `/app`?

vpavlin: Any particular reason you do not want to go with existing directory structure in `/`? I don't think it is a good practice in linux systems to put code in "random" dirs in `/`, but feel free to convince me otherwise:)

andreyvelich: Do you mean in `/examples/v1alpha3/metrics-with-timestamp` ?

vpavlin: > Do you mean in `/examples/v1alpha3/metrics-with-timestamp` ?

Is this answer to @johnugeorge'S question?:-)

andreyvelich: > Any particular reason you do not want to go with existing directory structure in `/`? I don't think it is a good practice in linux systems to put code in "random" dirs in `/`, but feel free to convince me otherwise:)

To this question :) 
Which `WORKDIR` you think will be better?

vpavlin: Ah, sorry:) as I suggested in my first comment - I would go for `/opt/mnist/mnist.py` or to make it extremely obvious what it is (although that is already captured in the image name, but you never know..) `/opt/metrics-with-timestamp/mnist.py` or something along those lines

andreyvelich: @johnugeorge @hougangliu What do you think about `WORKDIR` for the example as `/opt/metrics-with-timestamp/mnist.py`

andreyvelich:
I changed example from pytorch to mxnet.
I think that would be more useful for the Katib user.
Also, I use kubeflowkatib docker repository now.
/cc @johnugeorge @vpavlin 
/hold cancel

andreyvelich:
I changed `mxnet-mnist-timestamp` image to `mxnet-mnist` as we discussed here: https://github.com/kubeflow/katib/pull/1008#discussion_r364820947. 
I think it will be better to name it `mxnet-mnist` than `mxnet-mnist-example`. 
In case that some users use old yaml files for their Katib experiments with `mxnet-mnist-example` image and, if we push new image to `mxnet-mnist-example`, it will brake current Katib experiments.

I renamed images in according yaml files, please take a look.

/cc @johnugeorge @hougangliu @gaocegege

andreyvelich:
I changed CI cluster `num nodes` to 8.
It decreased CI test time ~ 30%.
What do you think about this change @johnugeorge @hougangliu @gaocegege ?

test/scripts/v1alpha3/create-cluster.sh
```diff
@@ -33,6 +33,9 @@ echo "Creating GPU cluster"
 gcloud --project ${PROJECT} beta container clusters create ${CLUSTER_NAME} \
     --zone ${ZONE} \
     --machine-type=n1-standard-8 \
+    --num-nodes=8 \
```
johnugeorge: This should not be changed since overall resources are fixed

PR #996, opened 2020-01-04
fix: Ignore the failure

gaocegege:

When katib-controller is unavailable, the mutation webhook will fail all the pods in the namespace. I think we should avoid it.

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/996)


johnugeorge:
When did you encounter this issue?

gaocegege:
For example, there is a namespace which enables the injections. When I created a TFJob and the katib-controller is not ready at that time, the TFJob will be rejected since the webhook fails.

PR #1008, opened 2020-01-08
Add NAS RL training container to kubeflowkatib repository

andreyvelich:
I deleted build for NAS training container from the CI and I pushed it to kubeflowkatib repository.
Also, I added in example README list of current Katib training container images with source links.
I think it would be great if we have all Katib related images in one place.
What do you think?

/cc @johnugeorge @hougangliu @gaocegege

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/1008)


examples/v1alpha3/README.md
```diff
@@ -308,3 +308,47 @@ Acsess to `http://127.0.0.1:8000/katib`
...
+
+# List of current Katib training container images
+
+- Mxnet mnist example, [source](https://github.com/apache/incubator-mxnet/blob/master/example/image-classification/train_mnist.py).
+
+```
+docker.io/kubeflowkatib/mxnet-mnist-example
+```
+
+- Mxnet mnist example with collecting metrics time, [source](https://github.com/kubeflow/katib/blob/master/examples/v1alpha3/mxnet-mnist-timestamp/mnist.py).
```
johnugeorge: Is it checked in? 
Is earlier one required? Should we keep this as the default and rename to mxnet-mnist-example?

andreyvelich: The PR https://github.com/kubeflow/katib/pull/970 is required for this.
We can keep it as default mxnet mnist example.
In that case, I will rename mxnet-mnist-timestamp to mxnet-mnist-example and we will collect metrics time in this example also. What do you think about it @johnugeorge ?
/cc @hougangliu @gaocegege

johnugeorge:
I think so. Let’s see if others have a different suggestion .

examples/v1alpha3/README.md
hougangliu: agree. thanks

PR #1011, opened 2020-01-13
E2E Test for NAS RL Suggestion

andreyvelich:
I added e2e test for NAS RL Suggestion.
It uses simple 1 layer CNN architecture with 1 epoch and CPU training.
That should not take much time in CI.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/1011)


andreyvelich:
NAS RL e2e test takes ~ 14 minutes, less than random example.

/assign @johnugeorge @hougangliu @gaocegege

Issue #906, opened 2019-10-31
[feature] Reconsider the design of Trial Template


gaocegege:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

We need to marshal the TFJob to JSON string then use it to create experiments if we are using K8s client-go. It is not good. And, go template is ugly, too.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

gaocegege:
/priority p0

/cc @johnugeorge @hougangliu @richardsliu 

We can discuss here.

hougangliu:
based on the proposal https://github.com/kubeflow/katib/issues/857#issuecomment-539849490, I refactor struct as below:

```
type TrialTemplate struct {
	Retain     bool        `json:"retain,omitempty"`
	Template *Template `json:"template,omitempty"`
}

type ConfigmapTemplate struct {
	ConfigMapName      string `json:"configMapName,omitempty"`
	ConfigMapNamespace string `json:"configMapNamespace,omitempty"`
	Path       string `json:"path,omitempty"`
}

type InjectParametersKind string
const (
	GoTemplate     InjectParametersKind = "goTemplate" // By Default
	EnvVar      InjectParametersKind = "envVar"
	Args         InjectParametersKind = "args"
)

type Template struct {
	Kind InjectParametersKind `json:"kind,omitempty"`
	ConfigmapTemplate *ConfigmapTemplate `json:"configmapTemplate,omitempty"`
	RawTemplate  string        `json:"rawTemplate,omitempty"`
}
```

gaocegege:
Should we support structured template? E.g. TFJob and PyTorchJob.

richardsliu:
@hougangliu Can you add examples of what a template would look like?

hougangliu:
Here list the `goTemplate`, `envVar`, and `args` kind

1. `goTemplate`: keep same as current behavior, the template from ConfigMap or rawTemplate will be handled in go template way by making hyperParameters set fill `.HyperParameters` placeholder.
```
apiVersion: "kubeflow.org/v1alpha3"
kind: Experiment
metadata:
 ...
spec:
  ...
  parameters:
    - name: --lr
      parameterType: double
      feasibleSpace:
        min: "0.01"
        max: "0.03"
    - name: --num-layers
      parameterType: int
      feasibleSpace:
        min: "2"
        max: "5"
    - name: --optimizer
      parameterType: categorical
[...]

PR #1023, opened 2020-01-20
record TrialNames group by stages in ExperimentStatus

sperlingxx:
In this PR, several list fields are added to record TrialNames in different stages owned by current Experiment, which facilitates trials tracking. Count fields are removed since trial counts of each stage can be figured out by the length of TrialNames.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/1023)

### DIFF
--- a/pkg/controller.v1alpha3/experiment/util/status_util.go
+++ b/pkg/controller.v1alpha3/experiment/util/status_util.go
@@ -115,7 +115,6 @@ func updateTrialsSummary(instance *experimentsv1alpha3.Experiment, trials *trial
 		bestTrial := trials.Items[bestTrialIndex]
 
 		sts.CurrentOptimalTrial.BestTrialName = bestTrial.Name
-
 		sts.CurrentOptimalTrial.ParameterAssignments = []commonv1alpha3.ParameterAssignment{}
 		for _, parameterAssigment := range bestTrial.Spec.ParameterAssignments {
 			sts.CurrentOptimalTrial.ParameterAssignments = append(sts.CurrentOptimalTrial.ParameterAssignments, parameterAssigment)
--- a/pkg/controller.v1alpha3/experiment/util/status_util.go
+++ b/pkg/controller.v1alpha3/experiment/util/status_util.go
@@ -50,10 +50,10 @@ func UpdateExperimentStatus(collector *ExperimentsCollector, instance *experimen
 
 func updateTrialsSummary(instance *experimentsv1alpha3.Experiment, trials *trialsv1alpha3.TrialList) bool {
 
-	var totalTrials int32
 	var bestTrialValue float64
 	sts := &instance.Status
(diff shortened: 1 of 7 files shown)


gaocegege:
Thanks for the PR. Not sure if we should do it, personally. Perfer to keep v1alpha3 API stable. I think we can add these new fields and keep the existing fields too.

WDYT @johnugeorge @hougangliu @richardsliu

gaocegege:
/ok-to-test

@sperlingxx BTW, could you please add your infomation to https://github.com/kubeflow/katib/blob/master/docs/community/adopters.md if you are evaluating katib? Thanks

sperlingxx:
> /ok-to-test
> 
> @sperlingxx BTW, could you please add your infomation to https://github.com/kubeflow/katib/blob/master/docs/community/adopters.md if you are evaluating katib? Thanks

It's done, please check this [PR](https://github.com/kubeflow/katib/pull/1025).

sperlingxx:
> Thanks for the PR. Not sure if we should do it, personally. Perfer to keep v1alpha3 API stable. I think we can add these new fields and keep the existing fields too.
> 
> WDYT @johnugeorge @hougangliu @richardsliu

I also agree to keep v1alpha3 API stable, and I will restore count fields.

johnugeorge:
Should we wait and move this to next api version(possible beta) bundled with other changes?

Related #906

gaocegege:
@johnugeorge I think we can merge it since it does not breack v1alpha3 API.

If we have a breaking change in API, we should discuss it before we implement it in the new API.

WDYT @hougangliu @sperlingxx @johnugeorge

hougangliu:
> @johnugeorge I think we can merge it since it does not break v1alpha3 API.
> 
> If we have a breaking change in API, we should discuss it before we implement it in the new API.
> 
> WDYT @hougangliu @sperlingxx @johnugeorge

SGTM

hougangliu:
/approve
thanks

pkg/controller.v1alpha3/experiment/util/status_util.go
```diff
@@ -50,26 +50,28 @@ func UpdateExperimentStatus(collector *ExperimentsCollector, instance *experimen
...
 
 	for index, trial := range trials.Items {
-		totalTrials++
+		sts.Trials++
 		if trial.IsKilled() {
-			trialsKilled++
+			sts.KilledTrialList = append(sts.KilledTrialList, trial.Name)
 		} else if trial.IsFailed() {
-			trialsFailed++
+			sts.KilledTrialList = append(sts.KilledTrialList, trial.Name)
```
andreyvelich: Should it be `sts.FailedTrialList` instead of `sts.KilledTrialList` ?

sperlingxx: @andreyvelich Yes, I've already fixed it.

andreyvelich:
Thanks.
/lgtm

Issue #1052, opened 2020-02-12
OOM errors when number of trials is high


richardsliu:
For bayesian optimization, after running 43-44 trials, the controller stops scheduling new trials.

Pod status:
```
Containers:
  suggestion:
    Container ID:   docker://f5a413a95e61db01770288b04961d694ed36e4a7a4051ee3b07954d6219e161a
    Image:          gcr.io/kubeflow-images-public/katib/v1alpha3/suggestion-skopt:v0.8.0
    Image ID:       docker-pullable://gcr.io/kubeflow-images-public/katib/v1alpha3/suggestion-skopt@sha256:a1795458ec4c31c2c5528a2cec0547b80bcf48e6098a80f815d589c1f22f8222
    Port:           6789/TCP
    Host Port:      0/TCP
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       OOMKilled
      Exit Code:    137
      Started:      Tue, 11 Feb 2020 16:50:17 -0800
      Finished:     Tue, 11 Feb 2020 16:54:41 -0800
    Ready:          False
    Restart Count:  16
    Limits:
      cpu:                500m
      ephemeral-storage:  5Gi
      memory:             100Mi
    Requests:
      cpu:                50m
      ephemeral-storage:  500Mi
      memory:             10Mi
    Liveness:             exec [/bin/grpc_health_probe -addr=:6789 -service=manager.v1alpha3.Suggestion] delay=10s timeout=1s period=120s #success=1 #failure=12
    Readiness:            exec [/bin/grpc_health_probe -addr=:6789 -service=manager.v1alpha3.Suggestion] delay=10s timeout=1s period=10s #success=1 #failure=3
    Environment:          <none>
```

In addition, the request latency for suggestions becomes high and requires a longer timeout.

johnugeorge:
the current defaults are not default for higher number of trials

andreyvelich:
I checked BO Suggestion code. Here: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1alpha3/skopt/base_skopt_service.py#L77 we run `skopt_optimizer.tell()` for each Trial every `GetSuggestion` call. 
After 30 Succeeded Trials this call takes around 3 seconds to run, so for 30 Trials this loop will take around 90 seconds. 

@gaocegege Do we need to run this method every time when we update `skopt_suggested` or we can run it after the loop, when we create `skopt_suggested` list with all Trials?
[...]

PR #1055, opened 2020-02-14
Change tell method for BO Suggestion

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1052.

/cc @richardsliu @hougangliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/1055)


johnugeorge:
Can you add PR description

andreyvelich:
> Can you add PR description

After this PR `skopt_suggested` will contains information of all Trials assignments and `loss_for_skopt` will be list of all Trials observation value. We will run `tell` method once with all Trials information instead of running it for every trial.

Issue #1039, opened 2020-01-31
Bayesian optimization experiment suggests same hyperparameters 


martham93:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

I've experienced the same hyperparameters being suggested by a Bayesian optimization experiment several times. 

- For example: with 1 parallel trial, and max trials of 4 the exact same hyperparameters are suggested for each trial 

![experiment_35](https://user-images.githubusercontent.com/13889100/73553683-95636380-4407-11ea-98f1-d234ef67420f.png)

part of experiment `yml` file: 
```
apiVersion: "kubeflow.org/v1alpha3"
kind: Experiment
metadata:
  namespace: kubeflow
  name: poi-experiment35
  labels:
    controller-tools.k8s.io: "1.0"
spec:
  objective:
    type: maximize
    goal: 0.99
    objectiveMetricName: fbeta_m
  metricsCollectorSpec:
    source:
      fileSystemPath:
        path: "/ml/models/abc/eval/"
        kind: Directory
    collector:
      kind: TensorFlowEvent
  algorithm:
    algorithmName: bayesianoptimization
    algorithmSettings:
      - name: "random_state"
        value: "5"
  parallelTrialCount: 1
  maxTrialCount: 4
  maxFailedTrialCount: 3
  parameters:
    - name: --tf_dense_size
      parameterType: int
      feasibleSpace:
        min: "128"
        max: "512"
    - name: --tf_dense_dropout_rate
      parameterType: double
      feasibleSpace:
        min: "0.01"
        max: "0.5"
    - name: --tf_learning_rate
      parameterType: double
      feasibleSpace:
        min: "0.00001"
        max: "0.001"
    - name: --tf_optimizer
      parameterType: categorical
      feasibleSpace:
        list:
        - sgd
        - rmsprop
        - adam
```

- Another example: with 3 parallel trials, and 9 max trials, different hyperparameters are suggested for the first three trials, but for the following 6 trials the hyper-parameter combination from the best performing trial is suggested again for all six of the remaining trials. 
![experiment_36](https://user-images.githubusercontent.com/13889100/73553701-9b594480-4407-11ea-9be6-f9b07729d975.png)

part of experiment `yml` file:
```
apiVersion: "kubeflow.org/v1alpha3"
kind: Experiment
metadata:
  namespace: kubeflow
  name: poi-experiment36
  labels:
    controller-tools.k8s.io: "1.0"
spec:
  objective:
[...]

PR #1057, opened 2020-02-17
Create Optimizer in BO Suggestion only for the first run

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1039.

In this PR I added `is_first_run` flag.
We must create `Optimizer` instance using `search_space` only for the first `GetSuggestions` call.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/1057)

### DIFF
--- a/pkg/suggestion/v1alpha3/skopt/base_skopt_service.py
+++ b/pkg/suggestion/v1alpha3/skopt/base_skopt_service.py
@@ -63,7 +63,6 @@ class BaseSkoptService(object):
         logger.info("New GetSuggestions call\n")
         skopt_suggested = []
         loss_for_skopt = []
-        return_trial_list = []
         if len(trials) > self.succeeded_trials or self.succeeded_trials == 0:
             self.succeeded_trials = len(trials)
             logger.info("Succeeded Trials changed or first call: {}\n".format(self.succeeded_trials))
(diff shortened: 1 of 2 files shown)


andreyvelich:
See comment: https://github.com/kubeflow/katib/issues/1039#issuecomment-586330937

andreyvelich:
@gaocegege Thanks!

andreyvelich:
/hold 
Hold until I test it on long-term experiment

andreyvelich:
After testing BO on big amount of Trials I found few problems.
``
1. Since we are using one `Optimizer` instance during whole experiment, each `GetSuggestion` call will run `tell` method which will add information about all Trials to `Optimizer`. Xi and yi will have duplicate values (See attributes here: https://scikit-optimize.github.io/modules/generated/skopt.Optimizer.html#skopt.Optimizer) which is not correct and slow performance.
To prevent it, I add `recorded_trials` dict, where key = loss value and value = List of Trial Assignment list for this value. Before running `tell` method, I am checking if this information already contains in the dict and I don't run `tell` method for the recorded Trials.

2. We don't need to iterate over trials if count of succeeded Trials is the same. For some cases, Katib controller can run `GetSuggestion` without new succeeded trials (For Example, some trials failed). We don't need to run `tell` method in that situation.

3. I tried to change `ask` method with `n_points=request_number`, see (https://scikit-optimize.github.io/modules/generated/skopt.Optimizer.html#skopt.Optimizer.ask). It indicates how many Trials do we want to create. Unfortunately, after 35 Trials this method takes so much time to run (more than 1 minute). I am not sure how we can fix it. Any thoughts @gaocegege ?

Also, I added few logging prints in BO.

/hold cancel

/cc @johnugeorge

andreyvelich:
According to the first point, instead of map it is easier to have list of recorded trials name, since they are unique.
Just forgot that we send Trial name to the Suggestion.

ever-cheng:
Doesn't SkoptService exist as a singleton instance for the system, to accept all GetSuggestions request?
How can it accept experiment change if it was bound to a fixed base_service?
        if self.is_first_run:
            self.base_service = BaseSkoptService(...)
            ...
            self.is_first_run = False

johnugeorge:
@ever-cheng No. algorithm service is deployed per experiment

Issue #1060, opened 2020-02-19
Katib controller goes into CrashLoopBackoff when no goal is specified


borremosch:
/kind bug

**What steps did you take and what happened:**
I created an experiment without a `goal`. The spec validated fine, but consequently the Katib controller entered a faulty state from which recovery seems impossible. I had to delete and re-install Katib entirely into my development cluster. Logs from the controller are as follows:

```{"level":"info","ts":1582106357.593974,"logger":"trial-controller","msg":"Trial  controller created"}
{"level":"info","ts":1582106357.5939763,"logger":"entrypoint","msg":"Setting up webhooks"}
{"level":"info","ts":1582106357.5940497,"logger":"entrypoint","msg":"Starting the Cmd."}
{"level":"info","ts":1582106357.694709,"logger":"kubebuilder.webhook","msg":"installing webhook configuration in cluster"}
{"level":"info","ts":1582106357.694735,"logger":"kubebuilder.controller","msg":"Starting Controller","controller":"suggestion-controller"}
{"level":"info","ts":1582106357.6947956,"logger":"kubebuilder.controller","msg":"Starting Controller","controller":"trial-controller"}
{"level":"info","ts":1582106357.6948807,"logger":"kubebuilder.controller","msg":"Starting Controller","controller":"experiment-controller"}
{"level":"info","ts":1582106357.794992,"logger":"kubebuilder.controller","msg":"Starting workers","controller":"trial-controller","worker count":1}
{"level":"info","ts":1582106357.794992,"logger":"kubebuilder.controller","msg":"Starting workers","controller":"suggestion-controller","worker count":1}
{"level":"info","ts":1582106357.7957547,"logger":"kubebuilder.controller","msg":"Starting workers","controller":"experiment-controller","worker count":1}
{"level":"info","ts":1582106357.7958603,"logger":"experiment-controller","msg":"Statistics","Experiment":"project-github-test-ae92s/experiment-run-workflow-fh6kw","requiredActiveCount":5,"parallelCount":5,"activeCount":0,"completedCount":0}
{"level":"info","ts":1582106357.7958694,"logger":"experiment-controller","msg":"CreateTrials","Experiment":"project-github-test-ae92s/experiment-run-workflow-fh6kw","addCount":5}
[...]

PR #1065, opened 2020-02-21
Enable to run Experiment without Goal

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1060.

After this PR users should be able to run Experiment without specifying Goal.

/assign @johnugeorge 
/cc @borremosch

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/1065)


johnugeorge:
Inorder to veerify in CI, shall we remove goal in one of the example yaml ? 
May be from tfjob example https://github.com/kubeflow/katib/blob/master/examples/v1alpha3/tfjob-example.yaml

andreyvelich:
@johnugeorge Sure

andreyvelich:
> Can we change or add a test case to test the feature?
> 
> Thanks

I added e2e test with this feature for TFJob Experiment.
Do you want to test it in the experiment_controller_test.go also?

johnugeorge:
I see a problem in merging this now. If examples are updated, people trying katib release of 1.0 will see issues as code fix will not be included.  We have to hold this for some time till images are added to manifests

andreyvelich:
@johnugeorge Sounds good to me.

andreyvelich:
In order to not confuse user with new example of TFJob, I removed changes in the example.
As @johnugeorge mentioned above, if user uses Katib images from current 1.0 `manifests` release and new example without the goal, user will get the error.

We can update examples in the future versions and merge these changes now.

/cc @johnugeorge @gaocegege @hougangliu

PR #1083, opened 2020-03-12
UI: Support Trial Templates in all namespaces and all configMaps

andreyvelich:
After this PR, user can view/add/edit/delete trial templates in all namespaces and all configMaps.
Katib supports to choose any configMap for TrialTemplate, not only default template (`trial-template`).

Also, I changed view of Templates to looks nicer.

During this change I noticed, that it would be great if user's configMaps with TrialTemplate will be labeled, for example with label `app: katib-trial-templates`.
I added one example of this configMap in `manifests`. Without this label it is impossible to find all configMaps with Templates for the user in the UI.

In the future we can implement this feature (adding label to configMap) to the katib-controller, for example, when user submits Experiment with specific ConfigMap for trialTemplate.

Please, take a look.
/assign @johnugeorge @gaocegege @hougangliu 
/cc @richardsliu 

I still need to fix some small issues and test it, I marked WIP for now.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/1083)


andreyvelich:
I think, this PR is ready. 
I will remove WIP status.
I tested new functionality and it seems works.

I added Filter Panel to Trial Template View.
Mar-13-2020 01-08-17

Also, I added select boxes for Namespace and ConfigMap for creating Experiments, because it should work with new design.

Templ2

Please, take a look.

manifests/v1alpha3/katib-controller/trialTemplateConfigmapLabeled.yaml
```diff
@@ -0,0 +1,95 @@
+apiVersion: v1
```
johnugeorge: I am reluctant to have them in v1alpha3 itself as this would confuse users. What do you think?

gaocegege:
LGTM. Awesome.

manifests/v1alpha3/katib-controller/trialTemplateConfigmapLabeled.yaml
andreyvelich: This change will not affect old trail template and controller. 
User just can't see old Trial Templates and Submit Experiment by Parameters from the UI, using not labeled trial templates.
Other functionality in the UI will work, so they can submit Experiment using yaml file and monitor Experiments

You think that it is better not include this trial template to manifests in v1alpha3?

nrchakradhar:
@andreyvelich I have not seen the code/changes. So please excuse for this question.
Can service account also be specified?

andreyvelich:
> @andreyvelich I have not seen the code/changes. So please excuse for this question.
> Can service account also be specified?

Do you mean specifying service account for the Trial Templates?

nrchakradhar:
Do you mean specifying service account for the Trial Templates?
Yes

andreyvelich:
> Do you mean specifying service account for the Trial Templates?
> Yes

Yes, you can specify service account for the Trials, since it is just a Template with Job, PytorchJob, TFJob yaml. You can make these templates as you wish.

Issue #1079, opened 2020-03-06
How to specify service account for an experiment


nrchakradhar:
/kind bug

**What steps did you take and what happened:**
Using kubectl when we create an experiment, how can we specify the "serivceaccount" to be used?
I understand that Experiment CR creates a Suggestion CR which creates a Deployment. For this Deployment, I want to specify a specific serviceaccount to be used? Is this possible.?

As per the code here, I do not see a way to set the serviceaccount.
https://github.com/kubeflow/katib/blob/82679294ecaecd83dba30d42f9ce9e36f732cd89/pkg/controller.v1alpha3/suggestion/composer/composer.go#L119

**What did you expect to happen:**
It should be possible to specify service account for Katib experiments.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: 0.7 and 1.0
- Minikube version:
- Kubernetes version: (use `kubectl version`): 1.15
- OS (e.g. from `/etc/os-release`):

andreyvelich:
@nrchakradhar Right now, all components regarding to Experiment is created by katib-controller with this serviceAccount: https://github.com/kubeflow/katib/blob/82679294ecaecd83dba30d42f9ce9e36f732cd89/manifests/v1alpha3/katib-controller/rbac.yaml#L71

I am not sure that we have functionality to add service account to Suggestion Deployment.
We can add functionality to add ServiceAccount to Suggestion Deployment using katib-config (https://github.com/kubeflow/katib/blob/82679294ecaecd83dba30d42f9ce9e36f732cd89/manifests/v1alpha3/katib-controller/katib-config.yaml#L24).

What do you think @gaocegege @johnugeorge ?

nrchakradhar:
Karib controller using its service account is fine. The suggestion cr which creates deployment in user namespace should also have a provision for using any service account provided by user. The service account would need role binding with required permissions.

andreyvelich:
@nrchakradhar Yes, I agree, it can be useful.

andreyvelich:
I think, as I mentioned above, `katib-config` can be the place where we can add ServiceAccount for the Suggestion deployment.
Also, we can include it to the Experiment yaml file, for example in `ExperimentSpec`.

What do you think @richardsliu  @gaocegege @hougangliu ?

hougangliu:
@andreyvelich agree

andreyvelich:
I will add this feature

/assign @andreyvelich

PR #1092, opened 2020-03-16
Enable to add Service Account Name in Katib config

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1079.

I added functionality to add Service Account Name using Katib config.

We can add Validation part, for example here: https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1alpha3/experiment/validation_webhook.go#L88, to Validate Katib config.

For example, users can't submit Experiment with Service Account, if they don't have appropriate Service Account in Experiment Namespace. I don't know if it useful or not for us.

What do you think @gaocegege @johnugeorge @hougangliu ?

/cc @nrchakradhar.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/katib/1092)

### DIFF
--- a/pkg/controller.v1alpha3/consts/const.go
+++ b/pkg/controller.v1alpha3/consts/const.go
@@ -73,6 +73,8 @@ const (
 	DefaultDiskRequest = "500Mi"
 	// LabelSuggestionImagePullPolicy is the name of suggestion image pull policy in configmap.
 	LabelSuggestionImagePullPolicy = "imagePullPolicy"
+	// LabelSuggestionServiceAccountName is the name of suggestion service account in configmap.
+	LabelSuggestionServiceAccountName = "serviceAccountName"
 	// DefaultImagePullPolicy is the default value for image pull policy.
 	DefaultImagePullPolicy = "IfNotPresent"
 	// LabelMetricsCollectorSidecar is the name of metrics collector config in configmap.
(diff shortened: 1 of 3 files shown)


nrchakradhar:
Can you also please update one example to show how to use this. Is it possible to use different service accounts for different experiments.

andreyvelich:
@nrchakradhar In the current implementation it is not possible to use different Service Account for different Experiment. You need to change katib-config every time you want to change service account for the algorithm.
We can add service account to Experiment API. In that case, each Experiment will have their own service account. 
What do you think @johnugeorge @gaocegege @hougangliu ?

gaocegege:
> In the current implementation it is not possible to use different Service Account for different Experiment.

The main problem is that suggestion does not support serviceaccount, right?

andreyvelich:
> > In the current implementation it is not possible to use different Service Account for different Experiment.
> 
> The main problem is that suggestion does not support serviceaccount, right?

Yes, suggestion Deployment for some cases can have Service Account.

gaocegege:
Then 

> We can add service account to Experiment API. In that case, each Experiment will have their own service account.

LGTM.

andreyvelich:
/hold

Hold until we decide the solution for this.

johnugeorge:
One problem I see is that Suggestions are hidden to a normal user. If it is just used for suggestion deployments, do we need to expose it?

nrchakradhar:
Service accounts are subjects. So even if some CRs are hidden, is it better they run as part of that service account providing greater isolation.

johnugeorge:
@nrchakradhar I meant, putting service account parameter in Experiment API vs Katib config.

andreyvelich:
For now, we can merge this implementation with `katib-config` changes. In the future, if more users need to have `serviceAccount` per Experiment, we will change it.

@nrchakradhar I will add documentation about parameters in `katib-config`.

johnugeorge:
Sounds good
/lgtm
/approve

Issue #981, opened 2019-12-20
ERROR:grpc._server:Exception calling application: Method not implemented!


devxoxo:
/kind bug

Hi, I'm having trouble using katib v1alpha3.
First, I installed katib by the followings
  1) git clone https://github.com/kubeflow/katib
  2) sh katib/scripts/v1alpha3/deploy.sh

And I tried to apply random-example.yaml
kubectl apply -f random-example.yaml
(example in katib/examples/v1alpha3)

**Results:**
kubectl get pods -n kubeflow
NAME                                     READY   STATUS    RESTARTS   AGE
katib-controller-6c6974678d-zsnlc        1/1     Running   1          24m
katib-db-558f649dc6-8cd9t                1/1     Running   0          24m
katib-manager-5f74bdff84-4d78z           1/1     Running   0          24m
katib-ui-6568bd6b44-qbq5k                1/1     Running   0          24m
random-example-random-846dc99654-bxb8j   1/1     Running   0          23m

kubectl get trials -n kubeflow
NAME                      TYPE      STATUS   AGE
random-example-drpkvb4b   Running   True     23m
random-example-k7xv6ktt   Running   True     23m
random-example-w6jlwdp2   Running   True     23m

kubectl get experiment -n kubeflow -oyaml
apiVersion: v1
items:
- apiVersion: kubeflow.org/v1alpha3
  kind: Experiment
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #1102, opened 2020-03-23
Fix custom Katib DB Manager env variables

andreyvelich:
Fixes this comment: https://github.com/kubeflow/katib/issues/981#issuecomment-600170773.

If users want to use their own DB Manager, they must specify these env variables in **Katib controller** and **Katib UI** manifests:

```
KATIB_DB_MANAGER_SERVICE_NAMESPACE = <db manager namespace>
KATIB_DB_MANAGER_SERVICE_IP = <db manager ip (service name)>
KATIB_DB_MANAGER_SERVICE_PORT = <db manager port>
```

Also, if user specifies `KATIB_DB_MANAGER_SERVICE_NAMESPACE=''`, we will use `KATIB_DB_MANAGER_SERVICE_IP:KATIB_DB_MANAGER_SERVICE_PORT` for DB manager address .

This might be useful, if user needs to send metrics to DataBase which is not deployed in Kubernetes.

I tried to deploy `katib-db-manager` in the different namespace than `katib-controller`. After set correct env, metrics got collected.

In the next PR I will create doc about all env variables that can be specifying in the manifest.

What do you think @johnugeorge @gaocegege @hougangliu ?

/cc @kunalyogenshah

### DIFF
--- a/pkg/common/v1alpha3/katib_manager_util.go
+++ b/pkg/common/v1alpha3/katib_manager_util.go
@@ -31,12 +31,15 @@ type katibDBManagerClientAndConn struct {
 
 // GetDBManagerAddr returns address of Katib DB Manager
 func GetDBManagerAddr() string {
+	dbManagerNS := consts.DefaultKatibDBManagerServiceNamespace
+	dbManagerIP := consts.DefaultKatibDBManagerServiceIP
+	dbManagerPort := consts.DefaultKatibDBManagerServicePort
 
-	if len(consts.DefaultKatibDBManagerNamespace) != 0 {
-		return consts.DefaultKatibDBManagerIP + "." + consts.DefaultKatibDBManagerNamespace + ":" + consts.DefaultKatibDBManagerPort
+	if len(dbManagerNS) != 0 {
+		return dbManagerIP + "." + dbManagerNS + ":" + dbManagerPort
 	}
 
-	return consts.DefaultKatibDBManagerIP + ":" + consts.DefaultKatibDBManagerPort
+	return dbManagerIP + ":" + dbManagerPort
 }
 
 func getKatibDBManagerClientAndConn() (*katibDBManagerClientAndConn, error) {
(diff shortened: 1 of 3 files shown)


pkg/common/v1alpha3/katib_manager_util.go
```diff
@@ -17,40 +17,29 @@ package v1alpha3
...
-		return KatibDBManagerService + "." + ns + ":" + KatibDBManagerPort
+	dbManagerNS := consts.DefaultKatibDBManagerServiceNamespace
+	dbManagerIP := consts.DefaultKatibDBManagerServiceIP
+	dbManagerPort := consts.DefaultKatibDBManagerServicePort
+
+	if len(dbManagerNS) != 0 {
+		return dbManagerIP + "." + dbManagerNS + ":" + dbManagerPort
 	}
+
+	return dbManagerIP + ":" + dbManagerPort
```
johnugeorge: Is this logic be used anytime?

johnugeorge: This will be used when namespace is explicitly set to empty,

pkg/controller.v1alpha3/consts/const.go
```diff
@@ -126,4 +135,11 @@ var (
 	DefaultKatibNamespace = env.GetEnvOrDefault(DefaultKatibNamespaceEnvName, "kubeflow")
 	// DefaultComposer is the default composer of katib suggestion.
 	DefaultComposer = env.GetEnvOrDefault(DefaultKatibComposerEnvName, "General")
+
+	// DefaultKatibDBManagerServiceNamespace is the default namespace of Katib DB Manager
+	DefaultKatibDBManagerServiceNamespace = env.GetEnvOrDefault(DefaultKatibDBManagerServiceNamespaceEnvName, DefaultKatibNamespace)
```
gaocegege: LGTM but I think we should illustrate the relationship between DefaultKatibNamespace and DefaultKatibDBManagerServiceNamespace in the doc in the future.

andreyvelich: Sure, I will create doc about all env variables in Katib components today

PR #1108, opened 2020-03-24
Fix logger namespace

c-bata:

See the following quotes from the Python official documentation.

> The name is potentially a period-separated hierarchical value, like foo.bar.baz (though it could also be just plain foo, for example). Loggers that are further down in the hierarchical list are children of loggers higher up in the list. For example, given a logger with a name of foo, loggers with names of foo.bar, foo.bar.baz, and foo.bam are all descendants of foo. The logger name hierarchy is analogous to the Python package hierarchy, and identical to it if you organise your loggers on a per-module basis using the recommended construction logging.getLogger(__name__).

https://docs.python.org/3/library/logging.html

None

None

**Release note**:

```release-note
NONE
```


gaocegege:
/lgtm
/approve
/ok-to-test

Thanks for the PR

PR #1110, opened 2020-03-24
Documentation for the Katib Config

andreyvelich:
I added documentation about Katib config. 

Please, check if everything is fine.

/assign @johnugeorge @gaocegege 

/cc @richardsliu @sarahmaddox

PR #1111, opened 2020-03-24
Documentation for the Katib env variables

andreyvelich:
I added documentation about all env variables that can be set in Katib components.
I hope, I added everything.

I am not sure that we need `DB_NAME` variable, since we support only `mysql` https://github.com/kubeflow/katib/blob/master/pkg/db/v1alpha3/db.go#L12.

Please take a look.

/assign @johnugeorge @gaocegege @hougangliu 
/cc @richardsliu @sarahmaddox

### DIFF
--- a/docs/components-env-variables.md
+++ b/docs/components-env-variables.md
@@ -1,10 +1,10 @@
 # Katib components environment variables
 
-Here you can find information about environment variables in each Katib component. If you want to deploy Katib not in default way, you can modify some of this variables.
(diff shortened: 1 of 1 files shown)


sarahmaddox:
@andreyvelich Thanks for contributing to the Katib docs! See my comment on PR https://github.com/kubeflow/katib/pull/1110#issuecomment-603436439 about moving the docs to kubeflow.org, rather than adding them to the katib repo.

andreyvelich:
/close
Close this, I created PR in the website: https://github.com/kubeflow/website/pull/1839.

PR #1116, opened 2020-03-31
Optimize Chocolate Suggestion

andreyvelich:
I changed Chocolate Suggestion to be consistent with other Suggestions (Hyperopt, skopt).
1. Chocolate suggestion should analyse search space only for the first run.
2. I changed mechanism of saving information to sqlite3 DB. We don't need to override all Trial data every GetSuggestions call. We just update current DB data with the new Trial loss and Trial name for appropriate `_chocolate_id`. For this, I use list of recorded dictionaries `self.created_trials` variable. I update this list when receive new information about Trials.
3. Because of that, Chocolate sampler generates correct `_chocolate_id` for each new assignment and Chocolate algorithms should work correct. User can check data in sqlite3 DB.
4. Added few logs steps.
5. Removed warnings from SQLAlchemy about different threads by adding: `check_same_thread=False` flag.

I tested this Suggestion on `grid`, `chocolate-quasirandom` and `chocolate-mocmaes`. All Experiments were succeeded. Also, I tested `grid` on 150 Trials and return Experiment functionality.

/assign @gaocegege 
/cc @johnugeorge 

While testing I found one thing. In very rare cases, Katib controller calls GetSuggestion and doesn't create appropriate jobs.

It might be because of these sequence of actions in the controller:
1. Katib controller calls `GetSuggestion` with `request_number` = 1.
2. Katib controller gets response from `GetSuggestion`
3. Katib controller gets these errors. grid-lg9jzkgw.- Previously created Trial.
```
"logger":"kubebuilder.controller","msg":"Reconciler error","controller":"suggestion-controller","request":"kubeflow/grid","error":"Operation cannot be fulfilled on suggestions.kubeflow.org \"grid\": the object has been modified; please apply your changes to the latest version and try again"
"logger": "trial-controller","msg":"Reconcile trial error","Trial":"kubeflow/grid-lg9jzkgw","error":"jobs.batch \"grid-lg9jzkgw\" not found"
```
4. At the same time Katib controller calls `GetSuggestion` with `request_number` = 2, because `activeCount`<`parallelCount` and got new suggestion with 2 Trials.
5. Katib controller created new Trials from the latest `GetSuggestion` response.

So we lose the first suggested Trial.

### DIFF
--- a/pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py
+++ b/pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py
@@ -9,24 +9,35 @@ from pkg.suggestion.v1alpha3.internal.trial import *
 
 logger = logging.getLogger(__name__)
 
+DB_ADDRESS = "sqlite:///my_db.db?check_same_thread=False"
+DB_FIELD_LOSS = "_loss"
+DB_FIELD_CHOCOLATE_ID = "_chocolate_id"
+DB_FIELD_TRIAL_NAME = "_trial_name"
+
 
 class BaseChocolateService(object):
     """
     Refer to https://chocolate.readthedocs.io/
     """
 
-    def __init__(self, algorithm_name=""):
-        self.algorithm_name = algorithm_name
+    def __init__(self, algorithm_name, search_space):
+        self.conn = choco.SQLiteConnection(DB_ADDRESS)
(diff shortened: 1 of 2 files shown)


andreyvelich:
CI passed. @gaocegege Can you take a look again, please?

Issue #1119, opened 2020-04-03
Step field does not work with int parameters


StefanoFioravanzo:
/kind bug

**What steps did you take and what happened:**
I create an experiment over two integer parameters, using grid search. The template looks like this:

```yaml
apiVersion: kubeflow.org/v1alpha3
kind: Experiment
metadata:
  labels:
    controller-tools.k8s.io: '1.0'
  name: katib-test
spec:
  algorithm:
    algorithmName: grid
  objective:
    additionalMetricNames:
    goal: 100
    objectiveMetricName: result
    type: maximize
  parameters:
  - feasibleSpace:
      max: '50'
      min: '1'
      step: '10'
    name: a
    parameterType: int
  - feasibleSpace:
      max: '50'
      min: '1'
      step: '9'
    name: b
    parameterType: int
  trialTemplate:
    goTemplate:
      rawTemplate: |
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: {{.Trial}}
          namespace: {{.NameSpace}}
        spec:
          template:
            metadata:
              annotations:
                sidecar.istio.io/inject: "false"
            spec:
              restartPolicy: Never
              containers:
                - name: {{.Trial}}
                  image: <myimage>
                  command:
                    - python3 -u -c "<some-command>"
```

The trial start as expected, but the `step` field doesn't have any effect. I see the parameters being increased by `1`, instead of `step`.

**What did you expect to happen:**
I would expect `step` to have effect on the input parameters of the trials

**Anything else you would like to add:**
When changing a parameter from `int` to `double`, the step has effect. Is this because this field works only for `double` type parameters? Is this documented somewhere?

**Environment:**

- Kubeflow version: 1.0
- Minikube version: 1.2.0 (MiniKF latests)
- Kubernetes version: 1.14

gaocegege:
It is by design, I think. step will be used only when the type is float.

StefanoFioravanzo:
@gaocegege is there a specific reason for this? It would make a lot of sense to be able to specify a step for integer parameters. As soon as we started to try out Katib we bumped into this.

gaocegege:
SGTM. I think we can support it. I think it is more like a historical problem.

andreyvelich:
I agree we can support it.
/assign @andreyvelich

PR #1123, opened 2020-04-03
Support step for int parameter in Chocolate and Hyperopt Suggestion

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1119.

I added parsing step in Internal Search Space function.
Default value for the int parameter step is 1.

/assign @gaocegege 
/cc @StefanoFioravanzo

### DIFF
--- a/pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py
+++ b/pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py
@@ -30,7 +30,7 @@ class BaseChocolateService(object):
             key = BaseChocolateService.encode(param.name)
             if param.type == INTEGER:
                 chocolate_search_space[key] = choco.quantized_uniform(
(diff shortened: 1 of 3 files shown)


StefanoFioravanzo:
@andreyvelich I see that the bot does not allow me to be a reviewer. I understand that I should add myself to https://github.com/kubeflow/internal-acls? Maybe to `devrel-team.members.txt`? cc @jlewi

andreyvelich:
@StefanoFioravanzo I believe you should be a Kubeflow member to left the review (https://github.com/kubeflow/internal-acls/blob/master/github-orgs/kubeflow/org.yaml#L23). 
Right @jlewi  ?

jlewi:
Yes to LGTM PRs you will need to become a member of the org. To become a member of the org you should have made contributions that you can point to in the PR applying to join.

PR #1133, opened 2020-04-09
Fix duplicated imports

c-bata:

* Add `__init__.py` to avoid using namespace package.
    * See https://dev.to/methane/don-t-omit-init-py-3hga.
* Fix duplicated import problems by `__all__` attribute.
    * To clarify this problem, I attached a screenshot of PyCharm.
    * ScreenShot 2020-04-09 23 15 58
    * `logging` is imported at L3 and used at L8. But Jetbrains IDE highlights `import logging` as gray. This means it is an unused line. The reason why it is highlighted as gray is using star imports ( `from pkg.suggestion.v1alpha3.internal.search_space import *` and `from pkg.suggestion.v1alpha3.internal.trial import *` ). ~To avoid this problem, I define ``__all__`` attribute. See https://docs.python.org/3/tutorial/modules.html#importing-from-a-package~

None

None

**Release note**:

```release-note
NONE
```


pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py
```diff
@@ -1,11 +1,8 @@
-import json
-import numpy as np
 import chocolate as choco
 import logging
 import base64
 
-from pkg.suggestion.v1alpha3.internal.search_space import *
-from pkg.suggestion.v1alpha3.internal.trial import *
+from pkg.suggestion.v1alpha3.internal import *
```
andreyvelich, line 5: In some files we are using directly functions/variables import like here: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1alpha3/chocolate_service.py#L8, but here we are using `import *`.
Maybe in all python files we should avoid `import *` and use only relevant imports, so we can easily see which method we are using from `search_space` and `trial` file?

What do you think?

c-bata: > Maybe in all python files we should avoid import * and use only relevant imports, so we can easily see which method we are using from search_space and trial file?

Yes, I agree you opinion. star-imports are known as anti-pattrens in Python.
OK, I'll remove star-imports.

c-bata:
It seems all star-imports are removed.

```
(venv) $ git co master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
(venv) $ git grep "import \*" | grep "\.py"
pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py:from pkg.suggestion.v1alpha3.internal.search_space import *
pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py:from pkg.suggestion.v1alpha3.internal.trial import *
pkg/suggestion/v1alpha3/hyperopt/base_hyperopt_service.py:from pkg.suggestion.v1alpha3.internal.search_space import *
pkg/suggestion/v1alpha3/hyperopt/base_hyperopt_service.py:from pkg.suggestion.v1alpha3.internal.trial import *
pkg/suggestion/v1alpha3/skopt/base_skopt_service.py:from pkg.suggestion.v1alpha3.internal.search_space import *
pkg/suggestion/v1alpha3/skopt/base_skopt_service.py:from pkg.suggestion.v1alpha3.internal.trial import *
(venv) $ git co fix-duplicated-imports
Switched to branch 'fix-duplicated-imports'
(venv) $ git grep "import \*" | grep "\.py"
(venv) $ 
```

pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py
```diff
@@ -1,11 +1,11 @@
-import json
-import numpy as np
 import chocolate as choco
 import logging
 import base64
 
-from pkg.suggestion.v1alpha3.internal.search_space import *
-from pkg.suggestion.v1alpha3.internal.trial import *
+from pkg.suggestion.v1alpha3.internal.search_space import (
+    INTEGER, DOUBLE, CATEGORICAL, DISCRETE, MAX_GOAL,
```
johnugeorge, line 6: Can we do this in a better way to look more clean? we might have more constants. May be wrap this?

c-bata: You mean that you want to import all constants, right?
Another solution is to define `__all__` attribute like [this revision](https://github.com/kubeflow/katib/commit/f99736db4c79009e8988b498090116b5b09230bb). But as the discussion of https://github.com/kubeflow/katib/pull/1133#discussion_r406294910, it is known as an anti-pattern in Python.

andreyvelich: @c-bata What do you think about creating file: `constant.py` under `/internal/` folder ?
And insert all constant there? I think it will be easier to have all constant in one file

c-bata: It sounds good 👍  @johnugeorge What do you think?

johnugeorge: Yes. SGTM

Issue #987, opened 2019-12-23
[feature] Show more information about best trial


gaocegege:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

https://github.com/kubeflow/katib/pull/986#issuecomment-568371597

> > We show metrics and parameter assignments. I am not sure if I get your idea. Could you please explain more?
> 
> for example, for now if objective metrics is `loss`, in experiment or trial status, it show the best `loss` metrics minimum value like below:
> 
> ```
> observation:
>   metrics:
>     - name: loss
>        value: 0.0001
> ```
> 
> we can detail it more for a better viewpoint like `ray`:
> 
> ```
> observation:
>  metrics:
>    - name: loss
>      min: 0.0001
>      max: 0.1234
>      latest: 0.1111
> ```

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

andreyvelich:
This was implemented.

Issue #1120, opened 2020-04-03
Suggestions pod does not fail when exception is raised


StefanoFioravanzo:
/kind bug

**What steps did you take and what happened:**
Created a Katib experiment, using `grid` search and two `int` parameters. CRD looks like this:

```yaml
apiVersion: kubeflow.org/v1alpha3
kind: Experiment
metadata:
  labels:
    controller-tools.k8s.io: '1.0'
  name: katib-simple-trial
spec:
  algorithm:
    algorithmName: grid
  parallelTrialCount: 1
  maxFailedTrialCount: 6
  maxTrialCount: 12
  objective:
    additionalMetricNames:
    goal: 100
    objectiveMetricName: result
    type: maximize
  parameters:
  - feasibleSpace:
      max: '50'
      min: '1'
      step: '10'
    name: a
    parameterType: int
  - feasibleSpace:
      max: '1'
      min: '50'
      step: '9'
    name: b
    parameterType: int
  trialTemplate:
    goTemplate:
      rawTemplate: |
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: {{.Trial}}
          namespace: {{.NameSpace}}
        spec:
          template:
            metadata:
              annotations:
                sidecar.istio.io/inject: "false"
            spec:
              restartPolicy: Never
              containers:
                - name: {{.Trial}}
                  image: <myimage>
                  command:
                    - python3 -u -c "<some_command>"
```
Since parameter `b` has `min=50` and `max=1`, I would expect the submission of the CRD to fail. 

**What did you expect to happen:**
What happens is that the suggestions pod is created and it starts to continuously produce the following error:

```
ERROR:grpc._server:Exception calling application: Low must be lower than high
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/grpc/_server.py", line 434, in _call_behavior
    response_or_iterator = behavior(argument, context)
  File "/usr/src/app/github.com/kubeflow/katib/pkg/suggestion/v1alpha3/chocolate_service.py", line 39, in GetSuggestions
    search_space, trials, request.request_number)
  File "/usr/src/app/github.com/kubeflow/katib/pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py", line 33, in getSuggestions
    int(param.min), int(param.max), 1)
  File "/usr/local/lib/python3.6/site-packages/chocolate/space.py", line 140, in __init__
    assert low < high, "Low must be lower than high"
[...]

PR #1140, opened 2020-04-13
extracting metric value in multiple ways

sperlingxx:
Currently, katib trial controller simply takes the best metric value among obversations. This approach will get the wrong objective value if training program doesn't preserve model (checkpoint)  with best performance and regard it as final result. Actually, a lot of training programs simply save the model after final step. 
So, it is necessary to support an another objective value extracting strategy: extracting the latest recorded value, which is the purpose of this PR.

------

After discussion,  I found that the approach raised in  #987  is better, in which we extract all min, max and latest metric value from observation logs.  When converting trial observations into `api_pb.Observation` for feeding suggestion service,  value will be chosen among min, max and latest metric according to `MetricStrategies`, a new field of `commonv1alpha3.ObjectiveSpec`.


sperlingxx:
@andreyvelich @gaocegege Could you help to review this?

pkg/apis/controller/trials/v1alpha3/trial_types.go
```diff
@@ -36,10 +36,21 @@ type TrialSpec struct {
 	// Whether to retain the trial run object after completed.
 	RetainRun bool `json:"retainRun,omitempty"`
 
+	// Describes how objective value will be extracted from collected metrics
+	ObjectiveExtract ObjectiveExtractType `json:"objectiveExtract,omitempty"`
+
```
andreyvelich, line 41: Actually, we have an issue for this https://github.com/kubeflow/katib/issues/987. Maybe instead of set up another env variable, in each run we can have `min`, `max` and `latest` in Objective.

andreyvelich: /cc @gaocegege

sperlingxx: @andreyvelich The idea extracting min, max and latest objective in each run looks good to me. But I think we also need an additional field like `ObjectiveExtractType ` to describe which kind of objective chosen to be compared with goal.
Of course env variable `DefaultObjectiveExtract` is not necessary since the default choice can be figured out by ObjectiveType.

andreyvelich: Then, maybe it should be part of Experiment? 
What do you think @johnugeorge ?

sperlingxx: @andreyvelich @johnugeorge 
After a detailed review, I found the modification of  `common_types.Observation` will cause enormous changes including corresponding interface of all existing suggestion services. 
Maybe we can keep `suggestionapi.Observation` (pb spec)  unchanged ? Only change   `common_types.Observation` to 
```yaml
observation:
 metrics:
   - name: loss
     min: 0.0001
     max: 0.1234
     latest: 0.1111
```

sperlingxx:
@johnugeorge  @andreyvelich @gaocegege   I think current PR is also ready.

pkg/apis/controller/common/v1alpha3/common_types.go
```diff
@@ -51,6 +51,8 @@ type ObjectiveSpec struct {
 	// This can be empty if we only care about the objective metric.
 	// Note: If we adopt a push instead of pull mechanism, this can be omitted completely.
 	AdditionalMetricNames []string `json:"additionalMetricNames,omitempty"`
+	// This field is allowed to missing, experiment defaulter (webhook) will fill it.
+	MetricStrategies map[string]MetricStrategy `json:"metricStrategies,omitempty"`
```
andreyvelich, line 55: After this change, I can see that you send Metrics to Suggestion according to MetricStrategies. Is it right way?
Maybe some Suggestions need min, max and latest Observation? Do we need to be consistent between manager: https://github.com/kubeflow/katib/blob/master/pkg/apis/manager/v1alpha3/api.proto#L177 and Trial API?
/cc @johnugeorge @gaocegege

sperlingxx: If we modify definition of observation in `api.proto`, we need to apply corresponding modifications on all suggestion services.

andreyvelich: Yes, I understand it. You think we don't need it in suggestions?

sperlingxx: @andreyvelich Yes. For suggestion service,  as far as I know, the reward of training program (F(x)) is the only necessary value.

gaocegege: I think it is a huge change to API. Can we place this change in the next release? Maybe v1beta1.

sperlingxx:
> The PR itself LGTM.
> 
> But prefer to propose in the next release.
> 
> @sperlingxx


I'm okay with that.

pkg/apis/controller/common/v1alpha3/common_types.go
andreyvelich: Agree with @gaocegege. Let's keep it for next release.

sperlingxx:
@gaocegege @andreyvelich I've migrated codes to `v1beta1`. And I merged modifications of `Metric` schema applied in current PR and  #1120 .  So, the `Metric` schema is 
```go
type Metric struct {
	Name   string  `json:"name,omitempty"`
	Min    float64 `json:"min,omitempty"`
	Max    float64 `json:"max,omitempty"`
	Latest string  `json:"latest,omitempty"`
}
``` 

`Latest` is changed to string-type in order to contain non-numeric values, while `Min` and `Max` are still float-type. Considering that max/min values become nonsence when metric values are non-numeric, they will be filled with `math.NaN.

gaocegege:
Thanks! I love the feature.

/lgtm

/cc @johnugeorge @andreyvelich

pkg/controller.v1beta1/trial/managerclient/managerclient.go
```diff
@@ -35,6 +35,31 @@ func (d *DefaultClient) GetTrialObservationLog(
...
 	}
+	if reply.ObservationLog == nil {
+		reply.ObservationLog = &api_pb.ObservationLog{}
+	}
+	if reply.ObservationLog.MetricLogs == nil {
+		reply.ObservationLog.MetricLogs = []*api_pb.MetricLog{}
+	}
+	// fetch additional metrics if exists
+	metricLogs := reply.ObservationLog.MetricLogs
+	for _, metricName := range instance.Spec.Objective.AdditionalMetricNames {
```
andreyvelich, line 46: Is that correct to add AdditionalMetricNames in `GetTrialObservationLog` function?
I believe this function is needed to add  `Observation` to Trial instance, right?
Currently, Trial `Observation` is related only to `ObjectiveMetricName`.

gaocegege: I think so.

sperlingxx: @andreyvelich Yes, I am trying to append addtional metrics into `Trial.Status.Observation`.  So, we can get full metrics of each trial via kubernetes API. Are you considering this modification will confuse our users/developers ?

andreyvelich: Ok, I am just worried if the Suggestions will work correct.
I checked here: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/internal/trial.py#L33, we add only objective metric to target metric. So, I think it should work fine.

andreyvelich:
@sperlingxx Can you submit one example with `metricsStrategies`, please?
You can show how to specify Strategy for various Experiment metrics.

pkg/controller.v1beta1/trial/trial_controller_util.go
```diff
@@ -171,34 +165,51 @@ func isJobSucceeded(jobCondition *commonv1.JobCondition) bool {
...
+			if math.IsNaN(metric.Max) || floatValue > metric.Max {
+				metric.Max = floatValue
 			}
 		}
+		currentTime, err := time.Parse(time.RFC3339Nano, metricLog.TimeStamp)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse timestamps %s: %e", metricLog.TimeStamp, err)
+		}
+		timestamp, _ := timestamps[metricLog.Metric.Name]
+		if timestamp == nil || !timestamp.After(currentTime) {
```
andreyvelich, line 201: What if we have more than one metric with the same timestamp? 
Maybe we can extract Latest metric using latest element from the `metricLogs` for each metric name, what do you think?

sperlingxx: I agree. And I think, with current implementation, katib will use latest element if there exists multiple metricLogs with same timestamp. 
The condition `!timestamp.After(currentTime)` ensure latest metric value will update only if current timestamp is not earlier than historical latest timestamp. So, when they are equal, the updation will still be applied.

sperlingxx:
> @sperlingxx Can you submit one example with `metricsStrategies`, please?
> You can show how to specify Strategy for various Experiment metrics.

Do you mean providing a full executable example like those in examples folder?

andreyvelich:
> > @sperlingxx Can you submit one example with `metricsStrategies`, please?
> > You can show how to specify Strategy for various Experiment metrics.
> 
> Do you mean providing a full executable example like those in examples folder?

Yes, just a simple one yaml example, you can use random algorithm.

pkg/controller.v1beta1/trial/trial_controller_util.go
andreyvelich: Should we change it to `!currentTime.After(timestamp)` ? 
Since `timestamp` <=  `currentTime`, because `currentTime` is the new value from `metricLog` and `timestamp` historical value that we have recorded ?

sperlingxx:
> > > @sperlingxx Can you submit one example with `metricsStrategies`, please?
> > > You can show how to specify Strategy for various Experiment metrics.
> > 
> > 
> > Do you mean providing a full executable example like those in examples folder?
> 
> Yes, just a simple one yaml example, you can use random algorithm.

Done.

sperlingxx:
I think this PR is ready for merging. @gaocegege @johnugeorge

examples/v1beta1/metric-strategy-example.yaml
```diff
@@ -0,0 +1,64 @@
...
+spec:
+  objective:
+    type: maximize
+    goal: 0.99
+    objectiveMetricName: Validation-accuracy
+    additionalMetricNames:
+      - Train-accuracy
+    metricStrategies:
+      Validation-accuracy: "max"
+      Train-accuracy: "latest"
```
gaocegege, line 17: Can we avoid `-` in name?

It is not popular in Kubernetes community. Maybe we can use:

```
  - name: Train-accuracy
    value: latest
  - name: Validation-accuracy
    value: max
```
```

sperlingxx: @gaocegege Okay, I've applied this suggestion.

pkg/controller.v1beta1/trial/trial_controller_util.go
andreyvelich: @sperlingxx Have you tried to test it?

andreyvelich:
@sperlingxx Thank you for doing this!
Overall lgtm, only this question: https://github.com/kubeflow/katib/pull/1140#discussion_r433827984

pkg/controller.v1beta1/trial/trial_controller_util.go
sperlingxx: Hi @andreyvelich , I missed this comment sereval days ago.
 I think `!timestamp.After(currentTime)` represents `not ({timestamp} > {currentTime})`, which equals to ` {timestamp} <= {currentTime}`.
And there is a test case on extracting latest records if their values are same. Here is the [link](https://github.com/kubeflow/katib/pull/1140/files#diff-89fc68ee85fb96116fd4b809f044ea8bR256).

andreyvelich: @sperlingxx Got it. Thank you

johnugeorge:
Sorry for being late. 
/lgtm
 
@sperlingxx can you rebase?

sperlingxx:
> Sorry for being late.
> /lgtm
> 
> @sperlingxx can you rebase?

It's Done!

PR #1141, opened 2020-04-13
katib ui: adapt environment in which cluster role is unavailable

sperlingxx:
Katib UI fails to get experiment list when cluster role is unavailable.  This problem is fixed in this PR  through additional environment variable representing available namespaces.


sperlingxx:
@andreyvelich Could you help to review this PR?

pkg/ui/v1alpha3/const.go
```diff
@@ -0,0 +1,21 @@
...
+const (
+	AvailableNameSpaceEnvName = "KATIB_UI_AVAILABLE_NS"
+	ClusterRoleKey            = "KATIB_UI_CLUSTER_ROLE"
+)
+
+var (
+	availableNameSpaces, hasClusterRole = func() ([]string, bool) {
+		ns := env.GetEnvOrDefault(AvailableNameSpaceEnvName, ClusterRoleKey)
+		if ns == ClusterRoleKey {
+			return []string{""}, true
```
c-bata, line 17: I'm not confident but do you mistake it for `[]string{}`?

And I think you can use `init()` function like:

```go
var (
    availableNameSpaces []string
    hasClusterRole      bool
)

func init() {
    ns := env.GetEnvOrDefault(AvailableNameSpaceEnvName, ClusterRoleKey)
    ...
}
```

sperlingxx: @c-bata Thanks for your suggestions!
1. It should be `[]string{""}` which means no namespace options(restriction). If `[]string{}`, it will use default name space defined by `controller.v1alpha3/consts.DefaultKatibNamespace`.
2. I have replaced with init function

c-bata: I see. Thank you for your explanation!

pkg/ui/v1alpha3/backend.go
```diff
@@ -232,17 +232,13 @@ func (k *KatibUIHandler) DeleteTemplate(w http.ResponseWriter, r *http.Request)
 
 func (k *KatibUIHandler) FetchNamespaces(w http.ResponseWriter, r *http.Request) {
 
-	namespaceList, err := k.katibClient.GetNamespaceList()
+	// Get all available namespaces
+	namespaces, err := k.getAvailableNamespaces()
 	if err != nil {
 		log.Printf("GetNamespaceList failed: %v", err)
```
johnugeorge, line 238: change error message

pkg/ui/v1alpha3/const.go
```diff
@@ -0,0 +1,30 @@
...
+	ClusterRoleKey = "KATIB_UI_CLUSTER_ROLE"
+)
+
+var (
+	availableNameSpaces []string
+	hasClusterRole      bool
+)
+
+func init() {
+	ns := env.GetEnvOrDefault(AvailableNameSpaceEnvName, ClusterRoleKey)
```
johnugeorge, line 21: Couple of questions here

1.  Is it right to get the namespaces during init function? What happens if env variable is changed later?
2.  GetEnvOrDefault has fallback value as the second argument.  Did you mean to keep ClusterRoleKey as the default value?

pkg/ui/v1alpha3/nas.go
```diff
@@ -13,8 +13,8 @@ import (
 
 func (k *KatibUIHandler) FetchAllNASJobs(w http.ResponseWriter, r *http.Request) {
 	//enableCors(&w)
-	// Use "" to get experiments in all namespaces.
-	jobs, err := k.getExperimentList("", JobTypeNAS)
+	// Get NAS-related experiments in all available namespaces.
+	jobs, err := k.getExperimentList(availableNameSpaces, JobTypeNAS)
```
johnugeorge, line 17: What do you expect when you have multiple namespaces passed?

johnugeorge:
Btwn, what is your use case? Can you explain the deployment requirement?

pkg/ui/v1alpha3/nas.go
andreyvelich: Currently, in Katib Client: https://github.com/kubeflow/katib/blob/master/pkg/util/v1alpha3/katibclient/katib_client.go#L210, we extract only first namespace from the list, so it will not give you all available namespaces.

sperlingxx: @andreyvelich Yes, but value of namespace is fixed on "" (https://github.com/kubeflow/katib/blob/master/pkg/ui/v1alpha3/hp.go#L22) which leads to list experiments in cluster scope.

sperlingxx:
> Btwn, what is your use case? Can you explain the deployment requirement?
In internal k8s clusters of Antfin,  service accounts of third party operators (like kubeflow operators) have no access to  `ClusterRole` and `ClusterRoleBinding` for security reasons.

pkg/ui/v1alpha3/nas.go
sperlingxx: @johnugeorge It seems only first namespace will be taken. It is a problem.

pkg/ui/v1alpha3/const.go
sperlingxx: @johnugeorge 
1.  To me, this env variable should be set via manifest yaml file. I can't see any motivation to modify this env variable later.
2. Yes.

andreyvelich:
> > Btwn, what is your use case? Can you explain the deployment requirement?
>> In internal k8s clusters of Antfin,  service accounts of third party operators (like kubeflow operators) have no access to  `ClusterRole` and `ClusterRoleBinding` for security reasons.

For this reason, do you have access only for the namespace where Katib UI was deployed?
Also, in your case, how Katib controller works? Since it is also required ClusterRole to deploy Jobs not in `DefaultNamespace`?

sperlingxx:
> > > Btwn, what is your use case? Can you explain the deployment requirement?
> > > In internal k8s clusters of Antfin,  service accounts of third party operators (like kubeflow operators) have no access to  `ClusterRole` and `ClusterRoleBinding` for security reasons.
> 
> For this reason, do you have access only for the namespace where Katib UI was deployed?
> Also, in your case, how Katib controller works? Since it is also required ClusterRole to deploy Jobs not in `DefaultNamespace`?

In my company, all machine learning related resources are shared the same namespace. So, cluster role is not required.

andreyvelich:
If Katib UI can't fetch all namespaces, just get Experiments and Namespaces only from namespace where Katib UI is deployed.
What do you think about it @sperlingxx ?

sperlingxx:
> If Katib UI can't fetch all namespaces, just get Experiments and Namespaces only from namespace where Katib UI is deployed.
> What do you think about it @sperlingxx ?


Good idea!

pkg/ui/v1alpha3/backend.go
sperlingxx: It's done

johnugeorge:
> > If Katib UI can't fetch all namespaces, just get Experiments and Namespaces only from namespace where Katib UI is deployed.
> > What do you think about it @sperlingxx ?
> 
> Good idea!

In this case, do you really need extra environment variable? If you do not have permissions,  get experiments from own namespace else do the current way.

sperlingxx:
@johnugeorge  I think we need whether has permission from env variable. In alternative, katib will try to get objects in cluster scope at first. If failed then fallback to get objects from own namespace.

johnugeorge:
> @johnugeorge I think we need whether has permission from env variable. In alternative, katib will try to get objects in cluster scope at first. If failed then fallback to get objects from own namespace.

+1 to this.

andreyvelich:
> In alternative, katib will try to get objects in cluster scope at first. If failed then fallback to get objects from own namespace.

I agree with it. Katib UI should try to get objects from all namespaces and if it fails, just get object from own namespace.

sperlingxx:
@johnugeorge @andreyvelich I think current PR is ready.

Issue #1124, opened 2020-04-03
Support string values for categorical parameter in Grid Suggestion


andreyvelich:
/kind feature

User can't specify string values for Grid suggestion in Categorical parameter.
Currently, our Chocolate suggestion supports only int or float values for the list in Categorical parameter. See here: https://github.com/kubeflow/katib/blob/master/examples/v1alpha3/grid-example.yaml#L37.

User should be able to specify any values in the list.

We can convert string values from the list to integers and save them in Chocolate DB.

StefanoFioravanzo:
Hey @andreyvelich I was testing some grid experiments and this has been my experience:

- Running a grid experiment with one `int` argument, one `double` argument and one `categorical` argument (with integers as strings), worked as expected. The trials were generated and the search space properly explored.
- When running a grid experiment with just one argument, of type categorical, just 2 trials get generated and then the experiment hangs. The suggestions pod produces this log `Chocolate db is exhausted, increase Search Space or decrease maxTrialCount!`, even though there are still some values in the list that have not been tried.

So I don't get what is the underlying behaviour, can you explain why this is happening?

andreyvelich:
> Hey @andreyvelich I was testing some grid experiments and this has been my experience:
> 
> * Running a grid experiment with one `int` argument, one `double` argument and one `categorical` argument (with integers as strings), worked as expected. The trials were generated and the search space properly explored.

Yes, for me it works as well. Only problem with the strings values. As I said above, we can convert these values to appropriate list of integer values and match them when we create Assignment for the Trial.

> * When running a grid experiment with just one argument, of type categorical, just 2 trials get generated and then the experiment hangs. The suggestions pod produces this log `Chocolate db is exhausted, increase Search Space or decrease maxTrialCount!`, even though there are still some values in the list that have not been tried.
> 
> So I don't get what is the underlying behaviour, can you explain why this is happening?

How many `maxTrialCount` have you specified? Can you show me your search space also.

This can happen because Chocolate suggestion has generated all possible combination of parameters from the Search Space.
This PR: https://github.com/kubeflow/katib/pull/1116 should fix some of these problems.

After this PR, inside suggestion sqlite3 DB, you should see all parameters that Chocolate suggestion was generated with correct Chocolate ID (ID of generated set of parameters).

StefanoFioravanzo:
[...]

PR #1149, opened 2020-04-16
Support Categorical string values in Chocolate Suggestion

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1124.
After this PR all Chocolate algorithms should support string values for Categorical parameters.

I insert index of the list to DB instead of list value.


sperlingxx:
LGTM

Issue #1152, opened 2020-04-16
Goptuna Suggestion Katib integration


andreyvelich:
/kind feature

For Katib integration with Goptuna we have to make these steps:

- [x] Suggestion source code

- [x] CMA-ES yaml example

- [x] Training container source code, if it is not possible to use mxnet example: https://github.com/kubeflow/katib/tree/master/examples/v1alpha3/mxnet-mnist

- [x] Build Goptuna image in CI and update Katib config

- [x] Create e2e test in CI

/cc @c-bata

c-bata:
Thank you!

/assign @c-bata

andreyvelich:
@c-bata I can help you with 4th and 5th step related to CI.
Can you submit yaml example and check training container, please?

c-bata:
Sure! Thank you.

c-bata:
I'm in two minds which example should I add:

## mxnet-mninst example

![ScreenShot 2020-04-17 12 54 52](https://user-images.githubusercontent.com/5564044/79530579-a76dad80-80aa-11ea-8c3f-3b318efec1ee.png)

<details>

<summary>experiment</summary>

This experiment is almost the same of `tpe-example`. Please note that Goptuna's CMA-ES does not support categorical search space. So `--optimizer` param is sampled by Goptuna independent sampler (default: random sampler).

```
apiVersion: "kubeflow.org/v1alpha3"
kind: Experiment
metadata:
  namespace: kubeflow
  labels:
    controller-tools.k8s.io: "1.0"
  name: cmaes-example
spec:
  objective:
    type: maximize
    goal: 0.99
    objectiveMetricName: Validation-accuracy
    additionalMetricNames:
      - Train-accuracy
  algorithm:
    algorithmName: cmaes
  parallelTrialCount: 3
  maxTrialCount: 12
  maxFailedTrialCount: 3
  parameters:
    - name: --lr
      parameterType: double
      feasibleSpace:
        min: "0.01"
        max: "0.03"
    - name: --num-layers
      parameterType: int
      feasibleSpace:
        min: "2"
        max: "5"
    - name: --optimizer
      parameterType: categorical
      feasibleSpace:
        list:
        - sgd
        - adam
        - ftrl
  trialTemplate:
    goTemplate:
        rawTemplate: |-
          apiVersion: batch/v1
          kind: Job
          metadata:
            name: {{.Trial}}
            namespace: {{.NameSpace}}
          spec:
            template:
              spec:
                containers:
[...]

PR #1154, opened 2020-04-16
Build Goptuna Suggestion image in CI

andreyvelich:
Part of https://github.com/kubeflow/katib/issues/1152.

I added Goptuna suggestion build to CI and updated `katib-config` with the new image.

/assign @c-bata 
/cc @johnugeorge @gaocegege

### DIFF
--- a/manifests/v1alpha3/katib-controller/katib-config.yaml
+++ b/manifests/v1alpha3/katib-controller/katib-config.yaml
@@ -48,7 +48,6 @@ data:
         }
       },
       "cmaes": {
-        "image": "gcr.io/kubeflow-images-public/katib/v1alpha3/suggestion-goptuna",
-        "imagePullPolicy": "Always"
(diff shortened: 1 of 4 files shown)


andreyvelich:
Also, I updated prow.
We should trigger CI if we make changes in `cmd/goptuna`

manifests/v1alpha3/katib-controller/katib-config.yaml
```diff
@@ -46,5 +46,9 @@ data:
             "memory": "200Mi"
           }
         }
+      },
+      "cmaes": {
+        "image": "gcr.io/kubeflow-images-public/katib/v1alpha3/suggestion-goptuna",
+        "imagePullPolicy": "Always"
```
c-bata, line 52: [nits] Most image pull policies of other suggestion service is `IfNotPresent`. So I think you can omit this line.

andreyvelich: Sure, deleted

PR #1155, opened 2020-04-17
Add an example of Goptuna suggestion service

c-bata:
Add an example of Goptuna suggestion service.

ScreenShot 2020-04-17 12 54 52

Part of #1152

1. This experiment is almost the same with `tpe-example`. I just change the name and algorithm name.
1. Please note that Goptuna's CMA-ES does not support categorical search space. So `--optimizer` param is sampled by Goptuna independent sampler (default: random sampler).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/examples/v1alpha3/cmaes-example.yaml
+++ b/examples/v1alpha3/cmaes-example.yaml
@@ -1,0 +1,61 @@
+apiVersion: "kubeflow.org/v1alpha3"
+kind: Experiment
+metadata:
+  namespace: kubeflow
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: cmaes-example
+spec:
+  objective:
+    type: maximize
+    goal: 0.99
+    objectiveMetricName: Validation-accuracy
+    additionalMetricNames:
+      - Train-accuracy
+  algorithm:
+    algorithmName: cmaes
+  parallelTrialCount: 3
+  maxTrialCount: 12
+  maxFailedTrialCount: 3
+  parameters:
+    - name: --lr
+      parameterType: double
+      feasibleSpace:
+        min: "0.01"
+        max: "0.03"
+    - name: --num-layers
+      parameterType: int
+      feasibleSpace:
+        min: "2"
+        max: "5"
+    - name: --optimizer
+      parameterType: categorical
+      feasibleSpace:
+        list:
+        - sgd
+        - adam
+        - ftrl
+  trialTemplate:
+    goTemplate:
+        rawTemplate: |-
+          apiVersion: batch/v1
+          kind: Job
+          metadata:
+            name: {{.Trial}}
+            namespace: {{.NameSpace}}
+          spec:
+            template:
+              spec:
+                containers:
+                - name: {{.Trial}}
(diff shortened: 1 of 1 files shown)


c-bata:
I'm in two minds which example should I add mxnet-mnist or Himmelblau-function:

**Himmelblau function**

https://en.wikipedia.org/wiki/Himmelblau%27s_function

This function is quite simple and we can evaluate so fast. But it's not a realistic problem and we need to add CI setting to build docker image.

```python
import argparse
import logging

logging.basicConfig(filename='/var/log/katib.log', level=logging.DEBUG)


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--x1", dest='x1', type=float)
    parser.add_argument("--x2", dest='x2', type=float)

    args = parser.parse_args()
    print(args.x1, args.x2)

    evaluation = (args.x1 ** 2 + args.x2 - 11.0) ** 2 + (args.x1 + args.x2 ** 2 - 7.0) ** 2

    metric_msg = f'{{metricName: evaluation, metricValue: {evaluation:.4f}}};'
    logging.info(metric_msg)
    print(metric_msg)


if __name__ == '__main__':
    main()
```

```
FROM python:3-alpine

ADD main.py /usr/src/main.py
WORKDIR /usr/src
```

<details>

<summary>experiment</summary>


```yaml
apiVersion: "kubeflow.org/v1alpha3"
kind: Experiment
metadata:
  namespace: kubeflow
  labels:
    controller-tools.k8s.io: "1.0"
  name: cmaes-example
spec:
  objective:
    type: minimize
    goal: 0.0001
    objectiveMetricName: evaluation
  algorithm:
    algorithmName: cmaes
  metricsCollectorSpec:
    source:
      filter:
        metricsFormat:
          - "{metricName: ([\\w|-]+), metricValue: ((-?\\d+)(\\.\\d+)?)}"
      fileSystemPath:
        path: "/var/log/katib.log"
        kind: File
    collector:
      kind: File
  parallelTrialCount: 2
  maxTrialCount: 250
  maxFailedTrialCount: 3
  parameters:
    - name: x1
      parameterType: double
      feasibleSpace:
        min: "-10"
        max: "10"
    - name: x2
      parameterType: iint
      feasibleSpace:
        min: "-10"
        max: "10"
        step: "2"
  trialTemplate:
    goTemplate:
        rawTemplate: |-
          apiVersion: batch/v1
          kind: Job
          metadata:
            name: {{.Trial}}
            namespace: {{.NameSpace}}
          spec:
            template:
              spec:
                containers:
                - name: {{.Trial}}
                  image: docker.io/cbata/hello-katib-quadratic-function
                  command:
                  - "python3"
                  - "main.py"
                  {{- with .HyperParameters}}
                  {{- range .}}
                  - "--{{.Name}} {{.Value}}"
                  {{- end}}
                  {{- end}}
                restartPolicy: Never
```

</details>

andreyvelich:
@c-bata I think, let's keep mxnet for now. If we decide that users needs another example, we can submit Himmelblau function training container.

c-bata:
I see. Thanks!

PR #1161, opened 2020-04-22
restructure algorithm configuration for hyperopt_service

sperlingxx:
Restructuring algorithm configuration related codes for hyperopt_service, which makes it easier to modify algorithm schema in future.

@andreyvelich @gaocegege @c-bata

### DIFF
--- a/pkg/suggestion/v1alpha3/hyperopt/base_service.py
+++ b/pkg/suggestion/v1alpha3/hyperopt/base_service.py
@@ -12,8 +12,15 @@ RANDOM_ALGORITHM_NAME = "random"
 
 
 class BaseHyperoptService(object):
-    def __init__(self, algorithm_name=TPE_ALGORITHM_NAME, random_state=None, search_space=None):
+    def __init__(self,
+                 algorithm_name=TPE_ALGORITHM_NAME,
+                 algorithm_conf=None,
+                 search_space=None):
         self.algorithm_name = algorithm_name
+        self.algorithm_conf = algorithm_conf or {}
+        # pop common configurations
+        random_state = self.algorithm_conf.pop('random_state', None)
+
         if self.algorithm_name == TPE_ALGORITHM_NAME:
             self.hyperopt_algorithm = hyperopt.tpe.suggest
         elif self.algorithm_name == RANDOM_ALGORITHM_NAME:
@@ -194,7 +201,8 @@ class BaseHyperoptService(object):
                     domain=self.fmin.domain,
                     trials=self.fmin.trials,
                     seed=random_state,
-                    n_startup_jobs=request_number)
+                    n_startup_jobs=request_number,
+                    **self.algorithm_conf)
                 self.is_first_run = False
             else:
                 new_trials = []
@@ -205,7 +213,8 @@ class BaseHyperoptService(object):
                         domain=self.fmin.domain,
                         trials=self.fmin.trials,
                         seed=random_state,
-                        n_startup_jobs=request_number)[0])
+                        n_startup_jobs=request_number,
+                        **self.algorithm_conf)[0])
 
         # Construct return advisor Trials from new hyperopt Trials
         list_of_assignments = []
--- a/pkg/suggestion/v1alpha3/hyperopt/base_service.py
+++ b/pkg/suggestion/v1alpha3/hyperopt/base_service.py
@@ -29,8 +29,6 @@ class BaseHyperoptService(object):
         #     self.hyperopt_algorithm = hyperopt.anneal.suggest_batch
         # elif algorithm_name == 'hyperopt-mix':
         #     self.hyperopt_algorithm = hyperopt.mix.suggest
-        else:
-            raise Exception('"Failed to create the algorithm: {}'.format(algorithm_name))
 
         self.search_space = search_space
         # New hyperopt variables
--- a/pkg/suggestion/v1alpha3/hyperopt/service.py
+++ b/pkg/suggestion/v1alpha3/hyperopt/service.py
@@ -55,8 +55,7 @@ class OptimizerConfiguration:
         'tpe': {
             'gamma': (lambda x: float(x), lambda x: 1 > float(x) > 0),
             'prior_weight': (lambda x: float(x), lambda x: float(x) > 0),
-            'n_EI_candidates': (lambda x: int(x),
-                                lambda x: float(x).is_integer() and int(x) > 0),
+            'n_EI_candidates': (lambda x: int(x), lambda x: int(x) > 0),
             "random_state": (lambda x: int(x), lambda x: x.isdigit()),
         },
         "random": {
--- a/pkg/suggestion/v1alpha3/hyperopt/service.py
+++ b/pkg/suggestion/v1alpha3/hyperopt/service.py
@@ -51,42 +51,71 @@ class HyperoptService(api_pb2_grpc.SuggestionServicer, HealthServicer):
 
 
 class OptimizerConfiguration:
-    __schema_dict = {
+    __conversion_dict = {
         'tpe': {
-            'gamma': (lambda x: float(x), lambda x: 1 > float(x) > 0),
-            'prior_weight': (lambda x: float(x), lambda x: float(x) > 0),
-            'n_EI_candidates': (lambda x: int(x), lambda x: int(x) > 0),
-            "random_state": (lambda x: int(x), lambda x: int(x) >= 0),
+            'gamma': lambda x: float(x),
+            'prior_weight': lambda x: float(x),
+            'n_EI_candidates': lambda x: int(x),
+            "random_state": lambda x: int(x),
         },
         "random": {
-            "random_state": (lambda x: int(x), lambda x: int(x) >= 0)
+            "random_state": lambda x: int(x),
         }
     }
 
     @classmethod
     def convert_algorithm_spec(cls, algorithm_spec):
         ret = {}
-        setting_schema = cls.__schema_dict[algorithm_spec.algorithm_name]
+        setting_schema = cls.__conversion_dict[algorithm_spec.algorithm_name]
         for s in algorithm_spec.algorithm_setting:
             if s.name in setting_schema:
-                ret[s.name] = setting_schema[s.name][0](s.value)
+                ret[s.name] = setting_schema[s.name](s.value)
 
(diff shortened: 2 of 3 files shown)


pkg/suggestion/v1alpha3/hyperopt/base_hyperopt_service.py
```diff
@@ -12,8 +12,15 @@
...
 class BaseHyperoptService(object):
-    def __init__(self, algorithm_name=TPE_ALGORITHM_NAME, random_state=None, search_space=None):
+    def __init__(self,
+                 algorithm_name=TPE_ALGORITHM_NAME,
+                 algorithm_conf=None,
+                 search_space=None):
         self.algorithm_name = algorithm_name
+        self.algorithm_conf = algorithm_conf or {}
+        # pop common configurations
+        random_state = self.algorithm_conf.pop('random_state', None)
```
c-bata, line 22: [nits] I prefer to use `.get('random_state', None)` because `.pop()` has a side-effect. I think it is not expected behavior from the callers.

pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -40,17 +41,53 @@ def GetSuggestions(self, request, context):
...
-    def __init__(self, random_state=None):
-        self.random_state = random_state
+class OptimizerConfiguration:
+    __schema_dict = {
+        'tpe': {
+            'gamma': (lambda x: float(x), lambda x: 1 > float(x) > 0),
+            'prior_weight': (lambda x: float(x), lambda x: float(x) > 0),
+            'n_EI_candidates': (lambda x: int(x),
+                                lambda x: float(x).is_integer() and int(x) > 0),
+            "random_state": (lambda x: int(x), lambda x: x.isdigit()),
```
c-bata, line 60: What's the differences between `lambda x: float(x).is_integer() and int(x) > 0)` and `lambda x: x.isdigit()`?

pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -40,17 +41,53 @@ def GetSuggestions(self, request, context):
...
+        },
+        "random": {
+            "random_state": (lambda x: int(x), lambda x: x.isdigit())
+        }
+    }
```
c-bata, line 65: I mind that validate functions ( `__schema_dict[algo_name][s.name][1]` ) cannot return the message why parameters are invalid.

c-bata: And I personally feels this dict is a bit tricky. I prefer the following code:

```python
@classmethod
def _validate_tpe_setting(algorithm_settings):
    for s in algorithm_settings:
        if s == 'gamma' and 0 < float(x) < 1:
            return False, "gamma should be in the range of [0, 1]"
        ...
    return True, ""

@classmethod
def validate_algorithm_spec(cls, algorithm_spec):
    algo_name = algorithm_spec.algorithm_name
    if algo_name == "tpe":
        return cls._validate_tpe_setting(algorithm_spec.algorithm_setting)
    elif algo_name == "random":
        return cls._validate_random_setting(algorithm_spec.algorithm_setting)
    return False, "invalid algorithm name"
```

But I want to hear opinions of maintainers ( @andreyvelich @gaocegege ).

pkg/suggestion/v1alpha3/hyperopt/base_hyperopt_service.py
sperlingxx: Pop here means to split common ones out of `algorithm_conf`. Then, we can regard the whole `algorithm_conf` dict as **kwargs of suggest method.

c-bata: OK.  I don't have a strong opinion about this, so please feel free to resolve this conversation 👍

pkg/suggestion/v1alpha3/hyperopt_service.py
sperlingxx: I'm okay with both code styles.

pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -40,17 +41,53 @@ def GetSuggestions(self, request, context):
...
-class OptimizerConfiguration(object):
-    def __init__(self, random_state=None):
-        self.random_state = random_state
+class OptimizerConfiguration:
+    __schema_dict = {
+        'tpe': {
+            'gamma': (lambda x: float(x), lambda x: 1 > float(x) > 0),
+            'prior_weight': (lambda x: float(x), lambda x: float(x) > 0),
+            'n_EI_candidates': (lambda x: int(x),
+                                lambda x: float(x).is_integer() and int(x) > 0),
```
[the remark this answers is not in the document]
sperlingxx: I've removed it.

pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -40,17 +41,53 @@ def GetSuggestions(self, request, context):
...
+        for s in algorithm_spec.algorithm_setting:
+            if s.name in setting_schema:
+                ret[s.name] = setting_schema[s.name][0](s.value)
+
+        return algorithm_spec.algorithm_name, ret
+
+    @classmethod
+    def validate_algorithm_spec(cls, algorithm_spec):
+        algo_name = algorithm_spec.algorithm_name
+        if algo_name not in cls.__schema_dict:
```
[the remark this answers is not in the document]
sperlingxx: Yes, I've removed it.

pkg/suggestion/v1alpha3/hyperopt_service.py
sperlingxx: I've changed it.

pkg/suggestion/v1alpha3/hyperopt_service.py
andreyvelich: I agree with @c-bata, I think for validation we should say why settings is not correct. Like we did with enas: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1alpha3/nas/enas_service.py#L218

pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -40,17 +41,52 @@ def GetSuggestions(self, request, context):
...
-class OptimizerConfiguration(object):
-    def __init__(self, random_state=None):
-        self.random_state = random_state
+class OptimizerConfiguration:
+    __schema_dict = {
+        'tpe': {
+            'gamma': (lambda x: float(x), lambda x: 1 > float(x) > 0),
+            'prior_weight': (lambda x: float(x), lambda x: float(x) > 0),
+            'n_EI_candidates': (lambda x: int(x), lambda x: int(x) > 0),
+            "random_state": (lambda x: int(x), lambda x: x.isdigit()),
```
andreyvelich, line 59: Maybe instead of `x.isdigit()` we can check `int(x) >= 0`, like we did above?

sperlingxx: @andreyvelich Done.

pkg/suggestion/v1alpha3/hyperopt_service.py
sperlingxx: @andreyvelich With current implementaion, we also tell users why settings is wrong: https://github.com/kubeflow/katib/pull/1161/files#diff-a9d0a87b0b01a9816c0583cd6a0003e2R244

sperlingxx: @c-bata @andreyvelich I've refactored codes of validation part.

pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -40,17 +41,81 @@ def GetSuggestions(self, request, context):
...
+                        return False, "gamma should be in the range of (0, 1)"
+                elif s.name == 'prior_weight':
+                    if not (float(s.value) > 0):
+                        return False, "prior_weight should be great than zero"
+                elif s.name == 'n_EI_candidates':
+                    if not (int(s.value) > 0):
+                        return False, "n_EI_candidates should be great than zero"
+                elif s.name == 'random_state':
+                    if not (int(s.value) >= 0):
+                        return False, "random_state should be great or equal than zero"
```
c-bata, line 101: [nits] I guess you can remove redundant brackets. It's OK to ignore this comment.

```suggestion
                    if not 1 > float(s.value) > 0:
                        return False, "gamma should be in the range of (0, 1)"
                elif s.name == 'prior_weight':
                    if not float(s.value) > 0:
                        return False, "prior_weight should be great than zero"
                elif s.name == 'n_EI_candidates':
                    if not int(s.value) > 0:
                        return False, "n_EI_candidates should be great than zero"
                elif s.name == 'random_state':
                    if not int(s.value) >= 0:
                        return False, "random_state should be great or equal than zero"
```

sperlingxx: I only found this commit suggestion can be applied directly after I pushed the same commit = =!

c-bata: No problem 👍 Thanks.

pkg/suggestion/v1alpha3/hyperopt_service.py
```diff
@@ -40,17 +41,81 @@ def GetSuggestions(self, request, context):
...
+                elif s.name == 'n_EI_candidates':
+                    if not int(s.value) > 0:
+                        return False, "n_EI_candidates should be great than zero"
+                elif s.name == 'random_state':
+                    if not int(s.value) >= 0:
+                        return False, "random_state should be great or equal than zero"
+                else:
+                    return False, "unknown setting %s for algorithm tpe" % s.name
+            except Exception as e:
+                return False, "failed to validate %s(%s): %s" % (s.name, s.value, e)
```
andreyvelich, line 105: Maybe use `{}.format()` instead of `%s` to be consistent with other prints in Python Suggestions. 
What do you think @sperlingxx ?

sperlingxx: Good suggestion, and it's applied.

sperlingxx:
@gaocegege  There exists a conflict with latest merged commit. And I've solved it.

Issue #1144, opened 2020-04-13
Suggestion services folder structure


andreyvelich:
/kind feature

Currently, we store all suggestion `<algorithm_name>_service.py` files under `/pkg/suggestion/v1alpha3` directory.

Is there are reason to have structure like that? For example, for Golang Algorithms we have to store service in the same package as other files for the algorithm (See comment: https://github.com/kubeflow/katib/pull/1131#discussion_r406737331).

I think we can add `<algorithm_name>_service.py` files to appropriate algorithm folder. 

For example, `hyperopt_service.py` to https://github.com/kubeflow/katib/tree/master/pkg/suggestion/v1alpha3/hyperopt.
What do you think @gaocegege @johnugeorge @c-bata ?

/cc @richardsliu

c-bata:
I agree with your opinion 👍 

Furthermore I think we can move `/pkg/suggestion/v1alpha3/bayesianoptimization/src` to  `/pkg/suggestion/v1alpha3/bayesianoptimization`.

johnugeorge:
SGTM

PR #1166, opened 2020-04-24
Refactor suggestion services folder structure

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1144.

- I moved all `<algorithm_name>_service.py` files to appropriate suggestion folder.

- I added `__init__.py` file to each suggestion folder, since we import them in `cmd/`.

- I made absolute path in all python files. For example: https://github.com/kubeflow/katib/compare/master...andreyvelich:issue-1144-suggestion-folder-structure?expand=1#diff-fe0ee80956c02a4ffe8ecf4881b2e40cR6. To be consistent between imports.

- I moved `base_health_service` to `internal` folder.

- I deleted `pkg/suggestion/test_requirements.txt ` file. @gaocegege Do we need this file? I didn't find any usage of this file.


andreyvelich:
> Thank you for your work!
> 
> It might be better to rename files like:
> 
> * `suggestion/chocolate/chocolate_service.py` -> `suggestion/chocolate/service.py`
> * `suggestion/hyperband/hyperbad_service.py` -> `suggestion/hyperbad/service.py`
> * `suggestion/skopt/skopt_service.py` -> `suggestion/skopt/service.py`
> * ...
> 
> But I have no strong opinion about this. And we can work on it in another pull request.
> 
> /lgtm

Good point. I will rename it. 
What do you think about Base Services: https://github.com/andreyvelich/katib/blob/issue-1144-suggestion-folder-structure/pkg/suggestion/v1alpha3/chocolate/base_chocolate_service.py ?
Maybe name it `base_service.py` ?

c-bata:
> Maybe name it base_service.py ?

Yeah, agree.

PR #1184, opened 2020-05-11
Fix Never Resume Policy for Experiment

andreyvelich:
I fixed some problems in `Never` resume policy.
1. `IsCompletedExperimentRestartable` should return `true` only if `ResumePolicy: LongRunning`.
2. Call `terminateSuggestion` only for `ResumePolicy: Never`.
3. Reconcile can call `terminateSuggestion` twice. I added check if Suggestion is already succeeded.
4. I believe we should call `g.Status().Update()` instead of `g.Update()` to update status of suggestion CR.
5. `terminateSuggestion` just returns error, I don't think we need to requeue Reconcile: `reconcile.Result{Requeue: true}`.
5. Fix some problems with Logging.
6. Add e2e test for Never Resume Experiment.

/cc @sperlingxx


pkg/controller.v1alpha3/experiment/experiment_controller.go
```diff
@@ -193,17 +193,24 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 	if instance.IsCompleted() {
 		// Check if completed instance is restartable
 		// Experiment is restartable only if it is in succeeded state by reaching max trials
+		// And Resume Policy is LongRunning
 		if util.IsCompletedExperimentRestartable(instance) {
```
sperlingxx, line 197: I think this paragraph could be modified to something below 
```go
	if instance.IsCompleted() {
		needRestart := false
		// Check if completed instance is restartable
		// Experiment is restartable only if it is in succeeded state by reaching max trials
		if util.IsCompletedExperimentRestartable(instance) {
			// Check if max trials is reconfigured
			if (instance.Spec.MaxTrialCount != nil &&
				*instance.Spec.MaxTrialCount != instance.Status.Trials) ||
				(instance.Spec.MaxTrialCount == nil && instance.Status.Trials != 0) {
				msg := "Experiment is restarted"
				instance.MarkExperimentStatusRestarting(util.ExperimentRestartingReason, msg)
				needRestart = true
			}
		}
		// If experiment who doesn't need to restart is completed without running trials, stop reconcile
		if !needRestart && !instance.HasRunningTrials() {
			if instance.Spec.ResumePolicy != experimentsv1alpha3.LongRunning {
				return r.terminateSuggestion(instance)
			}
			return reconcile.Result{}, nil
		}
	}
```
Otherwise, `terminateSuggestion` won't work when corresponding experiment has been completed for reaching max trials.

gaocegege:
@sperlingxx Everyone creates bugs. Do not need to worry about it.

pkg/controller.v1alpha3/experiment/experiment_controller.go
andreyvelich: @sperlingxx Why `terminateSuggestion` won't run when experiment finishes with reaching max trials?

I believe, Reconcile loop runs with Succeeded experiment state, after reaching this step: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1alpha3/experiment/util/status_util.go#L182.

I am wondering, do we actually need to check if Experiment has running trials in the controller @gaocegege @johnugeorge :
https://github.com/kubeflow/katib/blob/master/pkg/controller.v1alpha3/experiment/experiment_controller.go#L209 ?
Or we can just `return reconcile.Result{}, nil` without this check if Experiment is Completed?

sperlingxx: @andreyvelich I think `util.IsCompletedExperimentRestartable` will be true when experiment finishes with reaching max trials. But, for now, `terminateSuggestion` will only be called  when `util.IsCompletedExperimentRestartable` is false.

andreyvelich: Yeah, but `util.IsCompletedExperimentRestartable` return True only if `ResumePolicy: LongRunning`.
 
https://github.com/kubeflow/katib/pull/1184/files#diff-f0faa0b63a35acff95fe5e9d93d594ffR201

sperlingxx: @andreyvelich Oh, I got it!

andreyvelich:
I think it was a problem with new version of Scikit Learn 0.23.
I created PR to change version to 0.22 in this PR: https://github.com/kubeflow/katib/pull/1187.
I reverted this commit from that PR.

Issue #1188, opened 2020-05-14
additionalMetricNames does not seem to work for TensorFlowEvent


sadeel:
/kind bug

**What steps did you take and what happened:**
I have an experiment, which looks like this:
```yaml
apiVersion: "kubeflow.org/v1alpha3"
kind: Experiment
metadata:
  namespace: kubeflow
  labels:
    controller-tools.k8s.io: "1.0"
  name: hyperparameter-tuning
spec:
  objective:
    type: minimize
    goal: 0.00
    objectiveMetricName: epoch_loss
    additionalMetricNames:
    - epoch_binary_accuracy
  algorithm:
    algorithmName: random
  metricsCollectorSpec:
    source:
      fileSystemPath:
        path: /tmp/tensorboard_logs/validation
        kind: Directory
    collector:
      kind: TensorFlowEvent
```

Individual trials do not seem to report the objective metric or other the additional metrics. However, if I remove the additionalMetricNames altogether, it seems to work fine.
![none](https://user-images.githubusercontent.com/5696932/81990384-10dece80-960d-11ea-8dae-545ce508efb4.png)

**What did you expect to happen:**
All metrics to be populated.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version:
- Minikube version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

gaocegege:
Thanks for your report. We will have a look.

andreyvelich:
@sadeel Thank you for reporting this!

Yes, I think we have a bug there.
We add additional metrics to metrics collector container using ";" (https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1alpha3/pod/inject_webhook.go#L177).
In TF Event Metrics Collector we use "," to split these metrics: https://github.com/kubeflow/katib/blob/master/cmd/metricscollector/v1alpha3/tfevent-metricscollector/main.py#L42.

I will create PR to fix it.

PR #1191, opened 2020-05-18
Fix additional metrics in TF Event metrics collector

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1188.
I changed split metrics from "," to ";".
See comment: https://github.com/kubeflow/katib/issues/1188#issuecomment-630320234.

Do we need to modify TF Job example (https://github.com/kubeflow/katib/blob/master/examples/v1alpha3/tfjob-example.yaml) with additional metric, for example: `cross_entropy_1`?

/cc @sadeel

### DIFF
--- a/cmd/metricscollector/v1alpha3/tfevent-metricscollector/main.py
+++ b/cmd/metricscollector/v1alpha3/tfevent-metricscollector/main.py
@@ -39,7 +39,7 @@ if __name__ == '__main__':
 
     WaitOtherMainProcesses(completed_marked_dir=opt.dir_path)
 
-    mc = MetricsCollector(opt.metric_names.split(','))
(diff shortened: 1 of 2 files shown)


cmd/metricscollector/v1alpha3/tfevent-metricscollector/main.py
```diff
@@ -39,7 +39,7 @@ def parse_options():
 
     WaitOtherMainProcesses(completed_marked_dir=opt.dir_path)
 
-    mc = MetricsCollector(opt.metric_names.split(','))
+    mc = MetricsCollector(opt.metric_names.split(';'))
```
gaocegege, line 42: /lgtm
/approve

PR #1194, opened 2020-05-21
add clientset/lister/informer generation

sperlingxx:
This PR is to refine kubernetes clients generation for katib. The major contribution is generating codes of clientset/lister/informer via `k8s.io/code-generator/generate-groups.sh`.

In order to run `generate-groups.sh` properly,   project `k8s.io/code-generator` and `k8s.io/gengo` need to be unpruned. 

Could you help to review this PR?  @andreyvelich @gaocegege @johnugeorge


gaocegege:
Why do we need such a generation? Will you use it to build some upper-level projects?

sperlingxx:
> Why do we need such a generation? Will you use it to build some upper-level projects?

Yes, I want to preserve/track katib custom resources (experiments/suggestions/trials) with internal agent which requires listers and informers of CRDs. I think, as our internal agent, typed clients/listers/informers are usually essential, in order to interact with other kubernetes-native systems.

gaocegege:
/lgtm 

We are using k8s-sigs/controller-runtime/pkg/client to do such things. But I think this feature is necessary.

sperlingxx:
@gaocegege Yes, many early-built controllers (such as tf-operator) are based on client-go instead of kube-builder. Thanks for review!

hack/update-codegen.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+PROJECT_ROOT=${GOPATH}/src/github.com/kubeflow/katib
+
+${PROJECT_ROOT}/vendor/k8s.io/code-generator/generate-groups.sh \
+    all \
+    github.com/kubeflow/katib/pkg/client/controller \
+    github.com/kubeflow/katib/pkg/apis/controller \
+    "common:v1alpha3 experiments:v1alpha3 suggestions:v1alpha3 trials:v1alpha3" \
+    --go-header-file ${PROJECT_ROOT}/hack/boilerplate.go.txt
```
andreyvelich, line 32: Do we need `boilerplate.go.txt` file in /hack folder for header?

sperlingxx: @andreyvelich  Maybe we can change it to kubeflow specialized header?

andreyvelich: Should it be specific to Kubeflow licence, like here: https://github.com/kubeflow/katib/blob/master/hack/verify-gofmt.sh#L2-L14 ?

johnugeorge:
Instead of all files, why don't we just keep the generator files in the upstream? If someone needs it, it can run the generator files to generate all. Since Kubebuilder won't use this at all,  I feel that we should avoid it in the upstream

gaocegege:
> Instead of all files, why don't we just keep the generator files in the upstream? If someone needs it, it can run the generator files to generate all. Since Kubebuilder won't use this at all, I feel that we should avoid it in the upstream

I think it does not work since we cannot use vendor to install these packages. Then the users have to maintain a fork of katib. I think we should keep all the client and informer in the upstream, or just do not support it.

hack/update-openapigen.sh
```diff
@@ -19,40 +19,24 @@ set -o nounset
...
-KATIB_VERSION_V1ALPHA3="v1alpha3"
-SWAGGER_CODEGEN_FILE_V1ALPHA3="pkg/apis/v1alpha3/swagger.json"
-
-KATIB_VERSION_V1BETA1="v1beta1"
-SWAGGER_CODEGEN_FILE_V1BETA1="pkg/apis/v1beta1/swagger.json"
-
+PROJECT_ROOT=${GOPATH}/src/github.com/kubeflow/katib
+CODEGEN_PKG=${PROJECT_ROOT}/vendor/k8s.io/code-generator
+VERSION_LIST=(v1alpha3 v1beta1)
 SWAGGER_VERSION="0.1"
```
andreyvelich, line 28: For each Katib version Swagger Version will be v0.1, is that fine?

gaocegege: Is there any suggestion?

sperlingxx: @andreyvelich I've refined the schema of swagger version, which became `${katib_version}-{swagger_sub_version}`. So, with swagger version, we can figure out katib version who generated it.

andreyvelich: Thanks @sperlingxx, SGTM.

gaocegege:
@andreyvelich @johnugeorge PTAL

/lgtm

Issue #1081, opened 2020-03-10
Support argo Workflow CRD as new trial kind


terrykong:
/kind feature

**Describe the solution you'd like**
For more complicated jobs where we may execute things in separate containers I have noticed that the only job types we support in a trial are `Job`s, `TFJob`s and `PyTorchJob`s. Would it be possible to also support argo workflows?

**Anything else you would like to add:**

I see that https://github.com/kubeflow/katib/blob/master/docs/new-trial-kind.md is outdated, but it sounds like for a vanilla workflow, it should be possible to inject a metrics collector sidecar to each workflow container right?

gaocegege:
@terrykong It is possible. It is extensible. We have a doc for it https://github.com/kubeflow/katib/blob/master/docs/new-trial-kind.md although it may be outdated.

The problem is that whether it is an actual use case.

terrykong:
@gaocegege glad to hear it's possible. I don't know about others, but there are certainly convenient use cases covered by workflows. In particular optimizing a training -> benchmarking problem where the benchmarking doesn't need GPUs so the GPUs used by training pod can be freed up while benchmarking. At least to my knowledge this is definitely not possible with Jobs, and I'm guessing not with TFJobs and Pytorchjobs either.

nielsmeima:
@gaocegege for us this is a use case as well. We would like to be able to tune parameters of arbitrary chains of Docker containers, e.g. executed as an Argo workflow. In such a workflow we could easily mix various languages to achieve our needs, instead of relying on a single container or a specific language. I will try to come up with an implementation, and try to commit back / propose a design if it works out.

gaocegege:
@nielsmeima Thanks for your interest! Now we have a new abstraction `Provider` in pkg/job/v1beta1/provider.go. You just need to implement the provider for argo, then it should work.

nielsmeima:
@gaocegege Thanks, I will give it a go.

andreyvelich:
@nielsmeima We are also working on the new Trial Template implementation for the new version: https://github.com/kubeflow/katib/issues/906#issuecomment-632149834.
After that, you should be able to specify any type of resource as Trial Template.

nielsmeima:
[...]

PR #1202, opened 2020-06-02
New Trial Template API

andreyvelich:
This is new API for Trial Template.
See comment: https://github.com/kubeflow/katib/issues/906#issuecomment-632149834.

In `TrialParameters` user must define parameters that are replaced in Trial Template.
As @gaocegege proposed, I added `Reference` to make connection between `TrialParameters` and `Parameters` (search space).
Also, in `TrialSpec` `metadata.name` and `metadata.namespace` must be omitted, since we automatically added it in Katib controller.

I added ConfigMap specification to `TrialTemplate`, if user wants to read Trial Template from configMap.

In the future, we can extend `TrialTemplate` with custom settings (for example, how to get succeeded state of the Job), to support any Kubernetes CRDs.
In this CRDs we should be able to run TrialJobs, such as Argo, etc (https://github.com/kubeflow/katib/issues/1081).

/assign @gaocegege @jlewi @johnugeorge @sperlingxx
/cc @czheng94 @terrykong @nielsmeima

### DIFF
--- a/pkg/apis/controller/experiments/v1beta1/experiment_types.go
+++ b/pkg/apis/controller/experiments/v1beta1/experiment_types.go
@@ -19,6 +19,7 @@ import (
 	common "github.com/kubeflow/katib/pkg/apis/controller/common/v1beta1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 type ExperimentSpec struct {
@@ -187,8 +188,36 @@ type FeasibleSpace struct {
 }
 
 type TrialTemplate struct {
-	Retain     bool        `json:"retain,omitempty"`
+	// Retain indicates that Trial resources must be not cleanup
+	Retain bool `json:"retain,omitempty"`
+
 	GoTemplate *GoTemplate `json:"goTemplate,omitempty"`
+
+	// List of parameres that are used in Trial template
+	TrialParameters []TrialParameterSpec `json:"trialParameters,omitempty"`
+
+	// Trial spec contains Trial template in unstructured format
+	TrialSpec *unstructured.Unstructured `json:"trialSpec,omitempty"`
+
+	// Name of config map where Trial template is located
+	ConfigMapName string `json:"configMapName,omitempty"`
+
+	// Namespace of config map where Trial template is located
+	ConfigMapNamespace string `json:"configMapNamespace,omitempty"`
+
+	// Path in config map where Trial template is located
+	TemplatePath string `json:"templatePath,omitempty"`
+}
+
+type TrialParameterSpec struct {
+	// Name of the parameter that must be replaced in Trial template
+	Name string `json:"name,omitempty"`
+
+	// Description of the parameter
+	Description string `json:"description,omitempty"`
+
+	// Reference to the parameter in search space
+	Reference string `json:"reference,omitempty"`
 }
 
 type TemplateSpec struct {


pkg/apis/controller/experiments/v1beta1/experiment_types.go
```diff
@@ -187,8 +188,36 @@ type FeasibleSpace struct {
...
 	GoTemplate *GoTemplate `json:"goTemplate,omitempty"`
+
+	// List of parameres that are used in Trial template
+	TrialParameters []TrialParameterSpec `json:"trialParameters,omitempty"`
+
+	// Trial spec contains Trial template in unstructured format
+	TrialSpec *unstructured.Unstructured `json:"trialSpec,omitempty"`
+
+	// Name of config map where Trial template is located
+	ConfigMapName string `json:"configMapName,omitempty"`
```
gaocegege, line 203: The field is already in GoTemplate, should we remove it here?

pkg/apis/controller/experiments/v1beta1/experiment_types.go
```diff
@@ -187,8 +188,36 @@ type FeasibleSpace struct {
 }
 
 type TrialTemplate struct {
-	Retain     bool        `json:"retain,omitempty"`
+	// Retain indicates that Trial resources must be not cleanup
+	Retain bool `json:"retain,omitempty"`
+
 	GoTemplate *GoTemplate `json:"goTemplate,omitempty"`
```
gaocegege, line 194: Can we move it into TrialSpec?

sperlingxx:
Genrally, it looks great to me. Will we deprecate `rawTemplate` or keep supporting it?

pkg/apis/controller/experiments/v1beta1/experiment_types.go
johnugeorge, line 194: Are we supporting Gotemplate also?

andreyvelich: All current examples in v1beta1 are working with `GoTemplate`.
You think we should remove it and don't run e2e test until we implement new logic?

pkg/apis/controller/experiments/v1beta1/experiment_types.go
andreyvelich: Yeah, but I think when we implement new logic we will not support `GoTemplate` ?

pkg/apis/controller/experiments/v1beta1/experiment_types.go
andreyvelich: In that case what do you think about the name where `Unstructured` template will be located?
```
type TrialSpec struct {
<must be some name>     *unstructured.Unstructured `json:"<name>,omitempty"`

ConfigMapName string `json:"configMapName,omitempty"`
ConfigMapNamespace string `json:"configMapNamespace,omitempty"`
TemplatePath string `json:"templatePath,omitempty"`
}

**Comment by andreyvelich** [comment]
> Generally, it looks great to me. Will we deprecate `rawTemplate` or keep supporting it?

Thanks. I think we can deprecate it, when we implement new logic. Can't find any case when `rawTemplate` can give us additional functionality.

**Comment by jlewi** [comment]
Great to see this coming along. I'll let the Katib experts review this.

File: pkg/apis/controller/experiments/v1beta1/experiment_types.go (hunk 1 above)
Comment (gaocegege) [reply pkg/apis/controller/experiments/v1beta1/experiment_types.go:203]: SGTM, then can we just remove gotemplate?

File: pkg/apis/controller/experiments/v1beta1/experiment_types.go (hunk 2 above)
Comment (gaocegege) [reply pkg/apis/controller/experiments/v1beta1/experiment_types.go:194]: How about this:

```go
type TrialTemplate struct {
	Retain     bool        `json:"retain,omitempty"`

	TrialSource `json:",inline"`

	// List of parameres that are used in Trial template
	TrialParameters []TrialParameterSpec `json:"trialParameters,omitempty"`
}

type TrialSource struct {
	ConfigMap *ConfigMapSource `json:"configMap"`
	TrialSpec *unstructured.Unstructured `json:"trialSPec"`
}

type ConfigMapSource struct {
	// Name of config map where Trial template is located
	ConfigMapName string `json:"configMapName,omitempty"`

	// Namespace of config map where Trial template is located
	ConfigMapNamespace string `json:"configMapNamespace,omitempty"`
	
	// Path in config map where Trial template is located
	TemplatePath string `json:"templatePath,omitempty"`
}
```

gaocegege: I think we should remove and disable test if we decide to remove it.

johnugeorge: I agree.  Lets make all changes to trial template API in a single PR for better tracking . Once new TrialTemplate implementation is added, move all examples to new trial template.

andreyvelich: Ok, I will remove `GoTemplate` in this PR and disable all e2e tests.

andreyvelich: @gaocegege SGTM, it is nice to follow k8s APIs structure way.
What do you think @johnugeorge ?

johnugeorge: Looks great

andreyvelich:
If we delete `GoTemplate` API in this PR, v1alpha3 controller image also can't be built.

We need to fix problems in API package, for example: https://github.com/andreyvelich/katib/blob/new-trial-template-api/pkg/apis/controller/experiments/v1beta1/experiment_defaults.go#L57.

So maybe we can delete `GoTemplate` in the next PR, where we implement new logic for the controller.
In that case, we will not stop Katib development on other issues.

gaocegege:
> If we delete GoTemplate API in this PR, v1alpha3 controller image also can't be built.

I am not sure why. I think v1alpha3 is decoupled with v1beta1.

andreyvelich:
> > If we delete GoTemplate API in this PR, v1alpha3 controller image also can't be built.
> 
> I am not sure why. I think v1alpha3 is decoupled with v1beta1.

I think the problem is with this: https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/addtoscheme_katib_v1beta1.go#L26-L29.
Since v1beta1 and v1alpha3 in the same package `apis`, while building image for v1alpha3, it checks that we don't have errors in `v1beta1` package: https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/experiments/v1beta1/register.go#L23.

So the solution can be comment the lines for v1beta1, from `init()` function:
https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/addtoscheme_katib_v1beta1.go#L26-L29

Then, I can build image for v1alpha3.

andreyvelich:
I commited changes for the controller and examples in this PR to avoid problems in testing.
List of changes:
1. Change all examples in v1beta1 to support new Trial Template
2. Submit new API changes for Experiment API and Trial API (`RunSpec` now also unstructured).
3. Modify deepcopy and experiment_defaults to support new API
4. Delete old trial template configMap. I left only ConfigMap with trial template which is labeled: `app: katib-trial-templates`. So users can correct see them in the UI.
5. In Experiment Controller I think we should `createTrialInstance` only when Suggestion returns some trials. It can avoid unexpected errors with already created trials. https://github.com/kubeflow/katib/blob/cebec02ce21976e455eaf52a42266490578a6322/pkg/controller.v1beta1/experiment/experiment_controller.go#L353-L360
6. Modify all tests for Experiment and Trial controller. 
7. Add new logic in `generator.go` to support new Trial Template.
8. Add new test for `generator_test.go`.
9. Comment Validator for Trial Template. Need to modify it in the future PRs.

I think it is ready for review.
After this PR, we should pass the test for the new Trial Template. We still need to add validation and improve `generator_test.go` with more tests, which can be tracked here: https://github.com/kubeflow/katib/issues/1208.

/cc @gaocegege @johnugeorge

pkg/controller.v1beta1/experiment/experiment_controller.go
```diff
@@ -341,15 +340,24 @@ func (r *ReconcileExperiment) createTrials(instance *experimentsv1beta1.Experime
...
 	if err != nil {
 		logger.Error(err, "Get suggestions error")
 		return err
 	}
+	var trialNames []string
 	for _, trial := range trials {
-		if err = r.createTrialInstance(instance, &trial); err != nil {
-			logger.Error(err, "Create trial instance error", "trial", trial)
-			continue
+		trialNames = append(trialNames, trial.Name)
```
sperlingxx, line 351: I think this paragraph can be simplified to 
```go
for _, trial := range trials {
    if err = r.createTrialInstance(instance, &trial); err != nil {
        logger.Error(err, "Create trial instance error", "trial", trial)
        continue
    }
}			
```

andreyvelich: In that case we try to create trials even if  `trials==[ ] `. Is that correct?

sperlingxx: @andreyvelich If `trials == []`,  the foreach loop will be skipped?

andreyvelich: @sperlingxx Thanks, I got your point. Will make change.

Issue #1122, opened 2020-04-03
Chocolate service db exhausted


StefanoFioravanzo:
/kind bug

**What steps did you take and what happened:**
When creating experiments that use the `grid` search algorithm __sometimes__ trials stop being generated, even though the experiment is still seen as running.
By printing the logs of the suggestions pods I can see the following:

```
ERROR:sqlalchemy.pool.impl.NullPool:Exception during reset or similar
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/pool/base.py", line 680, in _finalize_fairy
    fairy._reset(pool)
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/pool/base.py", line 867, in _reset
    pool._dialect.do_rollback(self)
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 502, in do_rollback
    dbapi_connection.rollback()
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140636731225856 and this is thread id 140636053288704.
ERROR:sqlalchemy.pool.impl.NullPool:Exception closing connection <sqlite3.Connection object at 0x7fe8b4397650>
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/pool/base.py", line 680, in _finalize_fairy
    fairy._reset(pool)
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/pool/base.py", line 867, in _reset
    pool._dialect.do_rollback(self)
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 502, in do_rollback
    dbapi_connection.rollback()
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140636731225856 and this is thread id 140636053288704.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/pool/base.py", line 270, in _close_connection
    self._dialect.do_close(connection)
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 508, in do_close
    dbapi_connection.close()
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 140636731225856 and this is thread id 140636053288704.
[...]

PR #1205, opened 2020-06-03
Validate possible operations for Grid suggestion

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1122.

I added validation for Chocolate Grid suggestion to check if `max_trial_count` >= all possible search space combination.
Also, I moved `call_validate()` test function to utils to use it in Chocolate also.

/cc @sperlingxx

### DIFF
--- a/pkg/suggestion/v1beta1/chocolate/service.py
+++ b/pkg/suggestion/v1beta1/chocolate/service.py
@@ -18,11 +18,14 @@ import grpc
 from pkg.apis.manager.v1beta1.python import api_pb2
 from pkg.apis.manager.v1beta1.python import api_pb2_grpc
 
-from pkg.suggestion.v1beta1.internal.constant import DOUBLE
+from pkg.suggestion.v1beta1.internal.constant import (INTEGER, DOUBLE, CATEGORICAL, DISCRETE)
 from pkg.suggestion.v1beta1.internal.search_space import HyperParameterSearchSpace
 from pkg.suggestion.v1beta1.internal.trial import Trial, Assignment
 from pkg.suggestion.v1beta1.chocolate.base_service import BaseChocolateService
 from pkg.suggestion.v1beta1.internal.base_health_service import HealthServicer
+
+import numpy as np
+import itertools
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


elikatsis:
Hello @andreyvelich,

Me and @StefanoFioravanzo have been tinkering with Katib for some time now and have a different proposal on tackling #1122. We will describe it in detail as a comment to that issue by the end of the week.

Would you mind holding this PR until then and decide afterwards whether to merge this or not?

Thanks in advance!

andreyvelich:
Hi @elikatsis, sure.
Is this different from regular Validation Algorithm Settings?

pkg/suggestion/v1beta1/chocolate/service.py
```diff
@@ -4,12 +4,15 @@
 from pkg.apis.manager.v1beta1.python import api_pb2
 from pkg.apis.manager.v1beta1.python import api_pb2_grpc
 
-from pkg.suggestion.v1beta1.internal.constant import DOUBLE
+from pkg.suggestion.v1beta1.internal.constant import (INTEGER, DOUBLE, CATEGORICAL, DISCRETE)
```
gaocegege, line 7: I am not good at py, why do we need () for it?

gaocegege:
> By the way, I think we had better to support other common scales (like log-uniform scale) in feasible space.

I think so, would you mind opening an issue for it? @sperlingxx

pkg/suggestion/v1beta1/chocolate/service.py
andreyvelich: I think we can remove it from here, it uses when we need to use more than 1 line for importing

andreyvelich:
> /lgtm
> 
> By the way, I think we had better to support other common scales (like log-uniform scale) in feasible space.

In skopt we use log-uniform distribution: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/skopt/base_service.py#L44.
To define it, max and min DOUBLE value is not enough?

sperlingxx:
> > /lgtm
> > By the way, I think we had better to support other common scales (like log-uniform scale) in feasible space.
> 
> In skopt we use log-uniform distribution: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/skopt/base_service.py#L44.
> To define it, max and min DOUBLE value is not enough?

Other packages used by katib also support multiple distributions:
* hyperopt supports `loguniform`, `lognormal`, `normal`, `qnormal` ...
* optuna supports `loguniform` and `uniform`

So, maybe is it better to add an option to configure distribution of continuous search space in `FeasibleSpace` ? And we can validate whether the tuning algorithm supporting specific distribution in `ValidateAlgorithmSettings` method of suggestion service.

andreyvelich:
I believe @elikatsis wanted to hold this PR until end of this week.
We will merge it after.
/cc @gaocegege @sperlingxx

gaocegege:
Oh SGTM. Thanks for the hold in time.

andreyvelich:
I think we can merge these changes for Grid suggestion.
WDYT @anencore94 @johnugeorge @gaocegege ?

gaocegege:
SGTM!

anencore94 (commented):
`call_validate` call_validate to util class looks better! Thanks :) @andreyvelich

gaocegege:
LGTM

Issue #1208, opened 2020-06-05
Support new Trial Template in v1beta1


andreyvelich:
/kind feature

To support new Trial Template API in v1beta1 we need to make few changes:
- [x] Submit new API changes (https://github.com/kubeflow/katib/pull/1202).
- [x] Change Controller logic to support new Template. Also, modify Trial CR API (https://github.com/kubeflow/katib/pull/1202).
- [x] Change examples according to the new API (https://github.com/kubeflow/katib/pull/1202).
- [x] Enable testing for v1beta1 (https://github.com/kubeflow/katib/pull/1202).
- [x] Modify validation webhook. Make sure that `metadata.name` and `metadata.namespace` in `TrialSpec` is omitted (https://github.com/kubeflow/katib/pull/1215).
- [x] Add more test-cases for `generator.go` (https://github.com/kubeflow/katib/pull/1216).

If I missed something, please let me know.
/cc @gaocegege @johnugeorge

andreyvelich:
I will start work on the second step.

@gaocegege @johnugeorge What do you think about [`RunSpec`](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/trials/v1beta1/trial_types.go#L35) in Trial CR API, should we make it `unstructured` and store current Trial specification?

gaocegege:
I think we should use unstructured.

andreyvelich:
We can close this issue, since we fully implemented this.

PR #1215, opened 2020-06-12
New Trial Template validation

andreyvelich:
Part of: https://github.com/kubeflow/katib/issues/1208.

I added Validation for the New Trial Template. 
I believe I covered all test cases.
The most tricky part is how to validate that user specifies fully correct Job, TFJob or PyTorch Job (That all APIs are correct).

We can't just convert `unstructured` to according Kubernetes structure (e.g `batchv1.Job{}`) and convert it back to `unstructured`. And then use `DeepEqual` to compare with original.

After transforming Trial template to Kubernetes Job, some parameters will be added by default. For example, `metadata.creationTimestamp` will be `nil`. So JSON structure of original and transformed job is different.

My idea is to use JSON patch. After converting `unstructured` to Kubernetes structure we can try to use JSON patch on this structure with original `unstructured`. If all parameters from `unstructured` were transformed to Kubernetes Job, JSON patch will have only `remove` operations.

If not, we print to user which fields can't be converted.

I think that can work.

### DIFF
--- a/pkg/webhook/v1beta1/experiment/validator/validator.go
+++ b/pkg/webhook/v1beta1/experiment/validator/validator.go
@@ -1,6 +1,7 @@
 package validator
 
 import (
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"regexp"
@@ -20,10 +21,11 @@ import (
 	util "github.com/kubeflow/katib/pkg/controller.v1beta1/util"
 	jobv1beta1 "github.com/kubeflow/katib/pkg/job/v1beta1"
 	mccommon "github.com/kubeflow/katib/pkg/metricscollector/v1beta1/common"
-	batchv1 "k8s.io/api/batch/v1"
 
 	pytorchv1 "github.com/kubeflow/pytorch-operator/pkg/apis/pytorch/v1"
 	tfv1 "github.com/kubeflow/tf-operator/pkg/apis/tensorflow/v1"
+	jsonPatch "github.com/mattbaird/jsonpatch"
+	batchv1 "k8s.io/api/batch/v1"
 )
 
 var log = logf.Log.WithName("experiment-validating-webhook")
@@ -217,7 +219,7 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 
 	// Check if ApiVersion and Kind is specified
 	if runSpec.GetAPIVersion() == "" || runSpec.GetKind() == "" {
-		return fmt.Errorf("apiVersion and kind in spec.trialTemplate must be specified")
+		return fmt.Errorf("APIVersion and Kind in spec.trialTemplate must be specified")
 	}
 
 	// Check if Job is supported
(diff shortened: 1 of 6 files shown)


pkg/webhook/v1beta1/experiment/validator/validator.go
```diff
@@ -9,14 +10,22 @@ import (
...
 
 	commonapiv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/common/v1beta1"
 	experimentsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/experiments/v1beta1"
+	"github.com/kubeflow/katib/pkg/controller.v1beta1/consts"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/experiment/manifest"
+	util "github.com/kubeflow/katib/pkg/controller.v1beta1/util"
 	jobv1beta1 "github.com/kubeflow/katib/pkg/job/v1beta1"
 	mccommon "github.com/kubeflow/katib/pkg/metricscollector/v1beta1/common"
+
+	pytorchv1 "github.com/kubeflow/pytorch-operator/pkg/apis/pytorch/v1"
```
gaocegege, line 25: Can we merge this group to the second?

andreyvelich: Which group do you mean?

pkg/webhook/v1beta1/experiment/validator/validator.go
```diff
@@ -9,14 +10,22 @@ import (
...
+	tfv1 "github.com/kubeflow/tf-operator/pkg/apis/tensorflow/v1"
```
johnugeorge, line 26: Were we doing this validation before? Can we avoid operator imports somehow?

andreyvelich: Before that we didn't validate that Job was added properly: https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1alpha3/experiment/validator/validator.go#L161-L170. We just checked Kind.

Why do we want to avoid these imports?
That can help user to submit incorrect TFJobs, Jobs or PyTorch Jobs.

pkg/webhook/v1beta1/experiment/validator/validator.go
gaocegege: https://github.com/kubeflow/katib/pull/1215/files/26caeb72fc6c19673665b3d7725c8a922e44d734#diff-2a5dc923db6cdc05ea4f25ba0a8e8c73R15

pkg/webhook/v1beta1/experiment/validator/validator.go
johnugeorge: The general idea is to decouple katib from operators(if we can)

andreyvelich: In the future yes.
After we implement general provider for all K8s CRDs we can remove this validation.
I think for now we can keep this validation to make sure that current Trial jobs are correct

pkg/webhook/v1beta1/experiment/validator/validator.go
andreyvelich: Done

PR #1232, opened 2020-06-23
UI: Optimisation and enhancements for v1beta1 Katib UI 

andreyvelich:
I made few improvements to increase UI productivity and make it easier to understand source code:

1. Switch to `node v12` from `node v10` to make it more stable.
2. Switch to `react-scripts` from using `node` to build and start frontend.
3. Remove `webpack`, `jest`, `babel` configs/settings and remove `build`, `start`, `test` scripts, because `react-scripts` can handle all of these. If we need it, we can add them in the future.
4. Remove unused packages from `package.json` that helped to reduce `node_modules` size by 100Mb.
5. Modify source code to remove warnings.
6. Add `db-manager-address` flag that can help to properly deploy Katib UI locally.
7. Remove not necessary scripts imports from `index.html`.
8. Remove `serviceWorker`
9. Modify README.


andreyvelich:
@gaocegege @johnugeorge It seems I should commit `package-lock.json` file. Ref: https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5?answertab=votes#tab-top.
It can help to avoid problems with dependencies and user `npm ci` instead of `npm install`.

Issue #1227, opened 2020-06-20
Unable to parse NaN value in Trial CR


andreyvelich:
/kind bug

I tried to run Experiment with additional metric names that were not collected from the Trial pod.

Because of that, in Trial Controller we try to write [`math.NaN()`](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_util.go#L175-L176) value to observation and Trial can't be updated.

I got error [here](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller.go#L192-L199) while comparing Trial statuses.
I am not sure that Kubernetes objects supports `math.NaN()`, since it must be valid JSON.

We should think how we can handle empty metric results.

This issue can be related to https://github.com/kubeflow/katib/issues/889.

/cc @sperlingxx @gaocegege @johnugeorge

sperlingxx:
In addition, I think we should fail the trial if job is succeed without any ObservationLog, which are usually caused by incorrect metrics collecting configuration. Currently, trial status will be locked at pending under this situation.

andreyvelich:
@sperlingxx As you can see here: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_util.go#L126-L139, we convert Trial to succeeded state only if any observation was reported.

My suggestion for this issue is to convert all API metrics parameters to "string". 
1. As you can see here: https://github.com/kubernetes-sigs/controller-tools/issues/245 it is recommended by Kubernetes.
2. We can add [initial state](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_util.go#L173-L178) for max, min, latest as "not reported". So user can see which metrics were reported and which were not.

What do you think @sperlingxx @gaocegege @johnugeorge ?

gaocegege:
SGTM.

johnugeorge:
Sgtm

andreyvelich:
Will make this change.
/assign @andreyvelich

PR #1245, opened 2020-06-26
String type for metric values

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1227.

I changed metric value from `float64` to `string`.
If metrics can't be reported, `unavailable` value will be recorded.

I added condition that `metric.Latest` can't be `unavailable` here: https://github.com/andreyvelich/katib/blob/cd6b37b3614cd226781c0feac1dcf6d568d07f01/pkg/controller.v1beta1/trial/trial_controller_util.go#L133-L135.
If **Objective** metric values were not reported, Trial will be in `False` status and `Succeeded` condition with this message:
`Warning  MetricsUnavailable  2m11s (x2 over 2m11s)  trial-controller  Metrics are not available for Job`.

Experiment will be still running and waiting until at least one **Objective** metric value will be reported.

I think it is correct workflow, since user can see that **Objective** metric is not printed by training container, modify the training container image and manually restart the Trial, by deleting appropriate BatchJob, TFJob, etc.. .

What do you think @sperlingxx @gaocegege @johnugeorge ?

### DIFF
--- a/pkg/apis/controller/common/v1beta1/common_types.go
+++ b/pkg/apis/controller/common/v1beta1/common_types.go
@@ -83,10 +83,10 @@ type MetricStrategy struct {
 }
 
 type Metric struct {
-	Name   string  `json:"name,omitempty"`
-	Min    float64 `json:"min,omitempty"`
-	Max    float64 `json:"max,omitempty"`
-	Latest string  `json:"latest,omitempty"`
+	Name   string `json:"name,omitempty"`
+	Min    string `json:"min,omitempty"`
+	Max    string `json:"max,omitempty"`
+	Latest string `json:"latest,omitempty"`
 }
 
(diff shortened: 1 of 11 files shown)


sperlingxx:
This PR sounds good to me, except I still fill not good about pending trial if its job succeed but metrics unavailable. Considering no warning messages will be shown on __user interface__, when this condition happens.
/lgtm

andreyvelich:
> This PR sounds good to me, except I still fill not good about pending trial if its job succeed but metrics unavailable. Considering no warning messages will be shown on **user interface**, when this condition happens.
> /lgtm

Maybe we could notify user about this case in the Experiment monitor page in the Katib UI?

gaocegege:
SGTM. It can be in another PR.

PR #1248, opened 2020-06-29
Adding retries for gRPC calls

akirillov:
This PR adds retries for gRPC calls in `suggestionclient.go`. One of the specific problems resolved by retries is the timing issue in Katib controller pod when running on Istio 1.5+: when Experiment deployment is ready, Envoy sidecar health check doesn't treat it as a healthy upstream yet and `ValidateAlgorithmSettings` call from the controller fails with `no healthy upstream`.

In general, retries can improve resilience and help to avoid transient failures instead of marking experiment as failed after the first and the only attempt.

### DIFF
--- a/pkg/controller.v1alpha3/consts/const.go
+++ b/pkg/controller.v1alpha3/consts/const.go
@@ -1,6 +1,9 @@
 package consts
 
-import "github.com/kubeflow/katib/pkg/util/v1alpha3/env"
+import (
+	"github.com/kubeflow/katib/pkg/util/v1alpha3/env"
+	"time"
+)
 
 const (
 	// ConfigExperimentSuggestionName is the config name of the
@@ -34,6 +37,11 @@ const (
 	// DefaultGRPCService is the default service name in Suggestion,
 	// which is used to run healthz check using grpc probe.
 	DefaultGRPCService = "manager.v1alpha3.Suggestion"
+
+	// DefaultGRPCRetryAttempts is the the maximum number of retries for gRPC calls
+	DefaultGRPCRetryAttempts = 10
+	// DefaultGRPCRetryPeriod is a fixed period of time between gRPC call retries
+	DefaultGRPCRetryPeriod = 3 * time.Second
 
 	// DefaultKatibNamespaceEnvName is the default env name of katib namespace
 	DefaultKatibNamespaceEnvName = "KATIB_CORE_NAMESPACE"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -21,7 +21,7 @@
   version = "v33.2.0"
 
 [[projects]]
-  digest = "1:7dc53e012b9e58f6fc52dce9c201d38dbc3c0e573bad4a56fa0f861823934915"
+  digest = "0:"
   name = "github.com/Azure/go-autorest"
   packages = [
     "autorest",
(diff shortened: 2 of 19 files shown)


pkg/controller.v1alpha3/consts/const.go
```diff
@@ -35,6 +38,11 @@ const (
 	// which is used to run healthz check using grpc probe.
 	DefaultGRPCService = "manager.v1alpha3.Suggestion"
 
+	// DefaultGRPCRetryCount is the the maximum number of retries for gRPC calls
```
andreyvelich, line 41: Should we have DefaultGRPCRetryAttempts here?

akirillov: Good catch, thanks. Fixed that.

akirillov:
thanks for the review, @gaocegege, @andreyvelich! Speaking of adding the changes to both alpha and beta, I've noticed that even though the image tags in this PR https://github.com/kubeflow/manifests/pull/1239 were updated to 0.9.0 release, the images are still using `v1alpha3` version e.g. 
```
images:
  - name: gcr.io/kubeflow-images-public/katib/v1alpha3/katib-controller
    newTag: 917164a
```

Is it intentional?

andreyvelich:
> thanks for the review, @gaocegege, @andreyvelich! Speaking of adding the changes to both alpha and beta, I've noticed that even though the image tags in this PR [kubeflow/manifests#1239](https://github.com/kubeflow/manifests/pull/1239) were updated to 0.9.0 release, the images are still using `v1alpha3` version e.g.
> 
> ```
> images:
>   - name: gcr.io/kubeflow-images-public/katib/v1alpha3/katib-controller
>     newTag: 917164a
> ```
> 
> Is it intentional?


Yes, the current Katib release for Kubeflow 1.1 is v1alpha3 with `917164a` commit. This is the cut release: https://github.com/kubeflow/katib/releases/tag/v0.9.0.
Currently, we make all new changes only for v1beta1 version and we will try to release v1beta1 in the next Kubeflow release.
We update files for v1alpha3 version only with the urgent bug fixes.

akirillov:
Thanks for the clarification, @andreyvelich. So what's the decision, shall I revert the changes for `v1alpha3` and keep them only in `v1beta1`? Until `v1beta1` is released, the fix can be helpful for the users running Katib `v1alpha3` on Istio 1.5+ because, without it, all created Experiments are marked as failed after `ValidateAlgorithmSetting` gRPC call fails with 503 and no retry.

andreyvelich:
@akirillov What do you think about adding this change to v1alpha3 and v1beta1 version. 
I am not sure if we can include this change to the Katib v0.9.0, which deployed with Kubeflow 1.1, because it is already cut and manifest repository is updated. 

But if users come with this problem, they just need to update Katib controller image to the latest version and should be able to user Istio 1.5 +. They don't need to deploy v1beta1 version.

Let's see what others say.

akirillov:
thanks, @andreyvelich, this PR adds it to both versions and I think it is beneficial to have it that way to avoid waiting for `v1beta1` release. This is the use case my team is dealing with and it would be great to have `v1alpha3` image with the fix published to avoid living on the custom build from a fork until `v1beta1` is out.

andreyvelich:
@akirillov Got it.

Can we just update image tags for the master branch in manifest repository?
I believe, we have kfdef config, which is related on master branch.

gaocegege:
I have no objection to it. If users (akirillov 's team) need it we can do it.

/cc @johnugeorge

johnugeorge:
SGTM
/approve

PR #1256, opened 2020-07-07
controllers watch katib objects with specific APIVersion 

sperlingxx:
For now, both v1alpha3 and v1beta1 controllers can watch katib objects(experiments/suggestions/trials) of all versions. It will cause objects being watched by controller who is responsible for other versions. The problem occurs when we deploy both v1alpha3 controller and v1beta1 controller in same cluster. In my opinion, to update katib from v1alpha3 to v1beta1 without breaking existing workloads,  it is essential to deploy katib controllers of multiple versions in same cluster.

cc @andreyvelich @gaocegege @johnugeorge

### DIFF
--- a/pkg/controller.v1alpha3/experiment/experiment_controller.go
+++ b/pkg/controller.v1alpha3/experiment/experiment_controller.go
@@ -115,8 +115,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 // addWatch adds a new Controller to mgr with r as the reconcile.Reconciler
 func addWatch(mgr manager.Manager, c controller.Controller) error {
+	experimentType := experimentsv1alpha3.Experiment{}
+	experimentType.APIVersion = consts.APIVersionToWatch
+	trialType := trialsv1alpha3.Trial{}
+	trialType.APIVersion = consts.APIVersionToWatch
+	suggestionType := suggestionsv1alpha3.Suggestion{}
+	suggestionType.APIVersion = consts.APIVersionToWatch
+
 	// Watch for changes to Experiment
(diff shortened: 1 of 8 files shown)


pkg/controller.v1alpha3/experiment/experiment_controller.go
```diff
@@ -115,19 +115,26 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 // addWatch adds a new Controller to mgr with r as the reconcile.Reconciler
 func addWatch(mgr manager.Manager, c controller.Controller) error {
+	experimentType := experimentsv1alpha3.Experiment{}
+	experimentType.APIVersion = consts.APIVersionToWatch
+	trialType := trialsv1alpha3.Trial{}
+	trialType.APIVersion = consts.APIVersionToWatch
+	suggestionType := suggestionsv1alpha3.Suggestion{}
+	suggestionType.APIVersion = consts.APIVersionToWatch
```
gaocegege, line 123: Why do we need it? I think watch can infer the APIVersion from the type.

sperlingxx: @gaocegege After testing with controllers of both versions, I found v1beta1 experiments can also be watched by v1alpha3 controller. Vice versa.

gaocegege: Did you register two API versions?

gaocegege: Which one is served/storaged? Ref https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition

sperlingxx: @gaocegege  I found my problem, thanks for documentation! It's really helpful.

Issue #1214, opened 2020-06-10
[Feature] Modify Job provider to support any kind of Kubernetes CRDs


andreyvelich:
/kind feature

After migrating to the new Trial Template design (https://github.com/kubeflow/katib/pull/1202), we want to extend current Job provider to support any kind of Kubernetes CRDs that follow Trial job patterns (e.g Argo template: https://github.com/kubeflow/katib/issues/1081). 

Currently, Job provider supports only `batch` Jobs and Kubeflow Jobs.

We can extend Trial Template API with the custom settings to define:
1. How to get succeeded state of the Job.
2. How to properly mutate metrics collector on training pod.
3. How to specify `sidecar.istio.io/inject: false` annotation. (That can be done by user in advance).

Maybe we should define something more, this needs to be investigated.

Let's discuss about all required changes in this issue.

/cc @gaocegege @johnugeorge @czheng94 @nielsmeima

andreyvelich:
> I would like to contribute a bit to point 2 you mentioned @andreyvelich. I implemented support for the Argo Workflow CRD using the v1alpha3 `Provider` interface and discovered some issues which are also relevant to the new proposed API (great work btw, looks very exciting).

Thank you! You are more than welcome to contribute. This is great that Argo Workflow works with Katib. 
I think any new changes that we will make should be done in v1beta1. We will release v1alpha3 soon.

> I currently added a simple addition to the filter statement in `pns.go` (`pid == 1 || pid == thisPID || proc.PPid() != 0 || proc.Executable() == "argoexec"`) as a temporary solution. However, it is clear that we should introduce a mechanism to account for this.

Did you check what was the list in [`pids`](https://github.com/kubeflow/katib/blob/master/pkg/metricscollector/v1beta1/common/pns.go#L61). Why do you need to add `proc.Executable() == "argoexec"`?

> I have thought of the following potential solutions:
> 
> 1. Simply wrap all processes as we currently do with containers marked as training containers. This approach would be compatible with other tool that are integrating too: they might control their own sidecars. We simply watch till every sidecar completes before deleting the CRD in question.

I'd like this approach and I have the same suggestion in my mind. I think metrics collector should wait until all processes will be succeeded. Only few concerns that I see here:
1. What about Init containers. Will we see the init containers processes in the proc list, or they will be succeeded when we are injecting metrics collector container?

2. Is there any case, when any other processes will be running on Pod ones main training job is finished? How we should indicate them and delete from proc list?

[...]

Issue #692, opened 2019-07-19
support early stop feature


hougangliu:
Early stop should not only consider goal metrics, but also epoch, step and so on

gaocegege:
/0.7.0

andreyvelich:
Do we have any updates to support Early Stopping in Katib?
I think firstly we should try to implement Median Stopping Rule as @richardsliu noticed here https://github.com/kubeflow/katib/issues/936#issue-529639832.

According to Google Vizier paper, https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46180.pdf algorithm should analyse metrics from Running Trials and stop them.
In our case, Trial controller should process Trial metrics during training process and send them to Early Stopping service.

Right now, Metrics Collector parses metrics only once training process is finished (https://github.com/kubeflow/katib/blob/master/cmd/metricscollector/v1alpha3/file-metricscollector/main.go#L94).
So we should somehow send metrics from Trials to Early Stopping service.

Any thoughts @hougangliu @gaocegege @johnugeorge ?

johnugeorge:
There was some discussion regarding this before. I couldn't find the issue. @gaocegege

andreyvelich:
I have an idea how we can implement Early Stopping in current Katib functionality.

Maybe instead of creating independent service for Early Stopping, try to Mutate another container to Trial Pod, like we are doing with metrics collector, and stop main training process when Trial needs to be early stopped.

This is an example for median stopping rule:

1. Katib is training and reporting metrics for X steps.

2. After X steps Suggestion generates median rule from reported metrics.

3. Katib controller creates Trial job and injects to training pod Early Stopping container with appropriate parameters (e.g stop_rule = accuracy >= 0.7). We can use same injection webhook as now.

4. We add additional commands to Training Container to handle `kill` Linux process.
For example 
```
python3 mnist.py --num-layers=2 1 > /var/log/katib/metrics.log 
|| if test -f '/var/log/katib/early-stopping'; then 
         echo 'Training Container was Early Stopped'; 
      else 
         echo 'Training Container was failed'; 
         exit 1; 
    fi; 
&& echo completed > /var/log/katib/$$$$.pid
```

Ones python3 execution is failed container runs:
```
if test -f '/var/log/katib/early-stopping'; then 
         echo 'Training Container was Early Stopped'; 
      else 
         echo 'Training Container was failed'; 
         exit 1;
fi; 
```

[...]

PR #1273, opened 2020-07-18
Proposal: Support custom CRD in Trial Job

andreyvelich:
See comment: https://github.com/kubeflow/katib/issues/1214#issuecomment-659756130.
I added proposal for supporting any kind of CRD in Trial Spec.

Please take a look @gaocegege @johnugeorge @czheng94

/cc @sperlingxx @jlewi @nielsmeima @terrykong

### DIFF
--- a/docs/proposals/trial-custom-crd.md
+++ b/docs/proposals/trial-custom-crd.md
@@ -2,6 +2,7 @@
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
 ## Table of Contents
 
 - [Motivation](#motivation)
@@ -18,33 +19,34 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-Created by [doctoc](https://github.com/thlorenz/doctoc)
+Created by [doctoc](https://github.com/thlorenz/doctoc).
 
 ## Motivation
 
 Running trial is one of the essential step of running Katib experiments.
 We implemented new trial template design in Katib v1beta1 ([katib/pull#1202](https://github.com/kubeflow/katib/pull/1202)
 and [katib/pull#1215](https://github.com/kubeflow/katib/pull/1215)) to make
-experiments valid YAMLand make Katib more Kubernetes native.
+experiments valid YAML and make Katib more Kubernetes native.
 After migrating to the new API, users still can run only [BatchJob](https://kubernetes.io/docs/concepts/workloads/controllers/job/),
 [TFJob](https://github.com/kubeflow/tf-operator) or [PyTorchJob](https://github.com/kubeflow/pytorch-operator) as a trial job.
-If we want to support new CRD, we need to manually change Katib controller source code.
-This approach makes impossible to use other CRDs in trial template, even if they can be used in trial design.
-Number of various Kubernetes CRDs grows significantly and many users wants to use them in Katib
-(e.g, [katib/issue#1081 (Support Argo Workflow)](https://github.com/kubeflow/katib/issues/1081)).
-Another reason to design unify approach is that CRD controller can have go package dependencies versions
+If we would like to support new CRD, we need to manually change Katib controller source code.
+
+This approach makes impossible to use other CRDs in trial template, even if they can satisfies trial job design.
+Number of various Kubernetes CRDs grows significantly and many users would like to use them in Katib
+(e.g, [katib/issue#1081, support Argo Workflow)](https://github.com/kubeflow/katib/issues/1081)).
+Another reason to design unify approach, is that CRD controller can have go package dependencies versions
 that Katib controller doesn't support and make it not possible to build the controller
 (e.g, [katib/issue#1081](https://github.com/kubeflow/katib/issues/1081#issuecomment-635338276)).
 
-Thus we propose new design to support custom CRDs in template and make Katib usable for various Kubernetes resources.
-To make this possible we make changes in API, trial controller, job provider, mutation webhook, metrics collector.
+Thus we propose new controller design to support custom CRD as Trial job and make Katib usable for various Kubernetes resources.
+To make this possible, we make changes in API, trial controller, job provider, mutation webhook, metrics collector.
 
 ## Goals
 
-1. Setup trial controller to watch for the custom CRD changes.
+1. Watch for the custom CRD in trial controller.
 2. Inject Katib sidecar container on training pod.
-3. Find the container for Katib injection in list of training pod containers.
-4. Run metrics collector parser after completion of all pod processes.
+3. Indicate training container for metrics collector execution.
+4. Run metrics collector parser after all pod processes completion.
 5. Get succeeded status of running CRD.
 6. Verify that `sidecar.istio.io/inject: false` label is added.
 
@@ -56,8 +58,8 @@ To make this possible we make changes in API, trial controller, job provider, mu
 
 ## Implementation
 
-During implementation this feature we should not brake current Katib controller logic to
-make sure that CI is stable and it not blocks other Katib related tasks.
+During implementation this feature, we should not brake current Katib controller logic.
+And we need to make sure that CI is stable and it not blocks other Katib work tasks.
 After completion, we can clean-up redundant code.
 
 ### API
@@ -68,20 +70,17 @@ To achieve above goals, we introduce these `TrialTemplate` API changes.
 
 // TrialTemplate describes structure of Trial template
 type TrialTemplate struct {
-	// Retain indicates that Trial resources must be not cleanup
-	Retain bool `json:"retain,omitempty"`
+  // Retain indicates that Trial resources must be not cleanup
+  Retain bool `json:"retain,omitempty"`
 
-	// Source for Trial template (unstructured structure or config map)
-	TrialSource `json:",inline"`
+  // Source for Trial template (unstructured structure or config map)
+  TrialSource `json:",inline"`
 
-	// List of parameters that are used in Trial template
+  // List of parameters that are used in Trial template
   TrialParameters []TrialParameterSpec `json:"trialParameters,omitempty"`
 
   // Label that determines if pod needs to be injected by Katib sidecar container
   PrimaryPodLabel map[string]string `json:"primaryPodLabel,omitempty"`
-
-  // Name of training container where training is running
-  PrimaryContainerName string `json:"primaryContainerName,omitempty"`
 
   // Name of training container where training is running
   PrimaryContainerName string `json:"primaryContainerName,omitempty"`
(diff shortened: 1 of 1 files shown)


docs/proposals/trial-custom-crd.md
```diff
@@ -0,0 +1,189 @@
...
+  SucceededCondition string `json:"succeededCondition,omitempty"`
+
+}
+```
+
+### Trial controller watchers
+
+In the current design trial controller watches
+[three supported resource](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller.go#L94-L125).
+To generate these parameters dynamically when Katib starts, we add additional flag (`-trial-resource`)
```
sperlingxx, line 97: Maybe we can use configMap to define these trial-resources ?

docs/proposals/trial-custom-crd.md
```diff
@@ -0,0 +1,189 @@
...
+  Retain bool `json:"retain,omitempty"`
+
+  // Source for Trial template (unstructured structure or config map)
+  TrialSource `json:",inline"`
+
+  // List of parameters that are used in Trial template
+  TrialParameters []TrialParameterSpec `json:"trialParameters,omitempty"`
+
+  // Label that determines if pod needs to be injected by Katib sidecar container
+  PrimaryPodLabel map[string]string `json:"primaryPodLabel,omitempty"`
```
sperlingxx, line 82: I am curious about how will these extra fields be filled ?

docs/proposals/trial-custom-crd.md
andreyvelich: Currently, in `katib-config` ConfigMap we can set only Suggestion and Metrics collector settings.
Also, we added Watch only when controller starts: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller.go#L79.
I think later when we implement dynamically wathers update, we can think about better design how we can send these resources to the Controller.

docs/proposals/trial-custom-crd.md
andreyvelich: In this design it will look like this:
```yaml
PrimaryPodLabel:
  "label-key": "label-value"
```
Not sure if it is the best design.
We can follow the same API as [`metricStrategies`](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/common/v1beta1/common_types.go#L80-L83):

```go
. . .
PrimaryPodLabel *PrimaryPodLabel
. . .

type PrimaryPodLabel struct {
	Name  string `json:"name,omitempty"`
	Value string `json:"value,omitempty"`
}
```

Does it make sense, if we have can set 1 label currently?

andreyvelich:
> I think the new design is supposed to replace current jobPrivider intereface. With the help of new design, we can support arbitrary CRDs without adding go codes (implementing/registering a new jobPrivider).
> It looks fantastic to me except how we support `Provider.MutateJob` under new design.

Thanks, yes we should refactor `jobPrivider` and later we can left only 1 unify provider for every CRDs.

@sperlingxx Do we have a use-case when JobLevel injection might be useful?
I believe Job-level injection was introduced here: https://github.com/kubeflow/katib/blob/master/docs/proposals/metrics-collector.md#mutating-webhook.
I can't see that we use `ObjectSelector` for injection webhook: https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/webhook.go#L103-L114.

Also, `MutateJob` for Kubeflow and Job providers are currently empty: https://github.com/kubeflow/katib/blob/master/pkg/job/v1beta1/kubeflow.go#L80-L82 and
https://github.com/kubeflow/katib/blob/master/pkg/job/v1beta1/job.go#L72-L74.

sperlingxx:
> > I think the new design is supposed to replace current jobPrivider intereface. With the help of new design, we can support arbitrary CRDs without adding go codes (implementing/registering a new jobPrivider).
> > It looks fantastic to me except how we support `Provider.MutateJob` under new design.
> 
> Thanks, yes we should refactor `jobPrivider` and later we can left only 1 unify provider for every CRDs.
> 
> @sperlingxx Do we have a use-case when JobLevel injection might be useful?
> I believe Job-level injection was introduced here: https://github.com/kubeflow/katib/blob/master/docs/proposals/metrics-collector.md#mutating-webhook.
> I can't see that we use `ObjectSelector` for injection webhook: https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/webhook.go#L103-L114.
> 
> Also, `MutateJob` for Kubeflow and Job providers are currently empty: https://github.com/kubeflow/katib/blob/master/pkg/job/v1beta1/kubeflow.go#L80-L82 and
> https://github.com/kubeflow/katib/blob/master/pkg/job/v1beta1/job.go#L72-L74.

@andreyvelich In fact, we use `MutateJob` to do some adaption work in our origization. And it seems Job-Level injection can do the same job.

andreyvelich:
> @andreyvelich In fact, we use `MutateJob` to do some adaption work in our origization. And it seems Job-Level injection can do the same job.

In that case, I think we can follow 2 ways:
1. Add [`Mutate` and `Create`](https://github.com/kubeflow/katib/blob/master/pkg/job/v1beta1/kubeflow.go#L80-L87) functions to our dynamic provider and left it empty. If end-user needs to do some Job level mutation, he/she can modify that functions manually.

2. Create 2 providers: `default` and `custom`. In `default` provider we don't need to add JobLevel mutation, but in `custom` user can modify `Mutate` function in controller. Add new flag to Katib controller, which indicate which provider you want to use. 

Do you have any other ideas @sperlingxx ?

jlewi:
@andreyvelich Could you present this at an upcoming community meeting and we can do a design review?

andreyvelich:
> @andreyvelich Could you present this at an upcoming community meeting and we can do a design review?

Sure, thanks @jlewi.

sperlingxx:
> > @andreyvelich In fact, we use `MutateJob` to do some adaption work in our origization. And it seems Job-Level injection can do the same job.
> 
> In that case, I think we can follow 2 ways:
> 
> 1. Add [`Mutate` and `Create`](https://github.com/kubeflow/katib/blob/master/pkg/job/v1beta1/kubeflow.go#L80-L87) functions to our dynamic provider and left it empty. If end-user needs to do some Job level mutation, he/she can modify that functions manually.
> 2. Create 2 providers: `default` and `custom`. In `default` provider we don't need to add JobLevel mutation, but in `custom` user can modify `Mutate` function in controller. Add new flag to Katib controller, which indicate which provider you want to use.
> 
> Do you have any other ideas @sperlingxx ?

I prefer the second way, which seems to be more extensible. I suppose the dynamic provider will be the `default` one, and we also support adding custom providers.  I am not sure whether I have the right understanding.

docs/proposals/trial-custom-crd.md
gaocegege: I think we should provide a map or a slice here.

andreyvelich: Map looks good to me, since it follow k8s way: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L226-L231.

jlewi:
Thanks @andreyvelich A couple questions

Have you asked users of Katib for feedback?

How does this proposal compare to the approach Tekton is taking with [custom tasks](https://github.com/imjasonh/community/blob/custom-tasks/teps/0002-custom-tasks.md)? 

> Verify that sidecar.istio.io/inject: false label is added.

Why is Katib cotrollers getting involved here? Could a user control this by directly setting labels on their resource? e.g. for TFJob they could add the labels to the PodTemplateSpec in TFJob?

Would the proposal be different if we (Kubeflow/Kubernetes whatever) had a first class concept of inputs and outputs?

* The Kubernetes resource model is defined [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields)
* The fact that all Kubernetes objects have those fields enables a lot of functionality
   * e.g. the reason you can do bulk operations (kubectl apply -f ${DIR}) is because every YAML has apiVersion and kind so kubectl knows what type of object every object is
* What if we defined an extension of the KRM; call it KFRM which included fields like inputs and outputs so that outputs could be used to report things like metrics. Would that simplify things?

andreyvelich:
Thanks for the review @jlewi.

> Thanks @andreyvelich A couple questions
> 
> Have you asked users of Katib for feedback?

We have couple of requests to support [Argo](https://github.com/kubeflow/katib/issues/1081), [Kubeflow operators](https://github.com/kubeflow/katib/pull/1183). @czheng94 has many various CRD that his team want to use in Katib. Idea of this proposal is definitely on high demand.

> How does this proposal compare to the approach Tekton is taking with [custom tasks](https://github.com/imjasonh/community/blob/custom-tasks/teps/0002-custom-tasks.md)?

From my understanding, [custom tasks controller author](https://github.com/tektoncd/community/blob/master/teps/0002-custom-tasks.md#proposal) is responsible to proceed `Run` object and then execute their custom resource. User's controller should watch for the `Run` Object.
I didn't find in Tekton proposal that user can specify whole custom specification in Reference, only type of object:
```yaml
ref:
    apiVersion: example.dev/v0
    kind: Example
    name: my-example
```
Otherwise, here trial controller will be responsible to submit custom jobs and user doesn't need to modify source code and add watchers for the Katib Trial CR.
As well, we have to properly inject Katib sidecar container which is Katib controller responsibility.

> > Verify that sidecar.istio.io/inject: false label is added.
> 
> Why is Katib cotrollers getting involved here? Could a user control this by directly setting labels on their resource? e.g. for TFJob they could add the labels to the PodTemplateSpec in TFJob?

You are right. This is exactly what I want to say in this proposal. User has to specify this annotation in TrialSpec.

> Would the proposal be different if we (Kubeflow/Kubernetes whatever) had a first class concept of inputs and outputs?
> 
> * The Kubernetes resource model is defined [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields)
> * The fact that all Kubernetes objects have those fields enables a lot of functionality
>   
>   * e.g. the reason you can do bulk operations (kubectl apply -f ${DIR}) is because every YAML has apiVersion and kind so kubectl knows what type of object every object is
> * What if we defined an extension of the KRM; call it KFRM which included fields like inputs and outputs so that outputs could be used to report things like metrics. Would that simplify things?

Yes, I think that can simplify things, because metrics collector doesn't need to parse StdOut/File for the training container and metrics can be directly pushed to the DB, but we have various metrics collector (e.g. TFEvent) to get various metrics. 

This KRM should handle all of these functionality for the custom user's jobs.

Also, we should think how we can implement Early Stopping (https://github.com/kubeflow/katib/issues/692) in that approach without Katib sidecar container and without Katib SDK for early stopping, like in other opt framework.

jlewi:
Thanks. I don't consider myself an approver; just a passer by so this PR can be merged whenever the approvers LGTM it.

> From my understanding, custom tasks controller author is responsible to proceed Run object and then execute their custom resource. User's controller should watch for the Run Object.

I believe the way custom tasks work is as analogous to how Tasks and TaskRuns work today in Tekton

1. Your pipeline contains references to a resource that acts as a template; e.g. we might have TFJobTemplate

    * The template would define inputs and outputs to be passed at runtime

1. At runtime Tekton creates an instance of the run object e.g. TFJobRun

    * A run is a combination of a Template (e.g. TFJobTemplate) and specific values to be passed to be passed as inputs
    * The run controller then executes the job using those paramters; e.g. the run controller could substitute the parameters into the TFJobTemplate and to create a TFJob

andreyvelich:
> 1. Your pipeline contains references to a resource that acts as a template; e.g. we might have TFJobTemplate
>    

@jlewi If pipeline contains only reference on a template and input params, where you define template specification?
How Tekton can submit Job in runtime, if we pass only Reference to the object?

I haven't seen examples in Tekton, when we define template spec or maybe I misunderstand something ?

andreyvelich:
@gaocegege @johnugeorge @sperlingxx Do you have any other comment on this proposal or we can merge it and start the implementation.

johnugeorge:
it looks good to me.  

@gaocegege

Issue #1250, opened 2020-06-30
Save Suggestion state in persistent volume


andreyvelich:
/kind feature

To continue idea proposed here: https://github.com/kubeflow/katib/issues/1062.
We would like to attach persistent volume to every deployed suggestion to save suggestion state after corresponding pod is deleted.
To implement this we should follow these steps:

1. Extend [`ResumePolicyType`](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/experiments/v1beta1/experiment_types.go#L162-L165) with the new type. My idea, name it `VolumeSource`, any other ideas?

2. Add new Storage Class YAML to Katib deployed manifests. I am not sure what should be the default provisioner for us, since k8s doesn't support dynamic volume provisioning for local storage (https://kubernetes.io/blog/2019/04/04/kubernetes-1.14-local-persistent-volumes-ga/#limitations-of-ga).
We can be specific to **GKE** and use Persistent Disks (https://kubernetes.io/docs/concepts/storage/storage-classes/#gce-pd). Or use 3rd party local path provisioner (https://github.com/rancher/local-path-provisioner), this require additional controller.
What do you think ?

3. Implement new logic to the controller:
- Create PVC when user submits Experiment.
- Attach PVC to suggestion deployment.
- Delete suggestion resources when Experiment is succeeded.
- Restore suggestion resources when Experiment is resuming.

4. Extend `katib-config` with the new parameters for suggestion PVC.

What do you think @gaocegege @johnugeorge ?

/cc @sperlingxx @c-bata @jlewi 
/priority p0

jlewi:
Why use pickling to store data as opposed to YAMl?
Im lacking some backgeound. What precisely does persisting a suggestion mean? Is a suggestion just a set of hyperparameters?


If you use PDs are you going to have to manage disks?

Could you persist to object store or some other easily managed datastore?

Would it make sense to treat this data as metadata and store in metadata store?

johnugeorge:
Wrt to 1,  we will need  to have  consistent naming with other values
@gaocegege   Wrt to 2(Adding new storage class yaml), How is it handled?

andreyvelich:
Thanks for the comment @jlewi.
> Why use pickling to store data as opposed to YAMl?
> Im lacking some backgeound. What precisely does persisting a suggestion mean? Is a suggestion just a set of hyperparameters?
> 
[...]

PR #1275, opened 2020-07-21
Resume Experiment from Volume

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1250.

This is implementation for resuming experiment from the Volume.

Steps:

1. I added new `ResumePolicy=FromVolume` which represents resuming experiment from the volume. 
@gaocegege @johnugeorge @sperlingxx what do you think about the name? Should we rename it to `ResumePolicy=VolumeSource` or think about better name, maybe rename `LongRunning` also? 

2. I added `ResumePolicy` to Suggestion API to understand if controller needs to deploy volume for the Suggestion. 
Do we need to add some volume specification to Suggestion API also. 
For example: StorageClassName, PVC name ?

3. If `ResumePolicy=VolumeSource`, PVC is created. If `StorageClassName=DefaultSuggestionStorageClass`, PV with local cluster path is created also.

4. I didn’t attach Experiment labels to PVC and PV, like we did here: https://github.com/andreyvelich/katib/blob/issue-1250-resume-from-volume/pkg/controller.v1beta1/suggestion/composer/composer.go#L74 (it was done to connect service with deployment). 
Do we want to see these labels in the PVC/PV?
5. Currently, for the PV default cluster local volume path is:  `prefix + <suggestion name>-<suggestion namespace>`. We need to add name and namespace to not conflict with other submitted Experiments in various namespaces.
Prefix can be changed in `katib-config` later.

6. Currently, for suggestion container default path is `/opt/katib/data`. Also can be changed in `katib-config`.

7. PVC name = `<suggestion name>`

8. PV name = `<suggestion name>-<suggestion namespace>`

9. I added [new validation](https://github.com/andreyvelich/katib/blob/issue-1250-resume-from-volume/pkg/webhook/v1beta1/experiment/validation_webhook.go#L94-L110) for the webhook. That is needed to prevent submitting Experiment with the same name, if corresponding PV was not deleted after deleting Experiment. 
We can't `Watch` for the PV, like for the [`PVC`](https://github.com/andreyvelich/katib/blob/issue-1250-resume-from-volume/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L99-L105). And suggestion controller can't be triggered and create PV again, once PV is deleted from the cluster. 

10. I had to change suggestion status workflow to allow resuming experiment from the volume.
This change doesn't affect current workflow for `LongRunning` Suggestion.
For the `ResumePolicy=VolumeSource`:
When User submits Experiment, Suggestion status:

Created (status = True) -> DeploymentReady (status = False) -> DeploymentReady (status = True) -> Running (status = True) -> (experiment is finished) -> Running (status = False) ->  DeploymentReady (status = False) -> Succeeded (status = True) . 

When User restarts Experiment, Suggestion status:

Running (status = True, reason="Experiment is restarting") -> (remove succeeded condition) -> DeploymentReady (status = True) -> Running (status = True, reason="SuggestionRunning")) -> (experiment is finished) -> Running (status = False) ->  DeploymentReady (status = False) -> Succeeded (status = True).

This approach can avoid adding additional condition for Suggestion: `Restarting`.

I still need to make some testing, but this PR is ready for review.
I will add unit test and e2e test in the future PR.

/cc @jlewi

### DIFF
--- a/pkg/controller.v1beta1/experiment/experiment_util.go
+++ b/pkg/controller.v1beta1/experiment/experiment_util.go
@@ -155,11 +155,11 @@ func (r *ReconcileExperiment) restartSuggestion(instance *experimentsv1beta1.Exp
 		return err
 	}
 
-	logger.Info("Suggestion is restarting, suggestion status is not running")
+	logger.Info("Suggestion is restarting, suggestion Running status is false")
 	suggestion := original.DeepCopy()
-	reason := "Experiment is restarting, suggestion deployment is not ready"
+	reason := "Experiment is restarting"
 	msg := "Suggestion is not running"
-	// Mark suggestion status not running because deployment is not ready
+	// Mark suggestion status not running because experiment is restarting and suggestion deployment is not ready
 	suggestion.MarkSuggestionStatusRunning(corev1.ConditionFalse, reason, msg)
 
 	if err := r.UpdateSuggestionStatus(suggestion); err != nil {
--- a/pkg/controller.v1beta1/experiment/experiment_util.go
+++ b/pkg/controller.v1beta1/experiment/experiment_util.go
@@ -5,6 +5,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -12,7 +13,6 @@ import (
 	experimentsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/experiments/v1beta1"
 	suggestionsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/suggestions/v1beta1"
 	trialsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/trials/v1beta1"
-	suggestionController "github.com/kubeflow/katib/pkg/controller.v1beta1/suggestion"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/util"
 )
 
@@ -103,7 +103,8 @@ func (r *ReconcileExperiment) updateFinalizers(instance *experimentsv1beta1.Expe
 	}
 }
 
-func (r *ReconcileExperiment) terminateSuggestion(instance *experimentsv1beta1.Experiment) error {
+func (r *ReconcileExperiment) cleanupSuggestionResources(instance *experimentsv1beta1.Experiment) error {
+	logger := log.WithValues("Suggestion", types.NamespacedName{Name: instance.GetName(), Namespace: instance.GetNamespace()})
 	original := &suggestionsv1beta1.Suggestion{}
 	err := r.Get(context.TODO(),
 		types.NamespacedName{Namespace: instance.GetNamespace(), Name: instance.GetName()}, original)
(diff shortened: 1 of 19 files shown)


andreyvelich:
I tested example, everything was working.
I will remove WIP status.

Since we are using PV with local host path after deletion Suggestion PV and PVC, directory on cluster will be **not deleted.**
I am not sure that we can/should clean-up directory in host node's filesystem.

sperlingxx:
Hi @andreyvelich , after a careful review, I am still confused about how we restore/retrieve suggestion instance to/from PV.  Since in [restartSuggestion](https://github.com/kubeflow/katib/pull/1275/files#diff-fb096e014343ba37cb3679d228995e03R146) function, we fetch suggestion instance via kubernetes client directly.

andreyvelich:
@sperlingxx Thank you for the review.

When user submits Experiment, we deploy Suggestion service and deployment with attached PVC on it.
When Experiment is succeeded, we mark Suggestion succeeded and delete deployment and service: https://github.com/kubeflow/katib/blob/44b0ae646c5d29a6d13681ac664fa08e4313a3f4/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L147-L157.
We don't delete PVC and it saves information from the Suggestion deployment.
Ones user wants to restart Experiment, we change Suggestion status: https://github.com/kubeflow/katib/blob/44b0ae646c5d29a6d13681ac664fa08e4313a3f4/pkg/controller.v1beta1/experiment/experiment_util.go#L163.
Then it automatically runs `ReconcileSuggestion` https://github.com/kubeflow/katib/blob/44b0ae646c5d29a6d13681ac664fa08e4313a3f4/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L187 and it creates Deployment and Service that we have deleted before.
PV/PVC is not created again because we didn't delete it.

If user deletes Experiment, it automatically clean-up all Suggestion resources (deployment/service/pvc/pv) since we have owner reference for each object.

pkg/controller.v1beta1/experiment/experiment_util.go
```diff
@@ -113,15 +114,53 @@ func (r *ReconcileExperiment) terminateSuggestion(instance *experimentsv1beta1.E
...
 	}
-	log.Info("Start terminating suggestion")
+
+	logger.Info("Start cleanup suggestion resources")
+	suggestion := original.DeepCopy()
+
+	reason := "Experiment is succeeded"
+	// If ResumePolicy = Never, mark suggestion status succeeded, can't be restarted
+	if instance.Spec.ResumePolicy == experimentsv1beta1.NeverResume {
+		msg := "Suggestion is succeeded, can't be restarted"
```
johnugeorge, line 129: why do we add restart message when suggestion is terminated?

johnugeorge:
4. Why not?

sperlingxx:
@andreyvelich Thanks for detailed explanation. So, we assume suggestion programs can recover from disk (or maybe they are stateless) in `FromVolume` mode ?

andreyvelich:
> @andreyvelich Thanks for detailed explanation. So, we assume suggestion programs can recover from disk (or maybe they are stateless) in `FromVolume` mode ?

Yes, later add functionality to set StorageClassName from the `katib-config`. To allow user set [provisioner](https://kubernetes.io/docs/concepts/storage/storage-classes/#provisioner for the Suggestion volume.

pkg/controller.v1beta1/experiment/experiment_util.go
andreyvelich: Just for additional information after Suggestion is succeeded for the user.
@johnugeorge You think it is not necessary?

andreyvelich:
> 4. Why not?

@johnugeorge Not sure if it is needed for the Volume, since we have Owner Reference for the volume: https://github.com/andreyvelich/katib/blob/issue-1250-resume-from-volume/pkg/controller.v1beta1/suggestion/composer/composer.go#L288.

andreyvelich:
I renamed PVC to <suggestion name>-<suggestion algorithm name>, PV to <suggestion name>-<suggestion algorithm name>-<suggestion namespace> to be consistent with Suggestion's deployment and service.
I think user can easily understand which k8s resources related to Suggestion.

sperlingxx:
> StorageClassName from

Cool!
/lgtm

PR #1281, opened 2020-07-27
Unit test for resuming Experiment in controller reconcilers

andreyvelich:
I updated unit test for Experiment and Suggestion controllers.
I tried to make `suite_test` consistent between these controllers and cover more reconciler source code for resume Experiment feature.
For Experiment controller, I deleted `TestCreateExperiment`, because it doesn't add any additional tests with `TestReconcile`. 
Instead of that, I added `TestCompleteExperiment`.

Later, I will update Trial controller `suite_test` also.


andreyvelich:
@sperlingxx I deleted `buildTrialMetaForRunSpec` function from Experiment controller, I suppose in the new design you don't use it.

gaocegege:
Please fix the CI errors.

gaocegege:
Retest does not work for Travis. Please try to force push and trigger a new build.

sperlingxx:
> buildTrialMetaForRunSpec

Yes, it is useless.

andreyvelich:
@gaocegege I can see that test are passed.

andreyvelich:
I refactored Experiment test to make it more stable and it should be succeeded even if Controller fails [here](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/experiment/experiment_controller.go#L252-L257).
Now in `TestReconcile` I have added full workflow of Experiment. Created->Running->Succeeded->Restarting which makes more sense.

Do we actually need to traceback error [here](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/experiment/experiment_controller.go#L252), if `updateStatusHandler` failed. 
Maybe we could give `logger.Info` there and `return reconcile.Result{}, nil`, since controller will reconcile Experiment after this error?
That can improve Controller logging.

gaocegege:
If we do not retry, how to deal with temporary failures?

andreyvelich:
@gaocegege As I can see, this error can happen because resource version was modified for Experiment object, and it is more like Warning message (see comment: https://github.com/kubernetes/kubernetes/issues/28149#issuecomment-229469157).

If update status was unsuccessful, maybe we can `Get` Experiment object again and try to update it, instead of traceback this error and run whole reconcile loop again?

gaocegege:
I think resource version problems will be resolved in the next reconcile loop. Then we will just have one warning here instead of always reconciling it.

andreyvelich:
> I think resource version problems will be resolved in the next reconcile loop. Then we will just have one warning here instead of always reconciling it.

Currently, we print error message instead of warning. Do you think it makes sense to print error in case of not sync objects version?

gaocegege:
Can we requeue the request but not raise an error like this?

```
return reconcile.Result{
			Requeue: true,
		}, nil
```

andreyvelich:
> Can we requeue the request but not raise an error like this?
> 
> ```
> return reconcile.Result{
> 			Requeue: true,
> 		}, nil
> ```

Sure! Will make this change in the follow PR

PR #1297, opened 2020-08-06
Log update object status error

andreyvelich:
See comment: https://github.com/kubeflow/katib/pull/1281#issuecomment-666016563.
I changed error to log message to improve controller logging.
If `UpdateStatus` fails, we print the error and requeued reconciler.

As well, I noticed that `pod.Name` is not printed in inject webhook.
That is happening because object name is generated after Mutation unless appropriate controller has not set name explicitly (See comment: https://github.com/kubernetes/kubernetes/issues/46107#issuecomment-537592278.
I changed it to `GenerateName` instead.

### DIFF
--- a/pkg/controller.v1beta1/experiment/experiment_controller.go
+++ b/pkg/controller.v1beta1/experiment/experiment_controller.go
@@ -249,7 +249,7 @@ func (r *ReconcileExperiment) Reconcile(request reconcile.Request) (reconcile.Re
 		//assuming that only status change
 		err = r.updateStatusHandler(instance)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


andreyvelich:
I changed waiting for succeeded experiment in the same loop when we update experiment status in unit test which allow e:
https://github.com/kubeflow/katib/blob/efce83b4f21c31f168c8d9dfab3f12740578ef83/pkg/controller.v1beta1/experiment/experiment_controller_test.go#L270-L284


In some cases, controller doesn't reconcile after we update status of experiment and unit test can be failed:
https://travis-ci.org/github/kubeflow/katib/jobs/715545715#L371-L384

PR #1302, opened 2020-08-14
Custom CRD: Set dynamic watch from controller flags

andreyvelich:

I added possibility to set trial controller watchers from controller flags.
In controller deployment flags should be in format `Kind.version.group`.

For example:
```
args:
  - "-trial-resource=TFJob.v1.kubeflow.org".
  - "-trial-resource=Workflow.v1alpha1.argoproj.io"
  - "-trial-resource=Job.v1.batch"
```

/assign @gaocegege @sperlingxx @johnugeorge

### DIFF
--- a/cmd/katib-controller/v1beta1/main.go
+++ b/cmd/katib-controller/v1beta1/main.go
@@ -30,6 +30,7 @@ import (
 	apis "github.com/kubeflow/katib/pkg/apis/controller"
 	controller "github.com/kubeflow/katib/pkg/controller.v1beta1"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/consts"
+	trialutil "github.com/kubeflow/katib/pkg/controller.v1beta1/trial/util"
 	webhook "github.com/kubeflow/katib/pkg/webhook/v1beta1"
 )
 
@@ -44,6 +45,7 @@ func main() {
 	var injectSecurityContext bool
 	var serviceName string
 	var enableGRPCProbeInSuggestion bool
+	var trialResources trialutil.GvkListFlag
 
 	flag.StringVar(&experimentSuggestionName, "experiment-suggestion-name",
 		"default", "The implementation of suggestion interface in experiment controller (default)")
@@ -53,6 +55,7 @@ func main() {
 	flag.BoolVar(&injectSecurityContext, "webhook-inject-securitycontext", false, "Inject the securityContext of container[0] in the sidecar")
 	flag.StringVar(&serviceName, "webhook-service-name", "katib-controller", "The service name which will be used in webhook")
 	flag.BoolVar(&enableGRPCProbeInSuggestion, "enable-grpc-probe-in-suggestion", true, "enable grpc probe in suggestions")
+	flag.Var(&trialResources, "trial-resources", "The list of resources that can be used as trial template, in the form: Kind.version.group (e.g. TFJob.v1.kubeflow.org)")
 
 	flag.Parse()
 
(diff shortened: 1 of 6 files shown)


sperlingxx:
Shall we check whether there exists any overlap between dynamic GVKs and GVKs of `SupportedJobList`?

sperlingxx (commented):
Shall we check whether there exists any overlap between dynamic GVKs and GVKs of SupportedJobList?

andreyvelich:
> Shall we check whether there exists any overlap between dynamic GVKs and GVKs of SupportedJobList?

I think after complete custom CRD implementation we can remove SupportedJobList and by default add few `trial-resources` in Katib controller deployment for testing/examples.
What do you think @sperlingxx @gaocegege @johnugeorge ?

sperlingxx:
> > Shall we check whether there exists any overlap between dynamic GVKs and GVKs of SupportedJobList?
> 
> I think after complete custom CRD implementation we can remove SupportedJobList and by default add few `trial-resources` in Katib controller deployment for testing/examples.
> What do you think @sperlingxx @gaocegege @johnugeorge ?

Maybe we can keep `SupportedJobList` as a legacy and make current custom CRD implementation as default approach to watch and renconcile in trial level? In this way,  old-fashioned `SupportedJobList` could also be supported as a non-default option.

gaocegege:
> Maybe we can keep SupportedJobList as a legacy and make current custom CRD implementation as default approach to watch and renconcile in trial level? In this way, old-fashioned SupportedJobList could also be supported as a non-default option. In short, trial-controller will either watch SupportedJobList or custom CRDs, and the later one is default option.

SGTM

andreyvelich:
@sperlingxx Do you mean left `SupportedJobList` to not specify additional parameters in Trial template for default jobs (e.g. `PrimaryPodLabel`, `PrimaryContainerName`, `SucceededCondition`) ?
Do we have any other cases when `SupportedJobList` is useful ?

After implementation, Job provider can be used only for custom `MutateJob` function as we discussed here: https://github.com/kubeflow/katib/pull/1273#issuecomment-661533235, other functions can be omitted.

sperlingxx:
> @sperlingxx Do you mean left `SupportedJobList` to not specify additional parameters in Trial template for default jobs (e.g. `PrimaryPodLabel`, `PrimaryContainerName`, `SucceededCondition`) ?
> Do we have any other cases when `SupportedJobList` is useful ?
> 
> After implementation, Job provider can be used only for custom `MutateJob` function as we discussed here: [#1273 (comment)](https://github.com/kubeflow/katib/pull/1273#issuecomment-661533235), other functions can be omitted.

Yes, I supppose, due to `MutateJob` function, we should keep the whole stack of `SupportedJobList`?

andreyvelich:
@sperlingxx For `MutateJob` we don't actually need `SupportedJobList`. 
I think for custom [`MutateJob`](https://github.com/kubeflow/katib/blob/master/pkg/job/v1beta1/kubeflow.go#L81-L83) function you can just check the Kind of unstructured Job and make some changes if it is needed.
What do you think ?

sperlingxx:
> @sperlingxx For `MutateJob` we don't actually need `SupportedJobList`.
> I think for custom [`MutateJob`](https://github.com/kubeflow/katib/blob/master/pkg/job/v1beta1/kubeflow.go#L81-L83) function you can just check the Kind of unstructured Job and make some changes if it is needed.
> What do you think ?

Sry for slow reply. But @andreyvelich  how do we know which `MutateJob` function need to be invoked given an unstructured Job ?

andreyvelich:
@sperlingxx We invoke `MutateJob` with `desiredJob` object and we can get Kind of it.
Can we refactor Provider to have `MutateJob` per Kind?
In that case, users can re-use provider interface to define their own `MutateJob` function.
I believe all other functionality to handle Trial can be done without Provider.

sperlingxx:
> @sperlingxx We invoke `MutateJob` with `desiredJob` object and we can get Kind of it.
> Can we refactor Provider to have `MutateJob` per Kind?
> In that case, users can re-use provider interface to define their own `MutateJob` function.
> I believe all other functionality to handle Trial can be done without Provider.

SGTM

andreyvelich:
@gaocegege @johnugeorge @sperlingxx The test passed, can you approve it again, please?

johnugeorge:
After this change, Is supportedJobList required?

PR #1305, opened 2020-08-19
Custom CRD: Add primary pod labels

andreyvelich:

I named API: `PrimaryPodLabels` instead of `PrimaryPodLabel` because user can set more than 1 label there.
`PrimaryPodLabels` defines labels that mutated Pod must have.
I didn't remove `isMasterRole` to not brake current test examples.
Later we can update all examples with new APIs and remove redundant code.

Also I generated codegens and SDK.

### DIFF
--- a/pkg/apis/controller/experiments/v1beta1/experiment_types.go
+++ b/pkg/apis/controller/experiments/v1beta1/experiment_types.go
@@ -209,7 +209,7 @@ type TrialTemplate struct {
 	TrialParameters []TrialParameterSpec `json:"trialParameters,omitempty"`
 
 	// Labels that determines if pod needs to be injected by Katib sidecar container
-	PrimaryPodLabels map[string]string `json:"PrimaryPodLabels,omitempty"`
+	PrimaryPodLabels map[string]string `json:"primaryPodLabels,omitempty"`
 }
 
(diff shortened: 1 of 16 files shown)


pkg/apis/controller/experiments/v1beta1/experiment_types.go
```diff
@@ -207,6 +207,9 @@ type TrialTemplate struct {
 
 	// List of parameters that are used in trial template
 	TrialParameters []TrialParameterSpec `json:"trialParameters,omitempty"`
+
+	// Labels that determines if pod needs to be injected by Katib sidecar container
+	PrimaryPodLabels map[string]string `json:"primaryPodLabels,omitempty"`
```
johnugeorge, line 212: So for TFJob, user has to explicitly add primary labels ?

andreyvelich: User should specify it only in Experiment.
In TFJob for master pod, tfjob-controller automatically adds master labels.

As I said before we can set default values for `PrimaryPodLabels` for Kubeflow operators (TFJob, MPIJob).

johnugeorge: SGTM

PR #1306, opened 2020-08-20
Fix restart check in controller for completed experiments

andreyvelich:
We should check that Experiment is restartable in controller as well as in validation webhook.
For example, when Experiment is completed because optimal Trial was found, restart Experiment should not be triggered.

/assign @gaocegege @sperlingxx @johnugeorge


sperlingxx (approved):
/lgtm

andreyvelich (commented):
/approve

PR #1307, opened 2020-08-21
Custom CRD: Support dynamic Trial's jobs conditions

andreyvelich:
WIP until we merge this: https://github.com/kubeflow/katib/pull/1305, after that I will generate API clients, etc, but it would be great if you can start to review it.

I changed a bit API from proposal to support various success and failure conditions in CRD.
As I can see not all CRDs are using the same status approach with Conditions list as Kubeflow operators or BatchJobs. 
For example, [Argo](https://github.com/argoproj/argo/blob/master/pkg/apis/workflow/v1alpha1/workflow_types.go#L945) uses `.status.phase` to define whether Node is succeeded or failed.

To handle all of these cases, I proposed approach when user defines `SuccessCondition` and `FailureCondition` in Experiment.

This condition is [`gjson`](https://github.com/tidwall/gjson) expression which can directly point when job is succeeded or failed.

For example, for TFJob:
```
successCondition = "status.conditions.#(type=="Succeeded")#|#(status=="True")#"
failureCondition = "status.conditions.#(type=="Failed")#|#(status=="True")#"
```

For Argo Workflow:
```
successCondition = "status.[@this].#(phase=="Succeeded")"
failureCondition =  "status.[@this].#(phase=="Failed")"
```

I think this approach is extensible, since user can define any expression there when they want to fail Trial.
For example, when TFJob reaches 3 failed replicas.
`status.replicaStatuses.master.[@this].#(failed=="3")`

For Kubeflow operators or BatchJobs we can define these conditions in advance if we don't want to let user manually add them.

I renamed `UpdateTrialStatusCondition` to `UpdateTrialStatusConditionDeprecated`, since I had to refactor `UpdateTrialStatusCondition`.

In job_util.go I created `TrialJobStatus` which represents current Job status.

1. When Job is created/running status is Running.
2. If Job reaches successCondition status is Succeeded.
3. If Job reaches failureCondition status is Failed.

I also transfer Message and Reason from deployed Job to Trial if it is possible.

/cc @czheng94 @nielsmeima

### DIFF
--- a/pkg/controller.v1beta1/trial/trial_controller_util.go
+++ b/pkg/controller.v1beta1/trial/trial_controller_util.go
@@ -47,6 +47,7 @@ func (r *ReconcileTrial) UpdateTrialStatusCondition(instance *trialsv1beta1.Tria
 	}
 
 	timeNow := metav1.Now()
+
 	if jobStatus.Condition == trialutil.JobSucceeded {
 		if isTrialObservationAvailable(instance) {
 			msg := "Trial has succeeded "
@@ -54,10 +55,10 @@ func (r *ReconcileTrial) UpdateTrialStatusCondition(instance *trialsv1beta1.Tria
 
 			// Get message and reason from deployed job
 			if jobStatus.Message != "" {
-				msg = msg + ". Job message: " + jobStatus.Message
+				msg = fmt.Sprintf("%v. Job message: %v", msg, jobStatus.Message)
 			}
 			if jobStatus.Reason != "" {
-				reason = reason + ". Job reason: " + jobStatus.Reason
+				reason = fmt.Sprintf("%v. Job reason: %v", reason, jobStatus.Reason)
 			}
 
 			instance.MarkTrialStatusSucceeded(corev1.ConditionTrue, reason, msg)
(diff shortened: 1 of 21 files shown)


andreyvelich:
@gaocegege @johnugeorge @sperlingxx This PR is ready for review.

pkg/controller.v1beta1/trial/trial_controller_util.go
```diff
@@ -30,14 +30,92 @@ import (
...
+		eventMsg := fmt.Sprintf("Job %v is running", deployedJobName)
+		r.recorder.Eventf(instance, corev1.EventTypeNormal, JobRunningReason, eventMsg)
+		// TODO(gaocegege): Should we maintain a TrialsRunningCount?
+	}
+	// else nothing to do
+	return
+}
+
+// TODO (andreyvelich): Can be deleted after custom CRD is implemented
+func (r *ReconcileTrial) UpdateTrialStatusConditionDeprecated(instance *trialsv1beta1.Trial, deployedJob *unstructured.Unstructured, jobCondition *commonv1.JobCondition) {
```
gaocegege, line 118: Can we delete it now?

gaocegege (commented):
/assign @johnugeorge

pkg/controller.v1beta1/trial/trial_controller_util.go
andreyvelich: @gaocegege We can delete it after we finish Custom CRD implementation.
Controller still runs this function if successCondition and failureCondition is not set here: https://github.com/kubeflow/katib/blob/409b3a2676fa3971acd0850b3bc89655a0ca089e/pkg/controller.v1beta1/trial/trial_controller.go#L307

gaocegege (commented):
/lgtm

johnugeorge:
Please keep track of the cleanup required to remove the deprecated functions.

/approve

Issue #1314, opened 2020-08-25
TPE suggestion service failing to provide proposals for new trials


jimmy-hawkfish:
/kind bug

I am running an experiment with algorithm tpe and the following search space:

`[
  {
    "feasibleSpace": {
      "max": "1",
      "min": "0"
    },
    "name": "r1",
    "parameterType": "double"
  },
  {
    "feasibleSpace": {
      "max": "1",
      "min": "0"
    },
    "name": "r2",
    "parameterType": "double"
  },
  {
    "feasibleSpace": {
      "max": "1",
      "min": "0"
    },
    "name": "r3",
    "parameterType": "double"
  }
]`

and a `maxTrialCount` of 10,000. The experiment successfully runs 2,057 trials but fails to propose new ones. Looking at the suggestion pod logs I see:

`ERROR:grpc._server:Exception calling application: list index out of range
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/grpc/_server.py", line 434, in _call_behavior
    response_or_iterator = behavior(argument, context)
  File "/opt/katib/pkg/suggestion/v1alpha3/hyperopt/service.py", line 38, in GetSuggestions
    new_assignments = self.base_service.getSuggestions(trials, request.request_number)
  File "/opt/katib/pkg/suggestion/v1alpha3/hyperopt/base_service.py", line 220, in getSuggestions
    vals = new_trials[i]['misc']['vals']
IndexError: list index out of range`

**What did you expect to happen:**
I expected suggestions for the next 1000 trials to be proposed. It seems unlikely that the search space should be exhausted in ~2k trials.

**Environment:**
Running in Google Cloud Kubernetes engine
- Kubeflow version (`kfctl version`): 1.0.0
- Kubernetes version: (use `kubectl version`):
`Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #1315, opened 2020-08-26
Fix problem with Hyperopt Out of Range error

andreyvelich:
See comment: https://github.com/kubeflow/katib/issues/1314#issuecomment-681076525.
After this fix, we should not see `list index out of range` error.

/assign @gaocegege @sperlingxx 
/cc @jimmy-hawkfish


sperlingxx (approved):
/lgtm

gaocegege (commented):
/lgtm

PR #1316, opened 2020-08-27
Update CI test cluster version to 1.16

andreyvelich:
It seems that we can't create new GKE cluster with 1.14 version.

Ref: https://cloud.google.com/kubernetes-engine/docs/release-notes#august_27_2020_r28.
I update version to 1.16, KF serving [uses it](https://github.com/kubeflow/kfserving/blob/master/test/scripts/create-cluster.sh#L35) also.

/cc @yuzisun

### DIFF
--- a/test/scripts/v1alpha3/create-cluster.sh
+++ b/test/scripts/v1alpha3/create-cluster.sh
@@ -34,7 +34,7 @@ gcloud --project ${PROJECT} beta container clusters create ${CLUSTER_NAME} \
     --zone ${ZONE} \
     --machine-type=n1-standard-8 \
     --num-nodes=6 \
-    --cluster-version 1.14
+    --cluster-version 1.16
 echo "Configuring kubectl"
 gcloud --project ${PROJECT} container clusters get-credentials ${CLUSTER_NAME} \
     --zone ${ZONE}
--- a/test/scripts/v1beta1/create-cluster.sh
+++ b/test/scripts/v1beta1/create-cluster.sh
@@ -33,7 +33,7 @@ gcloud --project ${PROJECT} beta container clusters create ${CLUSTER_NAME} \
     --zone ${ZONE} \
     --machine-type=n1-standard-8 \
     --num-nodes=6 \
-    --cluster-version 1.14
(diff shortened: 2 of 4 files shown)


yuzisun:
looks like getting similar errors as KFServing was having eariler
```
W
Activating service-account
ERROR: (gcloud.auth.activate-service-account) There was a problem refreshing your current auth tokens: Unable to find the server at oauth2.googleapis.com
Please run:
 $ gcloud auth login
to obtain new credentials, or if you have already logged in with a
different account:
 $ gcloud config set account ACCOUNT
to select an already authenticated account to use.
```

andreyvelich:
@yuzisun Yes, sometime this error appears, I don't know why.
v1apha3 workflow was successful: http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-katib-presubmit-e2e-v1alpha3-1316-1e0684c-6208-8d16?tab=workflow

andreyvelich:
@yuzisun Did you fix it somehow for KF Serving ?

yuzisun:
> @yuzisun Did you fix it somehow for KF Serving ?

no, it is random, sometimes retry works, I think it is gcp issue.

andreyvelich:
@johnugeorge @gaocegege I added part to fix connection error in activate service account part.
The tests passed, should we try to merge this change?

johnugeorge:
Is this a gke bug of not getting activated?

andreyvelich:
@johnugeorge I think it is a GCP networking issue, KF Serving has the same: https://github.com/kubeflow/kfserving/pull/979#issuecomment-683411759

andreyvelich:
Instead of this adding try catch on Activate Service Account, we can try to add `retryStrategy` to Argo, like what they are trying to do in [KFServing](https://github.com/kubeflow/kfserving/pull/1063/files).

johnugeorge:
It looks much more clean right?

gaocegege:
I think it is better!

andreyvelich:
The test is passed with retry strategy. Can you approve it @gaocegege @johnugeorge ?

PR #1321, opened 2020-09-02
Switch test from kubeflow-ci to automl-ci project.

andreyvelich:
We should use `automl-ci` project to run our prow infra.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ jobs:
       script:
         - make check
         - make test
+      after_success:
         - goveralls -coverprofile=coverage.out
     - name: "Prettier frontend check"
       language: node_js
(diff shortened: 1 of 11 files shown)


andreyvelich:
I believe CI is working with the new project (automl-ci)
This PR blocked by: https://github.com/kubeflow/community-infra/pull/18, after that we should be able to restart test.

I added separate step to build images for file-metrics-collector and tfevent-metrics-collector in v1beta1 instead of running build in Katib controller script. It is better to create container for each building script, that's should speed up building part.

We also should refactor our scripts for e2e and builds since they are doing the same, but with different parameters.

andreyvelich:
The IAM was updated, we should be able to run test.

andreyvelich:
The tests are passed. I think we can finish migration to `automl-ci` project.
I changed GCP registry from `kubeflow-ci` to `automl-ci` for pre-submit to have everything related to test in one place.
Post-submit still pushes images to `kubeflow-images-public`

I am sure how we use this [params](https://github.com/kubeflow/katib/blob/master/test/workflows/components/params.libsonnet#L9-L23) since we parse only prow config here: https://github.com/kubeflow/testing/blob/master/images/run_workflows.sh#L17.
Do we need them?

/cc @gaocegege @johnugeorge @jlewi

gaocegege (commented):
/lgtm

Thanks!!

Issue #1320, opened 2020-09-02
Refactoring Supported Job List


andreyvelich:
/kind feature

After implementing custom CRD (https://github.com/kubeflow/katib/issues/1214), `SupportedJobList` parameter can be omitted since Katib supports any Kind of CRD.

In some cases users want to manually modify controller source code to Mutate Trial Job before submission, see discussion: https://github.com/kubeflow/katib/pull/1302#issuecomment-679452285 and https://github.com/kubeflow/katib/pull/1273#issuecomment-661533235.

Let's discuss here what is the best way to implement this in the controller.

/cc @sperlingxx @gaocegege @johnugeorge

PR #1333, opened 2020-09-10
Enhancement for Custom CRD

andreyvelich:
I was testing Custom CRD.
I was able to run Tekton `TaskRun` as Trial CR. I Will submit PR with few examples soon.
I found these problems while running:

- I disabled `MutateJob` in controller. We will track it in this issue: https://github.com/kubeflow/katib/issues/1320.

- I disabled validation for Job other than `SupportedJobList`.

- We should verify `PrimaryContainerName` in `MutateVolume` also.

- I changed `INSERT` in `RegisterObservationLog` function. It is better to insert all lines in one SQL query, instead of running separate query for each line. It helps to avoid unnecessary Trial updates.

- I added reconcile re-queue if metrics are not available. This use-case happens in Tekton job. Tekton task is succeeded once Training container is finished, but metrics can be not collected, yet. Because of that, controller reconciles Trial without observation and turns Trial status in `Metrics Unavailable`. To avoid it, I try to re-queue controller for  `maxRequeueCount` times to convert Trial to `Succeeded` status. @gaocegege @johnugeorge What do you think about this approach, can we do it in other way?

### DIFF
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -19,6 +19,7 @@ package trial
 import (
 	"context"
 	"fmt"
+	"time"
 
 	batchv1beta "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -45,6 +46,7 @@ import (
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/consts"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/trial/managerclient"
 	trialutil "github.com/kubeflow/katib/pkg/controller.v1beta1/trial/util"
+	"github.com/kubeflow/katib/pkg/controller.v1beta1/util"
 	jobv1beta1 "github.com/kubeflow/katib/pkg/job/v1beta1"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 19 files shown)


pkg/controller.v1beta1/trial/trial_controller.go
```diff
@@ -174,6 +175,12 @@ type ReconcileTrial struct {
 	collector *trialutil.TrialsCollector
 }
 
+// Map which contains number of requeuing for each trial if observation logs are not available
+// That is needed if Job is succeeded but metrics are not reported yet
+// Key = Trial name, value = requeue count
+var trialRequeueCount = make(map[string]int)
```
gaocegege, line 181: What will happen if the controller is restarted?

andreyvelich: @gaocegege If metrics are not available it tries to reconcile Trial for `maxRequeueCount` again after the restart.

gaocegege: OK, we are sure that there is no side effect right?

andreyvelich: Only one problem that I can see that `trialRequeueCount` is waisting some container resources, is that fine?

andreyvelich:
I figure out that we send all Trials to Suggestion. I think we don't need to send Trials with unavailable metrics to Suggestion service since we try to reconcile controller until metrics are reported.

gaocegege:
SGTM, we do not need to send these trials.

pkg/controller.v1beta1/trial/trial_controller.go
gaocegege: SGTM

andreyvelich:
@gaocegege We were discussing with @johnugeorge about my first solution with re-queuing controller and saving data in map.
It has few disadvantages.
1. We don't want to waste resources with this map and have problems with controller restarts as you mentioned before.
2. It is not a good way to monitor Reconcile restarts. We don't really know how many times Trial controller will be restarted and it depends on user's cluster.

I propose a bit different solution:
- If metrics collector doesn't find `objectiveMetricName` in the training logs, it reports `unavailable` value to the DB.
- If Trial CRD Job is succeeded, controller re-queues Trial object until DB records with appropriate trial name are empty.
- If DB has a record with `unavailable` metric value, Trial changes status to `MetricsUnavailable`

What do you think about this?

PR #1338, opened 2020-09-15
Remove istio sidecar annotation from all examples

andreyvelich:

I added `sidecar.istio.io/inject: "false"` to all examples.
Since we support custom CRD that should be added manually.


andreyvelich:
After discussion with @johnugeorge, we decided to delete istio annotation from the examples.
It makes easier for the user to read the example and this change requires only if Istio is used.

We should update documentation with this information, like we did for [PyTorch Job](https://www.kubeflow.org/docs/components/hyperparameter-tuning/hyperparameter/#pytorch-example).

/cc @gaocegege

gaocegege (commented):
/lgtm
/approve

Issue #1330, opened 2020-09-08
Early stopping implementation


andreyvelich:
/kind feature

To continue idea proposed here: https://github.com/kubeflow/katib/issues/692#issuecomment-626054238, we have to implement these steps:

1. I propose these API changes in [Suggestion](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/suggestions/v1beta1/suggestion_types.go):
```go

// SuggestionSpec defines the desired state of suggestion.
type SuggestionSpec struct {
	// Name of the algorithm that suggestion is used.
	AlgorithmName string `json:"algorithmName"`
       . . .
        // Name of the early stopping algorithm
        EarlyStoppingAlgorithmName string `json:"earlyStoppingAlgorithmName,omitempty"`
      . . .
}

// TrialAssignment is the assignment for one trial.
type TrialAssignment struct {
	// Suggestion results
	ParameterAssignments []common.ParameterAssignment `json:"parameterAssignments,omitempty"`

	//Name of the suggestion
	Name string `json:"name,omitempty"`

        // Parameters for early stopping techniques
	// Contains parameter name, value and comparison type
	EarlyStoppingRules []common.EarlyStoppingRule `json:"earlyStoppingRules,omitempty"`
}
```
I propose these API changes in [Trial](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/trials/v1beta1/trial_types.go):
```go
type TrialSpec struct {
         . . . 
	// Key-value pairs for hyperparameters and assignment values.
	ParameterAssignments []common.ParameterAssignment `json:"parameterAssignments"`

	// Parameters for early stopping techniques
	// Contains parameter name, value and comparison type
	EarlyStoppingRules []common.EarlyStoppingRule `json:"earlyStoppingRules,omitempty"`
        . . .
}
```
These API changes in [Common](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/common/v1beta1/common_types.go):
```go
type EarlyStoppingRule struct {
	Name       string         `json:"name,omitempty"`
	Value      string         `json:"value,omitempty"`
	Comparison ComparisonType `json:"comparison,omitempty"`
}

type ComparisonType string

const (
	Equal   ComparisonType = "equal"
	Less    ComparisonType = "less"
	Greater ComparisonType = "greater"
)
```

[...]

PR #1344, opened 2020-09-18
Early Stopping Implementation

andreyvelich:
Related: https://github.com/kubeflow/katib/issues/1330.

I started Early Stopping implementation with API changes.
@gaocegege @johnugeorge Please let me know what do you think about these APIs.

I will work on the controller changes in this PR also.


andreyvelich:
@gaocegege @johnugeorge 
I finished first step for EarlyStopping implementation. For provided example it's working:
Screenshot 2020-10-09 at 04 04 47

I still need to add/fix unit test and verify some e2e tests. In the meantime, can you start to review it, please ?
I also blocked by: https://github.com/kubeflow/katib/pull/1333, which helps me to test Tekton, MPI Jobs.

Few notes:

1. I made several changes in manager and controller APIs. Please, take a look and let me know about any thoughts/changes.
1. For Algorithm service I chose `port: 6789, portName: suggestion-api`. For EarlyStopping service I chose `port: 6788, portName: earlystop-api`.
1. My suggestion is to take Early Stopping data (image, resources, etc.. ) from the `katib-config`. We should do it in follow PR. Currently, I take [temp data](https://github.com/kubeflow/katib/pull/1344/files?file-filters%5B%5D=.go#diff-8a2a80b76ef44a6f38a2662e840c9d9dR219).
1. I didn't add median stopping rule in this PR. We can submit follow PR with the implementation. Currently, it simple stops Trials with `accuracy < 0.8` and `Epoch = 9`. Both condition should be met.
1. As mentioned here: https://github.com/kubeflow/katib/issues/1330#issuecomment-691229315, I attached `RBAC` to Suggestion deployment to change Trial status from Early Stopping service.
1. I use filters from `metricsFormat` field to parse early stopping rules metrics . You can check it [here](https://github.com/kubeflow/katib/pull/1344/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml#diff-e6f04cda420086f92536adc2eb1f9815R27-R28).
1. I made few changes in metrics collector command. Currently, the training command looks like this: 
```
python3 /opt/mxnet-mnist/mnist.py --num-epochs=10 1>/var/log/katib/metrics.log 2>&1 || 
      if test -f /var/log/katib/$$$$.pid && [ $(head -n 1 /var/log/katib/$$.pid) = early-stopped ]; then 
        echo Training Container was Early Stopped; 
      else 
        echo Training Container was Failed; 
        exit 1;
      fi && echo completed > /var/log/katib/$$$$.pid
```

If Python command is failed, it checks that `7.pid` file exists (if main proc = 7) and it contains `early-stopped` line.
If yes, Pod is completed. Otherwise, Pod is Failed with exit code 1.
This approach uses only one `.pid` file to monitor current status (`early-stopped` or `completed`)

So, the metrics collector workflow (with main proc PID = 7) is that:
1. Wait until all stop-rules are reached.
1. Create `7.pid` file with `early-stopped` line (after that  `&& echo completed > /var/log/katib/$$$$.pid` runs).
1. Kill the child process for 7 PID (Support only single child process).
1. Report metrics to the DB.
1. Dial EarlyStopping service with `SetTrialStatus` request to update Trial status.

Let me know what do you think.

andreyvelich:
After discussion with @johnugeorge, I add EarlyStopping calls (`GetEarlyStoppingRules` and `SetTrialStatus`) to `Suggestion` RPC service.
I think it should be clearer to understand because EarlyStopping is just another container under Suggestion's deployment.
It uses the same Kubernetes Service to expose it, but just a different ports.

/cc @gaocegege

gaocegege (commented):
Generally LGTM

andreyvelich:
After discussion with @johnugeorge, we decide to make few changes in APIs. To be between AutoML Algorithm and Early Stopping Experiment and Suggestion defines as:

```go

type ExperimentSpec struct {
	Algorithm     *common.AlgorithmSpec     `json:"algorithm,omitempty"`
	EarlyStopping *common.EarlyStoppingSpec `json:"earlyStopping,omitempty"`
}

type SuggestionSpec struct {
	Algorithm     *common.AlgorithmSpec     `json:"algorithm"`
	EarlyStopping *common.EarlyStoppingSpec `json:"earlyStopping,omitempty"`
}

type AlgorithmSpec struct {
	AlgorithmName     string             `json:"algorithmName,omitempty"`
	AlgorithmSettings []AlgorithmSetting `json:"algorithmSettings,omitempty"`
}

type AlgorithmSetting struct {
	Name  string `json:"name,omitempty"`
	Value string `json:"value,omitempty"`
}

type EarlyStoppingSpec struct {
	AlgorithmName     string                 `json:"algorithmName,omitempty"`
	AlgorithmSettings []EarlyStoppingSetting `json:"algorithmSettings,omitempty"`
}

type EarlyStoppingSetting struct {
	Name  string `json:"name,omitempty"`
	Value string `json:"value,omitempty"`
}

```

Original AutoML algorithm setting is located in `Suggestion.Spec.Algorithm.AlgorithmSettings`, modified AutoML algorithm settings is located in `Suggestion.Status.AlgorithmSettings`.

@gaocegege @johnugeorge What do you think about this APIs?

gaocegege:
LGTM The new API is clean.

andreyvelich:
I finished Median stop implementation, @gaocegege @johnugeorge Can you take a look, please?

I made few changes according to the paper: https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46180.pdf.
1. Add `start_step` parameter as fourth parameter in `-stop-rule` flag. Flag looks like this: 
`-stop-rule accuracy;0.8;less;4`. We start to apply the early stopping rule only after 4 metrics were reported. (This start step is optional parameter, without it we apply early stopping rule from the first reported metric).
1. `min_trials_required` setting - quantity of succeeded Trials before calculate median, `start_step` - quantity of reported metrics before applying rule.
1. Send `ObjectiveType` to the sidecar in`-o-type` flag. According to the paper, we should compare the best objective value at step S. So for the Objective metric I calculate only best optimal value and apply early stopping rule. Later we can add possibilities to compare `latest`, `min` or `max` value, if it's needed.

andreyvelich:
@johnugeorge @gaocegege If you don't have any objections on this PR, can you give your lgtm and we can make other changes in the following PRs?

gaocegege:
/lgtm

Tried it yesterday.

gaocegege:

PR #1355, opened 2020-10-15
Clean up code for custom CRD

andreyvelich:
I removed redundant code and fixed tests for new custom CRD implementation.

As well, I added some default values for `Job`, `PyTorchJob` and `TFJob` to make examples easier for the first look, as we discussed before.

In particular, for `Job`:
1. Set default `SuccessCondition` and `FailureCondition`.
I didn't add `PrimaryPodLabels` to `Job` examples since each `Job` has only one pod replica. If `PrimaryPodLabels` is omitted, metrics collector injected to all created pods.

For Kubeflow Job (`PytorchJob` and `TFJob`).
1. Set default `SuccessCondition` and `FailureCondition`.
2. Set default `PrimaryPodLabels`.

I manually added `PrimaryContainerName` to all examples. I think it is very easy to understand and user can explicitly see which container will be wrapped by metrics collector.

I can't set default values if Trial Template is set in ConfigMap, since it requires k8s client. We can think about it later.

I left validation for `Job`, `TFJob` and `PyTorchJob` for now. @gaocegege @johnugeorge What do you think about it ?

I deleted `job/v1beta1` since it is not used. Let's continue discussion here: https://github.com/kubeflow/katib/issues/1320 how we should implement `MutateJob`.


gaocegege (commented):
/lgtm

gaocegege (approved):
/lgtm

PR #1368, opened 2020-10-29
Remove default values for Trial template

andreyvelich:
I think it's better to remove default values for the `spec.trialTemplate.TrialSpec`.
It's make very unclear for the user that Experiment can be created without `trialTemplate` and which `spec.parameters` should be set.

WDYT @gaocegege @johnugeorge  ?


gaocegege (approved):
/lgtm
/approve

SGTM it is weird.

Issue #1367, opened 2020-10-26
AutoML WG needs to take ownership of its release infrastructure, process, and artifacts


jlewi:
Opening this issue to discuss the possibility of the @kubeflow/wg-automl-leads creating their own GitHub org.

Per kubeflow/testing#737 our current approach to relying on shared infrastructure is not scalable or resilient. I think to continue to grow each WG will need to position itself to take on more responsibility for its infrastructure.

This has come up recently for Katib in terms of both its release/test infra and now a Docker registry for publishing Katib images.

jlewi:
Related issue: kubeflow/katib#1332

PR #1371, opened 2020-10-30
Add Katib release script

andreyvelich:
I added release script to manually push images to `docker.io/kubeflowkatib` registry.
The image tag is `:v1beta1-<commit-SHA>`.
For example:
```
docker.io/kubeflowkatib/katib-controller:v1beta1-d5f5aa0b
```

Once we merge it, I will run the script to update the images.
Related: https://github.com/kubeflow/katib/issues/1367.

As well, I removed legacy release-workflow since we can't use `kubeflow-images-public` more.

/cc @jlewi


PatrickXYS:
After this PR merged, since Katib is not using Kubeflow Presubmit to release image.

We can also remove from [kubernetes/test-infra ](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubeflow/kubeflow-postsubmits.yaml#L148-L160)

andreyvelich:
> After this PR merged, since Katib is not using Kubeflow Presubmit to release image.
> 
> We can also remove from [kubernetes/test-infra ](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubeflow/kubeflow-postsubmits.yaml#L148-L160)

@PatrickXYS Yes, GCP prow post-submit is not needed to Katib more.

PR #1375, opened 2020-11-03
Add example with ConfigMap source for the Trial template

andreyvelich:
I added example for the ConfigMap source of Trial template.
I think, it's useful to point user to this example in the new trial template documentation.


gaocegege (commented):
/lgtm
/approve

PR #1384, opened 2020-11-09
select non-root process by user id instead of parent

### LINKED ISSUE #1383: Skip non-root user processes when waiting for main proces to end
/kind bug

Metrics cannot be collected while running trials with Istio enabled. The metrics pod keeps running without sending metrics to the db-manager. After debugging I found out that the function `GetMainProcesses()` waits forever for the Istio sidecar (envoy proxy) to finish. The istio sidecar does not run as root and should therefore be skipped in this proces. The parent process of istio however is the root user (which determines if a process is ran as root). We can check for user id instead of pid parent.

The istio sidecar should not be considered while checking for the main process to be finished.

I will make a PR with a fix.

**Environment:**

- Kubeflow version 1.0
- Minikube version (`minikube version`):
- Kubernetes version: (use `kubectl version`): 1.18.10
- OS (e.g. from `/etc/os-release`):

robbertvdg:
Enables running trials with Istio enabled, and any other process that run as non root.

Fixes #1383 

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/metricscollector/v1beta1/common/pns.go
+++ b/pkg/metricscollector/v1beta1/common/pns.go
@@ -69,14 +69,14 @@ func GetMainProcesses(completedMarkedDirPath string) (map[int]bool, int, error) 
 			return nil, 0, fmt.Errorf("Failed to create new Process from pid %v, error: %v", pid, err)
 		}
 
-		// Get parent process
-		ppid, err := proc.Ppid()
+		// Get process uids
+		pidUids, err := proc.Uids()
 		if err != nil {
-			return nil, 0, fmt.Errorf("Unable to get parent pid for pid: %v, error: %v", ppid, err)
+			return nil, 0, fmt.Errorf("Unable to get uids for pid: %v, error: %v", pidUids, err)
 		}
 
-		// Ignore the pause container, our own pid, and non-root processes (parent pid != 0)
-		if pid == 1 || pid == int32(thisPID) || ppid != 0 {
+		// Ignore the pause container, our own pid, and non-root processes (first pid user id != 0)
+		if pid == 1 || pid == int32(thisPID) || pidUids[0] != 0 {
 			continue
 		}
 


andreyvelich:
Thank you for this PR @robbertvdg!

1. If the metrics collector doesn't need to wait until Istio sidecar is finished, have you tried to set 
[`waitAll `](https://github.com/kubeflow/katib/blob/master/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L110) flag to false ? In that case, metrics collector starts after the training process is finished.

2. Do you know why the parent process of the istio sidecar is the root user ? Does some other PID execute the istio sidecar process ?

robbertvdg:
Hi @andreyvelich, the waitAll option seems to be what I need. I am struggling to configure it, do you have an example of how to set it?

> Do you know why the parent process of the istio sidecar is the root user ? Does some other PID execute the istio sidecar process ?
It is about the parent in the process tree of istio. The process parent is `init`, so it is a direct descendent of the startup process.

andreyvelich:
> Hi @andreyvelich, the waitAll option seems to be what I need. I am struggling to configure it, do you have an example of how to set it?

Unfortunately, currently we don't specify [the `-w` flag](https://github.com/kubeflow/katib/blob/master/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L110) in the [`getMetricsCollectorArgs`](https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/pod/inject_webhook.go#L291) for the metrics collector args.

I think, [Katib config](https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/katib-controller/katib-config.yaml#L10) is the good location to set this options for the metrics collector.
Since we already receive some [metrics collector data](https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/pod/inject_webhook.go#L215) from this config.

For example, we can extend [`MetricsCollectorConfig`](https://github.com/kubeflow/katib/blob/master/pkg/util/v1beta1/katibconfig/config.go#L30-L34) with the `WaitAllProcesses` parameter which is `true` or `false`.

It is useful for many use-cases.

What do you think @robbertvdg @gaocegege @johnugeorge  ?

robbertvdg:
I just tested the waitAll option in a fork, and that does the trick! The [MetricsCollectorConfig](https://github.com/kubeflow/katib/blob/master/pkg/util/v1beta1/katibconfig/config.go#L30-L34)  seems like a logical place, since it will be cluster dependent. I will close this MR and the corresponding issue, will you pick up the extension of the config if the others agree to the solution?

andreyvelich:
@robbertvdg It would be great if you can create PR with the updated Config to add this feature in Katib.

What do you think about the parameter name `WaitAllProcesses` @gaocegege @johnugeorge ?

robbertvdg:
@andreyvelich Will do, it will take me a bit longer since i'm not too familiar with the code.

andreyvelich:
@robbertvdg Thank you, no worries. Ping me on [Kubeflow Slack](https://kubeflow.slack.com/archives/DD5SB4Q1M) for any questions/help.

Issue #1385, opened 2020-11-09
Metrics Collector verifies one more metric after Early Stopping step


andreyvelich:
/kind bug

[Here](https://github.com/kubeflow/katib/blob/master/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L233) we should verify that `metricStartStep[metricName]` is not equal to `0`.
We have to compare metrics **at Step S**, but we make additional step before applying Early Stopping rule.

PR #1386, opened 2020-11-10
Fix compare step for the early stopping

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1385.
We should not wait one more step to compare values.
As well, I removed unnecessary params from the early stopping yaml.


gaocegege (approved):
LGTM :+1:
/lgtm

PR #1388, opened 2020-11-12
Remove new Trial kind doc

andreyvelich:
I removed new Trial template doc since it's outdated.
How to add new Trial kind can be found here: https://www.kubeflow.org/docs/components/katib/trial-template/#use-custom-kubernetes-resource-as-a-trial-template.

I will update the README in a separate PR.


gaocegege (commented):
LGTM :+1:
/lgtm

PR #1390, opened 2020-11-16
Add Stale config to close inactivity issues

andreyvelich:
I've added Stale config so we can better track inactive issues in Katib.

/cc @jlewi


gaocegege (commented):
/lgtm

PR #1391, opened 2020-11-16
Move Adopters file

andreyvelich:
I moved Adopters file to the root and more users can see it.
Also, let's keep the list in alphabetical order.


gaocegege (commented):
LGTM :+1:
/lgtm

PR #1392, opened 2020-11-17
Update docs for Katib 0.10

andreyvelich:
I've updated docs for Katib 0.10.
1. I removed `algorithm-settings.md`, `quick-start.md` and `user-guide.md`. I think we don't need these docs, since all actual info can be found in https://www.kubeflow.org/docs/components/katib/. What do you think @gaocegege @johnugeorge ?
2. I left only `workflow-design.md` doc since it has few important steps for Katib developers.
3. I modified README doc, please take a look.

Do we want to have [Katib concepts](https://github.com/kubeflow/katib#concepts-in-katib) under README, since we have them in [Kubeflow docs](https://www.kubeflow.org/docs/components/katib/overview/#katib-concepts) ?
The same question for the [Web UI](https://github.com/kubeflow/katib#web-ui), [Installation](https://github.com/kubeflow/katib#installation), etc. ?

What do we want to see in the our README ?


gaocegege (commented):
LGTM :+1:
/lgtm

andreyvelich:
@gaocegege @johnugeorge Can you check the README changes, please?

PR #1394, opened 2020-11-18
Feature/waitallprocesses config

robbertvdg:
Adds the waitAllProcesses option to the katib-config metricsCollector spec. It is made optional, defaulting to true.

See #1383 for discussion

**Release note**:

```release-note
Available waitAllProcesses option to katib-config's metrics-collector-sidecar spec.
```

### DIFF
--- a/cmd/metricscollector/v1beta1/file-metricscollector/main.go
+++ b/cmd/metricscollector/v1beta1/file-metricscollector/main.go
@@ -353,11 +353,7 @@ func main() {
 		go printMetricsFile(*metricsFilePath)
 	}
 
-	waitAll, err := strconv.ParseBool(*waitAllProcesses)
-	if err != nil {
-		klog.Errorf("Cannot parse %s to bool, defaulting to waitAllProcesses=%s", *waitAllProcesses, common.DefaultWaitAllProcesses)
-		waitAll, _ = strconv.ParseBool(common.DefaultWaitAllProcesses)
-	}
+	waitAll, _ := strconv.ParseBool(*waitAllProcesses)
 
 	wopts := common.WaitPidsOpts{
 		PollInterval:           *pollInterval,
--- a/cmd/metricscollector/v1beta1/file-metricscollector/main.go
+++ b/cmd/metricscollector/v1beta1/file-metricscollector/main.go
@@ -107,7 +107,7 @@ var (
 	metricFilters        = flag.String("f", "", "Metric filters")
 	pollInterval         = flag.Duration("p", common.DefaultPollInterval, "Poll interval between running processes check")
 	timeout              = flag.Duration("timeout", common.DefaultTimeout, "Timeout before invoke error during running processes check")
-	waitAll              = flag.Bool("w", common.DefaultWaitAll, "Whether wait for all other main process of container exiting")
+	waitAllProcesses     = flag.String("w", common.DefaultWaitAllProcesses, "Whether wait for all other main process of container exiting")
 	stopRules            stopRulesFlag
 	isEarlyStopped       = false
 )
@@ -353,10 +353,16 @@ func main() {
 		go printMetricsFile(*metricsFilePath)
 	}
 
+	waitAll, err := strconv.ParseBool(*waitAllProcesses)
+	if err != nil {
+		klog.Errorf("Cannot parse %s to bool, defaulting to waitAllProcesses=%s", *waitAllProcesses, common.DefaultWaitAllProcesses)
+		waitAll, _ = strconv.ParseBool(common.DefaultWaitAllProcesses)
+	}
+
 	wopts := common.WaitPidsOpts{
 		PollInterval:           *pollInterval,
 		Timeout:                *timeout,
-		WaitAll:                *waitAll,
+		WaitAll:                waitAll,
 		CompletedMarkedDirPath: filepath.Dir(*metricsFilePath),
 	}
 	if err := common.WaitMainProcesses(wopts); err != nil {
--- a/cmd/metricscollector/v1beta1/tfevent-metricscollector/main.py
+++ b/cmd/metricscollector/v1beta1/tfevent-metricscollector/main.py
@@ -38,7 +38,7 @@ if __name__ == '__main__':
     logger.addHandler(handler)
     logger.propagate = False
     opt = parse_options()
-    wait_all_processes = opt.wait_all_processes.lower() != "false"
+    wait_all_processes = opt.wait_all_processes.lower() == "true"
     db_manager_server = opt.db_manager_server_addr.split(':')
     if len(db_manager_server) != 2:
         raise Exception("Invalid Katib DB manager service address: %s" %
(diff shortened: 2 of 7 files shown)


robbertvdg:
@andreyvelich can you take a look at this?

cmd/metricscollector/v1beta1/file-metricscollector/main.go
```diff
@@ -107,7 +107,7 @@ var (
 	metricFilters        = flag.String("f", "", "Metric filters")
 	pollInterval         = flag.Duration("p", common.DefaultPollInterval, "Poll interval between running processes check")
 	timeout              = flag.Duration("timeout", common.DefaultTimeout, "Timeout before invoke error during running processes check")
-	waitAll              = flag.Bool("w", common.DefaultWaitAll, "Whether wait for all other main process of container exiting")
+	waitAllProcesses     = flag.String("w", common.DefaultWaitAllProcesses, "Whether wait for all other main process of container exiting")
```
andreyvelich, line 110: Do we need to keep this flag `String`, or we can use `Bool`?
I assume this flag can be only `true` or `false`.

cmd/metricscollector/v1beta1/file-metricscollector/main.go
```diff
@@ -353,10 +353,16 @@ func main() {
 		go printMetricsFile(*metricsFilePath)
 	}
 
+	waitAll, err := strconv.ParseBool(*waitAllProcesses)
```
andreyvelich, line 356: If we keep it `Bool`, we can skip conversion.

pkg/util/v1beta1/katibconfig/config.go
```diff
@@ -28,9 +28,10 @@ type SuggestionConfig struct {
 
 // MetricsCollectorConfig is the JSON metrics collector structure in Katib config.
 type MetricsCollectorConfig struct {
-	Image           string                      `json:"image"`
-	ImagePullPolicy corev1.PullPolicy           `json:"imagePullPolicy"`
-	Resource        corev1.ResourceRequirements `json:"resources"`
+	Image            string                      `json:"image"`
+	ImagePullPolicy  corev1.PullPolicy           `json:"imagePullPolicy"`
+	Resource         corev1.ResourceRequirements `json:"resources"`
+	WaitAllProcesses string                      `json:"waitAllProcesses"`
```
andreyvelich, line 34: Same as above, can we keep it `bool` ?

andreyvelich (commented):
Thank you for doing this @robbertvdg!
I left few comments.
/ok-to-test
/cc @gaocegege @johnugeorge

cmd/metricscollector/v1beta1/file-metricscollector/main.go
robbertvdg: I had problems with parsing booleans when a user has not specified the field in the katib-config. Do you know of a good way to do it?
I could leave it string in the config and parse it it in `GetMetricsCollectorConfigData`, would that be better?

andreyvelich: @robbertvdg It seems that empty `bool` in Go is `false`. So when user doesn't specify any values in config we send "false" to the metrics collector.
Maybe we can define `MetricsCollectorConfig` like this:
```go
type MetricsCollectorConfig struct {
	Image            string                      `json:"image"`
	ImagePullPolicy  corev1.PullPolicy           `json:"imagePullPolicy"`
	Resource         corev1.ResourceRequirements `json:"resources"`
	WaitAllProcesses *bool                      `json:"waitAllProcesses"`
}
```
In that case, you can set `-w` flag if `WaitAllProcesses != nil`. Otherwise the default value is set.

In Katib config I was able to define bool value like this:
```json
 "StdOut": {
        "image": "docker.io/andreyvelichkevich/file-metrics-collector",
        "imagePullPolicy": "Always",
        "waitAllProcesses": false
      },
```

robbertvdg: The problem lies with the `v1.Container.args` which is `string[]`. Therefore every container argument has to be a string, which I think is not compatible with the flag.Bool() argument (I tested it, doesn't work).

pkg/util/v1beta1/katibconfig/config.go
```diff
@@ -28,9 +28,10 @@ type SuggestionConfig struct {
 
 // MetricsCollectorConfig is the JSON metrics collector structure in Katib config.
 type MetricsCollectorConfig struct {
-	Image           string                      `json:"image"`
-	ImagePullPolicy corev1.PullPolicy           `json:"imagePullPolicy"`
-	Resource        corev1.ResourceRequirements `json:"resources"`
+	Image            string                      `json:"image"`
```
gaocegege, line 31: Should we add omitempty for the fieds?

cmd/metricscollector/v1beta1/file-metricscollector/main.go
andreyvelich: @robbertvdg Got it.
In that case can we keep `WaitAllProcesses *bool` in `MetricsCollectorConfig`, convert it to string here: 
https://github.com/kubeflow/katib/blob/3f5d81302a9ff2f43c30ab70bf9c6d0bc49b525e/pkg/webhook/v1beta1/pod/inject_webhook.go#L299
Then, use the String type like you did in [Metrics Collector](https://github.com/kubeflow/katib/blob/3f5d81302a9ff2f43c30ab70bf9c6d0bc49b525e/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L110) and convert it to Bool after.

In that case, we can avoid situation when user specify wrong values in Katib config.

What do you think @robbertvdg @gaocegege ?

pkg/util/v1beta1/katibconfig/config.go
andreyvelich: I think we can add it on all parameters but `Image`.

cmd/metricscollector/v1beta1/file-metricscollector/main.go
robbertvdg: @andreyvelich those were my thoughts as well, will update PR soon.

robbertvdg:
@andreyvelich @gaocegege updated PR, can you do a final check?

pkg/util/v1beta1/katibconfig/config.go
```diff
@@ -28,9 +28,10 @@ type SuggestionConfig struct {
...
+	ImagePullPolicy  corev1.PullPolicy           `json:"imagePullPolicy,omitempty"`
+	Resource         corev1.ResourceRequirements `json:"resources,omitempty"`
+	WaitAllProcesses *bool                       `json:"waitAllProcesses,omitempty"`
```
andreyvelich, line 34: Can you add `omitempty` to the `SuggestionConfig` also please ? It should be in all parameters but `Image`.

cmd/metricscollector/v1beta1/tfevent-metricscollector/main.py
```diff
@@ -38,6 +38,7 @@ def parse_options():
     logger.addHandler(handler)
     logger.propagate = False
     opt = parse_options()
+    wait_all_processes = opt.wait_all_processes.lower() == "true"
```
andreyvelich, line 41: Does this compare (`opt.wait_all_processes.lower() == "true"`) necessary ?
I assume `wait_all_processes` can be only "true" or "false"  since `WaitAllProcesses *bool` in Katib config?

robbertvdg: this is the python way of converting string to bool ;) should be fine since the string can only have two possibilities

andreyvelich (approved):
Thank you for your contribution @robbertvdg!
/lgtm
/cc @gaocegege

gaocegege:

PR #1252, opened 2020-07-03
[python sdk] add v1beta1 models

sperlingxx:
1.  Add replace v1alpha3 python models with v1beta1 python models.
2. Provide scripts to update python SDK.
3. Change default katib version in python SDK  from v1alpha3 to v1beta1.
4. Upgrade python SDK version to 0.0.3. 

What's your opinion about this PR, @prem0912 @andreyvelich @gaocegege @johnugeorge ?

PR #1396, opened 2020-11-18
Remove v1alpha3 version

andreyvelich:
Blocked by: https://github.com/kubeflow/katib/pull/1392.
I removed v1alpha3 version since we use the v1beta1 in Kubeflow manifests.
As well, I removed unused Go packages from `Gopkg.toml`.

For some reason, `v1_time.py` and `v1_unstructured_unstructured.py` is not generated for the v1beta1 SDK. @jinchihe I saw your comment here: https://github.com/kubeflow/katib/pull/1252#discussion_r450818372.
Do you know how we can include them to the SDK ?
Can we use [Swagger Import Mappings](https://github.com/kubeflow/katib/blob/master/hack/gen-python-sdk/swagger_config.json#L5) for it ?


andreyvelich:
I solved above problem by modifying `swagger_config.json` with this:
```json
 "typeMappings": {
    "V1Time": "datetime",
    "V1UnstructuredUnstructured": "dict()"
  }
```
I checked our Notebook example and it is working.
What do you think @jinchihe  ?
I believe Kubernetes Python client follows the same way for the date:

https://github.com/kubernetes-client/python/blob/f461dd42d48650a4ae1b41d630875cad9fcb68ad/kubernetes/client/models/apps_v1beta1_deployment_condition.py#L36-L37

andreyvelich:
After some tests, I think it's better to use `object` type instead of `dict()` for the unstructured. `"V1UnstructuredUnstructured": "object"`.
K8s Python lib also follows this type for some attributes, for example [here](https://github.com/kubernetes-client/python/blob/master/kubernetes/client/models/v1_watch_event.py#L36).

In that case, we can use [`deserialize`](https://github.com/kubeflow/katib/blob/9f5f65bdf8fe9c172efc88b4993ded1a5d060c01/sdk/python/v1beta1/kubeflow/katib/api_client.py#L221) method to convert object from the dict to Katib model. 
I think, we should modify our API to return Katib model objects instead of simple dict so user can easily manipulate them.
What do you think @gaocegege @johnugeorge @jinchihe ?

As well, it is useful in KFP Katib launcher. We can use default mechanism of serialisation de-serialisation instead of writing our own function, like I did here: https://github.com/kubeflow/pipelines/blob/7f101c5e41f1e9cef02e6a94225e97815fc2327f/components/kubeflow/katib-launcher/src/launch_experiment.py#L63-L193.

PR #1404, opened 2020-11-30
Fix Early Stopped Trials in Goptuna Suggestion

andreyvelich:
In Goptuna Suggestion we should transfer the `EarlyStopped` Trials to `Pruned`.
Is that correct @c-bata ?

Also, I added additional info in errors.

/assign @c-bata @gaocegege @johnugeorge

### DIFF
--- a/pkg/suggestion/v1beta1/goptuna/converter.go
+++ b/pkg/suggestion/v1beta1/goptuna/converter.go
@@ -1,7 +1,7 @@
 package suggestion_goptuna_v1beta1
 
 import (
-	"errors"
+	"fmt"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


gaocegege (commented):
/lgtm

c-bata:
> In Goptuna Suggestion we should transfer the EarlyStopped Trials to Pruned.
> Is that correct @c-bata ?

Correct 👍  Thanks you!

---

This is just a nit-picking comment but it is recommended to not capitalize the error string.

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.
> https://github.com/golang/go/wiki/CodeReviewComments#error-strings

andreyvelich:
> This is just a nit-picking comment but it is recommended to not capitalize the error string by Golang wiki.
> 
> > Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.
> > https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Got it, thank you for this info.
Sorry for these changes.

Issue #1399, opened 2020-11-25
Consider "using" duecredit


yarikoptic:
I saw katib listed among duecredit dependents: https://github.com/duecredit/duecredit/network/dependents?package_id=UGFja2FnZS01MjAzODE5OA%3D%3D  but apparently it is only present in the "frozen" requirements.txt specs.  So thought to alert that you might as well consider "using" it and provide citation , see https://github.com/duecredit/duecredit#duecredit for more info.

andreyvelich:
Hi @yarikoptic.
Thank you to point this.

@gaocegege @johnugeorge Do we need this package in our Suggestions ?
I think we can remove it, I was able to generate citation only for the `numpy` package.

gaocegege:
I have no objection to it.

Thanks

PR #1406, opened 2020-12-02
Remove duecredit pkg from the Suggestions

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1399.

I think we don't need `duecredit` package in our Suggestions.


gaocegege (commented):
/lgtm

PR #1409, opened 2020-12-09
Bump Prettier to 2.2.0 for the Katib UI

andreyvelich:
We should update the Prettier version for our Katib UI.
Even the [VSCode plugin](https://github.com/prettier/prettier-vscode) is using Prettier >= 2.0


gaocegege (commented):
/lgtm

PR #1414, opened 2021-01-07
Add SVG logo traced from bitmap logo

knkski:

Adds SVG logo traced from bitmap logo, for use in contexts that require vector graphics. Includes both just logo, and logo with "Katib" title below it. Also symlinks `icon.svg` to the new logo file for each charm.


docs/images/logo-title.svg
```diff
@@ -0,0 +1,17 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="762" width="548">
```
andreyvelich, line 1: Should we use your new logo instead of https://github.com/kubeflow/katib/blob/master/docs/images/Katib_Logo.png to be consistent ?

knkski:
I updated the PNG logo as well. I switched to the Urbanist font vs Calibri, as Calibri is not licensed for use on non-Windows machines. I chose Urbanist after looking around for fonts that had the same clean, straight design, but can switch it to another font if desired.

docs/images/logo-title.svg
andreyvelich, line 1: I think you should update the README link as well: https://github.com/kubeflow/katib/blame/master/README.md#L2.

knkski: Updated that link, and also added pngs of both SVG files as `logo.png` and `logo-title.png`, to match the SVG filenames.

For future reference, making the pngs was pretty easy, if they ever get edited again:

    cd docs/images/
    inkscape logo.svg -o logo.png
    inkscape logo-title.svg -o logo-title.png

gaocegege:
/lgtm

andreyvelich (approved):
Thank you for doing this @knkski!
/approve

Issue #1395, opened 2020-11-18
x509: certificate relies on legacy Common Name field


tenzen-y:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
1. Download release 0.10.0.
2. I installed Katib using the following commands.
```bash scripts/v1beta1/deploy.sh``` 
3. I have started the sample experiment by running the following commands.
```
$ kubectl apply -f example/v1beta1/tpe-example.yaml
Error from server (InternalError): error when creating "/Users/tenzen/katib-0.10.0/examples/v1beta1/tpe-example.yaml": Internal error occurred: failed calling webhook "mutating.experiment.katib.kubeflow.org": Post "https://katib-controller.kubeflow.svc:443/mutate-experiments?timeout=30s": x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0
```

**What did you expect to happen:**
I expected the experiment to start.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

I found related topics in the release note for 1.19.0
https://kubernetes.io/docs/setup/release/notes/
> Kubernetes is now built with golang 1.15.0-rc.1.
The deprecated, legacy behavior of treating the CommonName field on X.509 serving certificates as a host name when no Subject Alternative Names are present is now disabled by default. It can be temporarily re-enabled by adding the value x509ignoreCN=0 to the GODEBUG environment variable. (#93264, @justaugustus) [SIG API Machinery, Auth, CLI, Cloud Provider, Cluster Lifecycle, Instrumentation, Network, Node, Release, Scalability, Storage and Testing]

**Environment:**

- Kubeflow version (`kfctl version`): None
- Minikube version (`minikube version`): None
- Kubernetes version: (use `kubectl version`): 1.19.4
- OS (e.g. from `/etc/os-release`): ubuntu16.04(4.15.0-72-generic)

andreyvelich:
Hi @tenzen-y, thank you for creating this.

I think this problem happens because K8s 1.19 is using Go 1.15.
Which cluster are you using: GCP, AWS, on-prem ?
Also check this: https://github.com/golang/go/issues/39568#issuecomment-671424481.

/cc @gaocegege @johnugeorge

tenzen-y:
@andreyvelich 

> Which cluster are you using: GCP, AWS, on-prem ?

I'm using on-prem.

[...]

PR #1424, opened 2021-01-27
Allow disabling webhook installer

knkski:
Allows setting `DisableWebhookConfigInstaller` on the `webhook.ServerOptions` struct.

It allows configuring Katib to not install the webhook itself when starting up. This is useful for when an administrator is deploying Katib with a custom webhook configuration that Katib should not overwrite.

### DIFF
--- a/pkg/webhook/v1beta1/webhook.go
+++ b/pkg/webhook/v1beta1/webhook.go
@@ -32,10 +32,14 @@ import (
 	"github.com/kubeflow/katib/pkg/webhook/v1beta1/pod"
 )
 
-func AddToManager(m manager.Manager, port int32, serviceName string) error {
+func AddToManager(m manager.Manager, port int32, webhookInstaller bool, serviceName string) error {
 	so := webhook.ServerOptions{
 		CertDir: "/tmp/cert",
-		BootstrapOptions: &webhook.BootstrapOptions{
+		Port:    port,
+	}
+
+	if webhookInstaller {
+		so.BootstrapOptions = &webhook.BootstrapOptions{
 			Service: &webhook.Service{
 				Namespace: consts.DefaultKatibNamespace,
 				Name:      serviceName,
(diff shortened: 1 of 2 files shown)


pkg/webhook/v1beta1/webhook.go
```diff
@@ -32,10 +32,14 @@ import (
 	"github.com/kubeflow/katib/pkg/webhook/v1beta1/pod"
 )
 
-func AddToManager(m manager.Manager, port int32, serviceName string) error {
+func AddToManager(m manager.Manager, port int32, webhookInstaller bool, serviceName string) error {
 	so := webhook.ServerOptions{
```
andreyvelich, line 36: In case of custom WebHook configuration, should Katib controller create the webhook server ?

Maybe in case of custom webhooks we should disable adding to the manager here: 
https://github.com/kubeflow/katib/blob/master/cmd/katib-controller/v1beta1/main.go#L117-L121?

andreyvelich (commented):
Thank you for doing this @knkski!

I think maybe first of all we should add the YAMLs for the webhooks, because it requires refactoring `webhook.go` ?
If user wants to add custom webhook, they can easily modify appropriate YAML without setting the flag.

pkg/webhook/v1beta1/webhook.go
knkski: The scenario that I've got is that I'd like katib-controller to respond to webhook requests, but I'd like to set up the webhooks and Service resources myself, instead of katib-controller creating them. AFAICT, if we disabled adding to the manager as above, katib-controller wouldn't be listening to webhook requests. Is that correct?

andreyvelich: @knkski We can follow the [KFServing way](https://github.com/kubeflow/kfserving/blob/master/cmd/manager/main.go#L156-L178), but I believe we should update the controller-runtime pkg version for that.

Did you check my comment above: https://github.com/kubeflow/katib/pull/1424#pullrequestreview-579237771 ?
Maybe instead of adding another flag to the controller we should start with YAMLs for the webhooks?

/cc @gaocegege @johnugeorge

knkski: @andreyvelich: Is there an example somewhere else for using YAML files for the webhooks that you had in mind?

andreyvelich: @knkski For example, KFServing: https://github.com/kubeflow/kfserving/blob/master/config/webhook/manifests.yaml, Tekton Pipeline: https://github.com/tektoncd/pipeline/blob/master/config/500-webhooks.yaml.

gaocegege: If we need to update controller-runtime's version maybe we need to update k8s version first. tf-operator now depends on k8s 1.16, we are trying to remove direct deps to k8s. Ref https://github.com/kubeflow/tf-operator/pull/1235

andreyvelich: @gaocegege We currently use Kubernetes API from 1.12: https://github.com/kubeflow/katib/blob/master/Gopkg.toml#L57-L59.

I am fine with updating K8s version with controller-runtime, since currently Katib webhooks doesn't work with K8s 1.19, check issue here: https://github.com/kubeflow/katib/issues/1395#issuecomment-750373088.

knkski:
Superseded by #1450

PR #1432, opened 2021-02-15
Katib: Move manifests development upstream

yanniszark:
### Issue Resolved

Resolves: https://github.com/kubeflow/katib/issues/1431
Umbrella issue: https://github.com/kubeflow/manifests/issues/1740

### Description

As part of the work of wg-manifests for 1.3
(https://github.com/kubeflow/manifests/issues/1735), we are moving manifests
development in upstream repos. This gives the application developers full
ownership of their manifests, tracked in a single place.

This PR copies the manifests for application `Katib`
from path `apps/katib/upstream` of kubeflow/manifests to path
`manifests/v1beta1` of the upstream repo (https://github.com/kubeflow/katib).

cc @kubeflow/wg-automl-leads


manifests/v1beta1/katib-controller/base/katib-controller-deployment.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1
```
andreyvelich, line 1: Should we move all of the YAMLs from the `manifests/v1beta1/katib-controller/base` to the `/manifests/v1beta1/components/<appropriate-component-name/` ?
Since we left the old structure for the manifests to not brake `kfdef` configs before we transfer to the v3 manifests ?

manifests/v1beta1/katib-controller/overlays/application/application.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: app.k8s.io/v1beta1
```
andreyvelich, line 1: Do we need to maintain these overlays ?

andreyvelich (commented):
Thank you for doing this @yanniszark!
Do we have a unique structure for the kustomize manifests for each kubeflow component ?
Or every WG should decide how to store manifest by themselves?

manifests/v1beta1/katib-controller/base/katib-controller-deployment.yaml
yanniszark: I agree! I was suspecting something like this happened.

manifests/v1beta1/katib-controller/overlays/application/application.yaml
yanniszark: For the application overlays, we are discussing about moving away from them soon:
https://github.com/kubeflow/manifests/issues/1715

What we could do is put them in an `installs/kubeflow` kustomization, since they are used when Katib is installed alongside other Kubeflow components.

yanniszark:
@andreyvelich those are some good questions. Let me try to answer:

> Do we have a unique structure for the kustomize manifests for each kubeflow component ?
> Or every WG should decide how to store manifest by themselves?

We plan to propose a common way to organize manifests soon (step 4 in this plan https://github.com/kubeflow/manifests/issues/1735). The first goal is to get the content of kubeflow/manifests back to the respective upstream repos.

In general, applications are expected to provide kustomizations that can be used to install the application.

manifests/v1beta1/katib-controller/overlays/application/application.yaml
andreyvelich: @yanniszark So should we add the application overlay for Katib now or it's better to remove all labels and add them later with an Application, once we need it.
Since currently, many Katib labels are incorrect for the Application: https://github.com/kubeflow/manifests/pull/1669.

yanniszark: @andreyvelich in general, we are moving away from using the Application CR: https://github.com/kubeflow/manifests/issues/1715#issuecomment-779402779

For this PR, I would suggest leaving them as is. The initial plan is to just get the manifests under the control of the upstream repo (katib) and then we can make changes incrementally.

yanniszark:
@andreyvelich I have made the changed suggested here:

> Should we move all of the YAMLs from the manifests/v1beta1/katib-controller/base to the /manifests/v1beta1/components/<appropriate-component-name/ ?

Could you take another look?

andreyvelich:
@yanniszark Thank you for the update!
1. Do you need `istio-authz-policy-odl.yaml` and `istio-authz-policy-old.yaml ` in this PR ?
2. Please rebase to fix CI.

manifests/v1beta1/components/katib-controller/katib-ui-virtual-service.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - "*"
+  http:
+    - match:
+        - uri:
+            prefix: /katib/
+      rewrite:
+        uri: /katib/
+      route:
+        - destination:
+            host: katib-ui.$(katib-ui-namespace).svc.$(clusterDomain)
```
andreyvelich, line 18: Will we get these values from the appropriate stack.
For example: https://github.com/kubeflow/manifests/blob/master/distributions/stacks/gcp/kustomization.yaml ?

yanniszark: Yes exactly! Let's keep those as they are and we can introduce changes afterwards if we find it's not a good pattern.

yanniszark:
> Do you need istio-authz-policy-odl.yaml and istio-authz-policy-old.yaml in this PR ?
No, this was a mistaken `git add`.

> Please rebase to fix CI.
I rebased but the Github Actions CI seems broken. Here is the Github worker message:

> /home/runner/work/_temp/47c2af1d-bea5-4f4b-85f5-af8d0b594511.sh: line 1: juju: command not found

andreyvelich:
Thank you @yanniszark.
Yes there are some problems with CI as I mentioned here: https://github.com/kubeflow/katib/pull/1450#issuecomment-790930259.

yanniszark:
@andreyvelich thanks, so what are the next steps for the CI?
Wait for #1450 to be merged?

andreyvelich:
@yanniszark Yeah, please can you rebase and we can move forward?

yanniszark:
@andreyvelich great, I just rebased!

andreyvelich:
Thanks @yanniszark.
/lgtm
/approve
/retest

PR #1436, opened 2021-02-17
test: Pin manifests to stable v1.2-branch

### LINKED ISSUE #1433: e2e: Test breakage
/kind bug

E2E tests are not succeeding because they can't install the tf-operator manifests:
https://github.com/kubeflow/katib/blob/172999716efe76b73539391c2b028ebc5f0a837c/test/scripts/v1beta1/setup-katib.sh#L70

This is an artifact of using the manifests master branch to install the tf-operator.
A recent folder restructuring in manifests is what triggered it:
https://github.com/kubeflow/manifests/issues/1738
https://github.com/kubeflow/manifests/pull/1739

E2E tests should succeed.

This is an artifact of using the manifests master branch to install the tf-operator.
As master is not guarranteed to be stable or avoid any folder renames, the e2e tests for katib broke.

These are the possible solutions:
- Pin the manifests version to the latest stable release (v1.2). This will prevent it from breaking in the future and I think is the best choice. We can update it regularly as with other deps.
- Change the directories they use to the new ones.

I recommend the first approach. If you agree, I can make a quick PR.

cc @kubeflow/wg-automl-leads

yanniszark:
Fixes the e2e tests by pinning the kubeflow/manifests dependency to a stable branch.

Fixes #1433

### DIFF
--- a/test/workflows/components/workflows-v1beta1.libsonnet
+++ b/test/workflows/components/workflows-v1beta1.libsonnet
@@ -363,7 +363,7 @@
                 ],
                 env: prow_env + [{
                   name: "EXTRA_REPOS",
-                  value: "kubeflow/testing@HEAD;kubeflow/manifests@HEAD",
+                  value: "kubeflow/testing@HEAD;kubeflow/manifests@v1.2-branch",
                 }],
                 image: testWorkerImage,
                 volumeMounts: [


yanniszark:
@andreyvelich some seemingly unrelated tests are failing:
`build-suggestion-optuna`, `build-file-metrics-collector`, `build-db-manager`, `build-katib-controller`.
The logs are not very helpful. 
```
INFO[0046] Running: [/bin/sh -c if [ "$(uname -m)" = "ppc64le" ] || [ "$(uname -m)" = "aarch64" ]; then     apk --update add gcc musl-dev  &&     go build -o goptuna-suggestion
 ./v1beta1;     else     go build -o goptuna-suggestion ./v1beta1;     fi]
go: cannot find main module, but found Gopkg.lock in /go/src/github.com/kubeflow/katib
        to create a module there, run:
        cd ../../.. && go mod init
error building image: error building stage: failed to execute command: waiting for process to exit: exit status 1
```

Have you seen this before?

andreyvelich:
I will check it.
/retest

andreyvelich:
It seems that we have to switch to go modules to fix these errors.
I submitted PR: https://github.com/kubeflow/katib/pull/1438 to implement it.

I think we can close this PR, thank you for investigation @yanniszark !

yanniszark:
@andreyvelich thanks for taking the time to fix CI :)
Seems like it was a quite involved fix.

Issue #1405, opened 2020-12-01
Disable dynamic webhook creation in the Katib controller


andreyvelich:
/kind feature

Currently, Katib controller automatically [creates admission webhooks](https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/webhook.go#L74-L113).
As a result, we have several issues:
- https://github.com/kubeflow/manifests/issues/1379
- https://github.com/kubeflow/katib/issues/1395
- Kubebuilder 2.0 doesn't support dynamic webhooks.
- The Juju operator needs to manually create certificate. Check this: https://github.com/kubeflow/katib/pull/1403#discussion_r532925519.

We should disable the webhook creation completely and prepare appropriate manifests.
As well, we should implement the process for the certificate creation without `cert-manager`, if it is possible.

Let's think about the design here.

/cc @gaocegege @johnugeorge @knkski @jlewi 
/priority p1

gaocegege:
SGTM. One question here: How to setup a dev environment easily without this feature?

andreyvelich:
> SGTM. One question here: How to setup a dev environment easily without this feature?

Do you mean without `cert-manager` ?

yanniszark:
@andreyvelich is there any remaining work to figure out before moving certificate creation out of the Katib controller? Regarding:

> As well, we should implement the process for the certificate creation without cert-manager, if it is possible.

I don't think Katib should care about this. Katib should IMO provide a way to do this with `cert-manager` (as this is pretty standard and `kubebuilder` does it as well) and allow the consumer of the manifests to specify another way if required. This could be done by packaging the `cert-manager` functionality in a [Kustomize component](https://kubectl.docs.kubernetes.io/guides/config_management/components/), which allows creating re-usable and composable kustomizations.

So besides the manifests part, what is the work that needs to be done in order to disable webhook creation in the Katib code?

andreyvelich:
@yanniszark In that case all Katib users should install `cert-manager` on their clusters. In that case, we should maintain some version of this CRD and be consistent with other Kubeflow apps, e.g. KFServing.

>So besides the manifests part, what is the work that needs to be done in order to disable webhook creation in the Katib code?

Once we add webhooks to the manifests we should refactor controller to not create webhook during controller start: https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/webhook.go.

andreyvelich:
Any thoughts @gaocegege @johnugeorge ?

[...]

PR #1438, opened 2021-02-18
Migrate to Go modules

andreyvelich:
We should migrate to Go modules from `dep` since it is not longer supported from 2020: https://github.com/golang/dep#dep.
Also to avoid errors like this: https://github.com/kubeflow/katib/pull/1436#issuecomment-780755216.
Currently, I am using the same package versions with `k8s.io/kubernetes v1.13.3` and `sigs.k8s.io/controller-runtime v0.1.9` which is very old.

Unfortunately, I can't switch to the newer version because of this: https://github.com/kubeflow/katib/issues/1405. 
Controller runtime doesn't have dynamic webhook creation from v0.2.0.
We should disable dynamic webhook creation, update the versions and clean-up modules.

As well I fixed folder name in `Build Docker images` GitHub actions workflow. What do you think @DomFleischmann @RFMVasconcelos  ?

Please take a look @gaocegege @johnugeorge.


andreyvelich:
The tests are passed.
Can you check it @gaocegege @johnugeorge, please ?

gaocegege:
Should we use K8s v1.16.9 like tf-operator https://github.com/kubeflow/tf-operator/blob/master/go.mod#L33

andreyvelich:
> Should we use K8s v1.16.9 like tf-operator https://github.com/kubeflow/tf-operator/blob/master/go.mod#L33

I think once we fix problems with Webhooks we can directly switch to the latest version of controller-runtime and K8s.
E.g. v1.19 or v1.20

PR #1442, opened 2021-02-19
Enhance katib client by adding get_success_trial_details()

Adarsh2910:
Added API to get list of trials that succeded with hyperparam and metrics to KatibClient

```
get_success_trial_details(self, name, namespace) -> List of trial name with hyperparams and metrics
```

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -352,17 +352,17 @@ class KatibClient(object):
             result = []
             for i in katibtrial.get("items"):
                 output = {}
-                if i.get("metadata", {}).get("ownerReferences")[0].get("name") == name:                    
-                    output["status"] = i.get("status", {}).get(
-                    "conditions", [])[-1].get("type")
-                    output["name"] = i.get("metadata", {}).get("name")
-                    output["hyperparameters"] = i.get("spec", {}).get("parameterAssignments", [])
-                    output["metrics"] = (                            
-                        i.get("status", {})
-                        .get("observation", {})
-                        .get("metrics", [])
-                    )
-                    result.append(output)
+                if i.get("metadata", {}).get("ownerReferences")[0].get("name") == name:
+                    status = i.get("status", {}).get("conditions", [])[-1].get("type")
+                    if status == "Succeeded":
+                        output["name"] = i.get("metadata", {}).get("name")
+                        output["hyperparameters"] = i.get("spec", {}).get("parameterAssignments", [])
+                        output["metrics"] = (
+                            i.get("status", {})
+                            .get("observation", {})
+                            .get("metrics", [])
+                        )
+                        result.append(output)
         except multiprocessing.TimeoutError:
             raise RuntimeError("Timeout trying to getkatib experiment.")
         except client.rest.ApiException as e:
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -357,8 +357,7 @@ class KatibClient(object):
                     if status == "Succeeded":
                         output["name"] = i.get("metadata", {}).get("name")
                         output["hyperparams"] = i.get("spec", {}).get("parameterAssignments", [])
-                        output["metrics"] = (
-                            # FIXME: Support additional parameters
+                        output["metrics"] = (                            
                             i.get("status", {})
                             .get("observation", {})
                             .get("metrics", [])
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -334,7 +334,7 @@ class KatibClient(object):
         Get trial details that succeeded for an experiment
         :param name: existing experiment name
         :param namespace: defaults to current or default namespace
-        :return: trials name with hyperparams and metrics
+        :return: trials name with hyperparameters and metrics
         """
         if namespace is None:
             namespace = utils.get_default_target_namespace()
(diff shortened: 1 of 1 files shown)


Adarsh2910:
@googlebot I signed it!

gaocegege (commented):
/lgtm

/cc @andreyvelich

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -329,6 +329,52 @@ def list_trials(self, name=None, namespace=None):
           {2} ".format(name, namespace, e))
         return result
 
+    def get_success_trial_details(self, name=None, namespace=None):
```
andreyvelich, line 332: Maybe we should have an API which returns all Experiment's Trials objects instead of only Succeeded Trials with name, hyperparams, metrics?

Then user will have more freedom with this list of Trials ?
E.g. Get all Succeeded Trials, Get Early Stopped Trials, Get top <k> Trials.

WDYT @Adarsh2910 @gaocegege ?

andreyvelich (commented):
Thank you for this implementation @Adarsh2910!
I left one comment.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
gaocegege: SGTM, should we add it in this PR? or open an issue for it?

Adarsh2910: Do you think it's a good idea to give the user a way to filter the trials based on status, or just return the list of trials with hyperparams and metrics?

If it's the latter, we can also modify the `list_trials` method and just add hyperparams and metrics to the return.

andreyvelich: From my point of view, `list_trials` should return the list of `Trial` objects, instead of only name and status.
We can also add function: `get_success_trial_details` if it is needed in addition to `list_trials`.
What do you think @Adarsh2910 @gaocegege  ?

Adarsh2910: That sounds good to me
I think having a function to just get the success trials will be a helpful addition and we can raise another issue for patching `list_trials` to get the list of `Trial` objects

Adarsh2910: If we are all in agreement, Can we go ahead and merge this one in?
Thoughts? @andreyvelich @gaocegege

andreyvelich: Make sense, we can update `list_trials` in the next PR.
WDYT @gaocegege ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -329,6 +329,52 @@ def list_trials(self, name=None, namespace=None):
...
+        try:
+            katibtrial = thread.get(constants.APISERVER_TIMEOUT)
+            result = []
+            for i in katibtrial.get("items"):
+                output = {}
+                if i.get("metadata", {}).get("ownerReferences")[0].get("name") == name:
+                    status = i.get("status", {}).get("conditions", [])[-1].get("type")
+                    if status == "Succeeded":
+                        output["name"] = i.get("metadata", {}).get("name")
+                        output["hyperparams"] = i.get("spec", {}).get("parameterAssignments", [])
```
andreyvelich, line 359: Maybe use here not short word, what do you think @Adarsh2910  ?
```suggestion
                        output["hyperparameters"] = i.get("spec", {}).get("parameterAssignments", [])
```

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -329,6 +329,52 @@ def list_trials(self, name=None, namespace=None):
           {2} ".format(name, namespace, e))
         return result
 
+    def get_success_trial_details(self, name=None, namespace=None):
+        """
+        Get trial details that succeeded for an experiment
+        :param name: existing experiment name
+        :param namespace: defaults to current or default namespace
+        :return: trials name with hyperparams and metrics
```
andreyvelich, line 337:
```suggestion
        :return: trials name with hyperparameters and metrics
```

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
gaocegege: SGTM.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
Adarsh2910: yep, will fix it

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -329,6 +329,52 @@ def list_trials(self, name=None, namespace=None):
...
+            async_req=True)
+
+        katibtrial = None
+        try:
+            katibtrial = thread.get(constants.APISERVER_TIMEOUT)
+            result = []
+            for i in katibtrial.get("items"):
+                output = {}
+                if i.get("metadata", {}).get("ownerReferences")[0].get("name") == name:                    
+                    output["status"] = i.get("status", {}).get(
```
andreyvelich, line 356: Why did you delete check with succeeded status ?
Do you want to return hyperparameters and metrics for all Trials ?

Adarsh2910: ah, my bad
pushed a fix for it

andreyvelich (approved):
Thank you @Adarsh2910!
/lgtm
/approve

Issue #1435, opened 2021-02-17
http server panic, in katib controller with even the provided sample experiments.


ScrapCodes:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
On running the sample mnist experiment with kubeflow version 1.2.0 and katib version 0.10.1

While running the experiment, I am getting following error:
```
NFO:root:Creating Experiment: mnist in namespace: prashant
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/kubeflow/katib/api/katib_client.py", line 74, in create_experiment
    exp_object)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/apis/custom_objects_api.py", line 178, in create_namespaced_custom_object
    (data) = self.create_namespaced_custom_object_with_http_info(group, version, namespace, plural, body, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/apis/custom_objects_api.py", line 277, in create_namespaced_custom_object_with_http_info
    collection_formats=collection_formats)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 334, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 168, in __call_api
    _request_timeout=_request_timeout)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/api_client.py", line 377, in request
    body=body)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/rest.py", line 266, in POST
    body=body)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/rest.py", line 222, in request
    raise ApiException(http_resp=r)
kubernetes.client.rest.ApiException: (500)
Reason: Internal Server Error
HTTP response headers: HTTPHeaderDict({'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Date': 'Wed, 17 Feb 2021 08:37:23 GMT', 'Content-Length': '479'})
[...]

PR #1445, opened 2021-02-19
Verify nil objective in Experiment defaults

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1435.
We should verify nil objective before set the default values for metrics strategies.

/cc @ScrapCodes


gaocegege (commented):
/lgtm

PR #1457, opened 2021-03-08
Add Trial images build to the CI

andreyvelich:
Blocked by: https://github.com/kubeflow/testing/pull/923.

I added build of Katib Trial template images to the CI to make sure that each Trial template example can be built and executed.

As well, I fixed few problems with the current images:
- Fix https://github.com/pytorch/vision/issues/1938 problem in PyTorch mnist example.
- Add `--log-path` flag to PyTorch mnist to use this example in `File` and `StdOut` metrics collector mode.
- Use `docker.io/kubeflowkatib/pytorch-mnist` instead of `gcr.io/kubeflow-ci/pytorch-dist-mnist-test:v1.0` since it is similar images.
- Modify PyTorch mnist Dockerfile to be consistent with MXNet mnist
- Use `mxnet/python:latest_cpu_native_py3` base image in MXNet since it has all required MXNet setup.
- Use `tensorflow/tensorflow:1.15.4-py3` and `tensorflow/tensorflow:1.15.4-gpu-py3` for ENAS CNN Trial template. With that setup we can have only 1 requirements.txt for this image and we can cut the Dockerfile.

After we merge this PR, I will execute release script to push new images to the `kubeflowkatib` registry. Then I will update tags for the Katib Trial templates.

/cc @PatrickXYS


andreyvelich:
@gaocegege @johnugeorge I think the tests are working fine, can you take a look please ?

PR #1462, opened 2021-03-10
feat(db-manager): configurable DB name with extra params

Subreptivus:
- Demand non empty DB pass rather then hardcoded `mysql` DB name
- No hardcoded `mysql` DB name
- DB will be created if it not exists
- Ability to provide extra params for DB connection, for example: `"DB_EXTRA_PARAMS": "{\"tls\": \"true\"}"` for secure connection

I may have left EnvVariables names as is, yet now it looks more cleaner, aligned with other names and less confusing

**Release note**:
```release-note
Fully configurable external DB with ability to provide extra parameters with EnvVariable 'DB_EXTRA_PARAMS'
```


andreyvelich:
Thanks a lot for doing this @Subreptivus!
Let's include your change in the next release after the Katib 0.11 for Kubeflow 1.3 release which should happen in this week.
I am not sure that we have enough time to test it before the Katib 0.11 release.

We are still working on the manifest refactoring for the Katib: https://github.com/kubeflow/katib/pull/1464.
/ok-to-test

manifests/v1beta1/installs/katib-external-db/katib-db-manager-deployment.yaml
```diff
@@ -8,8 +8,6 @@ spec:
       containers:
         - name: katib-db-manager
           env:
-            - name: DB_NAME
-              value: mysql
```
andreyvelich, line 12: Maybe we can introduce another env variable, like `DB_TYPE: mysql`, to define type of DB?

cmd/db-manager/v1beta1/main.go
```diff
@@ -73,12 +73,12 @@ func (s *server) Check(ctx context.Context, in *health_pb.HealthCheckRequest) (*
 func main() {
 	flag.Parse()
 	var err error
-	dbNameEnvName := common.DBNameEnvName
-	dbName := os.Getenv(dbNameEnvName)
-	if dbName == "" {
-		klog.Fatal("DB_NAME env is not set. Exiting")
+	dbPasswordEnvName := common.DBPasswordEnvName
+	dbPassword := os.Getenv(dbPasswordEnvName)
```
andreyvelich, line 77: Maybe we can omit `dbPasswordEnvName` var ?
```suggestion
	dbPassword := os.Getenv(common.DBPasswordEnvName)
```

pkg/db/v1beta1/mysql/mysql.go
```diff
@@ -31,19 +30,73 @@ type dbConn struct {
...
+					}
+					return "", fmt.Errorf("Creation of Katib db '%s' failed: %v", dbName, err)
+				} else {
+					klog.Errorf("Open sql connection failed: %v", err)
+				}
+			case <-timeoutC:
+				return "", fmt.Errorf("Timeout waiting for DB conn successfully opened")
+			}
+		}
+	} else {
```
andreyvelich, line 96: Does the script reach this `else` since you are verify an empty password here: https://github.com/kubeflow/katib/blob/3e2d6fc72eecbb0668e2be8c4bbbc83ff1ba61fd/cmd/db-manager/v1beta1/main.go#L78

pkg/db/v1beta1/mysql/mysql.go
```diff
@@ -31,19 +30,73 @@ type dbConn struct {
...
+		ticker := time.NewTicker(connectInterval)
+		defer ticker.Stop()
+
+		timeoutC := time.After(connectTimeout)
+		for {
+			select {
+			case <-ticker.C:
+				if db, err := sql.Open(dbDriver, mysqlConfig.FormatDSN()); err == nil {
+					if _, err = db.Exec(fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s", dbName)); err == nil {
+						mysqlConfig.DBName = dbName
```
andreyvelich, line 85: Do you need to set `DBName` here since you are doing this in the `CreateMySQLConfig`: https://github.com/kubeflow/katib/blob/3e2d6fc72eecbb0668e2be8c4bbbc83ff1ba61fd/pkg/db/v1beta1/mysql/mysql.go#L48 ?

andreyvelich (commented):
@Subreptivus I apologies for waiting so long on this PR.
Since we finish our manifest refactoring, please can you rebase your PR ?

andreyvelich:
@gaocegege @johnugeorge Please take a look at these changes.

Subreptivus:
@andreyvelich I'll answer here to all those comments above, because it falls into single answer.
First of all, I wanted to introduce as little changes as possible, that's why there so many quirks in the whole flow. So in few words, there were hardcoded values all over the places, for `mysql` type of the DB and for password, etc.
Yet if it has a value in it, I can refactor it to be more flexible and address all risen questions.

andreyvelich:
> @andreyvelich I'll answer here to all those comments above, because it falls into single answer.
> First of all, I wanted to introduce as little changes as possible, that's why there so many quirks in the whole flow. So in few words, there were hardcoded values all over the places, for `mysql` type of the DB and for password, etc.
> Yet if it has a value in it, I can refactor it to be more flexible and address all risen questions.

I think the main problem is that `DB_NAME` env variable points to the DB Type currently.
And `KATIB_MYSQL_DB_DATABASE` env points to the Database name.
You can check the env variables definition here: https://www.kubeflow.org/docs/components/katib/env-variables/.
As I point here: https://github.com/kubeflow/katib/pull/1462#discussion_r618735448, we probably should rename `DB_NAME` to `DB_TYPE`.

In any case, we should design Katib DB Manager to be able to support multiply DB backend, not only MySQL.

What do you think @gaocegege @johnugeorge  ?

Issue #1443, opened 2021-02-19
[UI bug] Missing configuration fields in form


kimwnasptd:
/kind bug

The new Katib UI's form for submitting `Experiment` CRs is missing the following fields, which are present in the current UI's form:
- [x] [ResumePolicy](https://github.com/kubeflow/katib/blob/60f6c207ce8e1c88545450480c3a6f6dd0909ad5/pkg/apis/controller/experiments/v1beta1/experiment_types.go#L58) https://github.com/kubeflow/katib/pull/1463
- [x] [EarlyStopping](https://github.com/kubeflow/katib/blob/60f6c207ce8e1c88545450480c3a6f6dd0909ad5/pkg/apis/controller/experiments/v1beta1/experiment_types.go#L37) https://github.com/kubeflow/katib/pull/1463
- [x] [TrialParameters](https://github.com/kubeflow/katib/blob/60f6c207ce8e1c88545450480c3a6f6dd0909ad5/pkg/apis/controller/experiments/v1beta1/experiment_types.go#L217) https://github.com/kubeflow/katib/pull/1463

The most important one is the `TrialParameters` since without it the user cannot submit an Experiment by only using the form.
```
admission webhook "validating.experiment.katib.kubeflow.org" denied the request: spec.trialTemplate.trialParameters must be specified
```

/cc @andreyvelich

kimwnasptd:
I'll continue working on this now that the first iteration of our new UI #1421 is merged.

@andreyvelich @gaocegege @johnugeorge I see that the current UI has a way to automatically detect the Trial parameters. How exactly does this work? Does the UI parse the string content from the provided ConfigMaps and detects the `${trialParameters.<name>}`?

Also @andreyvelich could you also add this issue to our Kanban board?

andreyvelich:
> I'll continue working on this now that the first iteration of our new UI #1421 is merged.

Thanks @kimwnasptd!

> @andreyvelich @gaocegege @johnugeorge I see that the current UI has a way to automatically detect the Trial parameters. How exactly does this work? Does the UI parse the string content from the provided ConfigMaps and detects the `${trialParameters.<name>}`?

Yes:
https://github.com/kubeflow/katib/blob/081db0dd7f6f9470400f45dd4431c96927608845/pkg/ui/v1beta1/frontend/src/reducers/general.js#L88-L109

It just parses parameters for the given YAML Trial template.

kimwnasptd:
ACK, I'll mimic this logic then.

[...]

Issue #1421, opened 2021-01-25
Create a new Katib UI


kimwnasptd:
/kind feature

**Describe the solution you'd like**
Create a new Katib UI, as we've discussed in the [shared doc](https://docs.google.com/document/d/14cfGYJiTl32GWBUTbYQHzXC4UWA7r_ysgWYSWsxcDvU/edit).

We can use this as a tracking issue to track the progress of this effort as well as design decisions. I'll also be linking future issues to this one to keep everything organized and linked together.

**Anything else you would like to add:**
The switch should happen gradually. We should maintain both UIs until we ensure that this new version is on par, feature wise, with the functionality provided by the current UI.

cc @kubeflow/wg-automl-leads 
/assign @kimwnasptd

andreyvelich:
Thank you for your great work @kimwnasptd!
Sounds good to me.

gaocegege:
Thanks for your contribution! :tada: :+1:

andreyvelich:
@kimwnasptd I created Project for the new Katib UI: https://github.com/kubeflow/katib/projects/1.

Feel free to create issues that your first initial PR is currently missing and what should be integrated as part of your proposal.

kimwnasptd:
Adding a tracking comment to sum up the state of this effort.

The new UI is currently located under the [pkg/new-ui/v1beta1](https://github.com/kubeflow/katib/tree/5ed27d703b9e9401780c22d181fbdb8e2c1b4ee0/pkg/new-ui/v1beta1) and [cmd/new-ui/v1beta1](https://github.com/kubeflow/katib/tree/5ed27d703b9e9401780c22d181fbdb8e2c1b4ee0/cmd/new-ui/v1beta1) directories. Currently a user can use this UI to perform list, delete and create operations on the `Experiment` CRs of their cluster.

These are the things that we are working on and discussing:
* Define how we want the app to manage and handle the `TrialTemplates` https://github.com/kubeflow/katib/issues/1440. This might require us to redefine where these ConfigMaps should live and how the web app could enhance the UX of managing them
* Allow the app to work in standalone mode and not only with Kubeflow and the Central Dashboard https://github.com/kubeflow/katib/issues/1437
* Fully expose all the missing fields in the form https://github.com/kubeflow/katib/issues/1443
* Add CI tests for this new UI and CI/CD processes

There are also other ideas we have and track in our dedicated [kanban board](https://github.com/kubeflow/katib/projects/1)

[...]

PR #1463, opened 2021-03-10
Add missing form fields to new UI

kimwnasptd:
It adds support to the new UI's #1421 form for
* resumePolicy
* Early Stopping
* Trial Parameters

With this PR the form will be exposing all of the fields required to create a new Experiment. But, at the same time the only limitation I bumped into is the fact that the new UI doesn't currently allow us to edit the `source` of the Metric Collector. I'll comment this on our issue where we track the missing features. 

<details><summary>Videos of the changes</summary>
<p>

https://user-images.githubusercontent.com/11134742/110657178-4a4e7f00-81c9-11eb-9e2b-f90ccef6d649.mp4

https://user-images.githubusercontent.com/11134742/110657187-4cb0d900-81c9-11eb-98bb-61a0826dffae.mp4

https://user-images.githubusercontent.com/11134742/110657200-4e7a9c80-81c9-11eb-8395-2583ea44aef7.mp4

</p>
</details>

It's a first step for #1443. As mentioned above it won't close the issue because the new UI's form still doesn't allow us to configure the `source` of the Metrics Collector.

Since the PR is a bit big I'll also try to document my changes here, as well as in the commits.
* Early Stopping
    * a new module and component for this setting
    * needed to create a common component from the form's algorithm component for showing the algorithm settings
    * use the common component to show the Early Stopping algorithm's settings
* Resume Policy
    * just add one extra control to the Trial Thresholds
* Trial Parameters
    * create a component for each Trial Parameter
    * extend the component that shows the Trial Template to parse the yaml contents, when they change, and create the corresponding Trial Parameters
* Small fixes
    * previously the algorithm's settings would only load once the algorithm would change. This resulted in not showing the `random_state` setting for the `Random` HP algorithm [ which the value selected once the form loads ]
    * there were some places was still using the string value `None` for the `random_state`. I should use an empty value instead

/cc @andreyvelich @gaocegege @johnugeorge

### DIFF
--- a/pkg/new-ui/v1beta1/frontend/src/app/pages/experiment-creation/experiment-creation.component.html
+++ b/pkg/new-ui/v1beta1/frontend/src/app/pages/experiment-creation/experiment-creation.component.html
@@ -51,8 +51,8 @@
 
         <ng-container *ngIf="algorithmForm.value.type === 'hp'; else nasConfig">
           <!--Early Stopping-->
-          <mat-step [stepControl]="hyperParamsArray">
-            <form [formGroup]="hyperParamsArray">
+          <mat-step [stepControl]="earlyStoppingForm">
+            <form [formGroup]="earlyStoppingForm">
               <ng-template matStepLabel>Early Stopping</ng-template>
 
               <app-form-early-stopping [formGroup]="earlyStoppingForm">
--- a/pkg/new-ui/v1beta1/frontend/src/app/pages/experiment-creation/experiment-creation.component.html
+++ b/pkg/new-ui/v1beta1/frontend/src/app/pages/experiment-creation/experiment-creation.component.html
@@ -49,8 +49,18 @@
           </form>
         </mat-step>
 
-        <!--Parameters-->
         <ng-container *ngIf="algorithmForm.value.type === 'hp'; else nasConfig">
+          <!--Early Stopping-->
+          <mat-step [stepControl]="hyperParamsArray">
+            <form [formGroup]="hyperParamsArray">
+              <ng-template matStepLabel>Early Stopping</ng-template>
+
+              <app-form-early-stopping [formGroup]="earlyStoppingForm">
+              </app-form-early-stopping>
+            </form>
+          </mat-step>
+
+          <!--Parameters-->
           <mat-step [stepControl]="hyperParamsArray">
             <form [formGroup]="hyperParamsArray">
               <ng-template matStepLabel>Hyper Parameters</ng-template>
--- a/pkg/new-ui/v1beta1/frontend/src/app/services/experiment-form.service.ts
+++ b/pkg/new-ui/v1beta1/frontend/src/app/services/experiment-form.service.ts
@@ -220,6 +220,7 @@ export class ExperimentFormService {
       cmName: '',
       cmTrialPath: '',
       yaml: '',
+      trialParameters: this.builder.array([]),
     });
   }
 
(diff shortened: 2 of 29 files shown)


pkg/new-ui/v1beta1/frontend/src/app/pages/experiment-creation/experiment-creation.component.html
```diff
@@ -49,8 +49,18 @@
           </form>
         </mat-step>
 
-        <!--Parameters-->
         <ng-container *ngIf="algorithmForm.value.type === 'hp'; else nasConfig">
+          <!--Early Stopping-->
+          <mat-step [stepControl]="hyperParamsArray">
+            <form [formGroup]="hyperParamsArray">
```
andreyvelich, line 55: Should we change it to.
```suggestion
          <mat-step [stepControl]="earlyStoppingForm">
            <form [formGroup]="earlyStoppingForm">
```

pkg/new-ui/v1beta1/frontend/src/app/services/experiment-form.service.ts
```diff
@@ -42,6 +45,7 @@ export class ExperimentFormService {
       parallelTrialCount: 3,
       maxTrialCount: 12,
       maxFailedTrialCount: 3,
+      resumePolicy: 'Never',
```
andreyvelich, line 48: I think the default `resumePolicy` is `LongRunning`.

andreyvelich (commented):
Thanks for this update, it's looks great @kimwnasptd.
I left few comments.

pkg/new-ui/v1beta1/frontend/src/app/pages/experiment-creation/experiment-creation.component.html
kimwnasptd: Yeap, thanks! Never trust copy paste

pkg/new-ui/v1beta1/frontend/src/app/services/experiment-form.service.ts
kimwnasptd: Αhh indeed, I missed it. Thanks!

kimwnasptd: Pushed a commit for this

kimwnasptd:
Ι see the `Charmed Katib / Test` failing and this relevant, I believe, log
```
Error from server (BadRequest): container "katib-ui" in pod "katib-ui-7cff86f74f-9bvpp" is waiting to start: PodInitializing
```

I'll try to rebase on top of the latest master to see if this resolves the issue

kimwnasptd:
Hmm the error seems to persist. Why does the `katib-ui` test get triggered when we don't modify any "relevant" code? Is this expected?

Lastly, @andreyvelich could you also write the `I consent` command to google bot so that it doesn't complain about the CLA? The bot seems to error once I rebase and force push

andreyvelich:
@googlebot I consent.

andreyvelich:
@kimwnasptd GitHub actions should be fixed in this PR: https://github.com/kubeflow/katib/pull/1465.

kimwnasptd:
> @kimwnasptd GitHub actions should be fixed in this PR: #1465.

ACK!

I also pushed another small commit, because there was a small error with building the frontend. I've explained the reason behind it in the commit message as well, but the verdict is that Angular could not detect that 

```javascript
formGroup.get('trialParameters').controls
```
Is a FormArray and has a property controls. This happens when we use this in the html file.

In any case I believe this is another indication that the next thing to work on is the tests and having a CI process that at least checks that the UI can be build.

andreyvelich:
@kimwnasptd As I can see the CI passed.
Are we ready to merge it ?

kimwnasptd:
Yeap, we should be good to go!

andreyvelich (approved):
/lgtm
/approve

Issue #1453, opened 2021-03-04
[Juju] Update Juju operator with webhook manifests changes


andreyvelich:
/kind feature

After this PR: https://github.com/kubeflow/katib/pull/1450, we should update our Juju operators with the latest manifest changes. 

/cc @DomFleischmann @knkski @RFMVasconcelos

RFMVasconcelos:
Thank you for creating this issue @andreyvelich !! We'll take a look into it :)

andreyvelich:
@DomFleischmann @knkski @RFMVasconcelos I have released the latest commit: `c6c9172` in the `kubeflowkatib` registry: https://hub.docker.com/u/kubeflowkatib.
You should be able to use latest tag to get the latest changes for the controller now.

Issue #1415, opened 2021-01-08
katib-mysql: Access denied for user 'root'@'localhost' (using password: YES)


anneum:
/kind bug

**What steps did you take and what happened:**
Installed Kubeflow on an onpremise kubenetes cluster.
Multiple Pods stuck in CrashLoopBackOff
katib-db-manager
katib-mysql

**What did you expect to happen:**
Expect that the pods start normal

**Anything else you would like to add:**
As mentioned in #1156 I increased liveness probe and readiness probe.
I think, that the katib-db-manager also works if the katib-mysql is running.

```
kubectl -nkubeflow describe pod katib-mysql-98566c454-2kg4t
Name:         katib-mysql-98566c454-2kg4t
Namespace:    kubeflow
Priority:     0
Node:         
Start Time:   Fri, 08 Jan 2021 08:45:35 +0000
Labels:       app=katib
              app.kubernetes.io/component=katib
              app.kubernetes.io/name=katib-controller
              component=mysql
              pod-template-hash=98566c454
Annotations:  sidecar.istio.io/inject: false
Status:       Running
IP:           
IPs:
  IP:           
Controlled By:  ReplicaSet/katib-mysql-98566c454
Containers:
  katib-mysql:
    Container ID:  docker://8c14869f8a0cecdd6df1e6df2829838e11922167b66db1c6b5582d4ddbfb0818
    Image:         mysql:8
    Image ID:      docker-pullable://mysql@sha256:365e891b22abd3336d65baefc475b4a9a1e29a01a7b6b5be04367fcc9f373bb7
    Port:          3306/TCP
    Host Port:     0/TCP
    Args:
      --datadir
      /var/lib/mysql/datadir
    State:          Running
      Started:      Fri, 08 Jan 2021 08:45:37 +0000
    Ready:          False
    Restart Count:  0
    Liveness:       exec [/bin/bash -c mysqladmin ping -u root -p${MYSQL_ROOT_PASSWORD}] delay=30s timeout=5s period=10s #success=1 #failure=6
    Readiness:      exec [/bin/bash -c mysql -D ${MYSQL_DATABASE} -u root -p${MYSQL_ROOT_PASSWORD} -e 'SELECT 1'] delay=5s timeout=1s period=10s #success=1 #failure=6
    Environment:
      MYSQL_ROOT_PASSWORD:         <set to the key 'MYSQL_ROOT_PASSWORD' in secret 'katib-mysql-secrets'>  Optional: false
[...]

Issue #1212, opened 2020-06-09
katib-mysql pod :Access denied for user 'root'@'localhost' (using password: YES)


ijackcy:
/kind bug

**What steps did you take and what happened:**

When I deployed kubeflow1.0, katib-db-manager-849b858bc8-4h8c9  and katib-mysql-7f99dfd774-gc4qp  don't work well.

katib-db-manager-849b858bc8-4h8c9                              0/1     ImagePullBackOff   0          11h
katib-mysql-7f99dfd774-gc4qp                                   0/1     Running            1          11h was 

when i checked the log of katib-mysql-7f99dfd774-gc4qp  :

Events:
  Type     Reason     Age                   From                Message
  ----     ------     ----                  ----                -------
  Warning  Unhealthy  80s (x4130 over 11h)  kubelet, slaver003  Readiness probe failed: mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)

when i enter the katib-mysql-7f99dfd774-gc4qp :
root@katib-mysql-7f99dfd774-gc4qp:/# mysql -D ${MYSQL_DATABASE} -u root -p${MYSQL_ROOT_PASSWORD} -e 'SELECT 1'
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)
root@katib-mysql-7f99dfd774-gc4qp:/# mysql -D ${MYSQL_DATABASE} -e 'SELECT 1'                                 
ERROR 1049 (42000): Unknown database 'katib'

**What did you expect to happen:**

all of pod work well

**Anything else you would like to add:**

and then i try to delete the error pods, and modify the images(mysql 5) ,also the same error.
Access denied for user 'root'@'localhost' (using password: YES)
**Environment:**

- Kubeflow version:1.0
- Minikube version:
- Kubernetes version: (use `kubectl version`):1.15.1
- OS (e.g. from `/etc/os-release`):

gaocegege:
May I ask how you install the katib?

ijackcy:
> May I ask how you install the katib?

When I installed kubeflow, automatically  started katib pod

andreyvelich:
@ijackcy Can you check that `MYSQL_ROOT_PASSWORD` env is set on katib-mysql pod.
Execute on katib-mysql pod and run `echo $MYSQL_ROOT_PASSWORD`

[...]

Issue #1156, opened 2020-04-17
katib-mysql does not work


hjkkhj123:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

i just install kubeflow in manual

**What did you expect to happen:**
```
2020-04-17 06:28:38+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 8.0.19-1debian10 started.
2020-04-17 06:28:38+00:00 [Note] [Entrypoint]: Switching to dedicated user 'mysql'
2020-04-17 06:28:38+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 8.0.19-1debian10 started.
2020-04-17T06:28:38.815249Z 0 [Warning] [MY-011070] [Server] 'Disabling symbolic links using --skip-symbolic-links (or equivalent) is the default. Consider not using this option as it' is deprecated and will be removed in a future release.
2020-04-17T06:28:38.815363Z 0 [System] [MY-010116] [Server] /usr/sbin/mysqld (mysqld 8.0.19) starting as process 1
mysqld: Table 'mysql.plugin' doesn't exist
2020-04-17T06:28:43.673845Z 0 [ERROR] [MY-010735] [Server] Could not open the mysql.plugin table. Please perform the MySQL upgrade procedure.
2020-04-17T06:28:44.007568Z 0 [Warning] [MY-010015] [Repl] Gtid table is not ready to be used. Table 'mysql.gtid_executed' cannot be opened.
2020-04-17T06:28:44.391089Z 0 [Warning] [MY-010015] [Repl] Gtid table is not ready to be used. Table 'mysql.gtid_executed' cannot be opened.
2020-04-17T06:28:44.509101Z 0 [Warning] [MY-010068] [Server] CA certificate ca.pem is self signed.
2020-04-17T06:28:44.542561Z 0 [Warning] [MY-011810] [Server] Insecure configuration for --pid-file: Location '/var/run/mysqld' in the path is accessible to all OS users. Consider choosing a different directory.
2020-04-17T06:28:44.543088Z 0 [Warning] [MY-010441] [Server] Failed to open optimizer cost constant tables
2020-04-17T06:28:44.543392Z 0 [ERROR] [MY-013129] [Server] A message intended for a client cannot be sent there as no client-session is attached. Therefore, we're sending the information to the error-log instead: MY-001146 - Table 'mysql.component' doesn't exist
[...]

PR #1464, opened 2021-03-10
Refactor Kustomize manifests for Katib

andreyvelich:
I refactored Kustomize manifest for Katib.
I followed this approach:
1. All Katib manifests are located in `/components`.
2. All installs with kustomization files are located in `/installs`.

User can select required install Kustomize file, e.g. Katib standalone, Kubeflow, IBM, etc..
User also are able to use pure `kubectl` to deploy Katib, we can extend `deploy.sh` script for that later.

We can keep the `latest` tag for the Katib standalone installation and we can add Katib 0.11 tag for the Kubeflow Katib installation.

For the Katib config I added `patchesStrategicMerge` to the Kustomize file. I am not sure if it is possible to modify image tags in the other way.

For the Kubeflow installation I added: 
```
commonLabels:
  app.kubernetes.io/component: katib
```
I left only 1 application with this label for the Katib component.

Please take a look @yanniszark.

I am still updating some docs in this PR.

/cc @gaocegege @johnugeorge


andreyvelich:
I changed docs in the Katib repo.
/hold for the reviews

andreyvelich:
@DomFleischmann @knkski @RFMVasconcelos I disabled GitHub actions at this PR, since we should fix: https://github.com/kubeflow/katib/issues/1453 and add cert-generator to generator certs.
If you check logs from the controller: https://github.com/kubeflow/katib/pull/1464/checks?check_run_id=2080590497, manager fails because `/tmp/cert/tls.crt` file doesn't exist.

andreyvelich:
I also updated tags for the Trial images to `v1beta1-c6c9172` for these changes: https://github.com/kubeflow/katib/pull/1457.

yanniszark:
@andreyvelich to explore the diff, I built:
- `manifests/v1beta1/installs/katib-standalone` on master
- `manifests/v1beta1/installs/katib-with-kubeflow` on this branch

I'm uploading the old and new yaml, for completeness:
[old.yaml.log](https://github.com/kubeflow/katib/files/6124561/old.yaml.log)
[new.yaml.log](https://github.com/kubeflow/katib/files/6124562/new.yaml.log)

The main diff I saw is:
- The new YAML has the webhook configurations. Nice!
- The new YAML has the cert-generator script. Also nice, it would be good if we could somehow make it optional. But we can discuss later on this.
- The new YAML creates the following PersistentVolume:
```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  labels:
    app.kubernetes.io/component: katib
    type: local
  name: katib-mysql
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 10Gi
  hostPath:
    path: /tmp/katib
  storageClassName: katib
```
I don't think it's right to include a PersistentVolume, as it ties Katib installation to a particular type of storage (hostPath). It also presumes the existence of the Katib storageclass. By including just a PersistentVolumeClaim, we make storage pluggable. It's standard to include a PVC (not a PV) in other manifests as well. What do you think?

andreyvelich:
Thank you for the review @yanniszark!
> The new YAML has the cert-generator script. Also nice, it would be good if we could somehow make it optional. But we can discuss later on this.

I think in the next release (~ Katib 0.11.1 or Katib 0.12 ) we can make it optional. We should have a discussion how we can do it (cert-manager overlay for instance).


> I don't think it's right to include a PersistentVolume, as it ties Katib installation to a particular type of storage (hostPath). It also presumes the existence of the Katib storageclass. By including just a PersistentVolumeClaim, we make storage pluggable. It's standard to include a PVC (not a PV) in other manifests as well. What do you think?

We can exclude this PV from the Kubeflow installation.
Before the manifest migration, PV was included in the Katib installation, but PV was excluded in the Kubeflow installation.
We have seen a lot of problems when users use their own StorageClass to provision volume (https://github.com/kubeflow/katib/issues/1415, https://github.com/kubeflow/katib/issues/1156, https://github.com/kubeflow/katib/issues/1212, etc.)
That is why we included PV with `hostPath` in the Katib installation by default.

If users know that cluster Storage Class can properly provision volume, they can always modify the kustomization file.

WDYT @yanniszark ? cc @gaocegege @johnugeorge

yanniszark:
> We have seen a lot of problems when users use their own StorageClass to provision volume (#1415, #1156, #1212, etc.)
> That is why we included PV with hostPath in the Katib installation by default.

Thanks for including the related issues for context! I agree that it's good to try and make users' life easier. Since creating the PV is essentially an environment setup step (and may have security implications because it's using a hostPath volume, thus having side-effects on the host machine), I would include it in a separate manifest and add an optional step to install it.

For this PR, excluding it in the Kubeflow overlay seems ok. I'd like to open the discussion in the future if you agree!

andreyvelich:
> Thanks for including the related issues for context! I agree that it's good to try and make users' life easier. Since creating the PV is essentially an environment setup step (and may have security implications because it's using a hostPath volume, thus having side-effects on the host machine), I would include it in a separate manifest and add an optional step to install it.

> For this PR, excluding it in the Kubeflow overlay seems ok. I'd like to open the discussion in the future if you agree!

Sure, it makes sense. Let's keep for the Kubeflow 1.3 release as it was before (Katib installation - PV is included, Kubeflow installation - PV is excluded). I will create an issue and we can discuss how we want to manager it in the further release.

andreyvelich:
@yanniszark Also what do you think about this patch:
https://github.com/kubeflow/katib/blob/585c01df7e3991a119d27a855a9d9a7ba9906e6d/manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml#L49-L111

Does it sound good for you ?

yanniszark:
> Does it sound good for you ?

Yes, unfortunately you need to patch the whole ConfigMap :/
You can also use the `configMapGenerator` with behavior merge:
https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/configmapgenerator/

andreyvelich:
I believe `configMapGenerator` will add hash once it generates resource which doesn't work for us currently.
Let's continue with Patch for now.

@yanniszark If you are fine with other changes, please can you give your `lgtm` and we can move forward.

andreyvelich:
@yanniszark I was actually thinking about the patch more.
Since we try to follow release process as you mentioned here: https://github.com/kubeflow/katib/issues/1466, we don't need to have specific image tag in the `katib-with-kubeflow` install.

Thus, we can use [re-use patch](https://github.com/kubeflow/katib/blob/c7027143f8adf7085046c33b05cd6dbfed3b57f8/manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml#L49) for the Katib config and simplify kustomization files.

andreyvelich:
@gaocegege @johnugeorge I changed mxnet example to get MNIST data using mxnet API as described in [MXNet docs](https://mxnet.apache.org/versions/1.7.0/api/python/docs/tutorials/packages/gluon/image/mnist.html#Loading-Data).
As I can see downloading data from http://yann.lecun.com/exdb/mnist/ is impossible currently.

andreyvelich:
I changed MNIST dataset to FashionMNIST for Pytorch MNIST example.
I am not sure once http://yann.lecun.com/exdb/mnist/ become stable.
/hold cancel

PR #1480, opened 2021-03-17
Remove legacy controller flags from the developer guide

andreyvelich:
I removed flags from the developer guide that we currently disable: https://github.com/kubeflow/katib/blob/master/cmd/katib-controller/v1beta1/main.go#L57-L60.
If we decide to add them later, we should update the developer guide.

/cc @annajung @gaocegege @johnugeorge


gaocegege (commented):
LGTM :+1:
/lgtm

Issue #1437, opened 2021-02-17
[UI Feature] Standalone mode for the new Katib UI


andreyvelich:
/kind feature

The new Katib UI should work in the standalone mode when user deploys Katib without Kubeflow central dashboard.
Let's continue discussion here how we should do that.
Also check this comment: https://github.com/kubeflow/katib/pull/1427#discussion_r577589302.

/cc @kimwnasptd @gaocegege @johnugeorge

kimwnasptd:
First off, from the technical side of view, the frontend can figure out if it's in Kubeflow mode or in Standalone dynamically. So we won't need to extend the backend to provide this info to the frontend. I've tried to document the more technical details of why the above approach works in https://github.com/kubeflow/kubeflow/issues/5566#issuecomment-779278036.

The changes we will need to make to the UI, that will be aware of the mode, are:
1. Showing a namespace drop-down component in standalone mode
2. Have a side panel for app navigation in standalone mode [HP Tuning, NAS]
3. Rely on the CentralDashboard's side-panel for navigating into the app's different pages in Kubeflow mode

In both cases the UI will be checking if the `window.centraldashboard` object is present to detect if the app is exposed under the CentralDashboard

kimwnasptd:
For the side-panel we should try to avoid showing two distinct sidebars in Kubeflow mode https://github.com/kubeflow/kubeflow/issues/5566#issuecomment-782036247.

The Katib UI should only expose a hamburger-menu/sidebar only if the CentralDashboard is not present, which means no `window.centraldashboard` object exists in the app's window.

![Screenshot from 2021-02-19 13-42-11](https://user-images.githubusercontent.com/11134742/108503471-19abb180-72bd-11eb-853c-8b075aaa103a.png)

andreyvelich:
> For the side-panel we should try to avoid showing two distinct sidebars in Kubeflow mode kubeflow/kubeflow#5566 (comment).

The Katib UI should only expose a hamburger-menu/sidebar only if the CentralDashboard is not present, which means no window.centraldashboard object exists in the app's window.

Then how we will navigate between Trial template editor page, Katib Config and main page where user can create new Experiment?
Are you planing to integrate this pages in the central dashboard menu?

kimwnasptd:
> Are you planing to integrate these pages in the central dashboard menu?

[...]

PR #1483, opened 2021-03-18
new-ui: Show the Namespace Selector when app is deployed as standalone

kimwnasptd:
This PR updates the UI code, for the new UI #1421, to
1. detect if the dashboard is present
2. if it's not then show the namespace selector component

This is a first step for #1437. With this fix the new UI will be able to also be deployed as a standalone app and not always rely on the Central Dashboard to be present, to allow the users to navigate and inspect namespaces.

This brings the app at the same state and with the same behavior for the standalone mode with the current UI. Next steps would be to extend the current backend to use `SubjectAccessReviews` to perform authorization checks, if the app is deployed in kubeflow mode. But we can coordinate on this effort in our tracking issue #1437.

This is a first step for #1437

This PR also updates the base commit to use for the common ui code, that lives in `kubeflow/kubeflow`. In order to check the changes locally you will need to follow the [updated instructions](https://github.com/kubeflow/katib/blob/e7ea4476b98f85878f9870db8c5e12cc805ff288/pkg/new-ui/v1beta1/README.md#serve-only-the-frontend), which are based on the newer commit.

But you could also build the oci image from the updated dockerfile as well

### DIFF
--- a/pkg/new-ui/v1beta1/frontend/src/app/pages/experiments/experiments.component.html
+++ b/pkg/new-ui/v1beta1/frontend/src/app/pages/experiments/experiments.component.html
@@ -1,5 +1,5 @@
 <lib-namespace-select
-  *ngIf="!env.production"
+  *ngIf="(ns.dashboardConnected$ | async) === dashboardDisconnectedState"
   namespacesUrl="/katib/fetch_namespaces"
 ></lib-namespace-select>
 
--- a/cmd/new-ui/v1beta1/Dockerfile
+++ b/cmd/new-ui/v1beta1/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update && apt-get install git -y
 WORKDIR /kf
 RUN git clone https://github.com/kubeflow/kubeflow.git && \
     cd kubeflow && \
-    git checkout a349284
+    git checkout 24bcb8e
 
 # --- Build the frontend kubeflow library ---
 FROM node:12 AS frontend-kubeflow-lib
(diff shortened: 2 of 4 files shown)


pkg/new-ui/v1beta1/frontend/src/app/pages/experiments/experiments.component.html
```diff
@@ -1,5 +1,5 @@
 <lib-namespace-select
-  *ngIf="!env.production"
+  *ngIf="(ns.dashboardConnected$ | async) === dashboardDisconnectedState"
```
andreyvelich, line 2: Since you are not using `env.production` more, do you still need this: https://github.com/kubeflow/katib/tree/d9b4602280748e2fb322bffef95d02dda92664cb/pkg/new-ui/v1beta1/frontend/src/environments ?

andreyvelich (commented):
Great progress @kimwnasptd!
I left one comment.

pkg/new-ui/v1beta1/frontend/src/app/pages/experiments/experiments.component.html
kimwnasptd: While with this PR we don't use Angular's [environment mechanism](https://angular.io/guide/build#configure-environment-specific-defaults) anymore I would suggest that we keep these files.

They act as a central place of configuration we could use in the future.

andreyvelich: Got it, thanks.

andreyvelich (commented):
/lgtm
/approve

PR #1493, opened 2021-03-22
Fix release script to replace Katib image tags

andreyvelich:
I fixed incorrect replace for the Katib image tags in the release script.
The current script replaces `mysql` image tag for the [`katib-ibm` install](https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-ibm/kustomization.yaml#L9) and doesn't replace tags in the `katib-config-patch.yaml`, which is not correct.

I make the explicit replace in the `katib-external-db/kustomization.yaml`, `katib-config-patch.yaml`, `katib-standalone/kustomization.yaml` and `katib-with-kubeflow/kustomization.yaml` files to make sure that we don't replace tags in the other places.

Please take a look.

/assign @yanniszark @gaocegege @johnugeorge


gaocegege (commented):
/lgtm

PR #1498, opened 2021-03-30
fix kustomize manifests for kubeflow

DavidSpek:
When trying to build the manifests for the Kubeflow deployment the following error occurs:
```
Error: accumulating resources: accumulation err='accumulating resources from '../../components/controller/controller.yaml': security; file '/home/david/Documents/GitHub/katib/manifests/v1beta1/components/controller/controller.yaml' is not in or below '/home/david/Documents/GitHub/katib/manifests/v1beta1/installs/katib-with-kubeflow'': got file 'controller.yaml', but '/home/david/Documents/GitHub/katib/manifests/v1beta1/components/controller/controller.yaml' must be a directory to be a root
```

This PR creates `kustomization.yaml` files for each of the folders so that the `kustomization.yaml` in `katib-with-kubeflow` can reference the component folders properly. Similarly, the `katib-config-path.yaml` that was being referenced from `katib-standalone` has been copied into the ` katib-with-kubeflow` so that the manifests can be properly built with kustomize. 

/cc @andreyvelich


manifests/v1beta1/installs/katib-external-db/katib-config-patch.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: v1
```
andreyvelich, line 1: As we discussed on the meeting, we should move this file to `/manifests/v1beta1/components/controller` and remove current `katib-config.yaml` file.

manifests/v1beta1/installs/katib-external-db/kustomization.yaml
```diff
@@ -3,28 +3,17 @@ kind: Kustomization
 namespace: kubeflow
 resources:
   # Namespace.
-  - ../../components/namespace.yaml
+  - namespace.yaml
```
andreyvelich, line 6: Any ideas how we can keep namespace under `manifests/v1beta1/components` ?
Maybe create folder `components/namespace` and add `namespace.yaml` and `kustomization.yaml` there.
WDYT @DavidSpek  ? cc @yanniszark

manifests/v1beta1/installs/katib-with-kubeflow-cert-manager/patches/mysql-pvc.yaml
```diff
@@ -0,0 +1,6 @@
+[
```
andreyvelich, line 1: Do we have any options to combine this patch and https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-with-kubeflow/patches/mysql-pvc.yaml in the one file ? cc @yanniszark

manifests/v1beta1/installs/katib-with-kubeflow-cert-manager/ui-virtual-service.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: networking.istio.io/v1alpha3
```
andreyvelich, line 1: The same comment as for the `mysql-pvc.yaml` patch.

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
```diff
@@ -65,5 +52,7 @@ vars:
     kind: Deployment
     name: katib-ui
 
+commonLabels:
```
andreyvelich, line 55: This can be removed

andreyvelich (commented):
I left few comments @DavidSpek.
For the CI: we should change the path for the `katib-config` with image tags here: https://github.com/kubeflow/katib/blob/master/test/scripts/v1beta1/setup-katib.sh#L42.

For the release scripts: we should also modify the path for the `katib-config`: https://github.com/kubeflow/katib/blob/master/scripts/v1beta1/release.sh#L74.

andreyvelich:
It would be great if @yanniszark could also take a look at these changes.

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
DavidSpek: Are you sure? Kustomize sometimes uses this for the selectors of deployments or services.

yanniszark:
@DavidSpek thanks for this effort. I will take a look shortly but I should comment that I am not treating this PR as release blocking (I think we've discussed in the past, but also mentioning it now for clarity). The first thing I would check for a PR like this would be that the result of `kustomize build` before and after the PR is the same for every entry under `installs`. Could you confirm that this is the case? And of course, CI should be passing and @andreyvelich confirm that these changes won't break something we don't expect in Katib docs / examples.

DavidSpek:
@yanniszark That is indeed how I am testing things. I just got caught up with creating a PR for a fix to the JWA as something seemed to be broken but I am now continuing with working on these manifests. It isn't release blocking, but as the RC period is partly meant for fixing manifests if this PR is done in time for the release cut it should be included I think.

andreyvelich:
@yanniszark Since we are using `katib-with-kubeflow-cert-manager` install for the kubeflow installation maybe we can remove the overlay which is using the Katib cert generator for now ?
I think if any distribution owners will require this install for Katib, we can this overlay later.

manifests/v1beta1/installs/katib-external-db/kustomization.yaml
DavidSpek: I think this is the easiest way so I have implemented this.

manifests/v1beta1/installs/katib-external-db/katib-config-patch.yaml
DavidSpek: I've removed the `katib-config-patch.yaml` files in the installation directory and move the image tags to `/manifests/v1beta1/components/controller/katib-config.yaml`.

manifests/v1beta1/installs/katib-with-kubeflow-cert-manager/patches/mysql-pvc.yaml
DavidSpek: I've implemented `katib-with-kubeflow-cert-manager` to use `katib-with-kubeflow` as a base, which solves needing to duplicate this file.

manifests/v1beta1/installs/katib-with-kubeflow-cert-manager/ui-virtual-service.yaml
DavidSpek: I've implemented `katib-with-kubeflow-cert-manager` to use `katib-with-kubeflow` as a base, which solves needing to duplicate this file.

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
andreyvelich: I think yes, check: https://github.com/kubeflow/katib/pull/1507#discussion_r605749416, we decided with @yanniszark, that we don't need this labels more.

andreyvelich:
@DavidSpek @yanniszark I was actually think about the Katib installs more. What do you think about keeping these installs:
1. `/installs/katib-external-db`
1. `/installs/katib-ibm`
1. `/installs/katib-standalone`
1. `/installs/katib-cert-manager`
1. `/installs/katib-with-kubeflow`

Since Kubeflow installation uses cert manager instead of custom cert generator we don't need to have 2 installs for Katib with Kubeflow.
We can create `katib-cert-manager` with all Katib Standalone components without Cert generator.
This is useful in cases when our custom Cert Generator doesn't work, e.g. https://github.com/kubeflow/katib/issues/1512

WDYT @DavidSpek @yanniszark ?

DavidSpek:
@andreyvelich Sorry I completely missed your comment yesterday as I was very busy. Hopefully I will have time to continue wit this today (otherwise it'll be in the weekend).

andreyvelich:
> @andreyvelich Sorry I completely missed your comment yesterday as I was very busy. Hopefully I will have time to continue wit this today (otherwise it'll be in the weekend).

No worries, take your time.
Thank you for your effort on this!

DavidSpek:
@andreyvelich I've pushed a commit with the `katib-cert-manager` and `katib-with-kubeflow` uses this as a base. I've also had the cert-generator create around 10 pods before one was successful so I think this is a good move. I don't really have a way to test this deployment, but when looking at the resulting YAML it makes sense. The resulting kubeflow YAML is what is expected with this setup.

I've also updated the releasing and test scripts so that the image tags are updated in [katib-config.yaml](https://github.com/kubeflow/katib/pull/1498/files#diff-2e1aff4a1f1a5b4532e8a6a143f347aaee89ab48f4057d36499d25a00541aed7).

DavidSpek:
@andreyvelich I've pushed a commit with the `katib-cert-manager` and `katib-with-kubeflow` uses this as a base. I've also had the cert-generator create around 10 pods before one was successful so I think this is a good move. I don't really have a way to test this deployment, but when looking at the resulting YAML it makes sense. The resulting kubeflow YAML is what is expected with this setup.

I've also updated the releasing and test scripts so that the image tags are updated in [katib-config.yaml](https://github.com/kubeflow/katib/pull/1498/files#diff-2e1aff4a1f1a5b4532e8a6a143f347aaee89ab48f4057d36499d25a00541aed7).

DavidSpek:
@andreyvelich Is there a way to retest Charmed Katib? Or would you have an idea why it is failing and how to fix it?

andreyvelich:
@DomFleischmann @knkski @rfmvasconcelos any ideas why Action Juju Charm can fail ?

manifests/v1beta1/installs/katib-cert-manager/kustomization.yaml
```diff
@@ -0,0 +1,57 @@
...
+  - name: docker.io/kubeflowkatib/katib-controller
+    newName: docker.io/kubeflowkatib/katib-controller
+    newTag: latest
+  - name: docker.io/kubeflowkatib/katib-db-manager
+    newName: docker.io/kubeflowkatib/katib-db-manager
+    newTag: latest
+  - name: docker.io/kubeflowkatib/katib-ui
+    newName: docker.io/kubeflowkatib/katib-ui
+    newTag: latest
+  - name: docker.io/kubeflowkatib/cert-generator
```
andreyvelich, line 18: I think we can remove this image since we are not using cert-generator image in the cert-manager install.

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
```diff
@@ -2,51 +2,32 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
     newTag: latest
+  - name: docker.io/kubeflowkatib/katib-db-manager
     newName: docker.io/kubeflowkatib/katib-db-manager
-  - name: docker.io/kubeflowkatib/katib-ui
     newTag: latest
+  - name: docker.io/kubeflowkatib/katib-ui
     newName: docker.io/kubeflowkatib/katib-ui
-  - name: docker.io/kubeflowkatib/cert-generator
     newTag: latest
+  - name: docker.io/kubeflowkatib/cert-generator
```
andreyvelich, line 19: I think we can remove this image since we are not using cert-generator image in the Katib with Kubeflow install.

andreyvelich (commented):
Thank you for this update @DavidSpek.
I left few comments.

manifests/v1beta1/installs/katib-cert-manager/kustomization.yaml
DavidSpek: Good catch!

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
DavidSpek: Good catch! Just pushed a commit with these removed.

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
andreyvelich: Let's just remove these labels.
If we require them in the future we will update the Kustomization file.

manifests/v1beta1/installs/katib-cert-manager/kustomization.yaml
```diff
@@ -0,0 +1,54 @@
...
+    newName: docker.io/kubeflowkatib/katib-controller
+    newTag: latest
+  - name: docker.io/kubeflowkatib/katib-db-manager
+    newName: docker.io/kubeflowkatib/katib-db-manager
+    newTag: latest
+  - name: docker.io/kubeflowkatib/katib-ui
+    newName: docker.io/kubeflowkatib/katib-ui
+    newTag: latest
+
+patchesStrategicMerge:
```
andreyvelich, line 19: I think we should also create a patch to remove Katib Cert Generator RBAC: https://github.com/DavidSpek/katib/blob/fix-manifests/manifests/v1beta1/components/webhook/rbac.yaml.
ServiceAccount, ClusterRole, ClusterRoleBinding.
This RBAC is used for the `katib-cert-generator` Job.

DavidSpek: What about https://github.com/DavidSpek/katib/blob/fix-manifests/manifests/v1beta1/components/webhook/webhooks.yaml?

andreyvelich: These webhooks we should create in any install.
I believe we just use this patch: https://github.com/kubeflow/katib/blob/7054dff3fb1e29bb493c65dc6165c856a0128987/manifests/v1beta1/installs/katib-cert-manager/patches/katib-cert-injection.yaml#L6 to fill the `caBundle`.

DavidSpek: Cool, just wanted to be sure :).

DavidSpek: Just pushed a commit with these changes.

andreyvelich: Thanks @DavidSpek!
From my side only this comment left: https://github.com/kubeflow/katib/pull/1498#discussion_r611075289.

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
DavidSpek: @andreyvelich From what I saw all the selectors were setup properly so I removed this now.

andreyvelich:
Overall looks good to me, thank you @DavidSpek!
/lgtm
Please take a look @yanniszark @gaocegege @johnugeorge

yanniszark:
Thanks a lot for the effort @DavidSpek @andreyvelich.
I took a quick look and some small comments I have are:
- It would be good if the patches were declared in a universal way (separate files). Keep in mind that you can include multiple patches in one file, using `patchesStrategicMerge`, if the context is similar.
- Instead of patching to remove all cert-generator resources, how about creating a separate kustomization for the cert-generator?

And of course, before merging, I would validate that the result of each new kustomization is the same as the result of the respective previous kustomization.

DavidSpek:
Makes sense to me. @andreyvelich Let me know what you prefer and I'll make the needed changes.

andreyvelich:
I agree with @yanniszark.

Let's have 2 separate folders for the Webhooks:
1. `cert-generator` 
2. `webhook`

DavidSpek:
@andreyvelich I've just pushed a commit in which I have separated `webhook` from `cert-generator` and I'm not using `katib-standalone` as a base for `katib-cert-manager` to avoid needing all the patches that delete the `cert-generator` resources.

manifests/v1beta1/installs/katib-with-kubeflow/patches/remove-resources-patch.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: katib-mysql
+---
+$patch: delete
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: kubeflow
```
andreyvelich, line 11: Please add the new line.

andreyvelich:
Thank you @DavidSpek.
If the changes look good to you @yanniszark, can you add lgtm, please ?

manifests/v1beta1/installs/katib-with-kubeflow/patches/remove-resources-patch.yaml
DavidSpek: My bad, fixed it now.

gaocegege (commented):
/lgtm
/approve

Issue #1501, opened 2021-03-30
manifests: Disable Istio Injection


yanniszark:
/kind bug

**What steps did you take and what happened:**
In previous releases, Istio injection was disabled. At the moment, Katib won't work with Istio Injection enabled. Thus, add `sidecar.istio.io/inject: "false"` annotations in all workloads.

Issue #1500, opened 2021-03-30
manifests: Cert-generator script doesn't work for k8s < 1.18


yanniszark:
/kind bug

**What steps did you take and what happened:**
Installing the Katib manifests in K8s 1.16 makes the cert-generator script fail with:
```
error: error validating "STDIN": error validating data: ValidationError(CertificateSigningRequest.spec): unknown field "signerName" in io.k8s.api.certificates.v1beta1.CertificateSigningRequestSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

This is a client-side validation error because the `CertificateSigningRequest` API does not contain the `signerName` field in 1.16 (needs >= 1.18). @andreyvelich tried the script with `--validate=false` in 1.16 KIND and it worked. We also tried it with a cert-manager Certificate and it worked.

**Next Steps:**
- Add `--validate=false` to cert-generator script
- Add a `katib-with-kubeflow-cert-manager` overlay, which uses cert-manager to create the Certificate

Issue #1499, opened 2021-03-30
manifests: Kubeflow overlay has PVC with storageClassName katib


yanniszark:
/kind bug

**What steps did you take and what happened:**
The kubeflow overlay imports a PVC with storageClassName `katib`, which does not exist. This also ties in with the discussion in issue https://github.com/kubeflow/katib/issues/1479.

**What did you expect to happen:**
The PVC should not have a `storageClassName` set, so that it gets the cluster's default one.

**Anything else you would like to add:**
Since we need this for the 1.3 release, we will add a patch in the kubeflow overlay, to make it correct.
This is instead of solving https://github.com/kubeflow/katib/issues/1479, which needs additional changes in docs.

PR #1502, opened 2021-03-30
Katib manifests fixes for 1.3

yanniszark:

Fixes https://github.com/kubeflow/katib/issues/1499
Fixes https://github.com/kubeflow/katib/issues/1500
Fixes https://github.com/kubeflow/katib/issues/1501

@andreyvelich I made the changes needed for Kubeflow 1.3, as the aforementioned issues show.
In addition, I made sure that the VirtualService is deployable by default, as other components' manifests do.
After merging, I will create a cherry-pick PR.

### DIFF
--- a/hack/cert-generator.sh
+++ b/hack/cert-generator.sh
@@ -68,7 +68,9 @@ fi
 set -e
 echo "INFO: Creating CSR: ${csr_name}"
 
-cat <<EOF | kubectl create -f -
+# signerName is not supported in Kubernetes <= 1.17
+# See: https://github.com/kubeflow/katib/issues/1500
+cat <<EOF | kubectl create --validate=false -f -
 apiVersion: certificates.k8s.io/v1beta1
 kind: CertificateSigningRequest
 metadata:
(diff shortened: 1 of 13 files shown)


hack/cert-generator.sh
```diff
@@ -68,7 +68,7 @@ fi
 set -e
 echo "INFO: Creating CSR: ${csr_name}"
 
-cat <<EOF | kubectl create -f -
+cat <<EOF | kubectl create --validate=false -f -
```
andreyvelich, line 71: Please can you add the comment above.
```suggestion
# signerName is not supported in Kubernetes <= 1.17
cat <<EOF | kubectl create --validate=false -f -
```

andreyvelich (commented):
Thank you for this fix @yanniszark!
Just one small comment, let's keep track this validation flag.
/cc @gaocegege @johnugeorge

hack/cert-generator.sh
yanniszark: Makes sense! I did even better, and also added the issue 😛

andreyvelich (commented):
/lgtm
Please take a look @gaocegege @johnugeorge

gaocegege (commented):
/lgtm

PR #1525, opened 2021-04-29
Remove IBM install from Katib manifest 

andreyvelich:
See discussion: https://github.com/kubeflow/manifests/pull/1823#discussion_r616799381.
We don't need to support this install since IBM is using the `katib-with-kubeflow` install.

/cc @moficodes @Tomcli @gaocegege @johnugeorge


moficodes (approved):
/lgtm

Issue #1479, opened 2021-03-17
[Discussion] Inclusion PV for the MySQL in Katib standalone manifests


andreyvelich:
/kind feature

See discussion: https://github.com/kubeflow/katib/pull/1464#issuecomment-796876215.
Currently we exclude PV for MySQL in [`katib-with-kubeflow`](https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml#L18) installation and include PV in [`katib-standalone`](https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-standalone/kustomization.yaml#L23) installation.

Our users faced with many problems when they are using custom volume provisioner: https://github.com/kubeflow/katib/issues/1415, https://github.com/kubeflow/katib/issues/1156, https://github.com/kubeflow/katib/issues/1212.

We should discuss what is the easiest and stable way for the users to install Katib.

/cc @anneum @hjkkhj123 @Utkagr

heilerich:
One of your users here.

I think you don't have any problem with volume provisioners. I think what's causing all the issues you linked is this overzealous liveness probe.
https://github.com/kubeflow/katib/blob/f83df1b845fc1d7f5d736176a7487cd5ea28604e/manifests/v1beta1/components/mysql/mysql.yaml#L49-L57

In environments that have slightly slower storage, the mysql container get's killed in its initialisation phase and never recovers from that. I would suggest increasing the `initialDelaySeconds` or creating an additional `startupProbe` as per the official kubernetes docs (see [Protect slow starting containers with startup probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes)). Doing that should resolve all your users current problems.

See [[1]](https://github.com/kubeflow/katib/issues/1415#issuecomment-784104872), [[2]](https://github.com/kubeflow/katib/issues/1156#issuecomment-616899567). Also I can confirm from personal experience that the empty root password issue (#1212) with the mysql container is also cause by the container being killed before it has gotten rid of the temporary initialisation database.

Regarding providing an "easy and **stable**" way for your users to deploy katib: You should seriously reconsider supplying a static hostPath PV in *any* circumstance. As I understand it the mysql database does not store ephemeral data in this case. Supplying a static hostPath PV creates a bunch of problems:
* Scalability/Resource consumption: The database cannot be evicted/migrate to another host without data loss
[...]

PR #1527, opened 2021-05-04
Remove PV from MySQL component

andreyvelich:
Related: https://github.com/kubeflow/katib/issues/1479.

I removed PV from the MySQL component as we decided.
Also, I added `startupProbe` and left only `periodSeconds` and `failureThreshold` in the MySQL deployment as @heilerich proposed here: https://github.com/kubeflow/katib/issues/1479#issuecomment-804912359.

For the `katib-db-manager` I removed `readinessProbe`, we can add it later if we need it.

Please take a look @heilerich @maanur @gaocegege @johnugeorge @yanniszark @DavidSpek


manifests/v1beta1/components/mysql/kustomization.yaml
```diff
@@ -2,8 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- mysql.yaml
-- pv.yaml
```
gaocegege, line 6: Will we use the in-container FS to keep the db files?

andreyvelich: @gaocegege We mount PVC to the container FS here: https://github.com/kubeflow/katib/blob/bc7c1e19ee18e080b97b9ca0ed3d144c57a30c73/manifests/v1beta1/components/mysql/mysql.yaml#L68

It depends on the user's StorageClass, how it will store once container is deleted.

gaocegege (commented):
LGTM :+1:
/lgtm

PR #1552, opened 2021-06-04
Disable default PV for Experiment with resume from volume

andreyvelich:
Related: https://github.com/kubeflow/katib/issues/1479.

I disabled default PV creation for Experiment with from volume `resumePolicy`.

As discussed here: https://github.com/kubeflow/katib/issues/1479#issuecomment-805319555, it is very uncommon to create custom PV for the Experiment.

/cc @gaocegege @johnugeorge @maanur @heilerich


gaocegege:
LGTM

Issue #1559, opened 2021-06-17
Controller should allow an empty CPU resource limit


sravi999:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
The controller should allow an empty CPU resources limit when the CPU limit is empty. 

**What did you expect to happen:**
If the CPU limit is empty don't set a default value for the CPU. 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**
AWS EKS
- Kubeflow version (`kfctl version`): V1.1
- Minikube version (`minikube version`):
- Kubernetes version: (use `kubectl version`): V1.17
- OS (e.g. from `/etc/os-release`):

sravi999:
cc: @andreyvelich

andreyvelich:
Thank you for creating this @sravi999.
For all resources we can follow the Ephemeral Storage policy to remove them: https://github.com/kubeflow/katib/blob/master/pkg/util/v1beta1/katibconfig/config.go#L291.

/assign @andreyvelich

PR #1564, opened 2021-06-26
Allow empty resources for CPU and Memory in Katib config

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1559.
This will allow to specify empty resources for CPU and Memory in Katib config.

/cc @gaocegege @johnugeorge @sravi999


johnugeorge:
/lgtm
Do we have any tests just to check this?

PR #1492, opened 2021-03-20
Fix gRPC manager build script

andreyvelich:
We are currently unable to re-run `build.sh` script to generate gRPC manager APIs.
I am getting this error:
```
:/apiprotos protoc-gen-doc --doc_opt=markdown,api.md -I /protobuf -I /apiprotos api.proto
[libprotobuf WARNING ../../../../../src/google/protobuf/compiler/parser.cc:562] No syntax specified for the proto file: google/api/annotations.proto. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version. (Defaulted to proto2 syntax.)
google/api/annotations.proto:1:1: Expected top-level statement (e.g. "message").
api.proto: Import "google/api/annotations.proto" was not found or had errors.
```

As I can see, the annotations and `grpc-gateway_out` scripts were added as part of this PR: https://github.com/kubeflow/katib/pull/142.
We are not using REST manager more and that should be removed.
As well, I removed legacy Manager Swagger and Makefile.

/assign @gaocegege @johnugeorge

PR #1569, opened 2021-07-01
Add gRPC build script to CI

andreyvelich:
To avoid problems like this: https://github.com/kubeflow/katib/pull/1492, I added gRPC build script to the CI.


gaocegege (approved):
LGTM

Issue #1567, opened 2021-06-29
Update Katib SDK to be compatible with kfp-tekton


peterhaddad3121:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
Hello, I am utilizing Kubeflow with Kabit and kfp-tekton.

It is known that the Kabit SDK requires Kubernetes 10.0.1 and kfp-tekton requires 1.18 for Kubernetes.

**What did you expect to happen:**

I expect Kabit to be compatible with a later version of Kubernetes

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version (`kfctl version`): kfctl v1.2.0-0-gbc038f9
- Minikube version (`minikube version`): No using Minikube 
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

NAME="Ubuntu"
VERSION="18.04.5 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.5 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic

andreyvelich:
Thank you for creating this @peterhaddad3121.
Please try to use latest version of Katib SDK: 0.11.0.
In that version we should support the latest Kubernetes version.

peterhaddad3121:
Hi Audrey, I did try to use the latest version. Additionally when I install it says the following: Successfully installed `kubeflow-katib-0.11.0 kubernetes-10.0.1`

I run a kubernetes uninstall which removes kubernetes, but then I get the following error:
`AttributeError: 'V1beta1Experiment' object has no attribute 'openapi_types'`

This error gets resolved when downgrading back to the older kubernetes version. Not sure if this helps or not, but I am not too sure what else to troubleshoot.

Thank you for this!

andreyvelich:
@peterhaddad3121 What if you try to use `kubeflow-katib-0.11.0` Python package with the newer version of `kubernetes` package ?
Are you getting the error above ?

PR #1572, opened 2021-07-06
Update Katib SDK with OpenAPI generator

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1567.
I updated Katib SDK to be compatible with the latest Python Kubernetes Client version. I am using OpenAPI Generator instead of Swagger. Check this issue: https://github.com/kubernetes-client/python/issues/1112.

I modified requirements with `kubernetes>=12.0.0` version since KFP is using [`12.0.1`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/requirements.txt#L71) version.

I removed test from SDK since we are not using them right now and we need to investigate how to generate them properly.

/cc @jinchihe @peterhaddad3121 @gaocegege @johnugeorge


andreyvelich:
Hi @DomFleischmann, @knkski @rfmvasconcelos, do you have any ideas why Charmed Tests can failed ?

PR #1573, opened 2021-07-06
Remove Job Level Mutation from Katib Controller

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1320.
I think we can safely remove Job Level mutation from the controller.
The issue is created almost 1 year ago and our users don't complain about it.

/cc @gaocegege @johnugeorge


gaocegege (commented):
/lgtm

PR #1574, opened 2021-07-08
Switch to charmcraft snap for CI

knkski:
The snap store is now the preferred place for obtaining charmcraft, as opposed to pypi.org.

Fixes broken CI process


andreyvelich (approved):
Thank you for this fix @knkski!
/lgtm
/approve

PR #1575, opened 2021-07-09
Allow make pre-release tag

andreyvelich:
This allows make `v0.12.0-rc.0` releases for Katib.

Also, I fixed name for `katib-openshift` install.

/cc @gaocegege @johnugeorge @yanniszark


gaocegege (commented):
/lgtm

PR #1578, opened 2021-07-15
Fix grep in Tekton Experiment Doc

andreyvelich:
Our image names have been changed.
This command should be modified also.
/cc @gaocegege @johnugeorge


gaocegege (commented):
LGTM :+1:
/lgtm

PR #1593, opened 2021-07-28
Install charmcraft 1.0.0

DomFleischmann:
Install charmcraft 1.0.0 through pip for now until we move to pytest-operator tests.


andreyvelich (approved):
Thank you for fixing this @DomFleischmann!
/lgtm
/approve

PR #1594, opened 2021-07-29
fix mysql version in docker image

munagekar:

- This PR fixes the mysql version to a constant tag.

The current image `mysql:8` is mutable tag. Every time an update is made this points to a new version. I.e it could point to 8.0.1 or 8.0.2 or 8.0.26. [SQL upgrades are one-way it is not possible to downgrade](https://www.percona.com/blog/2020/01/10/mysql-8-minor-version-upgrades-are-one-way-only/). 

This leads to errors if the image is already cached on kubernetes node. For example node already has 8.0.2 image cached, but 8.0.26 on was used on another node.
```
[InnoDB] Cannot boot server version 80017 on data directory built by version 80018. Downgrade is not supported
```

Another possible alternative would be to set image pull policy to Always.

```
imagePullPolicy: Always
```

### DIFF
--- a/manifests/v1beta1/components/mysql/mysql.yaml
+++ b/manifests/v1beta1/components/mysql/mysql.yaml
@@ -22,7 +22,7 @@ spec:
     spec:
       containers:
         - name: katib-mysql
-          image: mysql:8
+          image: mysql:8.0.26
           args:
             - --datadir
             - /var/lib/mysql/datadir


manifests/v1beta1/components/mysql/mysql.yaml
```diff
@@ -22,7 +22,7 @@ spec:
     spec:
       containers:
         - name: katib-mysql
-          image: mysql:8
+          image: mysql:8.0.26
```
andreyvelich, line 25: Can we also add `imagePullPolicy: Always` here to avoid problems with caching data if users are running another `mysql` instance for other components ?
I remember that we had couple of problems with that.

andreyvelich (commented):
Thank you for updating this @munagekar!
I think it's very good point, we should keep the particular version for `mysql`.
/ok-to-test

gaocegege:
/retest

LGTM :+1:
/lgtm

manifests/v1beta1/components/mysql/mysql.yaml
munagekar: `imagePullPolicy: Always` is not required as long `mysql:8.0.26` tag is not updated, and cache can be used if available. 

However, I don't have any strong opinion, I can add it if you feel so.

andreyvelich: Do we have any use-case when `mysql` team can update previous image tag on the dockerhub ?

munagekar: Looking at [release history](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/My-sql])  & [dockerhub](https://hub.docker.com/_/mysql?tab=tags&page=1&ordering=last_updated) Mysql Team hasn't done it. I think it should be fine without `imagePullPolicy:"Always"`

A possible usecase would be when Dockerimages such as [ubuntu](https://hub.docker.com/_/ubuntu?tab=tags&page=1&ordering=last_updated&name=18.04) are rebuilt and pushed again to dockerhub when there are security fixes. Mysql doesn't seem to do so either.

Even if there were to be a security fix with mysql docker image. The mysql version would remain the same and this issue would not occur.
```
[InnoDB] Cannot boot server version 80017 on data directory built by version 80018. Downgrade is not supported
```

andreyvelich: Thank you for the explanation @munagekar!

andreyvelich (approved):
/lgtm
/approve

Issue #1596, opened 2021-07-29
Remove Kubeflow Training imports from go.mod


andreyvelich:
As we discussed in our WG community meeting we can remove packages for `kubeflow/common` and `kubeflow/tf-operator` from Katib controller.
That will help us to decrease dependencies.
We are using this only in Validation webhook: https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/experiment/validator/validator.go#L350-L371 which is not necessary.

/cc @gaocegege @johnugeorge 
/area release
/priority p1

/assign @andreyvelich

gaocegege:
SGTM!

Issue #1343, opened 2020-09-18
Metrics unavailable Trial condition


andreyvelich:
/kind feature

Currently, if metrics are unavailable we switch Trial to `Succeeded` condition with `Status: False`.
To be more clear, we should create another condition: `MetricsUnavailable` and set Trial to that condition.

andreyvelich:
We might also include `MetricsUnavailable` condition to [`IsCompleted()`](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/experiments/v1beta1/util.go#L93) function.
Check the comment here: https://github.com/kubeflow/katib/pull/1599#issue-700445805.

PR #1599, opened 2021-07-30
Remove Kubeflow Training dependencies from Katib

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1596.
I removed Kubeflow Training packages from Katib source code and made appropriate changes.

We should keep `k8s.io/code-generator` in `go.mod` to make sure that our generate scripts are running.

Also, I included `!instance.IsMetricsUnavailable()` check in [`reconcileTrial`](https://github.com/andreyvelich/katib/blob/48c69e454197c506160f9cb04eef662a0477e7df/pkg/controller.v1beta1/trial/trial_controller.go#L235). We should not try to update Trial status after Trial is in `MetricsUnavailable` status.

In addition to that, we can include `MetricsUnavailable` status in `IsCompleted()` Trial check.
I believe, we are not doing that currently because after Trial is complete, we [delete Trial's resources](https://github.com/andreyvelich/katib/blob/48c69e454197c506160f9cb04eef662a0477e7df/pkg/controller.v1beta1/trial/trial_controller.go#L304-L313). Maybe user wants to see why metrics have not been reported.
On the other hand, users can use `RetainRun` parameter to control it, so from my side [`IsCompleted()`](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/trials/v1beta1/util.go#L96) check should have `MetricsUnavailable` condition.
Any thoughts @gaocegege @johnugeorge ?

Let's discuss about it here: https://github.com/kubeflow/katib/issues/1343.


gaocegege (commented):
LGTM :+1:
/lgtm

PR #1609, opened 2021-08-09
Remove TFJob and PyTorchJob CRDs from unit tests

andreyvelich:
Since we are not using TFJob and PyTorchJob in our unit tests more, we can remove these CRDs templates.


gaocegege (commented):
/lgtm
/approve

PR #1611, opened 2021-08-10
Modify Labels for Katib Components

andreyvelich:
See discussion: https://github.com/kubeflow/common/issues/148#issuecomment-893826133.

I modified labels for Katib components to follow Kubernetes guidelines.

Probably, we should also think about renaming [`katib-metricscollector-injection: enabled`](https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/components/namespace/namespace.yaml#L6) to `katib.kubeflow.org/metricscollector-injection: true`

/cc @gaocegege @johnugeorge @alculquicondor

### DIFF
--- a/pkg/controller.v1beta1/consts/const.go
+++ b/pkg/controller.v1beta1/consts/const.go
@@ -142,9 +142,9 @@ const (
 	AnnotationIstioSidecarInjectValue = "false"
 
 	// LabelTrialTemplateConfigMapName is the label name for the Trial templates configMap
-	LabelTrialTemplateConfigMapName = "app"
+	LabelTrialTemplateConfigMapName = "katib.kubeflow.org/component"
 	// LabelTrialTemplateConfigMapValue is the label value for the Trial templates configMap
-	LabelTrialTemplateConfigMapValue = "katib-trial-templates"
+	LabelTrialTemplateConfigMapValue = "trial-templates"
 
 	// TrialTemplateParamReplaceFormat is the format to make substitution in Trial template from Names in TrialParameters
 	// E.g if Name = learningRate, according value in Trial template must be ${trialParameters.learningRate}
--- a/examples/v1beta1/trial-configmap-source.yaml
+++ b/examples/v1beta1/trial-configmap-source.yaml
@@ -48,6 +48,6 @@ spec:
         description: Training model optimizer (sdg, adam or ftrl)
         reference: optimizer
     configMap:
-      configMapName: trial-template
+      configMapName: trial-templates
       configMapNamespace: kubeflow
       templatePath: defaultTrialTemplate.yaml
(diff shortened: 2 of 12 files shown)


alculquicondor:
Looks good, but I don't have enough familiarity to know if the presubmit failure is related.

pkg/controller.v1beta1/consts/const.go
```diff
@@ -142,9 +142,9 @@ const (
 	AnnotationIstioSidecarInjectValue = "false"
 
 	// LabelTrialTemplateConfigMapName is the label name for the Trial templates configMap
-	LabelTrialTemplateConfigMapName = "app"
+	LabelTrialTemplateConfigMapName = "katib.kubeflow.org/component"
```
gaocegege, line 145: Should we use katib.kubeflow.org or automl.kubeflow.org?

andreyvelich: I am fine with both, it's better to take what is easier for user to see the Katib components.
WDYT @johnugeorge @alculquicondor @kubeflow/wg-training-leads ?

alculquicondor: Are there other projects that would potentially use `automl.kubeflow.org/component`?

andreyvelich: Currently, we have only Katib project in Kubeflow for AutoML

alculquicondor: I think the "component" part might be exclusive to katib.

Of course, you could do `automl.kubeflow.org/component: katib-ui`, but maybe it's overkill.

alculquicondor:
sidetopic: That's quite a bit of retests. Is there any understanding of the reasons for the flakiness?

pkg/controller.v1beta1/consts/const.go
andreyvelich: I agree, so in the future if we have more projects for AutoML we can extend it.

andreyvelich:
> sidetopic: That's quite a bit of retests. Is there any understanding of the reasons for the flakiness?

There is a problem with Docker pull limitation. We are getting this error in the CI:
```
ERRO[0002] Error while retrieving image from cache: mxnet/python:latest_cpu_native_py3 GET https://index.docker.io/v2/mxnet/python/manifests/latest_cpu_native_py3: TOOMANYREQUES
TS: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
``

pkg/controller.v1beta1/consts/const.go
andreyvelich: As we decided on the WG Meeting we will use `katib.kubeflow.org` for now since it is easier to indicate project for the user. 
@gaocegege @alculquicondor @johnugeorge if you are fine with the changes, please leave your `lgtm`

alculquicondor: lgtm

terrytangyuan (approved):
/lgtm

Issue #1549, opened 2021-06-04
Support multi-objective optimization


anencore94:
/kind feature

**Describe the solution you'd like**
- katib currently use only single-objective optimization, (even though katib could display additional metrics, the optimizing process focused on single objective function)
- However, as you might know well, **optuna** support [multi-objective optimization](https://optuna.readthedocs.io/en/v2.7.0/tutorial/20_recipes/002_multi_objective.html)
  - which means, for example, we can optimize the hyperparameters to maximize accuracy and minimize time-cost simultaneously.

- In such case, I know we can define a new objective function which combines two objectives, such as maximize `accuracy / time-cost` or something else more complicated.
- However, I wonder is there any plan to support such feature natively in katib.

andreyvelich:
Thank you for creating this @anencore94.
I think that feature will require some API modification since `type`, `goal` and `objectiveMetricName` can be unique: https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/common/v1beta1/common_types.go#L94-L103.

@anencore94 @gaocegege @johnugeorge Any ideas how we can implement this ?
 
@c-bata Does Goptuna support Multi-object optimization ?

c-bata:
> @c-bata Does Goptuna support Multi-object optimization ?

No. I'm not planning to implement multi-objective support in Goptuna because it requires some breaking changes. 

As @anencore94 said, I think Optuna is a good choice if Katib will support multi-objective optimizations.
Optuna supports a lot of multi-objective optimization algorithms like NSGA-II, multi-objective TPE, and the algorithms provided by BoTorch. And it is not difficult to implement Optuna based suggestion service since its internal design is almost similar with Goptuna.

g-votte:
Hi, I'm one of the maintainers of Optuna. Thank you for referring to Optuna as a possible backend of the multi-objective API.

As @c-bata mentions, Optuna supports several algorithms for multi-objective optimization, with almost the same interface as single objective optimization.

I think we can go forward by (1) implementing the Optuna suggestion service, and (2) fixing the interface of Katib's multi-objective functionality.

For (1), I created the PR #1613 to add the suggestion service. I appreciate your taking a look and providing any feedback.

anencore94:
Can we close this issue by #1613 ?

andreyvelich:
> Can we close this issue by #1613 ?

This issue still is progress. If we want to support multi-objective optimisation, we have to modify our Katib Experiment APIs.

[...]

PR #1613, opened 2021-08-11
Add Optuna based suggestion service

g-votte:
This PR proposes to add the suggestion service based on [Optuna](https://github.com/optuna/optuna). Optuna provides several sampling algorithms that have not been implemented in other suggestion services in Katib, such as **[multi-variate TPE](https://tech.preferred.jp/en/blog/multivariate-tpe-makes-optuna-even-more-powerful/)** and **[constant liar](https://nni.readthedocs.io/en/stable/CommunitySharings/ParallelizingTpeSearch.html)**.

In addition, as discussed in #1549, Optuna can offer the extension of **multi-objective optimization** when Katib supports that interface.

I've written the example of invoking multi-variate TPE in a [separated repository](https://github.com/g-votte/katib-optuna-example), so that we can test and discuss the new algorithm based on the Optuna service. Multi-variate TPE captures the dependencies among multiple inputs and shows [better performances than normal TPE](https://tech.preferred.jp/en/blog/multivariate-tpe-makes-optuna-even-more-powerful/) in many benchmark tasks. If the example looks fine, I'd also like to add it in another PR.

- Example: https://github.com/g-votte/katib-optuna-example

**Checklist:**

### DIFF
--- a/cmd/suggestion/optuna/v1beta1/Dockerfile
+++ b/cmd/suggestion/optuna/v1beta1/Dockerfile
@@ -1,0 +1,32 @@
+FROM python:3.6
+
+ENV TARGET_DIR /opt/katib
+ENV SUGGESTION_DIR cmd/suggestion/optuna/v1beta1
+
+RUN if [ "$(uname -m)" = "ppc64le" ] || [ "$(uname -m)" = "aarch64" ]; then \
+    apt-get -y update && \
+    apt-get -y install gfortran libopenblas-dev liblapack-dev && \
+    pip install cython; \
+    fi
+
+RUN GRPC_HEALTH_PROBE_VERSION=v0.3.1 && \
+    if [ "$(uname -m)" = "ppc64le" ]; then \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-ppc64le; \
+    elif [ "$(uname -m)" = "aarch64" ]; then \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-arm64; \
+    else \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64; \
+    fi && \
+    chmod +x /bin/grpc_health_probe
+
+ADD ./pkg/ ${TARGET_DIR}/pkg/
+ADD ./${SUGGESTION_DIR}/ ${TARGET_DIR}/${SUGGESTION_DIR}/
+WORKDIR  ${TARGET_DIR}/${SUGGESTION_DIR}
+RUN pip install --no-cache-dir -r requirements.txt
+
+RUN chgrp -R 0 ${TARGET_DIR} \
+  && chmod -R g+rwX ${TARGET_DIR}
+
+ENV PYTHONPATH ${TARGET_DIR}:${TARGET_DIR}/pkg/apis/manager/v1beta1/python:${TARGET_DIR}/pkg/apis/manager/health/python
+
+ENTRYPOINT ["python", "main.py"]
--- a/cmd/suggestion/optuna/v1beta1/Dockerfile
+++ b/cmd/suggestion/optuna/v1beta1/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6
+FROM python:3.9
 
 ENV TARGET_DIR /opt/katib
 ENV SUGGESTION_DIR cmd/suggestion/optuna/v1beta1
--- a/cmd/suggestion/optuna/v1beta1/Dockerfile
+++ b/cmd/suggestion/optuna/v1beta1/Dockerfile
@@ -5,8 +5,7 @@ ENV SUGGESTION_DIR cmd/suggestion/optuna/v1beta1
 
 RUN if [ "$(uname -m)" = "ppc64le" ] || [ "$(uname -m)" = "aarch64" ]; then \
     apt-get -y update && \
-    apt-get -y install gfortran libopenblas-dev liblapack-dev && \
-    pip install cython; \
+    apt-get -y install gfortran libopenblas-dev liblapack-dev; \
     fi
 
 RUN GRPC_HEALTH_PROBE_VERSION=v0.3.1 && \
--- a/pkg/suggestion/v1beta1/optuna/service.py
+++ b/pkg/suggestion/v1beta1/optuna/service.py
@@ -141,7 +141,7 @@ class OptunaService(api_pb2_grpc.SuggestionServicer, HealthServicer):
 
     def _get_assignments_key(self, assignments):
         assignments = sorted(assignments, key=lambda a: a.name)
-        assignments_str = [str(a) for a in assignments]
+        assignments_str = [f"{a.name}:{a.value}" for a in assignments]
         return ",".join(assignments_str)
 
     def _get_optuna_search_space(self):
--- a/pkg/suggestion/v1beta1/optuna/service.py
+++ b/pkg/suggestion/v1beta1/optuna/service.py
@@ -147,7 +147,7 @@ class OptunaService(api_pb2_grpc.SuggestionServicer, HealthServicer):
                     # A new trial object is created and reported using study.add_trial() with the assignments and the search space.
                     optuna_trial = optuna.create_trial(
                         params={a.name:self._get_casted_assignment_value(a) for a in trial.assignments},
-                        distributions=self._get_optuna_serch_space(),
+                        distributions=self._get_optuna_search_space(),
                         value=value,
                     )
                     self.study.add_trial(optuna_trial)
--- a/pkg/suggestion/v1beta1/optuna/service.py
+++ b/pkg/suggestion/v1beta1/optuna/service.py
@@ -79,6 +79,7 @@ class OptunaService(api_pb2_grpc.SuggestionServicer, HealthServicer):
                     raise ValueError("Unknown name for {}: {}".format(name, k))
 
             kwargs["multivariate"] = name == "multivariate-tpe"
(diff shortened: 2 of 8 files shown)


g-votte:
@googlebot I signed it!

pkg/suggestion/v1beta1/optuna/service.py
```diff
@@ -0,0 +1,183 @@
...
+
+                if len(optuna_trial_numbers) != 0:
+                    # The trial has been suggested by the Optuna study.
+                    # The objective value is reported using study.tell() with the corresponding trial number.
+                    trial_number = optuna_trial_numbers.pop(0)
+                    self.study.tell(trial_number, value)
+                else:
+                    # The trial has not been suggested by the Optuna study.
+                    # A new trial object is created and reported using study.add_trial() with the assignments and the search space.
+                    optuna_trial = optuna.create_trial(
```
andreyvelich, line 149: Please can you tell me the use-case when "the Trial has not been suggested by the Optuna study"?

andreyvelich (commented):
Thank you for this awesome contribution @g-votte!
I left few comments.
Please can you also add one example, for instance of using multivariate TPE ?

@gaocegege @johnugeorge Since we are planing to cut the release on the next week, do we want to include this feature in Katib 0.12 ?

pkg/suggestion/v1beta1/optuna/service.py
g-votte: Thanks for the question. This block is to deal with the change of suggestion logic during an experiment, e.g. Optuna-based search after a certain number of bayesianoptimization trials, but does it happen in Katib?

If it is guaranteed that the suggestion service is unchangeable during an experiment, I will change the logic so that an error is raised instead of creating and adding Optuna trials.

gaocegege:

https://github.com/g-votte/katib-optuna-example is missing

gaocegege:
> Since we are planing to cut the release on the next week, do we want to include this feature in Katib 0.12 ?

This PR will not break existing features, thus I think we can have it. Maybe mark it alpha or beta, WDYT @andreyvelich

cmd/suggestion/optuna/v1beta1/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM python:3.6
+
+ENV TARGET_DIR /opt/katib
+ENV SUGGESTION_DIR cmd/suggestion/optuna/v1beta1
+
+RUN if [ "$(uname -m)" = "ppc64le" ] || [ "$(uname -m)" = "aarch64" ]; then \
+    apt-get -y update && \
+    apt-get -y install gfortran libopenblas-dev liblapack-dev && \
+    pip install cython; \
```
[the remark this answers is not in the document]
g-votte: Actually, we don't need Cython to run Optuna. I removed the Cython installation.

gaocegege (commented):
LGTM.

In this PR, additionalMetricNames is used to provide multiple metrics to katib, I think we should have a new CRD design to support it.

WDYT @andreyvelich @johnugeorge

pkg/suggestion/v1beta1/optuna/service.py
```diff
@@ -0,0 +1,178 @@
...
+                        params={a.name:self._get_casted_assignment_value(a) for a in trial.assignments},
+                        distributions=self._get_optuna_search_space(),
+                        value=value,
+                    )
+                    self.study.add_trial(optuna_trial)
+
+    def _get_assignments_key(self, assignments):
+        assignments = sorted(assignments, key=lambda a: a.name)
+        assignments_str = [str(a) for a in assignments]
+        return ",".join(assignments_str)
```
c-bata, line 154: [Question] I have two questions.

1. Does the string representation of Assigment object holds both of parameter names and parameter values?
1. You defined `assignments_to_optuna_number` as `defaultdict(list)`. I guess the reason why you use list is for duplicated hyperparameters, right?

c-bata (commented):
@g-votte Thank you for your pull request. Overall looks good 💯  I put some review comments.

johnugeorge:
@gaocegege our definition of `additionalMetricName` is different. This seems to be a new CRD design. 

Since this PR doesn't affect anything else, this can go in this release also.

pkg/suggestion/v1beta1/optuna/service.py
g-votte: 1. Yes; this should contain both the assignment's name and value, using the string representation of the Assignment class.
https://github.com/kubeflow/katib/blob/abbc9c9d34ef0b4ef6b165bb7dc3d6b91ccb5a1d/pkg/suggestion/v1beta1/internal/trial.py#L88-L89 

2. As you mention, this is to handle duplications of assignments.

cmd/suggestion/optuna/v1beta1/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM python:3.6
```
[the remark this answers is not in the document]
g-votte: Sounds good. I updated the Python version. Some dependencies in requirements.txt are also updated for that purpose.

Commit: https://github.com/kubeflow/katib/pull/1613/commits/a4ae6e2531071e48f14e994fe07dcdec3c047c15

gaocegege:
SGTM

g-votte:
@andreyvelich 
Thanks for your comment. I copied the example to `examples/v1beta1/`.
Commit: https://github.com/kubeflow/katib/pull/1613/commits/a119fb260215b825139411bb0f224757c4edc6d2

Please let me know if we should not include this until `katib-config.yaml` is updated. I can separate the PR in that case.

CC: @c-bata @gaocegege

pkg/suggestion/v1beta1/optuna/service.py
c-bata: Great! Sounds like it works 👍 

This is a very nit-picking comment but I'd say the following code is more clear to ensure that contains both the parameter name and the value (and a bit memory efficient).

```python
assignments_str = [f"{a.name}:{a.value}" for a in assignments]
```

pkg/suggestion/v1beta1/optuna/service.py
```diff
@@ -0,0 +1,178 @@
...
+                elif k == "ei_candidates":
+                    kwargs["n_ei_candidates"] = int(v)
+                elif k == "random_state":
+                    kwargs["seed"] = int(v)
+                else:
+                    raise ValueError("Unknown name for {}: {}".format(name, k))
+
+            kwargs["multivariate"] = name == "multivariate-tpe"
+
+            sampler = optuna.samplers.TPESampler(**kwargs)
```
c-bata, line 83: @g-votte I think it's reasonable to set `kwargs["constant_liar"] = True` by default. Because you know, Katib allows us to run distributed optimization easily. What do you think?

c-bata: Or we can provide an algorithm setting to set `constant_liar=True`.

andreyvelich:
> Please let me know if we should not include this until katib-config.yaml is updated. I can separate the PR in that case.

It's fine that we have this example, once we push `docker.io/kubeflowkatib/suggestion-optuna` to the Docker hub, we can update Katib config.

pkg/suggestion/v1beta1/optuna/service.py
andreyvelich: In that case, should we remove this part for now?
In the future if we decide to add this feature in the controller, we can extend the Suggestion logic.
What do you think @g-votte @c-bata ?

c-bata: > In that case, should we remove this part for now?

Agree. Goptuna suggestion service doesn't also support such a use case now.

g-votte: @andreyvelich @c-bata 
Thanks for your comments. I changed the logic so that it raises an error for unknown assignments. (I also changed the test case because the previous test passes the assignments externally created as the initial trials.)
Commit: https://github.com/kubeflow/katib/pull/1613/commits/1c0e15ff5b9f1b52abb9f6b984dde589f66074b4

pkg/suggestion/v1beta1/optuna/service.py
g-votte: Agree. I changed the line.
https://github.com/kubeflow/katib/pull/1613/commits/b2085a6340cc8e1ef57c70a8956facb296972e6b

pkg/suggestion/v1beta1/optuna/service.py
g-votte: Good catch. There is less reason to disable constant liar especially with parallel optimization.
Added a line to turn on the option by default.
https://github.com/kubeflow/katib/pull/1613/commits/1d8188534de9e10a028fa7253f33cbf5de7eea87

pkg/suggestion/v1beta1/optuna/service.py
```diff
@@ -0,0 +1,183 @@
...
+        Main function to provide suggestion.
+        """
+        with self.lock:
+            if self.study is None:
+                self.search_space = HyperParameterSearchSpace.convert(request.experiment)
+                self.study = self._create_study(request.experiment.spec.algorithm, self.search_space)
+
+        trials = Trial.convert(request.trials)
+
+        with self.lock:
```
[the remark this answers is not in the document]
g-votte: Agree; we don’t need the separated blocks. I merged them.

pkg/suggestion/v1beta1/optuna/service.py
```diff
@@ -0,0 +1,178 @@
...
+
+        study = optuna.create_study(sampler=sampler, direction=direction)
+
+        return study
+
+    def _create_sampler(self, algorithm_spec):
+        name = algorithm_spec.algorithm_name
+        settings = {s.name:s.value for s in algorithm_spec.algorithm_settings}
+
+        if name == "tpe" or name == "multivariate-tpe":
```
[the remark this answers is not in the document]
g-votte: Thanks for your information. Since changing those components is user-facing, I'd like to work on that in a separated PR.

g-votte:
@andreyvelich @c-bata @gaocegege @johnugeorge 
Thanks for your swift and detailed reviews!
I think all comments are reflected. PTAL.

andreyvelich:
Thank you for implementing this @g-votte!
/lgtm
Others please take a look @c-bata @gaocegege @johnugeorge.

c-bata:
LGTM :tada:

PR #1621, opened 2021-08-13
Modify labels for controller resources

andreyvelich:
To be consistent with this PR: https://github.com/kubeflow/katib/pull/1611, we should also change our labels for Trials and Suggestion Deployment's Pod.

Please take a look @alculquicondor @gaocegege @johnugeorge 

/hold for the review

### DIFF
--- a/pkg/controller.v1beta1/consts/const.go
+++ b/pkg/controller.v1beta1/consts/const.go
@@ -50,11 +50,11 @@ const (
 	ConfigTrialResources = "trial-resources"
 
 	// LabelExperimentName is the label of experiment name.
-	LabelExperimentName = "experiment"
+	LabelExperimentName = "katib.kubeflow.org/experiment"
 	// LabelSuggestionName is the label of suggestion name.
-	LabelSuggestionName = "suggestion"
+	LabelSuggestionName = "katib.kubeflow.org/suggestion"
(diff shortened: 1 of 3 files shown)


pkg/controller.v1beta1/consts/const.go
```diff
@@ -50,11 +50,11 @@ const (
...
 
 	// LabelExperimentName is the label of experiment name.
-	LabelExperimentName = "experiment"
+	LabelExperimentName = "katib.kubeflow.org/experiment"
 	// LabelSuggestionName is the label of suggestion name.
-	LabelSuggestionName = "suggestion"
+	LabelSuggestionName = "katib.kubeflow.org/suggestion"
 	// LabelDeploymentName is the label of deployment name.
-	LabelDeploymentName = "deployment"
+	LabelDeploymentName = "katib.kubeflow.org/deployment"
```
gaocegege, line 57: Will it be released in next version?

andreyvelich: I think it's fine to release it in 0.12, since it's not breaking change.
These labels are used internally in Katib Controller.

gaocegege: SGTM.

/lgtm

PR #1623, opened 2021-08-14
Modify XGBoostJob example for the new Controller

andreyvelich:
Since we will release Kubeflow Training operator within Kubeflow 1.4, let's make appropriate change for XGBoost example in Katib.

/cc @kubeflow/wg-training-leads

### DIFF
--- a/examples/v1beta1/xgboost-lightgbm.yaml
+++ b/examples/v1beta1/xgboost-lightgbm.yaml
@@ -58,7 +58,7 @@ spec:
                     image: docker.io/kubeflowkatib/xgboost-lightgbm:1.0
                     ports:
                       - containerPort: 9991
-                        name: xgboost-port
+                        name: xgboostjob-port
                     imagePullPolicy: Always
                     args:
                       - --job_type=Train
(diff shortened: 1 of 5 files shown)


examples/v1beta1/xgboost-lightgbm.yaml
```diff
@@ -59,11 +54,11 @@ spec:
             template:
               spec:
                 containers:
-                  - name: xgboostjob
+                  - name: xgboost
                     image: docker.io/kubeflowkatib/xgboost-lightgbm:1.0
                     ports:
                       - containerPort: 9991
-                        name: xgboostjob-port
+                        name: xgboost-port
```
Jeffwan, line 61: https://github.com/kubeflow/tf-operator/blob/08c0d0374cada2c18d19ce9fc07bb598e62b6ccf/pkg/apis/xgboost/v1/constants.go#L22

default container port name is still `xgboostjob-port`?

terrytangyuan (commented):
LGTM besides @Jeffwan's comment above

examples/v1beta1/xgboost-lightgbm.yaml
andreyvelich: You are right, thanks!

gaocegege (commented):
Thanks for your contribution!

LGTM

Jeffwan (approved):

PR #1629, opened 2021-08-18
new-ui: Add devDependency - prettier

seong7:
What : 
It adds 'Prettier' as devDependency to `package.json` of 'new-ui/frontend' app.

Why :
Currently the style formatter 'Prettier' is not included in the `package.json`. Which makes the style formatting of 'new-ui/frontend' app on local dev environment depend on whether `Prettier` is installed globally on each developer's computer. It may confuse them while setting the IDE integration for style formatting.
Even though it's being checked on Github Actions, checking and fixing the style errors on Actions take way much longer than formatting on local with features of IDE such as 'auto format on save'.
So I suggest adding `Prettier` to the project's `package.json` so every developer can easily set the style formatting process on their local.

none

**Checklist:**

### DIFF
--- a/.github/workflows/test-node.yaml
+++ b/.github/workflows/test-node.yaml
@@ -20,5 +20,5 @@ jobs:
 
       - name: Run Node test
         run: |
-          npm install --global prettier@2.2.0
+          npm install --global prettier@2.3.2
           make prettier-check
--- a/.github/workflows/test-node.yaml
+++ b/.github/workflows/test-node.yaml
@@ -20,5 +20,5 @@ jobs:
 
       - name: Run Node test
         run: |
-          npm install --global prettier@2.3.2
+          cd pkg/new-ui/v1beta1/frontend && npm install prettier
           make prettier-check
--- a/.github/workflows/test-node.yaml
+++ b/.github/workflows/test-node.yaml
@@ -20,5 +20,6 @@ jobs:
 
       - name: Run Node test
         run: |
-          cd pkg/new-ui/v1beta1/frontend && npm install prettier
+          version=$(cd pkg/new-ui/v1beta1/frontend && node -p "require('prettier/package.json').version")
+          npm install --global prettier@${version}
           make prettier-check
--- a/.github/workflows/test-node.yaml
+++ b/.github/workflows/test-node.yaml
@@ -20,6 +20,5 @@ jobs:
 
       - name: Run Node test
         run: |
-          version=$(cd pkg/new-ui/v1beta1/frontend && node -p "require('prettier/package.json').version")
-          npm install --global prettier@${version}
+          npm install prettier --prefix ./pkg/new-ui/v1beta1/frontend
           make prettier-check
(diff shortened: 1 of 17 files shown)


gaocegege (commented):
/ok-to-test

andreyvelich (commented):
Thank you for this update @seong7!
Can we also update GitHub actions: https://github.com/kubeflow/katib/blob/master/.github/workflows/test-node.yaml#L23, to take prettier version from this config ?

andreyvelich:
cc @kimwnasptd

seong7:
> Thank you for this update @seong7!
Can we also update GitHub actions: https://github.com/kubeflow/katib/blob/master/.github/workflows/test-node.yaml#L23, to take prettier version from this config ?

Hi @andreyvelich :)
I just pushed a new commit that update prettier version on Github Actions.

.github/workflows/test-node.yaml
```diff
@@ -20,5 +20,5 @@ jobs:
 
       - name: Run Node test
         run: |
-          npm install --global prettier@2.2.0
+          npm install --global prettier@2.3.2
```
andreyvelich, line 23: Can we try to take version from package.json ?
For example: `(cd pkg/new-ui/v1beta1/frontend && npm install prettier)`
I think it will be easier to maintain.
WDYT @seong7 ?

andreyvelich: If this will not work (with `--global` flag I unable to install specific version from `package.json`), try to use this command:
```
version=$(cd pkg/new-ui/v1beta1/frontend && node -p "require('prettier/package.json').version")
npm install --global prettier@${version}
```

andreyvelich:
Also, please modify this command: https://github.com/kubeflow/katib/blob/master/Makefile#L82-L83.
We have to use `pkg/new-ui/v1beta1/frontend` prefix there.

andreyvelich:
@seong7 I've tried to make some tests: https://github.com/andreyvelich/katib/pull/8.
As I can see, `npm install prettier --prefix ./pkg/new-ui/v1beta1/frontend` in GitHub Actions should work without `--global` flag.
Please try to make these changes.

seong7:
> @seong7 I've tried to make some tests: [andreyvelich#8](https://github.com/andreyvelich/katib/pull/8).
> As I can see, `npm install prettier --prefix ./pkg/new-ui/v1beta1/frontend` in GitHub Actions should work without `--global` flag.
> Please try to make these changes.

@andreyvelich Thank you so much for your work 🙏 I was stuck on that issue 😅 
Your solution worked on my forked repo too: [seong7#1](https://github.com/seong7/katib/pull/1).
Hope we can see it works here this time!

andreyvelich (commented):
Thank you for this update @seong7!
/lgtm
/assign @kimwnasptd Please take a look.

seong7:
Can we get updated on the progress? 😃

andreyvelich (approved):
> Can we get updated on the progress? 😃

Yes, sorry for the delay. I think we can merge it.
/approve

PR #1630, opened 2021-08-18
SDK: change list apis to return objects as default

### LINKED ISSUE #1447: [SDK] list_trials and list_experiments should return list of objects
/kind feature

See discussion https://github.com/kubeflow/katib/pull/1442#discussion_r579503300.
`list_trials` and `list_experiments` should return list of corresponding objects instead of names and statuses to give user more functionality.

/cc @gaocegege @johnugeorge @Adarsh2910

anencore94:
- change list_trials, list_experiments to return list of objects as a default

Fixes #1447

**Checklist:**

**Questions To Members**
- Is there any recommended way to add unit test codes for python sdk, except this [example notebook](https://github.com/kubeflow/katib/tree/master/examples/v1beta1/sdk)?

### DIFF
--- a/hack/gen-python-sdk/post_gen.py
+++ b/hack/gen-python-sdk/post_gen.py
@@ -50,6 +50,8 @@ def _rewrite_helper(input_file, output_file, rewrite_rules):
         lines.append("from kubernetes.client import V1ListMeta\n")
         lines.append("from kubernetes.client import V1Container\n")
         lines.append("from kubernetes.client import V1HTTPGetAction\n")
+        lines.append("from kubernetes.client import V1ManagedFieldsEntry\n")
+        lines.append("from kubernetes.client import V1OwnerReference\n")
 
     with open(output_file, 'w') as f:
         f.writelines(lines)
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -14,10 +14,11 @@
 
 import multiprocessing
 
-from kubernetes import client, config
-
+from kubeflow.katib import V1beta1Experiment
+from kubeflow.katib import V1beta1Trial
 from kubeflow.katib.constants import constants
 from kubeflow.katib.utils import utils
+from kubernetes import client, config
 
 
 class KatibClient(object):
@@ -245,16 +246,14 @@ class KatibClient(object):
                 "Exception when calling CustomObjectsApi->delete_namespaced_custom_object:\
          %s\n" % e)
 
-    def list_experiments(self, namespace=None, in_short=False):
+    def list_experiments(self, namespace=None):
         """List all Katib Experiments.
 
         :param namespace: Experiments namespace.
         If the namespace is None, it takes "default" namespace.
-        :param in_short: List Experiments in summary or not
-        If the in_short is True, it will shows only name and status of each Experiments.
 
-        :return: List of Experiment names with the statuses.
-        :rtype: list[dict]
+        :return: List of Experiment objects.
+        :rtype: list[V1beta1Experiment]
         """
         if namespace is None:
             namespace = utils.get_default_target_namespace()
@@ -269,16 +268,17 @@ class KatibClient(object):
         katibexp = None
         try:
             katibexp = thread.get(constants.APISERVER_TIMEOUT)
-            result = []
-            if in_short:
-                for i in katibexp.get("items"):
-                    output = {}
-                    output["name"] = i.get("metadata", {}).get("name")
-                    output["status"] = i.get("status", {}).get(
-                        "conditions", [])[-1].get("type")
-                    result.append(output)
-            else:
-                result = katibexp.get("items")
+            result = [
+                V1beta1Experiment(
+                    api_version=item.get("apiVersion"),
+                    kind=item.get("kind"),
+                    metadata=item.get("metadata"),
+                    spec=item.get("spec"),
+                    status=item.get("status")
+                )
+                for item in katibexp.get("items")
+            ]
+
         except multiprocessing.TimeoutError:
             raise RuntimeError("Timeout trying to get katib experiment.")
         except client.rest.ApiException as e:
@@ -322,17 +322,15 @@ class KatibClient(object):
             name, namespace=namespace)
         return experiment_status.lower() == "succeeded"
 
-    def list_trials(self, name=None, namespace=None, in_short=False):
+    def list_trials(self, name=None, namespace=None):
         """List all Experiment's Trials.
 
         :param name: Experiment name.
         :param namespace: Experiments namespace.
         If the namespace is None, it takes "default" namespace.
-        :param in_short: List Trials in summary or not
-        If the in_short is True, it will shows only name and status of each Trials.
 
-        :return: List of Trial names with the statuses.
-        :rtype: list[dict]
+        :return: List of Trial objects
+        :rtype: list[V1beta1Trial]
         """
         if namespace is None:
             namespace = utils.get_default_target_namespace()
@@ -347,17 +345,16 @@ class KatibClient(object):
         katibtrial = None
         try:
             katibtrial = thread.get(constants.APISERVER_TIMEOUT)
-            result = []
-            if in_short:
-                for i in katibtrial.get("items"):
-                    output = {}
-                    if i.get("metadata", {}).get("ownerReferences")[0].get("name") == name:
-                        output["name"] = i.get("metadata", {}).get("name")
-                        output["status"] = i.get("status", {}).get(
-                            "conditions", [])[-1].get("type")
-                        result.append(output)
-            else:
-                result = katibtrial.get("items")
+            result = [
+                V1beta1Trial(
+                    api_version=item.get("apiVersion"),
+                    kind=item.get("kind"),
+                    metadata=item.get("metadata"),
+                    spec=item.get("spec"),
+                    status=item.get("status")
+                )
+                for item in katibtrial.get("items")
+            ]
         except multiprocessing.TimeoutError:
             raise RuntimeError("Timeout trying to getkatib experiment.")
         except client.rest.ApiException as e:
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -46,7 +46,7 @@ class KatibClient(object):
             self.in_cluster = True
 
         self.api_instance = client.CustomObjectsApi()
-        self.katib_api_client = api_client.ApiClient()
+        self.deserializer = utils.Deserializer()
 
     def _is_ipython(self):
         """Returns whether we are running in notebook."""
@@ -270,7 +270,7 @@ class KatibClient(object):
         try:
             katibexp = thread.get(constants.APISERVER_TIMEOUT)
             result = [
-                self.katib_api_client.deserialize_data(item, V1beta1Experiment)
+                self.deserializer.deserialize(item, V1beta1Experiment)
                 for item in katibexp.get("items")
             ]
 
@@ -341,7 +341,7 @@ class KatibClient(object):
         try:
             katibtrial = thread.get(constants.APISERVER_TIMEOUT)
             result = [
-                self.katib_api_client.deserialize_data(item, V1beta1Trial)
+                self.deserializer.deserialize(item, V1beta1Trial)
                 for item in katibtrial.get("items")
             ]
         except multiprocessing.TimeoutError:
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -14,7 +14,7 @@
 
 import multiprocessing
 
-from kubeflow.katib import V1beta1Experiment, api_client
+from kubeflow.katib import V1beta1Experiment
 from kubeflow.katib import V1beta1Trial
 from kubeflow.katib.constants import constants
 from kubeflow.katib.utils import utils
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -16,6 +16,7 @@ import multiprocessing
 
 from kubeflow.katib import V1beta1Experiment
 from kubeflow.katib import V1beta1Trial
+from kubeflow.katib.api_client import ApiClient
 from kubeflow.katib.constants import constants
 from kubeflow.katib.utils import utils
 from kubernetes import client, config
@@ -46,7 +47,7 @@ class KatibClient(object):
             self.in_cluster = True
 
         self.api_instance = client.CustomObjectsApi()
-        self.deserializer = utils.Deserializer()
+        self.api_client = ApiClient()
 
     def _is_ipython(self):
         """Returns whether we are running in notebook."""
@@ -270,7 +271,7 @@ class KatibClient(object):
         try:
             katibexp = thread.get(constants.APISERVER_TIMEOUT)
             result = [
-                self.deserializer.deserialize(item, V1beta1Experiment)
+                self.api_client.deserialize(utils.FakeResponse(item), V1beta1Experiment)
                 for item in katibexp.get("items")
             ]
 
@@ -341,7 +342,7 @@ class KatibClient(object):
         try:
             katibtrial = thread.get(constants.APISERVER_TIMEOUT)
             result = [
-                self.deserializer.deserialize(item, V1beta1Trial)
+                self.api_client.deserialize(utils.FakeResponse(item), V1beta1Trial)
                 for item in katibtrial.get("items")
             ]
         except multiprocessing.TimeoutError:
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -245,11 +245,13 @@ class KatibClient(object):
                 "Exception when calling CustomObjectsApi->delete_namespaced_custom_object:\
          %s\n" % e)
 
-    def list_experiments(self, namespace=None):
+    def list_experiments(self, namespace=None, in_short=False):
         """List all Katib Experiments.
 
         :param namespace: Experiments namespace.
         If the namespace is None, it takes "default" namespace.
+        :param in_short: List Experiments in summary or not
+        If the in_short is True, it will shows only name and status of each Experiments.
 
         :return: List of Experiment names with the statuses.
         :rtype: list[dict]
@@ -268,12 +270,15 @@ class KatibClient(object):
         try:
             katibexp = thread.get(constants.APISERVER_TIMEOUT)
             result = []
-            for i in katibexp.get("items"):
-                output = {}
-                output["name"] = i.get("metadata", {}).get("name")
-                output["status"] = i.get("status", {}).get(
-                    "conditions", [])[-1].get("type")
-                result.append(output)
+            if in_short:
+                for i in katibexp.get("items"):
+                    output = {}
+                    output["name"] = i.get("metadata", {}).get("name")
+                    output["status"] = i.get("status", {}).get(
+                        "conditions", [])[-1].get("type")
+                    result.append(output)
+            else:
+                result = katibexp.get("items")
         except multiprocessing.TimeoutError:
             raise RuntimeError("Timeout trying to get katib experiment.")
         except client.rest.ApiException as e:
@@ -282,8 +287,8 @@ class KatibClient(object):
           %s\n" % e)
         except Exception as e:
             raise RuntimeError(
-                "There was a problem to get experiments in namespace {1}. Exception: \
-          {2} ".format(namespace, e))
+                "There was a problem to get experiments in namespace {0}. Exception: \
+          {1} ".format(namespace, e))
         return result
 
     def get_experiment_status(self, name, namespace=None):
@@ -317,12 +322,14 @@ class KatibClient(object):
             name, namespace=namespace)
         return experiment_status.lower() == "succeeded"
 
-    def list_trials(self, name=None, namespace=None):
+    def list_trials(self, name=None, namespace=None, in_short=False):
         """List all Experiment's Trials.
 
         :param name: Experiment name.
         :param namespace: Experiments namespace.
         If the namespace is None, it takes "default" namespace.
+        :param in_short: List Trials in summary or not
+        If the in_short is True, it will shows only name and status of each Trials.
 
         :return: List of Trial names with the statuses.
         :rtype: list[dict]
@@ -341,13 +348,16 @@ class KatibClient(object):
         try:
             katibtrial = thread.get(constants.APISERVER_TIMEOUT)
             result = []
-            for i in katibtrial.get("items"):
-                output = {}
-                if i.get("metadata", {}).get("ownerReferences")[0].get("name") == name:
-                    output["name"] = i.get("metadata", {}).get("name")
-                    output["status"] = i.get("status", {}).get(
-                        "conditions", [])[-1].get("type")
-                    result.append(output)
+            if in_short:
+                for i in katibtrial.get("items"):
+                    output = {}
+                    if i.get("metadata", {}).get("ownerReferences")[0].get("name") == name:
+                        output["name"] = i.get("metadata", {}).get("name")
+                        output["status"] = i.get("status", {}).get(
+                            "conditions", [])[-1].get("type")
+                        result.append(output)
+            else:
+                result = katibtrial.get("items")
         except multiprocessing.TimeoutError:
             raise RuntimeError("Timeout trying to getkatib experiment.")
         except client.rest.ApiException as e:
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -14,7 +14,7 @@
 
 import multiprocessing
 
-from kubeflow.katib import V1beta1Experiment
+from kubeflow.katib import V1beta1Experiment, api_client
 from kubeflow.katib import V1beta1Trial
 from kubeflow.katib.constants import constants
 from kubeflow.katib.utils import utils
@@ -46,6 +46,7 @@ class KatibClient(object):
             self.in_cluster = True
 
         self.api_instance = client.CustomObjectsApi()
+        self.katib_api_client = api_client.ApiClient()
 
     def _is_ipython(self):
         """Returns whether we are running in notebook."""
@@ -269,13 +270,7 @@ class KatibClient(object):
         try:
             katibexp = thread.get(constants.APISERVER_TIMEOUT)
             result = [
-                V1beta1Experiment(
-                    api_version=item.get("apiVersion"),
-                    kind=item.get("kind"),
-                    metadata=item.get("metadata"),
-                    spec=item.get("spec"),
-                    status=item.get("status")
-                )
+                self.katib_api_client.deserialize_data(item, V1beta1Experiment)
                 for item in katibexp.get("items")
             ]
 
@@ -346,13 +341,7 @@ class KatibClient(object):
         try:
             katibtrial = thread.get(constants.APISERVER_TIMEOUT)
             result = [
-                V1beta1Trial(
-                    api_version=item.get("apiVersion"),
-                    kind=item.get("kind"),
-                    metadata=item.get("metadata"),
-                    spec=item.get("spec"),
-                    status=item.get("status")
-                )
+                self.katib_api_client.deserialize_data(item, V1beta1Trial)
                 for item in katibtrial.get("items")
             ]
         except multiprocessing.TimeoutError:
--- a/sdk/python/v1beta1/kubeflow/katib/api_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api_client.py
@@ -12,8 +12,6 @@ from __future__ import absolute_import
 
 import atexit
 import datetime
-
-import kubernetes
 from dateutil.parser import parse
 import json
 import mimetypes
@@ -24,7 +22,6 @@ import tempfile
 
 # python 2 and python 3 compatibility library
 import six
-from kubeflow.katib.rest import RESTResponse
 from six.moves.urllib.parse import quote
 
 from kubeflow.katib.configuration import Configuration
@@ -269,7 +266,6 @@ class ApiClient(object):
 
         :return: deserialized object.
         """
-        assert isinstance(response, RESTResponse)
         # handle file downloading
         # save response body into a tmp file and return the instance
         if response_type == "file":
@@ -282,18 +278,6 @@ class ApiClient(object):
             data = response.data
 
         return self.__deserialize(data, response_type)
-
-    def deserialize_data(self, data, data_type):
-        """Deserializes data into an object.
-
-        :param data: object to be deserialized.
-        :param data_type: class literal for
-            deserialized object, or string of class name.
-
-        :return: deserialized object.
-        """
-        assert isinstance(data, (dict, list, str))
-        return self.__deserialize(data, data_type)
 
     def __deserialize(self, data, klass):
         """Deserializes dict, list, str into an object.
@@ -320,12 +304,8 @@ class ApiClient(object):
             # convert str to class
             if klass in self.NATIVE_TYPES_MAPPING:
                 klass = self.NATIVE_TYPES_MAPPING[klass]
-            elif klass in dir(kubeflow.katib.models):
-                klass = getattr(kubeflow.katib.models, klass)
-            elif klass in dir(kubernetes.client.models):
-                klass = getattr(kubernetes.client.models, klass)
             else:
-                raise ValueError(f"type: {klass} is not supported to deserialized")
+                klass = getattr(kubeflow.katib.models, klass)
 
         if klass in self.PRIMITIVE_TYPES:
             return self.__deserialize_primitive(data, klass)
--- a/sdk/python/v1beta1/kubeflow/katib/api_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api_client.py
@@ -12,6 +12,8 @@ from __future__ import absolute_import
 
 import atexit
 import datetime
+
+import kubernetes
 from dateutil.parser import parse
 import json
 import mimetypes
@@ -22,6 +24,7 @@ import tempfile
 
 # python 2 and python 3 compatibility library
 import six
+from kubeflow.katib.rest import RESTResponse
 from six.moves.urllib.parse import quote
 
 from kubeflow.katib.configuration import Configuration
@@ -266,6 +269,7 @@ class ApiClient(object):
 
         :return: deserialized object.
         """
+        assert isinstance(response, RESTResponse)
         # handle file downloading
         # save response body into a tmp file and return the instance
         if response_type == "file":
@@ -278,6 +282,18 @@ class ApiClient(object):
             data = response.data
 
         return self.__deserialize(data, response_type)
+
+    def deserialize_data(self, data, data_type):
+        """Deserializes data into an object.
+
+        :param data: object to be deserialized.
+        :param data_type: class literal for
+            deserialized object, or string of class name.
+
+        :return: deserialized object.
+        """
+        assert isinstance(data, (dict, list, str))
+        return self.__deserialize(data, data_type)
 
     def __deserialize(self, data, klass):
         """Deserializes dict, list, str into an object.
@@ -304,8 +320,12 @@ class ApiClient(object):
             # convert str to class
             if klass in self.NATIVE_TYPES_MAPPING:
                 klass = self.NATIVE_TYPES_MAPPING[klass]
-            else:
+            elif klass in dir(kubeflow.katib.models):
                 klass = getattr(kubeflow.katib.models, klass)
+            elif klass in dir(kubernetes.client.models):
+                klass = getattr(kubernetes.client.models, klass)
+            else:
+                raise ValueError(f"type: {klass} is not supported to deserialized")
 
         if klass in self.PRIMITIVE_TYPES:
             return self.__deserialize_primitive(data, klass)
--- a/sdk/python/v1beta1/kubeflow/katib/utils/utils.py
+++ b/sdk/python/v1beta1/kubeflow/katib/utils/utils.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from __future__ import absolute_import
-
 import json
 import os
 
--- a/sdk/python/v1beta1/kubeflow/katib/utils/utils.py
+++ b/sdk/python/v1beta1/kubeflow/katib/utils/utils.py
@@ -16,9 +16,6 @@ import json
 import os
 
 
-# python 2 and python 3 compatibility library
-
-
 def is_running_in_k8s():
     return os.path.isdir('/var/run/secrets/kubernetes.io/')
 
--- a/sdk/python/v1beta1/kubeflow/katib/utils/utils.py
+++ b/sdk/python/v1beta1/kubeflow/katib/utils/utils.py
@@ -14,15 +14,11 @@
 
 from __future__ import absolute_import
 
-import datetime
+import json
 import os
-import re
 
-import kubeflow.katib.models
+
 # python 2 and python 3 compatibility library
-import six
-from dateutil.parser import parse
-from kubeflow.katib import rest
 
 
 def is_running_in_k8s():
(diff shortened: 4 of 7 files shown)


sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -245,11 +245,13 @@ def delete_experiment(self, name, namespace=None):
...
          %s\n" % e)
 
-    def list_experiments(self, namespace=None):
+    def list_experiments(self, namespace=None, in_short=False):
         """List all Katib Experiments.
 
         :param namespace: Experiments namespace.
         If the namespace is None, it takes "default" namespace.
+        :param in_short: List Experiments in summary or not
+        If the in_short is True, it will shows only name and status of each Experiments.
```
andreyvelich, line 254: Do we really need `in_short` parameter ? Is it common practice to have this functionality with Kubernetes API ?
What do you think @johnugeorge @gaocegege @anencore94 ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -268,12 +270,15 @@ def list_experiments(self, namespace=None):
...
-                result.append(output)
+            if in_short:
+                for i in katibexp.get("items"):
+                    output = {}
+                    output["name"] = i.get("metadata", {}).get("name")
+                    output["status"] = i.get("status", {}).get(
+                        "conditions", [])[-1].get("type")
+                    result.append(output)
+            else:
+                result = katibexp.get("items")
```
andreyvelich, line 281: Maybe we can cast the result to [`[]V1beta1Trial`](https://github.com/kubeflow/katib/blob/778e4fdf8d995a00f37ba06e4e5c3b768152f1ed/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial.py#L21) list of objects ?
What do you think ?

andreyvelich (commented):
Thank you for taking this @anencore94!
I left few comments.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: In fact, I was wondering whether this feature is necessary or not too.
But since I'm not sure why the original api returns only name and status, I just thought there must be some reason before and some people still wants to use api with such responses. (to match backward compatibility)
However, I'm afraid it would be not that common usage

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: That would be better for safe usage!
However, then don't we have to change return types for other similar cases ?
i.e. `get_experiment`, `get_suggestion`, `list_experiments`

anencore94: For now, I changed list_experiments and list_trials to return List of v1beta1_objects rather than dict

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: For now, I removed `in_short` parameter since I couldn't find any practice to use such parameter within k8s modules

andreyvelich: Nice suggestion about compatibility.
Maybe we can log that `in_short` parameter will be deprecated in the next SDK releases and just remove it after 1-2 SDK release?
What do you think ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: > However, then don't we have to change return types for other similar cases ?
i.e. get_experiment, get_suggestion, list_experiments

Yeah, that should be our final goal 😄 
We can change `get_experiment`, `get_suggestion`, etc. in the following PRs.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: Yeap, that would be better to care someone who wants to list experiments only with name and status as before.

By the way, 
>Maybe we can log that in_short parameter will be deprecated in the next SDK releases and just remove it after 1-2 SDK release?

Did you mean `log those` on docstring? or real log ? @andreyvelich

andreyvelich:
/hold Please let's merge this PR after Katib 0.12 release.

After this change: https://github.com/kubeflow/katib/pull/1572, we still need to do some work to modify our [SDK](https://github.com/kubeflow/katib/tree/master/examples/v1beta1/sdk) and [KFP](https://github.com/kubeflow/pipelines/tree/master/samples/contrib/kubeflow-katib) examples to support the latest SDK changes.

Since this PR will add some additional functionality, we can include it in the next release (e.g. Katib 0.13).

What do you think @gaocegege @johnugeorge @anencore94 ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: That makes sense :)
Then I'll change return type to Union[List[dict], List[V1beta1Experiment]] until `in_short` parameter will be deprecated

anencore94:
> /hold Please let's merge this PR after Katib 0.12 release.

I agree. I'm not sure when Katib 0.12 will be released, but this feature change the API of sdk, this feature should change the minor version

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: I was thinking about including the log message somewhere here: https://github.com/anencore94/katib/blob/778e4fdf8d995a00f37ba06e4e5c3b768152f1ed/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L274-L279.

Maybe it's redundant, what do others think @gaocegege @johnugeorge ?

johnugeorge:
Since release branch is created, we can merge it to master

anencore94:
Is there any reason why **kubeflow-katib-presubmit** fail sometimes ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -267,13 +268,17 @@ def list_experiments(self, namespace=None):
...
-                output["name"] = i.get("metadata", {}).get("name")
-                output["status"] = i.get("status", {}).get(
-                    "conditions", [])[-1].get("type")
-                result.append(output)
+            result = [
+                V1beta1Experiment(
+                    api_version=item.get("apiVersion"),
+                    kind=item.get("kind"),
+                    metadata=item.get("metadata"),
+                    spec=item.get("spec"),
```
andreyvelich, line 276: In that case what will be the type for `experiment.spec` ?
I was thinking maybe we should use [`deserialize`](https://github.com/kubeflow/katib/blob/427ca71006103627230a1532352357c02cd32e7c/sdk/python/v1beta1/kubeflow/katib/api_client.py#L260) method to convert result to the List of objects.
What do you think @anencore94 @gaocegege @johnugeorge ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -251,8 +252,8 @@ def list_experiments(self, namespace=None):
         :param namespace: Experiments namespace.
         If the namespace is None, it takes "default" namespace.
 
-        :return: List of Experiment names with the statuses.
-        :rtype: list[dict]
+        :return: List of Experiment objects.
+        :rtype: list[V1beta1Experiment]
```
andreyvelich, line 256: Please could you also modify this doc: https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/docs/KatibClient.md#list_experiments.

andreyvelich:
> Is there any reason why kubeflow-katib-presubmit fail sometimes ?

Sometime tests can be flaky, we have to restart it couple of times.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: Yeap, I should've done that.
By the way, I guess the `make generate` doesn't work for generating docs/KatibClient.md since `hack/gen-python-sdk/post_gen.py` doesn't handle /docs/KatibClient.md. Is it intended ?

andreyvelich: Unfortunately, we don't have script to automatically generate `KatibClient.md` doc, currently we manually update it. We might need to modify `post_gen.py` in the future.

anencore94: DONE! Thanks for the explanation!

sdk/python/v1beta1/kubeflow/katib/api_client.py
```diff
@@ -304,8 +320,12 @@ def __deserialize(self, data, klass):
             # convert str to class
             if klass in self.NATIVE_TYPES_MAPPING:
                 klass = self.NATIVE_TYPES_MAPPING[klass]
-            else:
+            elif klass in dir(kubeflow.katib.models):
                 klass = getattr(kubeflow.katib.models, klass)
+            elif klass in dir(kubernetes.client.models):
```
anencore94, line 325: I have to add this feature to support deserializing kuberentest's api resources such as V1ObjectMeta, ...

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: That would be better! Thanks

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: Maybe just deprecate old API style with sufficient explanation in release note would be fine.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: With new method, I could check V1beta1Experiment object's nested type was filled as I intended.
Here is the test case:
- Python code
```
    resp = client.list_experiments(namespace='jaeyeon-kim')
    print(resp[0])
    print(type(resp[0]))
    print(type(resp[0].metadata))
    print(type(resp[0].spec))
    print(type(resp[0].spec.algorithm))

    resp2 = client.list_trials(name='katib-hpo-test', namespace='jaeyeon-kim')
    print(resp2[0])
    print(type(resp2[0]))
    print(type(resp2[0].spec.objective))
```

- The stdout with some omitted
```
{'api_version': 'kubeflow.org/v1beta1',
 'kind': 'Experiment',
 'metadata': {'annotations': {'kubectl.kubernetes.io/last-applied-configuration': '{"apiVersion":"kubeflow.org/v1beta1","kind":"Experiment","metadata":{"annotations":{},"name":"katib-hpo-test","namespace":"jaeyeon-kim"},"spec":{"algorithm":{"algorithmName":"tpe"},"maxFailedTrialCount":2,"maxTrialCount":5,"metricsCollectorSpec":{"collector":{"kind":"File"},"source":{"fileSystemPath":{"kind":"File","path":"/var/log/katib/mrx-metrics.log"}}},"objective":{"goal":0.001,"objectiveMetricName":"loss","type":"minimize"},"parallelTrialCount":2,"parameters":[{"feasibleSpace":{"max":"2.0","min":"0.0"},"name":"x","parameterType":"double"},{"feasibleSpace":{"max":"3.0","min":"0.0"},"name":"y","parameterType":"double"},{"feasibleSpace":{"max":"5.0","min":"1.0"},"name":"z","parameterType":"double"}],"trialTemplate":{"failureCondition":"status.conditions.#(type==\\"Failed\\")#|#(status==\\"True\\")#","primaryContainerName":"katib-helper","successCondition":"status.conditions.#
......
......
              'name': 'katib-hpo-test',
              'namespace': 'jaeyeon-kim',
              'owner_references': None,
              'resource_version': '422218026',
              'self_link': '/apis/kubeflow.org/v1beta1/namespaces/jaeyeon-kim/experiments/katib-hpo-test',
              'uid': '7f930fc1-9de6-474c-9417-5a41c9339498'},
 'spec': {'algorithm': {'algorithm_name': 'tpe', 'algorithm_settings': None},
          'early_stopping': None,
          'max_failed_trial_count': 2,
          'max_trial_count': 5,
          'metrics_collector_spec': {'collector': {'custom_collector': None,
                                                   'kind': 'File'},
                                     'source': {'file_system_path': {'kind': 'File',
                                                                     'path': '/var/log/katib/mrx-metrics.log'},
                                                'filter': None,
                                                'http_get': None}},
          'nas_config': None,
          'objective': {'additional_metric_names': None,
                        'goal': 0.001,
                        'metric_strategies': [{'name': 'loss', 'value': 'min'}],
                        'objective_metric_name': 'loss',
                        'type': 'minimize'},
          'parallel_trial_count': 2,
          'parameters': [{'feasible_space': {'list': None,
                                             'max': '2.0',
                                             'min': '0.0',
                                             'step': None},
                          'name': 'x',
                          'parameter_type': 'double'},
                         {'feasible_space': {'list': None,
                                             'max': '3.0',
                                             'min': '0.0',
                                             'step': None},
                          'name': 'y',
                          'parameter_type': 'double'},
                         {'feasible_space': {'list': None,
                                             'max': '5.0',
                                             'min': '1.0',
                                             'step': None},
                          'name': 'z',
                          'parameter_type': 'double'}],
          'resume_policy': 'LongRunning',
          'trial_template': {'config_map': None,
                             'failure_condition': 'status.conditions.#(type=="Failed")#|#(status=="True")#',
                             'primary_container_name': 'katib-helper',
                             'primary_pod_labels': None,
                             'retain': None,
......
......
                                                                                           'imagePullPolicy': 'IfNotPresent',
                                                                                           'name': 'katib-helper'}],
                                                                           'imagePullSecrets': [{'name': 'regcred-ghcr'}],
                                                                           'restartPolicy': 'Never',
                                                                           'serviceAccountName': 'katib-account'}}}}}},
 'status': {'completion_time': datetime.datetime(2021, 8, 21, 14, 9, 27, tzinfo=tzutc()),
            'conditions': [{'last_transition_time': datetime.datetime(2021, 8, 21, 14, 6, 5, tzinfo=tzutc()),
                            'last_update_time': datetime.datetime(2021, 8, 21, 14, 6, 5, tzinfo=tzutc()),
                            'message': 'Experiment is created',
                            'reason': 'ExperimentCreated',
                            'status': 'True',
                            'type': 'Created'},
                           {'last_transition_time': datetime.datetime(2021, 8, 21, 14, 9, 27, tzinfo=tzutc()),
                            'last_update_time': datetime.datetime(2021, 8, 21, 14, 9, 27, tzinfo=tzutc()),
                            'message': 'Experiment is running',
                            'reason': 'ExperimentRunning',
                            'status': 'False',
                            'type': 'Running'},
                           {'last_transition_time': datetime.datetime(2021, 8, 21, 14, 9, 27, tzinfo=tzutc()),
                            'last_update_time': datetime.datetime(2021, 8, 21, 14, 9, 27, tzinfo=tzutc()),
                            'message': 'Experiment has succeeded because max '
                                       'trial count has reached',
                            'reason': 'ExperimentMaxTrialsReached',
                            'status': 'True',
                            'type': 'Succeeded'}],
            'current_optimal_trial': {'best_trial_name': 'katib-hpo-test-g4wkct7w',
                                      'observation': {'metrics': [{'latest': '0.16119065270234822',
                                                                   'max': '0.16119065270234822',
                                                                   'min': '0.16119065270234822',
                                                                   'name': 'loss'}]},
                                      'parameter_assignments': [{'name': 'x',
                                                                 'value': '0.8376333862283576'},
                                                                {'name': 'y',
                                                                 'value': '2.3292919169041912'},
                                                                {'name': 'z',
                                                                 'value': '3.0852625163721057'}]},
            'early_stopped_trial_list': None,
            'failed_trial_list': None,
            'killed_trial_list': None,
            'last_reconcile_time': None,
            'pending_trial_list': None,
            'running_trial_list': None,
            'start_time': datetime.datetime(2021, 8, 21, 14, 6, 5, tzinfo=tzutc()),
            'succeeded_trial_list': ['katib-hpo-test-b67wvc6l',
                                     'katib-hpo-test-g4wkct7w',
                                     'katib-hpo-test-nfpb5k75',
                                     'katib-hpo-test-znjhw7rt',
                                     'katib-hpo-test-xphcc7hm'],
            'trials': 5,
            'trials_early_stopped': None,
            'trials_failed': None,
            'trials_killed': None,
            'trials_pending': None,
            'trials_running': None,
            'trials_succeeded': 5}}
<class 'kubeflow.katib.models.v1beta1_experiment.V1beta1Experiment'>
<class 'kubernetes.client.models.v1_object_meta.V1ObjectMeta'>
<class 'kubeflow.katib.models.v1beta1_experiment_spec.V1beta1ExperimentSpec'>
<class 'kubeflow.katib.models.v1beta1_algorithm_spec.V1beta1AlgorithmSpec'>

---------------------------------------------------------------------------------------------
{'api_version': 'kubeflow.org/v1beta1',
 'kind': 'Trial',
 'metadata': {'annotations': None,
              'cluster_name': None,
              'creation_timestamp': datetime.datetime(2021, 8, 21, 14, 5, 16, tzinfo=tzutc()),
              'deletion_grace_period_seconds': None,
              'deletion_timestamp': None,
              'finalizers': ['clean-metrics-in-db'],
              'generate_name': None,
              'generation': 1,
              'labels': {'experiment': 'katib-hpo-test-2'},
              'managed_fields': [{'api_version': 'kubeflow.org/v1beta1',
                                  'fields_type': 'FieldsV1',
                                  'fields_v1': {'f:metadata': {'f:finalizers': {},
                                                               'f:labels': {'.': {},
                                                                            'f:experiment': {}},
                                                               'f:ownerReferences': {}},
                                                'f:spec': {'.': {},
                                                           'f:failureCondition': {},
                                                           'f:metricsCollector': {'.': {},
                                                                                  'f:collector': {'.': {},
                                                                                                  'f:kind': {}},
                                                                                  'f:source': {'.': {},
                                                                                               'f:fileSystemPath': {'.': {},
                                                                                                                    'f:kind': {},
                                                                                                                    'f:path': {}}}},
                                                           'f:objective': {'.': {},
                                                                           'f:goal': {},
                                                                           'f:metricStrategies': {},
                                                                           'f:objectiveMetricName': {},
                                                                           'f:type': {}},
                                                           'f:parameterAssignments': {},
                                                           'f:primaryContainerName': {},
                                                           'f:runSpec': {'.': {},
                                                                         'f:apiVersion': {},
                                                                         'f:kind': {},
                                                                         'f:metadata': {'.': {},
                                                                                        'f:creationTimestamp': {},
                                                                                        'f:labels': {'.': {},
                                                                                                     'f:app.kubernetes.io/part-of': {},
                                                                                                     'f:controller-uid': {},
                                                                                                     'f:job-name': {}},
                                                                                        'f:managedFields': {},
                                                                                        'f:name': {},
                                                                                        'f:namespace': {},
                                                                                        'f:ownerReferences': {},
                                                                                        'f:resourceVersion': {},
                                                                                        'f:selfLink': {},
                                                                                        'f:uid': {}},
                                                                         'f:spec': {'.': {},
                                                                                    'f:backoffLimit': {},
                                                                                    'f:completions': {},
                                                                                    'f:parallelism': {},
                                                                                    'f:selector': {'.': {},
                                                                                                   'f:matchLabels': {'.': {},
                                                                                                                     'f:controller-uid': {}}},
                                                                                    'f:template': {'.': {},
                                                                                                   'f:metadata': {'.': {},
                                                                                                                  'f:annotations': {'.': {},
                                                                                                                                    'f:sidecar.istio.io/inject': {}},
                                                                                                                  'f:creationTimestamp': {},
                                                                                                                  'f:labels': {'.': {},
                                                                                                                               'f:app.kubernetes.io/part-of': {},
                                                                                                                               'f:controller-uid': {},
                                                                                                                               'f:job-name': {}}},
                                                                                                   'f:spec': {'.': {},
                                                                                                              'f:containers': {},
                                                                                                              'f:dnsPolicy': {},
                                                                                                              'f:imagePullSecrets': {},
                                                                                                              'f:restartPolicy': {},
                                                                                                              'f:schedulerName': {},
                                                                                                              'f:securityContext': {},
                                                                                                              'f:serviceAccount': {},
                                                                                                              'f:serviceAccountName': {},
                                                                                                              'f:terminationGracePeriodSeconds': {}}}},
                                                                         'f:status': {}},
                                                           'f:successCondition': {}},
                                                'f:status': {'.': {},
                                                             'f:completionTime': {},
                                                             'f:conditions': {},
                                                             'f:observation': {'.': {},
                                                                               'f:metrics': {}},
                                                             'f:startTime': {}}},
                                  'manager': 'katib-controller',
                                  'operation': 'Update',
                                  'time': datetime.datetime(2021, 8, 21, 14, 6, 22, tzinfo=tzutc())}],
              'name': 'katib-hpo-test-2-b9pf6btg',
              'namespace': 'jaeyeon-kim',
              'owner_references': [{'api_version': 'kubeflow.org/v1beta1',
                                    'block_owner_deletion': True,
                                    'controller': True,
                                    'kind': 'Experiment',
                                    'name': 'katib-hpo-test-2',
                                    'uid': '83155259-8944-42f2-a288-0afc4b2dfb68'}],
              'resource_version': '422207197',
              'self_link': '/apis/kubeflow.org/v1beta1/namespaces/jaeyeon-kim/trials/katib-hpo-test-2-b9pf6btg',
              'uid': '18e5ac5c-ca24-46d7-9acb-6a886bf2f09a'},
 'spec': {'early_stopping_rules': None,
          'failure_condition': 'status.conditions.#(type=="Failed")#|#(status=="True")#',
          'metrics_collector': {'collector': {'custom_collector': None,
                                              'kind': 'File'},
                                'source': {'file_system_path': {'kind': 'File',
                                                                'path': '/var/log/katib/mrx-metrics.log'},
                                           'filter': None,
                                           'http_get': None}},
          'objective': {'additional_metric_names': None,
                        'goal': 0.001,
                        'metric_strategies': [{'name': 'loss', 'value': 'min'}],
                        'objective_metric_name': 'loss',
                        'type': 'minimize'},
          'parameter_assignments': [{'name': 'x',
                                     'value': '0.8240284763063495'},
                                    {'name': 'y',
                                     'value': '1.8777500963698879'},
                                    {'name': 'z',
                                     'value': '4.132651796238958'}],
          'primary_container_name': 'katib-helper',
          'primary_pod_labels': None,
          'retain_run': None,
          'run_spec': {'apiVersion': 'batch/v1',
                       'kind': 'Job',
                       'metadata': {'name': 'katib-hpo-test-2-b9pf6btg',
                                    'namespace': 'jaeyeon-kim'},
                       'spec': {'template': {'metadata': {'annotations': {'sidecar.istio.io/inject': 'true'},
                                                          'labels': {'app.kubernetes.io/part-of': 'katib-experiment'}},
                                             'spec': {'containers': [{'command': ['until '
                                                                                  '$(curl '
                                                                                  '--output '
                                                                                  '/dev/null '
                                                                                  '--silent '
                                                                                  '--head '
                                                                                  '--fail '
                                                                                  'http://localhost:15021/healthz/ready); '
                                                                                  'do '
                                                                                  'echo '
                                                                                  '. '
                                                                                  '; '
                                                                                  'sleep '
                                                                                  '5; '
                                                                                  'done; '
                                                                                  'kfp '
                                                                                  'run '
                                                                                  'submit '
                                                                                  '-t '
                                                                                  '300 '
                                                                                  '-e '
                                                                                  '1442bc8a-bc0e-482d-8658-2553abf7257e '
                                                                                  '-p '
                                                                                  '780544e6-40e9-45ab-9987-322988b91edf '
                                                                                  'x=0.8240284763063495 '
                                                                                  'y=1.8777500963698879 '
                                                                                  'z=4.132651796238958 '
                                                                                  '> '
                                                                                  '/tmp/run-submit-result; '
                                                                                  'export '
                                                                                  'RUN_ID=$(awk '
                                                                                  "'/Succeeded/' "
                                                                                  '/tmp/run-submit-result '
                                                                                  '| '
                                                                                  'awk '
                                                                                  '-F '
                                                                                  "'| "
                                                                                  "' "
                                                                                  "'{ "
                                                                                  'print '
                                                                                  '$2 '
                                                                                  "}'); \n"
                                                                                  '    '
                                                                                  '(echo '
                                                                                  '"import '
                                                                                  'kfp";\n'
                                                                                  '     '
                                                                                  'echo '
                                                                                  '"import '
                                                                                  'os";\n'
                                                                                  '     '
                                                                                  'echo '
                                                                                  '"import '
                                                                                  'json";\n'
                                                                                  '     '
                                                                                  'echo '
                                                                                  '"client=kfp.Client()";\n'
                                                                                  '     '
                                                                                  'echo '
                                                                                  '"run_response '
                                                                                  '= '
                                                                                  'client.get_run(os.getenv(\'RUN_ID\'))";\n'
                                                                                  '     '
                                                                                  'echo '
                                                                                  '"for '
                                                                                  'metric '
                                                                                  'in '
                                                                                  'run_response.run.metrics: '
                                                                                  'print(f\'{metric.name}={metric.number_value}\')"\n'
                                                                                  '     '
                                                                                  ') '
                                                                                  '| '
                                                                                  'python '
                                                                                  '>> '
                                                                                  '/var/log/katib/mrx-metrics.log\\\n'
                                                                                  '     '],
                                                                      'image': 'ghcr.io/makinarocks/katib-helper:v1.0.2',
                                                                      'imagePullPolicy': 'IfNotPresent',
                                                                      'name': 'katib-helper'}],
                                                      'imagePullSecrets': [{'name': 'regcred-ghcr'}],
                                                      'restartPolicy': 'Never',
                                                      'serviceAccountName': 'katib-account'}}}},
          'success_condition': 'status.conditions.#(type=="Complete")#|#(status=="True")#'},
 'status': {'completion_time': datetime.datetime(2021, 8, 21, 14, 6, 22, tzinfo=tzutc()),
            'conditions': [{'last_transition_time': datetime.datetime(2021, 8, 21, 14, 5, 16, tzinfo=tzutc()),
                            'last_update_time': datetime.datetime(2021, 8, 21, 14, 5, 16, tzinfo=tzutc()),
                            'message': 'Trial is created',
                            'reason': 'TrialCreated',
                            'status': 'True',
                            'type': 'Created'},
                           {'last_transition_time': datetime.datetime(2021, 8, 21, 14, 6, 22, tzinfo=tzutc()),
                            'last_update_time': datetime.datetime(2021, 8, 21, 14, 6, 22, tzinfo=tzutc()),
                            'message': 'Trial is running',
                            'reason': 'TrialRunning',
                            'status': 'False',
                            'type': 'Running'},
                           {'last_transition_time': datetime.datetime(2021, 8, 21, 14, 6, 22, tzinfo=tzutc()),
                            'last_update_time': datetime.datetime(2021, 8, 21, 14, 6, 22, tzinfo=tzutc()),
                            'message': 'Trial has succeeded',
                            'reason': 'TrialSucceeded',
                            'status': 'True',
                            'type': 'Succeeded'}],
            'last_reconcile_time': None,
            'observation': {'metrics': [{'latest': '1.1150389600458972',
                                         'max': '1.1150389600458972',
                                         'min': '1.1150389600458972',
                                         'name': 'loss'}]},
            'start_time': datetime.datetime(2021, 8, 21, 14, 5, 16, tzinfo=tzutc())}}
<class 'kubeflow.katib.models.v1beta1_trial.V1beta1Trial'>
<class 'kubeflow.katib.models.v1beta1_objective_spec.V1beta1ObjectiveSpec'>

Process finished with exit code 0

```

sdk/python/v1beta1/kubeflow/katib/api_client.py
andreyvelich: Please can you try to run `make generate` to check if `api_client.py` will be properly generated with your change ?
We might need to do some modification in [`post_gen.py`](https://github.com/kubeflow/katib/blob/master/hack/gen-python-sdk/post_gen.py).

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Thank you @anencore94!
> That would be better! Thanks. But I have to add similar method : deserialize_data, to deserialize json, not RESTResponse object from thread.get()

Do you know, what is the type in [`item` here](https://github.com/kubeflow/katib/blob/f0308997475a3787668aedc21b63b37eb270d57e/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L274) ?
It is not `RESTResponse` ?

Have you tried to run `deserialize(list[V1beta1Experiment])` ?

anencore94: > Do you know, what is the type in item here ?
It is not RESTResponse ?

Testing with following code in  [item here ](https://github.com/kubeflow/katib/blob/f0308997475a3787668aedc21b63b37eb270d57e/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L274)
```
        katibexp = None
        try:
            katibexp = thread.get(constants.APISERVER_TIMEOUT)

            print(type(katibexp))
            for item in katibexp.get("items"):
                print(type(item))
                print(item)
                sys.exit(1)
```

The stdout was as follows:
```
<class 'dict'>
<class 'dict'>
{'apiVersion': 'kubeflow.org/v1beta1', 'kind': 'Experiment', 'metadata': {'annotations': {'kubectl.kubernetes.io/last-applied-configuration': '{"apiVersion":"kubeflow.org/v1beta1","kind":"Experiment","metadata":{"annotations":{},"name":"hpo-quadratic","namespace":"jaeyeon-kim"},"spec":{"algorithm":{"algorithmName":"tpe"},"maxFailedTrialCount":2,"maxTrialCount":3,"metricsCollectorSpec":{"collector":{"kind":"File"},"source":{"fileSystemPath":{"kind":"File","path":"/var/log/katib/mrx-metrics.log"}}},"objective":{"goal":0.5,"objectiveMetricName":"loss","type":"minimize"},"parallelTrialCount":1,"parameters":[{"feasibleSpace":{"max":"1.0","min":"0.0"},"name":"x","parameterType":"double"},{"feasibleSpace":{"max":"3.0","min":"0.0"},"name":"y","parameterType":"double"},{"feasibleSpace":{"max":"65.0","min":"0.0"},"name":"z","parameterType":"double"}],"trialTemplate":{"failureCondition":"status.conditions.#(type==\\"Failed\\")#|#(status==\\"True\\")#","primaryContainerName":"katib-helper","successCondition":"status.conditions.#(type==\\"Complete\\")#|#(status==\\"True\\")#","trialParameters":[{"name":"x","reference":"x"},{"name":"y","reference":"y"},{"name":"z","reference":"z"}],"trialSpec":{"apiVersion":"batch/v1","kind":"Job",
.....
// omitted
}
```

anencore94: Since both instances' type was dict, not RESTResponse, I couldn't use existing [deserialize](https://github.com/kubeflow/katib/blob/427ca71006103627230a1532352357c02cd32e7c/sdk/python/v1beta1/kubeflow/katib/api_client.py#L260) function

sdk/python/v1beta1/kubeflow/katib/api_client.py
anencore94: Thanks @andreyvelich ,
I tried `make generate` and it erase all differences within this PR, as you might worried...

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Yeah, I think it's common issue: https://github.com/kubernetes-client/python/issues/340.
I also tried this: https://github.com/kubernetes-client/python/issues/574#issuecomment-405400414 with `_preload_content=False`, but `list_namespaced_custom_object` can't be converted still.
And it requires `kubernetes.client.models` converter for it as you did.

Maybe we can create our own `deserialize` method in utils: https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/kubeflow/katib/utils/utils.py ?

What do you think?

anencore94: Yeap, that makes sense @andreyvelich , Thanks for checking this further :)
Then you mean the original [deserialize](https://github.com/kubeflow/katib/blob/427ca71006103627230a1532352357c02cd32e7c/sdk/python/v1beta1/kubeflow/katib/api_client.py#L260) function should be as before, but we make another method something like `deserialize(data: Union[dict, list, str], data_type: Type[T]) -> T` right ?

andreyvelich: Yes, we can make another method as you made [here](https://github.com/kubeflow/katib/blob/f0308997475a3787668aedc21b63b37eb270d57e/sdk/python/v1beta1/kubeflow/katib/api_client.py#L286).

The problem with this conversion: https://github.com/kubeflow/katib/blob/f0308997475a3787668aedc21b63b37eb270d57e/sdk/python/v1beta1/kubeflow/katib/api_client.py#L323-L326.

Is there any chance to remove it and make deserialization?
We defined Kubernetes models in Katib models: https://github.com/kubeflow/katib/blob/f0308997475a3787668aedc21b63b37eb270d57e/sdk/python/v1beta1/kubeflow/katib/models/__init__.py#L59-L63.
So it's strange why `getattr` doesn't work.

anencore94: I changed the location of `deserialize` method to utils.py, so I need to copy all existing private methods in api_client.py.

Also, I still needs to check `dir(kubernetes.client.models)` even if they were in `dir(kubeflow.katib.models)` as you mentioned.

anencore94: I think I found the reason why.
There were missing class in katib/sdk/python/v1beta1/kubeflow/katib/models/__init__.py :
```
from kubernetes.client import V1ManagedFieldsEntry
from kubernetes.client import V1OwnerReference
```

However, even though I tried to add them manually, `make generate` automatically removes such lines.
Do I have to specify those import packages at [swagger_config.json](https://github.com/kubeflow/katib/blob/21580f66972640ee60aea0f86d74a4f496ee98c3/hack/gen-python-sdk/swagger_config.json#L5) or somewhere else ?

andreyvelich: >Do I have to specify those import packages at swagger_config.json or somewhere else ?

Exactly, let's add these import to swagger config.

hack/gen-python-sdk/post_gen.py
```diff
@@ -50,6 +50,8 @@ def _rewrite_helper(input_file, output_file, rewrite_rules):
         lines.append("from kubernetes.client import V1ListMeta\n")
         lines.append("from kubernetes.client import V1Container\n")
         lines.append("from kubernetes.client import V1HTTPGetAction\n")
+        lines.append("from kubernetes.client import V1ManagedFieldsEntry\n")
+        lines.append("from kubernetes.client import V1OwnerReference\n")
```
anencore94, line 54: I need to add these lines, not `hack/gen-python-sdk/swagger_config.json` to control `sdk/python/v1beta1/kubeflow/katib/models/__init__.py`
@andreyvelich

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
```diff
@@ -34,3 +44,164 @@ def set_katib_namespace(katib):
     katib_namespace = katib.metadata.namespace
     namespace = katib_namespace or get_default_target_namespace()
     return namespace
+
+
+class Deserializer:
```
andreyvelich, line 49: Since after this change: https://github.com/kubeflow/katib/pull/1630#discussion_r697873613, your `__deserialize` is the same as [`__deserialize` from api_client](https://github.com/kubeflow/katib/blob/4a453d8e1e2a4ff390b4691811f0a71c6c2d7366/sdk/python/v1beta1/kubeflow/katib/api_client.py#L282), maybe we could try not to duplicate all of the functions ?
I understand that we can't execute `__deserialize` from the utils, but maybe we can try to do some hack, as here: https://github.com/kubernetes-client/python/issues/977#issuecomment-592030030.

For example, in utils.py follow this:

```
class FakeResponse:
    def __init__(self, obj):
        self.data = obj         ​

```

And in Katib client just do this:

```
katibexp = thread.get(constants.APISERVER_TIMEOUT)
            result = [
                api_client.deserialize(FakeResponse(item), V1beta1Experiment)
                for item in katibexp.get("items")
            ]
```
WDYT @anencore94 ?

anencore94: Wow :open_mouth: , That looks much better!
I feel not clear for copying the whole private deserialize methods :joy: 

I think using some another name rather than `FakeResponse` will be more readable.
Is there any candidates? @andreyvelich

andreyvelich: We can name it as `RESTResponse`, as an official name of this [response object](https://github.com/kubeflow/katib/blob/4a453d8e1e2a4ff390b4691811f0a71c6c2d7366/sdk/python/v1beta1/kubeflow/katib/api_client.py#L559).
What are your thoughts ?

anencore94: I felt a bit confused we have to specify `kubernetes` or not on naming.
Maybe one of FakeResponse, FakeRESTResponse, FakeKubernetesRESTResponse, ...

anencore94: By the way, we need to use json.dumps : `self.data = json.dumps(obj)` as https://github.com/kubernetes-client/python/issues/977#issuecomment-592030030 did.
Since our response type is already dict, not str or bytes.

andreyvelich: I think it should work without `dumps`.
Check here: https://github.com/kubeflow/katib/blob/4a453d8e1e2a4ff390b4691811f0a71c6c2d7366/sdk/python/v1beta1/kubeflow/katib/api_client.py#L275-L278.
In case of error: `data = response.data`

andreyvelich: Let's keep it `FakeResponse`, we can rename it later if it requres.

anencore94: When such case occurs, it raise TypeError, not ValueError

```
Traceback (most recent call last):
  File "/home/kjy/git/katib/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py", line 274, in list_experiments
    for item in katibexp.get("items")
  File "/home/kjy/git/katib/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py", line 274, in <listcomp>
    for item in katibexp.get("items")
  File "/home/kjy/git/katib/sdk/python/v1beta1/kubeflow/katib/api_client.py", line 276, in deserialize
    data = json.loads(response.data)
  File "/home/kjy/.pyenv/versions/3.6.9/lib/python3.6/json/__init__.py", line 348, in loads
    'not {!r}'.format(s.__class__.__name__))
TypeError: the JSON object must be str, bytes or bytearray, not 'dict'

During handling of the above exception, another exception occurred:
```

anencore94: Could we change the **auth-generated** `api_client.py` to catch ValueError and TypeError too ? Or just add skipping parts when `data`'s type is already `dict`

andreyvelich: Hmm, I am not sure that we can modify this auto-generated file.
Let's keep it `self.data = json.dumps(obj)` as you said before.
Does it deserialize properly in that case ?

anencore94: Yes, with your FakeResponse class and `json.dumps` as in the last commit

```python
    resp = client.list_experiments(namespace='jaeyeon-kim')
    # print(resp)
    print(type(resp[0]))
    print(type(resp[0].metadata))
    print(type(resp[0].spec))
    print(type(resp[0].spec.algorithm))

    resp2 = client.list_trials(name='katib-hpo-test', namespace='jaeyeon-kim')
    # print(resp2[0])
    print(type(resp2[0]))
    print(type(resp2[0].spec.objective))
```
- output : 
```
<class 'kubeflow.katib.models.v1beta1_experiment.V1beta1Experiment'>
<class 'kubernetes.client.models.v1_object_meta.V1ObjectMeta'>
<class 'kubeflow.katib.models.v1beta1_experiment_spec.V1beta1ExperimentSpec'>
<class 'kubeflow.katib.models.v1beta1_algorithm_spec.V1beta1AlgorithmSpec'>
<class 'kubeflow.katib.models.v1beta1_trial.V1beta1Trial'>
<class 'kubeflow.katib.models.v1beta1_objective_spec.V1beta1ObjectiveSpec'>
```

I'm afraid how to add unit-tests for these. I see there isn't any unit-test codes yet for [sdk](https://github.com/kubeflow/katib/tree/master/sdk/python/v1beta1/kubeflow/katib) right ?

andreyvelich: That's great, thank you!
Please can you also check that `status.startTime` was properly created ? I am worried about this comment: https://github.com/kubernetes-client/python/issues/977#issuecomment-594045477.

Yeah, we don't have tests for the SDK currently: https://github.com/kubeflow/katib/blob/master/hack/gen-python-sdk/post_gen.py#L87.
We can think about it in the following PRs.

> Yes, with your FakeResponse class and json.dumps as in the last commit

anencore94: > Yeah, we don't have tests for the SDK currently: https://github.com/kubeflow/katib/blob/master/hack/gen-python-sdk/post_gen.py#L87.
We can think about it in the following PRs.

Thanks for your support !

anencore94: > Please can you also check that status.startTime was properly created ? I am worried about this comment: kubernetes-client/python#977 (comment).

The output looks good :)
```python
    print(type(resp[0].status))
    print(type(resp[0].status.start_time))
    print(resp[0].status.start_time)
```

- output
```
<class 'kubeflow.katib.models.v1beta1_experiment_status.V1beta1ExperimentStatus'>
<class 'datetime.datetime'>
2021-08-28 12:00:30+00:00
```

anencore94: For make sure, here's the test for `list_trials`
```python
    resp2 = client.list_trials(name='katib-hpo-test', namespace='jaeyeon-kim')
    print(type(resp2[0]))
    print(type(resp2[0].status))
    print(type(resp2[0].status.start_time))
    print(resp2[0].status.start_time)

```

```
<class 'kubeflow.katib.models.v1beta1_trial.V1beta1Trial'>
<class 'kubeflow.katib.models.v1beta1_trial_status.V1beta1TrialStatus'>
<class 'datetime.datetime'>
2021-08-28 12:01:56+00:00
```

andreyvelich: Great, thank you so much for taking this!

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
```diff
@@ -12,9 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import absolute_import
+
```
andreyvelich, line 16: Do you need this ? 
```suggestion
```

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
```diff
@@ -12,9 +12,15 @@
...
+import json
 import os
 
 
+# python 2 and python 3 compatibility library
+
+
```
andreyvelich, line 23:
```suggestion
```

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
anencore94: Not really :) Thanks

andreyvelich (commented):
Thank you @anencore94!
/lgtm
@gaocegege @johnugeorge Please take a look.

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
anencore94: It was a long road! Thank you so much for bring me here :+1:

Issue #1649, opened 2021-08-31
Go Test failed sometimes


anencore94:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

One of github action for `Go Test` failed sometimes, but it succeed when retest.

I've check for some cases:
- https://github.com/kubeflow/katib/runs/3444147420?check_suite_focus=true
- https://github.com/kubeflow/katib/runs/3451003535?check_suite_focus=true
- https://github.com/kubeflow/katib/runs/3451166076?check_suite_focus=true
- https://github.com/kubeflow/katib/runs/3435251159?check_suite_focus=true
- https://github.com/kubeflow/katib/runs/3413375074?check_suite_focus=true

All case failed for either one of these:
```
    trial_controller_test.go:261: 
        Timed out after 40.000s.
        Expected
            <bool>: false
        to be true
FAIL
```

```
    experiment_controller_test.go:350: 
        Timed out after 40.000s.
        Expected
            <bool>: false
        to be true
```

**What did you expect to happen:**

If these issue occured just from the network issue, what about increasing the timeout ?
Does it make sense ? WYDT ?

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version (`kfctl version`):
- Minikube version (`minikube version`):
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

andreyvelich:
Thank you for creating this @anencore94!
Yes, this unit test is flaky.
[envtest](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_suite_test.go#L42) sometime can fail in the [Controller test](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_test.go#L89).

Any ideas how to improve this and keep the same coverage level would be appreciated!

anencore94:
@andreyvelich Thanks :)
If the envtest you linked failed, could you link the action log in this issue ?
So we can keep track of it, then anybody could fix it

andreyvelich:
Let's continue discussion: https://github.com/kubeflow/katib/pull/1654#pullrequestreview-747137194 for improving controller unit test here.
/priority p2

gaocegege:
> Maybe we should split it in 3 different unit tests with separate test managers.

I am wondering why it works. Splitting test cases just makes it fail fast, I think.

[...]

PR #1654, opened 2021-09-03
[bugfix]: increase test timeout

anencore94:
- It might fix unit test fail issue described in #1649

#1649

**Checklist:**

### DIFF
--- a/pkg/controller.v1beta1/experiment/experiment_controller_test.go
+++ b/pkg/controller.v1beta1/experiment/experiment_controller_test.go
@@ -54,7 +54,7 @@ const (
 	namespace        = "default"
 	primaryContainer = "training-container"
 
-	timeout = time.Second * 40
+	timeout = time.Second * 80
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


andreyvelich (commented):
@anencore94 Thank you for taking this!
Are you sure that increasing timeout will help to avoid flaky unit tests ?

For example in this failed test: https://github.com/kubeflow/katib/runs/3523610062?check_suite_focus=true.
Trial status was updated to `Metrics Unavailable`, but tests were failed on the [previous step](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_test.go#L254-L261), where we compare observation results.

I think to fix it, we might need to split running Experiments in the different unit tests.

For example, currently in the [`TestReconcileBatchJob`](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_test.go#L89) we run 3 Experiments: 
1. Trial run with "Failed" BatchJob
2. Trail with "Complete" BatchJob and Available metrics.
3. Trail with "Complete" BatchJob and Unavailable metrics.

Maybe we should split it in 3 different unit tests with separate test managers.

WDYT @anencore94 @gaocegege @johnugeorge ?

gaocegege (commented):
/lgtm
/approve

gaocegege:
Sorry for the quick merge.

I think we can increase it and see If it works. In the best case, we should split the test cases. Should we create an issue for it?

andreyvelich:
No worries @gaocegege :)
Yes, let's continue the discussion under https://github.com/kubeflow/katib/issues/1649.

Issue #1635, opened 2021-08-24
Add version for Mock Generator


andreyvelich:
/kind feature

We should add new instruction to [Makefile](https://github.com/kubeflow/katib/blob/master/Makefile) to run [`mockgen.sh`](https://github.com/kubeflow/katib/blob/master/scripts/mockgen.sh) script with specific mock version.

That will help us to keep files consistent.

/good-first-issue

PR #1658, opened 2021-09-06
Add mockgen version to gen script

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1635.

I added v1.6.0 version for mockgen.
Also, I moved mockgen to hack, similar to Kubernetes: https://github.com/kubernetes/kubernetes/tree/master/hack.
Let's keep only user-facing scripts under `./scripts/v1beta1`

cc @gaocegege @johnugeorge


gaocegege (commented):
/lgtm

PR #1659, opened 2021-09-06
Remove README toc script

andreyvelich:
Since GitHub added default Table of Contents to the markdown files, we can remove this script.

cc @gaocegege @johnugeorge


gaocegege (commented):
/lgtm

Issue #1570, opened 2021-07-01
Refactoring README information


andreyvelich:
/kind discussion

As we discussed in the community meetings, we should keep Katib README simple with only necessary information.
My Suggestion that we remove the following sections:
1. [Getting Started](https://github.com/kubeflow/katib#getting-started)
2. [Name](https://github.com/kubeflow/katib#getting-started)
3. [Concepts in Katib](https://github.com/kubeflow/katib#concepts-in-katib), since it is similar to: https://www.kubeflow.org/docs/components/katib/overview/#katib-concepts.
For the `Search Algorithms` we can create a table with supported AutoML techniques, algorithms. @gaocegege @johnugeorge What do you think ?
4. [Components in Katib](https://github.com/kubeflow/katib#components-in-katib) - We can add this section to kubeflow.org docs.
5. [Web UI](https://github.com/kubeflow/katib#web-ui) - similar to https://www.kubeflow.org/docs/components/katib/overview/#hyperparameters-and-hyperparameter-tuning.
6. [GRPC API Documentation](https://github.com/kubeflow/katib#grpc-api-documentation) - Move it to kubeflow.org docs.
7. [Katib SDK](https://github.com/kubeflow/katib#katib-sdk) - Described in https://www.kubeflow.org/docs/components/katib/overview/#katib-interfaces.
7. [Quick Start](https://github.com/kubeflow/katib#quick-start)

[Installation](https://github.com/kubeflow/katib#installation) section should be simplified. Remove it completely and add link to: https://www.kubeflow.org/docs/components/katib/hyperparameter/#installing-katib or explain how to install particular version of Katib.

[Contributing](https://github.com/kubeflow/katib#contributing) and [Citation](https://github.com/kubeflow/katib#contributing) section can be moved to [Community](https://github.com/kubeflow/katib#community). 

What do you think @gaocegege @johnugeorge @jbottum @RFMVasconcelos ?

PR #1667, opened 2021-09-15
Refactor README 

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1570.

I made some changes in the README to leave only necessary information for our users.

Few questions:

1. How do you feel about Algorithm table in this PR with HP, NAS and EarlyStopping headers ?
1. Do we want to add `implementation` in the Algorithm table with type of framework that we use (e.g. `hyperopt`, `optuna`, `goptuna`) ?
1. Any other information that we can add in the README (e.g. `Why Katib ?`) ? 

Please take a look and give your feedback @gaocegege @johnugeorge @kimwnasptd @jbottum @RFMVasconcelos @anencore94 @tenzen-y @c-bata @g-votte


README.md
```diff
@@ -17,310 +17,205 @@ Katib supports
...
+## Release Version
+
+For the specific Katib release (for example `v0.11.1`) run this command:
+
+```
+kubectl apply -k "github.com/kubeflow/katib.git/manifests/v1beta1/installs/katib-standalone?ref=v0.11.1"
+```
+
+Learn more about various Katib installs in the
+[Kubeflow guide](https://www.kubeflow.org/docs/components/katib/hyperparameter/#katib-setup).
```
anencore94, line 158: What do you think about moving this sentence to right after `#Installation`, and add explanation following Installation is for stand-alone or for quick-start ?

anencore94 (commented):
Table looks nice :)

tenzen-y (changes requested):
Thanks for adding the supported K8s version to this documentation.

It might be better to use the `kustomize` binary instead of the built-in `kustomize`, since the `kustomize` version used by each user may be very different, and the built-in kustomize does not immediately bug fixes.

What do you think @andreyvelich ?

README.md
andreyvelich: That is a good point, thanks @anencore94

andreyvelich:
Thank you for the review @tenzen-y.

> It might be better to use the kustomize binary instead of the built-in kustomize, since the kustomize version used by each user may be very different, and the built-in kustomize does not immediately bag fixes.

Before the installation we have the requirements that user should install `kubectl >= 1.21` to deploy Katib.
Should we make sure that our Standalone `kustomize` manifest is compatible with that version of `kubectl` ?

That will remove additional step for the new users to install `kustomize` binary.

For most advance users they can always follow the [Kubeflow Katib guide](https://www.kubeflow.org/docs/components/katib/hyperparameter/#installing-katib) to manipulate `kustomize` manifests.

WDYT @tenzen-y @gaocegege @johnugeorge ?

tenzen-y:
@andreyvelich 
> For most advance users they can always follow the Kubeflow Katib guide to manipulate kustomize manifests.

It sounds good.  I was relieved to see that the detailed installation instructions are still available.

README.md
```diff
@@ -17,310 +17,213 @@ Katib supports
...
-```
-
-You can access the Katib UI using this URL: `http://localhost:8080/katib/`.
-
-### Katib SDK
+- Know more about various Katib interfaces in the
+  [introduction guide](https://www.kubeflow.org/docs/components/katib/overview/#katib-interfaces).
 
-Katib supports Python SDK:
+- TODO (andreyvelich): Add Katib components section from the website.
```
johnugeorge, line 194: Should we remove this till we add this section?

johnugeorge:
One thing that we have to add is , Why Katib is a better choice ?  We can take this up later

README.md
andreyvelich: Sure, does it look good to you: https://github.com/kubeflow/katib/blob/e2434e3c3c4001280d8ca1eb9d0af74fd17aed75/README.md#documentation ?
I made the similar sentences.

andreyvelich:
> One thing that we have to add is , Why Katib is a better choice ? We can take this up later

Sure, I created issue to track this.

gaocegege (commented):
/lgtm

Issue #1376, opened 2020-11-04
Remove use of ksonnet in tests #1608


jlewi:
/kind bug
ksonnet was deprecated over a year ago the tests should no longer be using ksonnet to define tests.

Related to #774

Issue #1601, opened 2021-08-01
Add Linter and Formatter for python source codes


anencore94:
/kind feature

**Describe the solution you'd like**

In kubeflow/katib, there are lots of source codes written in python. Even though there are well-defined CI processes for source codes written in Golang, however for Python codes, only charmed-katib seems to have such process now.

There should be a fixed linter, formatter and tester with some rule-config files(.flake8, .pylintrc, ...) and the github action workflow to check those all. For example, there are lots of unit test codes for suggestion, but current CI process doesn't check if they are failed or check the test coverage decreased in every PR.
If there are any convention checking tools, please share in katib community :)

**Anything else you would like to add:**
If you guys agree to this, I'd like to contribute to this feature in some parts.

andreyvelich:
That is a great suggestion and it will be very useful for us.
Thank you for your help @anencore94!

In addition to that, I think we should move our current Python tests from AWS CI Workflow to GitHub actions: https://github.com/kubeflow/katib/blob/master/test/scripts/v1beta1/python-tests.sh.

Currently, we are not using AWS Kubernetes cluster requirement to perform this workflow step.
We can create separate GitHub Actions Workflow for all Python tests (linter, formatter, unit), similar to Go: https://github.com/kubeflow/katib/tree/master/.github/workflows.

cc @gaocegege @johnugeorge ?

anencore94:
Using mypy to type check would be nice too.

gaocegege:
SGTM, I am not quite familiar with Python, I think we need it.

sandipanpanda:
I would like to work on this.
/assign

sagnik3788:
@sandipanpanda still working on it ?

sandipanpanda:
Sorry, this fell off my radar. @sagnik3788 feel free to work on this.
/unassign

forsaken628:
Which formater is better, [black](https://github.com/psf/black), [autopep8](https://github.com/hhatto/autopep8) or something else?

tenzen-y:
> Which formater is better, [black](https://github.com/psf/black), [autopep8](https://github.com/hhatto/autopep8) or something else?

We are supposed to use the flake8 and black.

Souradip121:
@sagnik3788 still working on it??

tenzen-y:
This has already been fixed by https://github.com/kubeflow/katib/pull/2407.
/close

/assign @Ygnas

PR #1674, opened 2021-09-23
Create Python script to run e2e Argo Workflow

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1376.

I added script to run our AWS Argo Workflow in python instead of deprecated `ksonnet`.
I removed [Python Unit Test](https://github.com/kubeflow/katib/blob/master/test/scripts/v1beta1/python-tests.sh) from the Workflow.
We will use GitHub Action to run these tests: https://github.com/kubeflow/katib/issues/1601

/cc @gaocegege @johnugeorge


andreyvelich:
Test seems to be working.
@gaocegege @johnugeorge Please take a look.
cc  @thunderboltsid

prow_config.yaml
```diff
@@ -45,12 +46,9 @@ workflows:
...
       - test/e2e/v1beta1/*
-      - test/scripts/v1beta1/*
-      - test/suggestion/v1beta1/*
-      - test/workflows/*
       - manifests/v1beta1/*
       - scripts/v1beta1/*
       - prow_config.yaml
       - Makefile
-    params:
+    kwargs:
```
johnugeorge, line 53: This got changed?

andreyvelich: @johnugeorge Check here: https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/run_e2e_workflow.py#L12-L17.
It uses `kwargs` in Python scripts.

PR #1678, opened 2021-09-24
Add envtest to check `reconcileRBAC`

tenzen-y:

Currently, there are not any tests to check the `reconcileRBAC` in the following coverage reports. So, I think it might be better to add tests to check it.

- [suggestion_controller_util.go#L95](https://coveralls.io/builds/43047408/source?filename=pkg%2Fcontroller.v1beta1%2Fsuggestion%2Fsuggestion_controller_util.go#L95)
- [suggestion_controller.go#L235](https://coveralls.io/builds/43047408/source?filename=pkg%2Fcontroller.v1beta1%2Fsuggestion%2Fsuggestion_controller.go#L235)


pkg/apis/controller/common/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
andreyvelich, line 1: Did you run `make generate` to update these files ?
What version of Go are you using ?

pkg/controller.v1beta1/suggestion/suggestion_controller_test.go
```diff
@@ -255,6 +246,84 @@ func TestReconcile(t *testing.T) {
 	// Test 4 - Update status condition for empty experiment
 	g.Expect(r.updateStatusCondition(&suggestionsv1beta1.Suggestion{}, oldS)).To(gomega.HaveOccurred())
 
+	// test 5 - Early Stopping Suggestion run
+	// Create early stopping config
+	jsonConfigEarlyStopping := map[string]katibconfig.EarlyStoppingConfig{
```
andreyvelich, line 251: Maybe we can add EarlyStopping in the Test 1, just extending [`katib-config`](https://github.com/kubeflow/katib/blob/9f6266084d357e664ee9080ae124e347bee4177f/pkg/controller.v1beta1/suggestion/suggestion_controller_test.go#L330) and [`Suggestion spec`](https://github.com/kubeflow/katib/blob/9f6266084d357e664ee9080ae124e347bee4177f/pkg/controller.v1beta1/suggestion/suggestion_controller_test.go#L319)?
Then, 1 test will covert: Deployment, Service and RBAC creation and we don't need to create another Suggestion in envtest ?
WDYT @tenzen-y ?

andreyvelich (commented):
Thank you for taking this @tenzen-y!
It's useful.

pkg/apis/controller/common/v1beta1/zz_generated.deepcopy.go
tenzen-y: > Did you run make generate to update these files ?

I didn't run `make generate`.

> What version of Go are you using ?

I'm using Go 1.17.1.

According to [Go 1.17 release note](https://golang.org/doc/go1.17#gofmt), if a file only has `// +build lines`, `//go:build` lines will be added.

> gofmt (and go fmt) now synchronizes //go:build lines with // +build lines. If a file only has // +build lines, they will be moved to the appropriate location in the file, and matching //go:build lines will be added. Otherwise, // +build lines will be overwritten based on any existing //go:build lines. For more information, see https://golang.org/design/draft-gobuild.

Should I remove `//go:build !ignore_autogenerated` ? @andreyvelich 
If we should remove it, it might be better to change the required Go version in [the developer guide](https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md#requirements).

pkg/controller.v1beta1/suggestion/suggestion_controller_test.go
tenzen-y: It makes sense! @andreyvelich 
I'd like to fix it!

pkg/apis/controller/common/v1beta1/zz_generated.deepcopy.go
andreyvelich: Makes sense, let's remove it for now.

We are using Go 1.15 currently: https://github.com/kubeflow/katib/blob/master/go.mod#L3.
I remember that we had issues with Go 1.16 and our packages.
I will submit PR soon to use Go 1.17.

tenzen-y: > I will submit PR soon to use Go 1.17.

It sounds good!
I'd like to remove it.

andreyvelich (commented):
Thank you for the updates @tenzen-y!
/lgtm

pkg/controller.v1beta1/suggestion/suggestion_controller_test.go
```diff
@@ -19,6 +19,9 @@ package suggestion
 import (
 	"context"
 	"encoding/json"
+	"github.com/kubeflow/katib/pkg/controller.v1beta1/util"
```
gaocegege, line 22: Can we move the katib deps to the third group and the third party deps to the second group?

tenzen-y: It makes sense.
I moved it!

gaocegege (commented):
/lgtm

PR #1683, opened 2021-09-28
Update Go version to 1.17

andreyvelich:
Check discussion: https://github.com/kubeflow/katib/pull/1678#discussion_r715949137.

I updated Go version to 1.17.
Go 1.17 explicitly adds indirect dependencies (See release notes: https://golang.org/doc/go1.17#go-command).

Also, I fixed problem with `code-generator` by creating `tools.go` file with required imports and moving it to `require`.
I believe, other projects follow the similar way.

/assign @tenzen-y @gaocegege @johnugeorge @anencore94


hack/tools.go
```diff
@@ -0,0 +1,30 @@
+// +build tools
```
tenzen-y, line 1:
```suggestion
//go:build tools
// +build tools
```
I think it needs to add `//go:build tools`.

https://github.com/kubeflow/katib/pull/1683/checks?check_run_id=3733490414#step:5:1295

tenzen-y (changes requested):
Thank you for your quick work! @andreyvelich

hack/tools.go
andreyvelich: Sure, makes sense.
I also updated `go get` to `go install` for `golangci-lint` since we are using Go 1.16+ (ref: https://golangci-lint.run/usage/install/#install-from-source).

tenzen-y: Sounds good!

tenzen-y:
LGTM
/lgtm

PR #1701, opened 2021-10-07
Fix same set for HyperParameters in Bayesian Optimization algorithm

### LINKED ISSUE #1700: Multiple trials with same set of parameters using bayesianoptimization and parallelTrialCount > 1
/kind bug

Using bayesianoptimization and parallelTrialCount > 1 will result in multiple trials having suggested the same set of parameters. This is due to the way https://github.com/kubeflow/katib/blob/2db65b2c6b349cfd4901bfe9ee1f308f7ea6444e/pkg/suggestion/v1beta1/skopt/base_service.py#L121-L125 are structured

fabianvdW:
Fixes #1700

### DIFF
--- a/pkg/suggestion/v1beta1/skopt/base_service.py
+++ b/pkg/suggestion/v1beta1/skopt/base_service.py
@@ -118,11 +118,11 @@ class BaseSkoptService(object):
 
         return_trial_list = []
 
-        for i in range(request_number):
-            skopt_suggested = self.skopt_optimizer.ask()
-            logger.info("New suggested parameters for Trial: {}".format(skopt_suggested))
+        skopt_suggested = self.skopt_optimizer.ask(n_points = request_number)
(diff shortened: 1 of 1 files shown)


pkg/suggestion/v1beta1/skopt/base_service.py
```diff
@@ -118,11 +118,11 @@ def getSuggestions(self, trials, request_number):
 
         return_trial_list = []
 
-        for i in range(request_number):
-            skopt_suggested = self.skopt_optimizer.ask()
```
johnugeorge, line 122: Won't this retrieve new unique value each time?

andreyvelich (commented):
It's a great fix, thank you @fabianvdW!
I think it was initially implemented with multiple calls for `ask`: https://github.com/kubeflow/katib/pull/777

PR #1704, opened 2021-10-07
[Bug Fix - new ui] : Fix bug on list type hp

seong7:

This PR fixes a bug described below.

When user edits hyper parameters' type from `'int' or 'double'` to `'Categorical' or 'Discrete'` and save the values,
image

The page falls into infinite error loop.
image

This bug has been caused by [this line](https://github.com/kubeflow/katib/blob/2db65b2c6b349cfd4901bfe9ee1f308f7ea6444e/pkg/new-ui/v1beta1/frontend/src/app/shared/params-list/parameter/parameter.component.ts#L85) in `parameter.component.ts`.

```typescript
this.paramCtrl
    .get('feasibleSpace')
    .setValue(group.get('feasibleSpace').value);
```

The method [setValue()](https://github.com/angular/angular/blob/a92a89b0eb127a59d7e071502b5850e57618ec2d/packages/forms/src/model.ts#L1547) of `FormGroup` throws when the new value is not exactly in the same structure of the existing [FormControl](https://angular.io/api/forms/FormControl#formcontrol).

This case, the existing `Control` has already been initialized as the structure below.
(This is the difference compared to 'adding a new param'. When adding, the error does not occur.)
```typescript
{
  min: 1,
  max: 1, 
  step: ''
}
```
But the new value has a different structure as below.
```typescript
{
  list: [ '111', '222' ]
}
``` 
To replace a `Control` in a `FormGroup`, we should use the method [`setControl()`](https://github.com/angular/angular/blob/6e30b83267cdb4be62cc2c9737aeac62d14ed513/packages/forms/src/model.ts#L1322) instead of `setValue()`.

Fixes #

**Checklist:**


gaocegege:

/assign @andreyvelich

pkg/new-ui/v1beta1/frontend/package-lock.json
```diff
@@ -1,8 +1,15797 @@
 {
```
andreyvelich, line 1: Why the `package-lock` was re-generated ?

andreyvelich (commented):
Thank you for this fix @seong7!
/assign @kimwnasptd

pkg/new-ui/v1beta1/frontend/package-lock.json
seong7: I think it's just how a newer version of NPM (over 7) works. (ref: [stackoverflow](https://stackoverflow.com/a/68479189/11717184))

andreyvelich: Does it mean that we have to update `npm` version in our UI development guide: https://github.com/kubeflow/katib/tree/master/pkg/new-ui/v1beta1#requirements ?

seong7: Can we know the progress of this PR? 😄 

I think I can just remove the change of the package-lock.json from this PR, since this PR is nothing to do with the dependency tree.

kimwnasptd: @seong7 apologies for the huge delay on this one. I'll take a closer look either later today or tomorrow tops.

kimwnasptd: I've left some comments on how to handle the `package-lock.json`. @seong7 indeed let's drop the changes in that file in this PR.

I'm looking at the other changes now, for which I'll leave discrete review comments in different messages later on if needed.

kimwnasptd:
The changes LTGM, also @seong7 very good job on describing the problem in the first message!

Could you only remove the changes in the `package-lock.json` file?

kimwnasptd:
Everything looks good, thanks for the work @seong7 and apologies for the delay!

/lgtm
/approve

PR #1721, opened 2021-10-25
Enhance/UI feasible space

seong7:

In parameter modal ui,
- this PR restricts the type of `Min`, `Max` and `Step` inputs as `number`, since `string` type of input must not be passed.
- this PR prevents the `Step` input from being `0` or `0.0000...`, since the step `0` does not make sense in the min-max range.

Fixes #

**Checklist:**

### DIFF
--- a/pkg/new-ui/v1beta1/frontend/src/app/shared/params-list/add-modal/add-modal.component.scss
+++ b/pkg/new-ui/v1beta1/frontend/src/app/shared/params-list/add-modal/add-modal.component.scss
@@ -21,3 +21,8 @@
     margin: 0 12px;
   }
 }
+
+input[type='number']::-webkit-outer-spin-button,
+input[type='number']::-webkit-inner-spin-button {
+  display: none;
+}
--- a/pkg/new-ui/v1beta1/frontend/src/app/shared/params-list/add-modal/add-modal.component.scss
+++ b/pkg/new-ui/v1beta1/frontend/src/app/shared/params-list/add-modal/add-modal.component.scss
@@ -21,8 +21,3 @@
     margin: 0 12px;
   }
 }
-
-input[type='number']::-webkit-outer-spin-button,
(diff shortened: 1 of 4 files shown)


seong7:
Can we know the progress of this PR? 😄

andreyvelich:
Sorry for the long waiting time @seong7.

@kimwnasptd Please can you check this PR also ?

pkg/new-ui/v1beta1/frontend/src/app/shared/params-list/add-modal/add-modal.component.scss
```diff
@@ -21,3 +21,8 @@
     margin: 0 12px;
   }
 }
+
+input[type='number']::-webkit-outer-spin-button,
```
kimwnasptd, line 25: Could we remove the use of `-webkit-outer-spin-button`?

It's non-standard https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-outer-spin-button, plus I also believe it's nice to include it. It makes it explicit that we are expecting a number.

Did you have a reason to remove the two arrows from the input, or was it just to ensure it will be the same as previously?

kimwnasptd (commented):
Only one small comment @seong7, thanks for the work!

pkg/new-ui/v1beta1/frontend/src/app/shared/params-list/add-modal/add-modal.component.scss
seong7: I just meant to keep it same as previous.

Thank you for pointing out the standard!
I just removed the lines and pushed.

kimwnasptd: Then we are good to roll!

kimwnasptd:
Thanks for helping improving the web app @seong7!

/lgtm
/approve

Issue #1382, opened 2020-11-09
Support for PBT (Population Based Training)


romeokienzler:
/kind feature

**Support for PBT (Population Based Training)**
Ray tune recently came out with support for PBT, Deepmind showed exceptional performance. Are we considering to support PBT in Katib as well?

https://arxiv.org/abs/1711.09846
https://deepmind.com/blog/article/population-based-training-neural-networks

andreyvelich:
Thank you for this information @romeokienzler, I think it's very exciting.

We definitely should investigate it and try to adopt it in Katib.
What do you think @gaocegege @johnugeorge ?

gaocegege:
My colleague @ezioliao is working on our internal AutoML system and he is interested in this issue.

ezioliao:
> My colleague @ezioliao is working on our internal AutoML system and he is interested in this issue.

PBT is already implemented in our own AutoML system and  I'm willing to solve it.

andreyvelich:
That would be great!
Thank you @ezioliao, let us know if you need any help.

andreyvelich:
@ezioliao Do you have time resources to work on this in 2021 ?
We can include PBT support in the 2021 Roadmap.

hongjunGu2019:
@andreyvelich Hello, I am very interested in this proposal. May I ask when it will be launched and how can I participate

romeokienzler:
@ezioliao

johnugeorge:
@hongjunGu2019 Please see #1833

hongjunGu2019:
> @hongjunGu2019 Please see #1833

Thanks

hongjunGu2019:
> @hongjunGu2019 Please see #1833

Thanks

david-thrower:
One enhancement to this which, I propose on this is an addition of a parameter to this, which a user can use to configure the training task to filter out invalid permutations of parameter values from each generation's trials, given rules like `minimum_skip_connection_depth` must be < `maximum_skip_connection_depth` , so as to filter out planned trials that would throw ValueErrors before they are executed. This would make more elaborate algorithms practical to run with the training op.

johnugeorge:
@david-thrower We have an interface to add validation checks per suggestion algorithm. Ref: https://github.com/kubeflow/katib/pull/1924

david-thrower:
[...]

PR #1726, opened 2021-10-29
fix: check if parameter references exist in experiment parameters

henrysecond1:

Hi, we experienced that the users tend to make mistakes like below when creating experiments using new UI (Thanks for the nice UI update!)

When users set parameter references in trial templates like below,

image

they need to enter the parameter reference as same as the hyperparameter defined in the experiment spec. 
The experiment is created without errors even if the users mistakenly set the wrong parameter reference.

I think it would be better to check the wrong parameter reference in the validating webhook
so that experiments are not created with the wrong parameter reference.

Please take a look, thanks!

### DIFF
--- a/pkg/webhook/v1beta1/experiment/validator/validator.go
+++ b/pkg/webhook/v1beta1/experiment/validator/validator.go
@@ -288,12 +288,13 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 		if _, ok := trialParametersRefs[parameter.Reference]; ok {
 			return fmt.Errorf("parameter reference %v can't be duplicated in spec.trialTemplate.trialParameters: %v", parameter.Reference, trialTemplate.TrialParameters)
 		}
-		// Check if parameter references exist in experiment parameters
-		if _, ok := experimentParameterNames[parameter.Reference]; !ok {
-			return fmt.Errorf("parameter reference %v does not exists in spec.Parameters: %v", parameter.Reference, instance.Spec.Parameters)
-		}
 		trialParametersNames[parameter.Name] = true
 		trialParametersRefs[parameter.Reference] = true
+
+		// Check if parameter reference exist in experiment parameters
+		if _, ok := experimentParameterNames[parameter.Reference]; !ok {
+			return fmt.Errorf("parameter reference %v does not exist in spec.parameters: %v", parameter.Reference, instance.Spec.Parameters)
+		}
 
 		// Check if trialParameters contains all substitution for Trial template
 		if !strings.Contains(trialTemplateStr, fmt.Sprintf(consts.TrialTemplateParamReplaceFormat, parameter.Name)) {
--- a/pkg/webhook/v1beta1/experiment/validator/validator.go
+++ b/pkg/webhook/v1beta1/experiment/validator/validator.go
@@ -265,6 +265,11 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 		return fmt.Errorf("unable to parse spec.trialTemplate: %v", err)
 	}
 
+	experimentParameterNames := make(map[string]bool)
+	for _, parameter := range instance.Spec.Parameters {
+		experimentParameterNames[parameter.Name] = true
+	}
+
 	trialParametersNames := make(map[string]bool)
 	trialParametersRefs := make(map[string]bool)
 
@@ -282,6 +287,10 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 		// Check if parameter references are not duplicated
 		if _, ok := trialParametersRefs[parameter.Reference]; ok {
 			return fmt.Errorf("parameter reference %v can't be duplicated in spec.trialTemplate.trialParameters: %v", parameter.Reference, trialTemplate.TrialParameters)
+		}
+		// Check if parameter references exist in experiment parameters
+		if _, ok := experimentParameterNames[parameter.Reference]; !ok {
+			return fmt.Errorf("parameter reference %v does not exists in spec.Parameters: %v", parameter.Reference, instance.Spec.Parameters)
 		}
 		trialParametersNames[parameter.Name] = true
 		trialParametersRefs[parameter.Reference] = true
--- a/pkg/webhook/v1beta1/experiment/validator/validator.go
+++ b/pkg/webhook/v1beta1/experiment/validator/validator.go
@@ -292,8 +292,10 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 		trialParametersRefs[parameter.Reference] = true
 
 		// Check if parameter reference exist in experiment parameters
-		if _, ok := experimentParameterNames[parameter.Reference]; !ok {
-			return fmt.Errorf("parameter reference %v does not exist in spec.parameters: %v", parameter.Reference, instance.Spec.Parameters)
+		if len(experimentParameterNames) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/webhook/v1beta1/experiment/validator/validator.go
```diff
@@ -283,6 +288,10 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 		if _, ok := trialParametersRefs[parameter.Reference]; ok {
 			return fmt.Errorf("parameter reference %v can't be duplicated in spec.trialTemplate.trialParameters: %v", parameter.Reference, trialTemplate.TrialParameters)
 		}
+		// Check if parameter references exist in experiment parameters
+		if _, ok := experimentParameterNames[parameter.Reference]; !ok {
+			return fmt.Errorf("parameter reference %v does not exists in spec.Parameters: %v", parameter.Reference, instance.Spec.Parameters)
+		}
```
anencore94, line 294: IMHO, How about move these codes after
```go
trialParametersNames[parameter.Name] = true
trialParametersRefs[parameter.Reference] = true
```

It may not important, but I feel like these should be right after codes right after checking duplicates.

pkg/webhook/v1beta1/experiment/validator/validator.go
```diff
@@ -283,6 +288,10 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 		if _, ok := trialParametersRefs[parameter.Reference]; ok {
 			return fmt.Errorf("parameter reference %v can't be duplicated in spec.trialTemplate.trialParameters: %v", parameter.Reference, trialTemplate.TrialParameters)
 		}
+		// Check if parameter references exist in experiment parameters
+		if _, ok := experimentParameterNames[parameter.Reference]; !ok {
+			return fmt.Errorf("parameter reference %v does not exists in spec.Parameters: %v", parameter.Reference, instance.Spec.Parameters)
```
anencore94, line 293: nit: does not exist

pkg/webhook/v1beta1/experiment/validator/validator.go
```diff
@@ -283,6 +288,10 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 		if _, ok := trialParametersRefs[parameter.Reference]; ok {
 			return fmt.Errorf("parameter reference %v can't be duplicated in spec.trialTemplate.trialParameters: %v", parameter.Reference, trialTemplate.TrialParameters)
 		}
+		// Check if parameter references exist in experiment parameters
```
anencore94, line 291: nit: parameter reference exist

pkg/webhook/v1beta1/experiment/validator/validator_test.go
```diff
@@ -524,6 +526,16 @@ spec:
...
+		// Wrong parameter reference in trialParameters
+		{
+			Instance: func() *experimentsv1beta1.Experiment {
+				i := newFakeInstance()
+				i.Spec.TrialTemplate.TrialParameters[1].Reference = "wrong-ref"
+				return i
+			}(),
+			Err:             true,
+			testDescription: "Wrong reference in Trial parameters",
+		},
```
anencore94, line 538: nit: How about make this test case as `invalidReferenceParameterTemplate` or `unknownReferenceParameterTemplate` rather than `validTemplate5`?
Also, how about more detailed description? something like `Trial template contains Trial Parameters which weren't referenced from spec.parameters`

anencore94 (commented):
Thank you for such a nice contribution :) I think this feature helps many others

pkg/webhook/v1beta1/experiment/validator/validator.go
henrysecond1: Thanks for the suggestion, done :)

pkg/webhook/v1beta1/experiment/validator/validator.go
henrysecond1: done

pkg/webhook/v1beta1/experiment/validator/validator.go
henrysecond1: done

pkg/webhook/v1beta1/experiment/validator/validator_test.go
henrysecond1: I don't like the name `validTemplate5` either and I agree that `invalidReferenceParameterTemplate` is better.. but I guess that some rules are applied to mock template names (maybe I misunderstood it?).

- Mock templates with name `validTemplate*` have a valid job, while mock templates with specific names like `missedParameterTemplate`, `invalidParameterTemplate` do not have a valid job.

So I named it as a `validTemplate5` (since it has a valid job)... Please let me know if you still think it's better to name it as `invalidReferenceParameterTemplate` instead of `validTemplate5`?

I changed the description as you suggested, thanks :)

henrysecond1:
@anencore94 Thanks for the review! PTAL

pkg/webhook/v1beta1/experiment/validator/validator.go
```diff
@@ -286,6 +291,11 @@ func (g *DefaultValidator) validateTrialTemplate(instance *experimentsv1beta1.Ex
 		trialParametersNames[parameter.Name] = true
 		trialParametersRefs[parameter.Reference] = true
 
+		// Check if parameter reference exist in experiment parameters
+		if _, ok := experimentParameterNames[parameter.Reference]; !ok {
+			return fmt.Errorf("parameter reference %v does not exist in spec.parameters: %v", parameter.Reference, instance.Spec.Parameters)
+		}
```
andreyvelich, line 297: I think for NAS Experiments that doesn't work.
For example, for the [ENAS Experiment](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/nas/enas-cpu.yaml#L126-L132), the Suggestion returns `architecture` and `nn_config` which don't exist in `spec.Parameters`.
For HP Experiments, currently all parameters that Suggestions return exist in `spec.parameters`, but I am not sure if for other AutoML techniques that always be true (for example for [PBT](https://github.com/kubeflow/katib/issues/1382), AutoMC, etc.).

I think for now, we can check if  `experimentParameterNames` is not empty than we can apply this check.

WDYT @henrysecond1 @anencore94 @gaocegege @johnugeorge ?

andreyvelich (commented):
Thank you for driving this @henrysecond1!
I left comment.

pkg/webhook/v1beta1/experiment/validator/validator.go
gaocegege: SGTM!

henrysecond1: @andreyvelich Thank you for pointing it out. Sounds good to me, too :)

anencore94: Yes that make sense ! Thanks

pkg/webhook/v1beta1/experiment/validator/validator_test.go
anencore94: Since the validation logic was changed to check only it's empty, maybe we could change the test code :)
And in that case, yes the variable name `validTemplate5` would be better.

henrysecond1:
@anencore94 @andreyvelich I updated the test code (for the case when `spec.parameters` is empty). Could you re-check PR? Thanks for your time.

andreyvelich:
Thank you for updating this @henrysecond1!
/lgtm

gaocegege (approved):
/lgtm

Issue #1637, opened 2021-08-24
Rename request_number parameter in gRPC manager


andreyvelich:
/kind feature

This PR: https://github.com/kubeflow/katib/pull/1633 introduces `total_request_number ` parameter.
To be consistent, we should rename `request_number` to `current_request_number`.
After that, all suggestions have to be modified.

Does it sound good to you @johnugeorge @gaocegege ?

johnugeorge:
LGTM 

/assign

gaocegege:
SGTM

PR #1728, opened 2021-11-02
Modify gRPC API with Current Request Number

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1637.

I changed `request_number` to `current_request_number` to be consistent.

/hold for the review

Please take a look.

/assign @gaocegege @johnugeorge @tenzen-y @anencore94


anencore94:
Maybe it' ok to do not change the original proposal [here](https://github.com/andreyvelich/katib/blob/issue-1637-rename-request-number/docs/proposals/suggestion.md#grpc-api) ?

pkg/suggestion/v1beta1/hyperband/service.py
```diff
@@ -48,7 +48,7 @@ def GetSuggestions(self, request, context):
                 # Hyperband outlerloop has finished
                 return reply
             # This is a hack to get request number.
```
anencore94, line 50: nit: how about change this comment to add the word current ?

pkg/suggestion/v1beta1/skopt/base_service.py
```diff
@@ -118,13 +118,13 @@ def getSuggestions(self, trials, request_number):
...
 
-        skopt_suggested = self.skopt_optimizer.ask(n_points=request_number)
+        skopt_suggested = self.skopt_optimizer.ask(n_points=current_request_number)
         for suggestion in skopt_suggested:
             logger.info("New suggested parameters for Trial: {}".format(suggestion))
             return_trial_list.append(
                 BaseSkoptService.convert(self.search_space, suggestion))
 
-        logger.info("GetSuggestions return {} new Trials\n\n".format(request_number))
+        logger.info("GetSuggestions returns {} new Trials\n\n".format(current_request_number))
```
anencore94, line 127: maybe it's not about this PR directly, but isn't this log should be `len(skopt_suggested)` or `len(return_trial_list)` rather than `current_request_number` ?

pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go
```diff
@@ -102,19 +102,19 @@ func (g *General) SyncAssignments(
...
 	}
 
 	// Get new suggestions
 	responseSuggestion, err := rpcClientSuggestion.GetSuggestions(ctx, requestSuggestion)
 	if err != nil {
 		return err
 	}
-	logger.Info("Getting suggestions", "endpoint", endpoint, "Number of request parameters", requestNum, "Number of response parameters", len(responseSuggestion.ParameterAssignments))
-	if len(responseSuggestion.ParameterAssignments) != requestNum {
+	logger.Info("Getting suggestions", "endpoint", endpoint, "Number of request parameters", currentRequestNum, "Number of response parameters", len(responseSuggestion.ParameterAssignments))
```
anencore94, line 116: nit: Number of request parameters -> Number of current request parameters

anencore94 (commented):
SGTM !

gaocegege (commented):
LGTM :+1:

Thanks!

tenzen-y (approved):
LGTM!
/lgtm

andreyvelich:
> Maybe it' ok to do not change the original proposal [here](https://github.com/andreyvelich/katib/blob/issue-1637-rename-request-number/docs/proposals/suggestion.md#grpc-api) ?

I think we decide to change API name to be consistent with `total_request_number`.
In original proposal we don't have it.

Maybe we can keep `request_number` gRPC API as `deprecated` for 1 release (until Katib 0.14 will be released) ? 
Some Katib users might deploy their own custom Algorithm service where they are using this API.

For our Suggestions we can rename `request_number` to `current_request_number` in this PR.

WDYT @gaocegege @anencore94 @johnugeorge ?

pkg/suggestion/v1beta1/skopt/base_service.py
andreyvelich: Nice catch!

gaocegege:
> Maybe we can keep request_number gRPC API as deprecated for 1 release (until Katib 0.14 will be released) ?
Some Katib users might deploy their own custom Algorithm service where they are using this API.


SGTM

anencore94:
>Maybe we can keep request_number gRPC API as deprecated for 1 release (until Katib 0.14 will be released) ?
Some Katib users might deploy their own custom Algorithm service where they are using this API.

As we discussed last working group meeting, I also agree with this suggestion. What's left now is to decide the way to notify the deprecation of this API. Is it enough to announce on Katib v0.13's Release note and [here](https://github.com/kubeflow/katib/blob/66d3d1f8a5c7ba8c5a14365dbed58f849cf60989/pkg/apis/manager/v1beta1/api.proto#L292)

andreyvelich:
As we discussed, I left `request_number` with `[deprecated=true]` parameter in gRPC API. We will also leave a note in the Katib 0.13 release notes about it.
As I can see, some of the Kubernetes API also follows this: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto#L298.

Please take a look.

tenzen-y:
> I left request_number with [deprecated=true] parameter in gRPC API. We will also leave a note in the Katib 0.13 release notes about it.

SGTM

anencore94:
> As we discussed, I left request_number with [deprecated=true] parameter in gRPC API. We will also leave a note in the Katib 0.13 release notes about it.
As I can see, some of the Kubernetes API also follows this: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto#L298.

It make sense :) Thanks for sharing @andreyvelich 
/lgtm

andreyvelich:
Thanks everyone for the review!
/hold cancel

Issue #1636, opened 2021-08-24
Grid Algorithm fails for int parameters


andreyvelich:
/kind bug

Grid fails with `Chocolate db is exhausted, increase Search Space or decrease maxTrialCount!` error when running this example:
```yaml
apiVersion: "kubeflow.org/v1beta1"
kind: Experiment
metadata:
  namespace: kubeflow-user-example-com
  name: grid-example
spec:
  objective:
    type: maximize
    goal: 0.99
    objectiveMetricName: Validation-accuracy
    additionalMetricNames:
      - Train-accuracy
  algorithm:
    algorithmName: grid
  parallelTrialCount: 6
  maxTrialCount: 12
  maxFailedTrialCount: 3
  parameters:
    - name: num-layers
      parameterType: int
      feasibleSpace:
        min: "3"
        max: "6"
    - name: optimizer
      parameterType: categorical
      feasibleSpace:
        list:
          - sgd
          - adam
          - ftrl
  trialTemplate:
    primaryContainerName: training-container
    trialParameters:
      - name: numberLayers
        description: Number of training model layers
        reference: num-layers
      - name: optimizer
        description: Training model optimizer (sdg, adam or ftrl)
        reference: optimizer
    trialSpec:
      apiVersion: batch/v1
      kind: Job
      spec:
        template:
          metadata:
            annotations:
              sidecar.istio.io/inject: "false"
          spec:
            containers:
              - name: training-container
                image: docker.io/kubeflowkatib/mxnet-mnist:v1beta1-45c5727
                command:
                  - "python3"
                  - "/opt/mxnet-mnist/mnist.py"
                  - "--batch-size=64"
                  - "--num-layers=${trialParameters.numberLayers}"
                  - "--optimizer=${trialParameters.optimizer}"
                  - "--num-epochs=1"
            restartPolicy: Never
```

We should verify how [quantized_uniform](https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/chocolate/base_service.py#L62-L67) distribution works in Chocolate.
For some reason, not all parameters are generated.

cc @johnugeorge

johnugeorge:
It looks fine as high is not counted in Chocolate

https://chocolate.readthedocs.io/en/latest/api/space.html#chocolate.quantized_uniform

PR #1732, opened 2021-11-11
Fix Range for Int and Double values in Grid

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1636.

In Chocolate Suggestion [`quantized_uniform`](https://chocolate.readthedocs.io/en/latest/api/space.html#chocolate.quantized_uniform) distribution uses half-open interval.

/assign @tenzen-y @gaocegege @johnugeorge


gaocegege (commented):
LGTM!

Thanks for the fix.

/lgtm

PR #1743, opened 2021-11-23
Fix Status Handling in Charmed Operators

DomFleischmann:
- Handle status of operators only in set_pod_spec function.
- Centralize all the logic in set_pod_spec and add helper checks.


ca-scribner (approved):
lgtm

andreyvelich (approved):
/approve

PR #1746, opened 2021-11-25
Add Workflow to Publish Katib Images

andreyvelich:
After this PR, images will be automatically published to our Docker Hub after every commit to the `master` branch.
So we don't need to manually run [`make push-latest`](https://github.com/kubeflow/katib/blob/master/Makefile#L58) after each PR.
We still need to update repository settings with Docker Hub token.

GitHub Actions has limitation of 20 concurrent jobs: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits in Free plan, that is why I split the workflows to: `Core`, `Algorithms`, and `Trials`.

Please take a look at this workflow.


.github/workflows/publish-algorithm-images.yaml
```diff
@@ -0,0 +1,99 @@
+name: Publish AutoML Algorithm Images
+
+on:
+  push:
+    branches:
+      - master
```
anencore94, line 6: Since this workflow runs when push event for master branch (when PR merged), We couldn't prevent docker build failed case previously.

How about make another template with `push:false` option, and run such events for `on.pull_requests`?
```
      - name: Build and Push
        uses: docker/build-push-action@v2
        with:
          context: .
          file: ${{ inputs.dockerfile }}
          push: false <<< only build
```

This could make sure that master branch is always stable for building docker image.

anencore94:
>GitHub Actions has limitation of 20 concurrent jobs: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits in Free plan, that is why I split the workflows to: Core, Algorithms, and Trials.

I think my suggestion method is somewhat bothersome stuff, but we could save limitation if we split each image's trigger point to only dependent PATH changed.

For example:
```
# for suggestion-hyperopt image,
on:
  push(or pull_request):
    paths:
      - "cmd/suggestions/hyperopt"
      - "pkg/apis/manager"
      - "pkg/suggestion/v1beta1/hyperopt"

# for enas-cnn-cifar10 image,
on:
  push(or pull_request):
    paths:
      - "examples/v1beta1/trial-images/enas-cnn-cifar10"
```

.github/workflows/publish-algorithm-images.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  suggestion-darts:
+    name: Suggestion DARTS
+    uses: kubeflow/katib/.github/workflows/template-publish-image.yaml@master
+    with:
+      image: docker.io/kubeflowkatib/suggestion-darts
+      dockerfile: cmd/suggestion/nas/darts/v1beta1/Dockerfile
+    secrets:
+      dockerhub_username: ${{ secrets.DOCKERHUB_USERNAME }}
+      dockerhub_token: ${{ secrets.DOCKERHUB_TOKEN }}
```
tenzen-y, line 88: Should we use [`composite-run-steps`](https://github.com/actions/runner/blob/main/docs/adrs/0549-composite-run-steps.md) and [`matrix`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) to further reuse action scripts?

```suggestion
jobs:
  # Suggestion Images.
  suggestion:
    runs-on: ubuntu-20.04
    strategy:
      matrix:
        suggestion-name:
          - "hyperopt"
          - "chocolate"
          - "hyperband"
          - "skopt"
          - "goptuna"
          - "optuna"
        basedir: ["suggestion"]
        include:
          - suggestion-name: "enas"
            basedir: "suggestion/nas"
          - suggestion-name: "darts"
            basedir: "suggestion/nas"
    steps:
      - name: Checkout
        uses: actions/checkout@v2
        with:
          fetch-depth: 1
      - name: Suggestion ${{ matrix.suggestion-name }}
        uses: ./.github/workflows/template-publish-image
        with:
          image: docker.io/kubeflowkatib/suggestion-${{ matrix.suggestion-name }}
          dockerfile: cmd/${{ matrix.basedir }}/${{ matrix.suggestion-name }}/v1beta1/Dockerfile
```

.github/workflows/template-publish-image.yaml
```diff
@@ -0,0 +1,54 @@
...
+        required: true
+      dockerhub_token:
+        required: true
+
+jobs:
+  publish-image:
+    name: Publish Image
+    # Trigger Publish Workflow only for kubeflow/katib repository.
+    if: github.repository == 'kubeflow/katib'
+    runs-on: ubuntu-latest
```
tenzen-y, line 23:
```suggestion
runs:
  using: "composite"
```
Same comment.

tenzen-y (commented):
This feature is very useful!
I left some comments.

.github/workflows/publish-algorithm-images.yaml
andreyvelich: @anencore94 I think our AWS CI workflow verifies that all images can be built: https://github.com/kubeflow/katib/blob/master/test/e2e/v1beta1/argo_workflow.py#L327-L339.
Which means PRs that are merged to the `master` has valid docker build.
Does it make sense ?

andreyvelich:
> I think my suggestion method is somewhat bothersome stuff, but we could save limitation if we split each image's trigger point to only dependent PATH changed.

@anencore94 I think currently we give our users experience when they can use images with specific commit for all Katib components.
For example: `v1beta1-0e17302`, [Katib Controller image](https://hub.docker.com/layers/kubeflowkatib/katib-controller/v1beta1-0e17302/images/sha256-a1a4751b3646a02806e326004f3b0f62f1569e60badb32b5b09ab3856ff6f920?context=repo), [MXNet Mnist](https://hub.docker.com/layers/kubeflowkatib/mxnet-mnist/v1beta1-0e17302/images/sha256-32a12aa9eaf69a6b9dfe1126f63fe5000485809c1fda80553ee7b6c1614034ab?context=repo).

I think that approach simplify Katib usage to understand which changes our users try to deploy.
WDYT @anencore94  ?

.github/workflows/publish-algorithm-images.yaml
andreyvelich: In that approach each image will be built in a separate `step`, right ?
I use separate job to build each image in favour on this feature: https://github.com/github/roadmap/issues/271
Once it will be implemented we can easily restart the specific job that was failed (for example connection issue), instead of re-run the whole workflow.
What do you think @tenzen-y ?

tenzen-y: > In that approach each image will be built in a separate step, right ?

Maybe, I think my suggestion script will build in a separate `job` in the following.
Is the following different from the `job` you are expecting? @andreyvelich 

https://github.com/tenzen-y/katib/actions/runs/1506407092

スクリーンショット 2021-11-27 0 55 22

andreyvelich: As I can see it also creates the separate Job for each build, thank you for the testing.
Let me try to test it on my end also.

andreyvelich: @tenzen-y Do you know how to pass secret in steps once we apply reusable workflow on the step level ?
As I can see, secrets can be accessible only in job level: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idsecrets.

tenzen-y: If we use `composite action`, we can pass token and username using the global env field in the following.

- publish-algorithm-images.yaml

```yaml
env:
  dockerhub_username: ${{ secrets.DOCKERHUB_USERNAME }}
  dockerhub_token: ${{ secrets.DOCKERHUB_TOKEN }}
on: push

jobs:
  # Suggestion Images.
...
```

- action.yaml

```yaml
...
  using: "composite"
  steps:

    - name: Set Up Docker Buildx
      uses: docker/setup-buildx-action@v1

    - name: Docker Login
      uses: docker/login-action@v1
      with:
        username: ${{ env.dockerhub_username }}
        password: ${{ env.dockerhub_token }}
...
```

The following is the example that I actually run actions in my forked repository.

- [actions log](https://github.com/tenzen-y/katib/runs/4337719723?check_suite_focus=true#step:3:142)

```shell
...
Run docker/login-action@v1
  with:
    username: ***
    password: ***
    logout: true
  env:
    dockerhub_username: ***
    dockerhub_token: ***
Logging into Docker Hub...
Login Succeeded!
...
```

- Settings/Secrets

スクリーンショット 2021-11-27 6 19 46

andreyvelich: Thank you for this information @tenzen-y!
I refactored the workflow using composite run.
To be consistent and easy to read, I use similar `matrix` structure for Core, Algorithms, and Trials workflow.
In that case, we don't need to create separate job for Early Stopping, Metrics Collector, etc.
Please let me know what do you think.

tenzen-y: It sounds good to me! @andreyvelich

tenzen-y (commented):
Thank you for the updating! @andreyvelich 
/lgtm

.github/workflows/publish-algorithm-images.yaml
anencore94: Yes, I didn't know there were an AWS CI workflows.
If so, how about create an **AWS CI-success badge** to master branch's README.md, if it is possible to integrate ?

anencore94:
>@anencore94 I think currently we give our users experience when they can use images with specific commit for all Katib components.
For example: v1beta1-0e17302, Katib Controller image, MXNet Mnist.
I think that approach simplify Katib usage to understand which changes our users try to deploy.
WDYT @anencore94 ?

Yeah, This could make users feel much easier to try on. Thanks :)

.github/workflows/publish-algorithm-images.yaml
tenzen-y: Hi, @anencore94.
We can see E2E running in AWD CI [here](https://argo.kubeflow-testing.com/workflows/).
For example, E2E for this PR is [this](https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-katib-presubmit-e2e-v1beta1-1746-bc60821-8000-497b).

andreyvelich:
@james-jwu @zijianjoy @Bobgy Please take a look at this PR.
We have to update Katib repository settings with our Docker Hub token.

.github/workflows/template-publish-image/action.yaml
```diff
@@ -0,0 +1,40 @@
...
+  using: composite
+  steps:
+    - name: Set Up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
+    - name: Docker Login
+      uses: docker/login-action@v1
+      with:
+        username: ${{ env.DOCKERHUB_USERNAME }}
+        password: ${{ env.DOCKERHUB_TOKEN }}
```
zijianjoy, line 21: Should it be `secrets.DOCKERHUB_TOKEN` instead?

andreyvelich:
We updated the token with @zijianjoy, let's try to merge this PR.
/hold cancel

PR #1751, opened 2021-11-29
Update links for issue template

andreyvelich:
Users don't need to check Kubeflow version by using `kfctl`.
I think Katib version should be enough for us.

Also, I added links to our docs and Slack channel to the issue config.

/assign @tenzen-y @gaocegege @johnugeorge


tenzen-y (commented):
LGTM👍
/lgtm

Issue #1754, opened 2021-12-01
Recent metrics filter change broke metrics collection for non-scientific values


RDarrylR:
/kind bug

**What steps did you take and what happened:**
The DefaultFilter value for the File Metrics Collector was recently changed to fix issues parsing scientific notation values but with the change multi digit number values are no longer parsed correctly and are truncated at the first digit.

**What did you expect to happen:**
Parsing of normal number values would work correctly.

**Anything else you would like to add:**

My metrics values that used to parse correctly are now truncated at the first digit. You can see it using a tool like https://regex101.com/

Old filter = `([\w|-]+)\s*=\s*((-?\d+)(\.\d+)?)`

New filter = `([\w|-]+)\s*=\s*([+-]?\d(\.\d+)?([Ee][+-]?\d+)?)`

Metrics like the following don't get parsed correctly with the new filter - they only match the first digit

metric1=888
metric2=888.33

metrics collector will get just 8 for these metrics with the new filter

**Environment:**

- Katib version (check the Katib controller image version): latest from master 
- Kubernetes version: (`kubectl version`): 1.18.20
- OS (`uname -a`): ubuntu 20.04

---

Impacted by this bug? Give it a 👍 We prioritize the issues with the most 👍

andreyvelich:
Thank you so much for finding this @RDarrylR!
I think we forget to add `+` in the regex.

I would suggest that we use this regex to support `-.111` format also.
```
([\w|-]+)\s*=\s*([+-]?\d*(\.\d+)?([Ee][+-]?\d+)?)
```

WDYT @RDarrylR  @kubeflow/wg-automl-leads @anencore94  ?

/assign @andreyvelich 
/priority p0

RDarrylR:
@andreyvelich That new version seems to work for all the cases I'm using.

PR #1755, opened 2021-12-01
Fix the default Metrics Collector regex

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1754.
I fixed the default regex format.
I will create separate PR to update the website doc.

/assign @kubeflow/wg-automl-leads 
/cc @RDarrylR @anencore94


anencore94:
That's my bad. Thanks so much for the fix 
/lgtm

Issue #1554, opened 2021-06-11
Use release tags for the Trial images


andreyvelich:
/kind feature

As discussed in the AutoML meeting at 2021-03-24, we decided to use the latest tag in the `master` branch and release tag in the `release-` branch for the [Trial images](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/hp-tuning/random.yaml#L56).
Then, users can deploy examples for the specific Katib version and use the latest changes.

/cc @gaocegege @johnugeorge @yanniszark

gaocegege:
LGTM!

PR #1757, opened 2021-12-02
Use release tags for Trial images 

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1554.
I added latest tag for Trial images.
For SDK examples, I use release tags, since we are installing specific version of Katib SDK.

Also, I modified `update-images` script to re-use it in release and e2e tests.
Currently, this script updates all Katib images that we maintain.

Please take a look.

/assign @gaocegege @tenzen-y @anencore94 @johnugeorge


tenzen-y:
LGTM 👍
/lgtm
/hold

tenzen-y (commented):
LGTM 👍
/lgtm
/hold

andreyvelich:
This PR should be ready.
/hold for the review

anencore94 (approved):
LGTM!
just a small question: Is there some special reason to use semantic version in notebook files rather than using `latest` ?

andreyvelich:
> Is there some special reason to use semantic version in notebook files rather than using `latest`

Since we install the specific version of Katib SDK in our notebooks, it's better to be consistent with the Trial images.

andreyvelich:
@ca-scribner @DomFleischmann Any ideas why Charms can fail ?

PR #1784, opened 2022-01-19
Update webhook failurePolicy from Ignore to Fail

knkski:

Updates `katib-controller` webhook for operator and manifests to set `failurePolicy` to `Fail`. This allows for greater visibility on errors. Also updates webhook creation handling in `katib-controller` operator.

### DIFF
--- a/manifests/v1beta1/components/webhook/webhooks.yaml
+++ b/manifests/v1beta1/components/webhook/webhooks.yaml
@@ -5,7 +5,6 @@ metadata:
 webhooks:
   - name: validator.experiment.katib.kubeflow.org
     sideEffects: None
-    failurePolicy: Fail
     admissionReviewVersions:
       - v1
     clientConfig:
@@ -32,7 +31,6 @@ metadata:
 webhooks:
   - name: defaulter.experiment.katib.kubeflow.org
     sideEffects: None
-    failurePolicy: Fail
     admissionReviewVersions:
       - v1
     clientConfig:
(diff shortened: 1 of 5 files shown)


gaocegege (commented):
/cc @andreyvelich

andreyvelich (commented):
Thank you for this update @knkski!
Do we want to include this change to Katib 0.13 release ?

andreyvelich:
It seems that this change doesn't work on K8s 1.18.
@knkski Have you tested this change on that version ?

knkski:
I've only tested as far back as 1.21. It does look like mutating/validating webhooks in 1.18 have a `failurePolicy` field that allows for `Fail` or `Ignore`, so I'm not sure offhand why it wouldn't work. The docs say that the field defaults to `Fail`, which is what this change does anyways, so I can try just removing the field altogether.

As far as which version to release this in, no strong preference. Whichever works best for you.

manifests/v1beta1/components/webhook/webhooks.yaml
```diff
@@ -53,7 +51,6 @@ webhooks:
           - experiments
   - name: mutator.pod.katib.kubeflow.org
     sideEffects: None
-    failurePolicy: Ignore
```
tenzen-y, line 56: Hi, @knkski !
I think that we need to specify `Ignore` in `mutator.pod.katib.kubeflow.org` since the katib-cert-generator job generates a Pod before the certificate is injected into `MutatingWebhookConfiguration`.

knkski:
Apologies for the delay in responding, @DomFleischmann will be taking this over

PR #1793, opened 2022-01-25
Bump Katib Python SDK to 0.13.0rc0 version

andreyvelich:
We should update our SDK version to the latest in the `master` branch.

/assign @gaocegege @tenzen-y @johnugeorge

### DIFF
--- a/sdk/python/v1beta1/setup.py
+++ b/sdk/python/v1beta1/setup.py
@@ -19,7 +19,7 @@ with open('requirements.txt') as f:
 
 setuptools.setup(
     name='kubeflow-katib',
-    version='0.12.0',
+    version='0.13.0rc0',
     author="Kubeflow Authors",
     author_email='premnath.vel@gmail.com',
     license="Apache License Version 2.0",


tenzen-y (commented):
Thank you for this update @andreyvelich !
I think it would be better to add Python 3.8 ~ 3.10 to the following field.
WDYT? @andreyvelich 

https://github.com/kubeflow/katib/blob/458eee70c5738c931ba9276ee30c22224ac955ee/sdk/python/v1beta1/setup.py#L38-L42

andreyvelich:
> Thank you for this update @andreyvelich ! I think it would be better to add Python 3.8 ~ 3.10 to the following field. WDYT? @andreyvelich
> 
> https://github.com/kubeflow/katib/blob/458eee70c5738c931ba9276ee30c22224ac955ee/sdk/python/v1beta1/setup.py#L38-L42

It makes sense.
Do we want to support Python 3.6 and 3.7 in our SDK ?

I think we can update it in the following PR.

tenzen-y:
> > Thank you for this update @andreyvelich ! I think it would be better to add Python 3.8 ~ 3.10 to the following field. WDYT? @andreyvelich
> > https://github.com/kubeflow/katib/blob/458eee70c5738c931ba9276ee30c22224ac955ee/sdk/python/v1beta1/setup.py#L38-L42
> 
> It makes sense. Do we want to support Python 3.6 and 3.7 in our SDK ?
> 
> I think we can update it in the following PR.

I think it needs to support Python 3.7 in our SDK since Python 3.7 will be kept until 2023-06-27 according to this [document](https://devguide.python.org/#status-of-python-branches).

PR #1796, opened 2022-01-26
Update mysql registry and tag to get security updates

juliusvonkohout:

Please use proper registry names and a proper tag to get security updates for mysql 8.

### DIFF
--- a/manifests/v1beta1/components/mysql/mysql.yaml
+++ b/manifests/v1beta1/components/mysql/mysql.yaml
@@ -21,7 +21,7 @@ spec:
     spec:
       containers:
         - name: katib-mysql
-          image: mysql:8.0.26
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
@andreyvelich i do not know whether is is a build problem again, but you really should use mysql with security fixes.

andreyvelich:
@juliusvonkohout Please check this PR: https://github.com/kubeflow/katib/pull/1594.
We want to use the specific mysql image to avoid errors like:
```
[InnoDB] Cannot boot server version 80017 on data directory built by version 80018. Downgrade is not supported
```
If `8.0.26` version has some security issues, we can update the patch version.

/cc @kubeflow/wg-automl-leads @munagekar

PR #1797, opened 2022-01-26
Update supported Python version for kubeflow-katib SDK

tenzen-y:
As discussed in https://github.com/kubeflow/katib/pull/1793#pullrequestreview-862307507, we should update the supported Python version for our SDK.

/assign @andreyvelich @gaocegege @johnugeorge


sdk/python/v1beta1/setup.py
```diff
@@ -38,8 +38,10 @@
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
```
andreyvelich, line 40: I think we can also remove these Python <= 3 versions, since they are not longer maintained.

andreyvelich (commented):
Thank you for this update @tenzen-y!
We can also cherry pick this change on `release-0.13` branch.

sdk/python/v1beta1/setup.py
tenzen-y: Done.

andreyvelich:
Thank you for the update @tenzen-y!
/lgtm
/approve

PR #1817, opened 2022-02-16
manifests: Change the image tags to special versions

### LINKED ISSUE #1816: Proposal: change the image tags in the manifests from latest to special versions
/kind feature

See here, https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml#L9-L18
I think we should change the image tags to special versions, such as `v0.13.rc.0` and update the tags when we are releasing.
It will break the users' clusters if they deploy their clusters with Argo CD and import the resource from the GitHub url.

---

haoxins:
Fixes #1816 

**Checklist:**


andreyvelich:
/hold
Please check my comment here: https://github.com/kubeflow/katib/issues/1816#issuecomment-1041535196

PR #1822, opened 2022-02-19
Fix the Dockerfile for API documentation generation

tenzen-y:
I fix the Dockerfile for API documentation generation since the base image of `pseudomuto/protoc-gen-doc` has been changed from `debian:jessie-slim` to `alpine:3.15.0` in v1.5.1.

Ref: https://github.com/pseudomuto/protoc-gen-doc/blob/v1.5.1/Dockerfile#L1

Fixes #

**Checklist:**

### DIFF
--- a/pkg/apis/manager/v1beta1/gen-doc/Dockerfile
+++ b/pkg/apis/manager/v1beta1/gen-doc/Dockerfile
@@ -1,17 +1,15 @@
 FROM pseudomuto/protoc-gen-doc
-RUN apt-get -q -y update && apt-get -q -y install curl && \
-        mkdir -p /protobuf/google/protobuf && \
-        for f in any duration descriptor empty struct timestamp wrappers; do \
-        curl -L -o /protobuf/google/protobuf/${f}.proto https://raw.githubusercontent.com/google/protobuf/master/src/google/protobuf/${f}.proto; \
-        done && \
-        mkdir -p /protobuf/google/api && \
-        for f in annotations http; do \
-        curl -L -o /protobuf/google/api/${f}.proto https://raw.githubusercontent.com/grpc-ecosystem/grpc-gateway/master/third_party/googleapis/google/api/${f}.proto; \
-        done && \
-        mkdir -p /protobuf/github.com/gogo/protobuf/gogoproto && \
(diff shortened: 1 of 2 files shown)


tenzen-y:
It seems that `protoc-gen-doc` uses `-` instead of `.` from v1.5.1.

Ref: https://github.com/pseudomuto/protoc-gen-doc/commit/13340ab057c1c18a8e5f7725138c28524b26c76c

pkg/apis/manager/v1beta1/gen-doc/Dockerfile
```diff
@@ -1,17 +1,15 @@
 FROM pseudomuto/protoc-gen-doc
```
anencore94, line 1: I checked to build this image with tags `1.5.1`, the expected error was printed :
image

Also, with tags `1.4.1`, such error was not printed as expected, since it is based on Debian as you mentioned.
image

What do you think of whether is it necessary to specify the exact tag here for possible future update on this image or not?

Refs: https://hub.docker.com/r/pseudomuto/protoc-gen-doc/tags

anencore94 (commented):
/lgtm

pkg/apis/manager/v1beta1/gen-doc/Dockerfile
tenzen-y: As you say, we have to specify the image tag in images used by end users, but our `protoc-gen-doc` image is only used by developers.

So, I think it might be better not to specify a tag for `pseudomuto/protoc-gen-doc` since specifying the exact tag increases the management cost.

What do you think @kubeflow/wg-automl-leads @anencore94 ?

anencore94: I agree with @tenzen-y :) The potential management const is unnecessary for this docker image. Since we always check the validation with CI process already.

johnugeorge: @anencore94  Do we have this validation in CI process ?

anencore94: We check from make check(which includes make generate) from go test

PR #1828, opened 2022-03-03
Chore: Uses Istio v1beta1 api

haoxins:
There is nothing different between `v1alpha3` and `v1beta1` currently IMO.
But it's better to use `v1beta1` for the Istio upgrade in the future.

### DIFF
--- a/manifests/v1beta1/installs/katib-with-kubeflow/ui-virtual-service.yaml
+++ b/manifests/v1beta1/installs/katib-with-kubeflow/ui-virtual-service.yaml
@@ -1,4 +1,4 @@
-apiVersion: networking.istio.io/v1alpha3
+apiVersion: networking.istio.io/v1beta1
 kind: VirtualService
 metadata:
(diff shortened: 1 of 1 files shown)


tenzen-y (commented):
Thank you for upgrading this @haoxins!

/ok-to-test

johnugeorge:
we should hold this PR for a while. 
The manifests folder is synced to upstream Kubeflow/manifests for every release. Hence, it is better to match Istio version for other components as well.  

https://github.com/kubeflow/manifests/tree/master/apps/katib/upstream

/hold

haoxins:
> we should hold this PR for a while. The manifests folder is synced to upstream Kubeflow/manifests for every release. Hence, it is better to match Istio version for other components as well.
> 
> https://github.com/kubeflow/manifests/tree/master/apps/katib/upstream
> 
> /hold

Agreed, I will work on upgrade the Istio manifests into `1.13+` at first.

PR #1833, opened 2022-03-08
Population based training

a9p:

Support the discovery of modulated hyperparameters rather than attempting to find a fixed set over the entire training process. The [paper](https://arxiv.org/abs/1711.09846) has more details about the technique.

This PR provides some initial support for PBT within Katib (#1382).

**Checklist:**


a9p:
Comments and open questions:
- The learned hyperparameter schedule is not accessible in the experiment object, only the last generation hyperparameters (as best trial).
- The learned hyperparameter schedule is not presented in the UI. A new view may be necessary; or if there is a way to increment the trial generation field, we could create a plot in the frontend based on that. Suggestions?
- Is it possible to/should the final learned model be added to the artifact store from the suggestion service on shutdown?

/cc @andreyvelich

Arhell (commented):
/ok-to-test

johnugeorge:
Please comment when PR is complete to be reviewed

pkg/suggestion/v1beta1/pbt/service.py
```diff
@@ -0,0 +1,392 @@
...
+    HyperParameterSearchSpace,
+)
+import pkg.suggestion.v1beta1.internal.constant as constant
+from pkg.suggestion.v1beta1.internal.trial import Trial, Assignment, Annotations
+from pkg.suggestion.v1beta1.internal.base_health_service import HealthServicer
+
+logger = logging.getLogger(__name__)
+
+
+_REQUIRED_SETTINGS = ["checkpoint_dir", "n_population", "truncation_threshold"]
```
andreyvelich, line 35: Since `checkpoint_dir` will be common across various algorithms, we can move it to [constant.py](https://github.com/kubeflow/katib/blob/280eaed8cc27e25241fdcd0517354ed5deb732d9/pkg/suggestion/v1beta1/internal/constant.py)

pkg/suggestion/v1beta1/pbt/service.py
```diff
@@ -0,0 +1,392 @@
...
+                )
+                request_count = len(self.job_queue)
+            elif len(self.job_queue.running) > 0:
+                logger.info(
+                    "Waiting for trials to complete before spawning next generation: {}".format(
+                        list(self.job_queue.running.keys())
+                    )
+                )
+                self.job_queue.verify_running(request.trials)
+                return api_pb2.GetSuggestionsReply(parameter_assignments=[])
```
andreyvelich, line 127: Won't we get an error here when [assignments != current request number](https://github.com/kubeflow/katib/blob/280eaed8cc27e25241fdcd0517354ed5deb732d9/pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go#L125-L128) ?
Or this use-case should not be possible ?

pkg/suggestion/v1beta1/pbt/service.py
```diff
@@ -0,0 +1,392 @@
...
+    def from_hyperparameter(hp: HyperParameter):
+        return HyperParameterSampler(hp.name, hp.type, hp.min, hp.max, hp.list, hp.step)
+
+
+class PbtJob(object):
+    def __init__(self, assignment_list, generation=0, parent=None, uid=None):
+        if uid is None:
+            self.uid = str(uuid.uuid4())
+        else:
+            self.uid = uid
```
andreyvelich, line 189: Are we using UUID here to define Trial name from the Suggestion ?
For some use-cases Suggestion has to know the generated Trial name, I created the issue before: https://github.com/kubeflow/katib/issues/610.
Do we want to move forward with this feature or we use UUID for something else @a9p ?
cc @gaocegege @johnugeorge

pkg/suggestion/v1beta1/pbt/service.py
```diff
@@ -0,0 +1,392 @@
...
+            api_pb2.TrialStatus.TrialConditionType.RUNNING,
+        ):
+            return
+        for i in range(self.generation + 1):
+            if uid in self.completed[i]:
+                return
+
+        if not uid in self.running:
+            logger.error("Unable to find trial match in PbtJobQueue.running!")
+            return
```
andreyvelich, line 311: Do we ever have this use-case ?

pkg/suggestion/v1beta1/pbt/service.py
```diff
@@ -0,0 +1,392 @@
...
+                api_pb2.TrialStatus.TrialConditionType.EARLYSTOPPED,
+            )
+            and Trial.convertTrial(trial) is None
+        ):
+            # Trial invalid, retry
+            logger.error(
+                "Unable to convert trial {} (status: {}), re-adding to task queue".format(
+                    trial.name, trial.status.condition
+                )
+            )
```
andreyvelich, line 330: I think you can avoid this use-case, since it's the Controller responsibility that all Succeeded and EarlyStopped Trials have reported metrics.
WDYT @a9p @johnugeorge @gaocegege ?

andreyvelich (commented):
Thank you for this great contribution @a9p, I left my first comments.

You can also create PR, similar to this: https://github.com/kubeflow/testing/pull/970.
We should add PBT Suggestion and Trial images to our CI.

/assign @gaocegege @johnugeorge @tenzen-y @anencore94

pkg/apis/controller/suggestions/v1beta1/suggestion_types.go
```diff
@@ -78,12 +78,15 @@ type TrialAssignment struct {
...
-	//Name of the suggestion
+	// Name of the suggestion
 	Name string `json:"name,omitempty"`
 
 	// Rules for early stopping techniques
 	// Contains rule name, value and comparison type
 	EarlyStoppingRules []common.EarlyStoppingRule `json:"earlyStoppingRules,omitempty"`
+
+	// Suggestion metadata to attach to Trial job
+	Annotations map[string]string `json:"annotations,omitempty"`
```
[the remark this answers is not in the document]
a9p: You are correct about uid and generation annotations being used by the suggestion service to track a spawned trial to completion! I'm not sure I follow the algorithmSettings solution - isn't that field static once the Suggestion is created?

pkg/webhook/v1beta1/pod/inject_webhook.go
```diff
@@ -316,3 +324,79 @@ func (s *SidecarInjector) getMetricsCollectorArgs(trial *trialsv1beta1.Trial, me
...
+	}
+	suggestion := &suggestionsv1beta1.Suggestion{}
+	if err := s.client.Get(context.TODO(), apitypes.NamespacedName{Name: experimentName, Namespace: trial.Namespace}, suggestion); err != nil {
+		return err
+	}
+
+	// Check if mutation is needed
+	checkpointPath := ""
+	for _, s := range suggestion.Spec.Algorithm.AlgorithmSettings {
+		if s.Name == "checkpoint_dir" && s.Value != "" {
```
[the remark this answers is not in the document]
a9p: I'll run with `suggestion_trial_dir` for now, but happy to swap if the team has a differing preference.

pkg/suggestion/v1beta1/pbt/service.py
a9p: It will error, but the client keeps retrying with increasing delay. This in the worst case will cause a job to pause for an extended duration while the retry timeout runs down. I experimented with changing `!=` to `>` in an earlier iteration, but wasn't sure if that would break other suggestion service assumptions (it works fine for a few E2E tests I manually went through, but breaks many unit tests; a few I wasn't able to easily track down).

pkg/suggestion/v1beta1/pbt/service.py
a9p: Correct, the uid is used to track the Trial from spawn to result. `I think, we should provide functionality, when user can specify special name to create Trial.` in that referenced issue would be necessary for PBT to be able to use that mechanism instead of the uid annotation.

pkg/webhook/v1beta1/pod/inject_webhook.go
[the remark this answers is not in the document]
a9p: Were you thinking:
```golang
// SuggestionVolumeMountKey specifies the AlgorithmSettings key used to toggle Suggestion managed trial storage
SuggestionVolumeMountKey = "suggestion_trial_dir"
```
and then:
```golang
if s.Name == consts.SuggestionVolumeMountKey && s.Value != "" { /*...*/ }
```

examples/v1beta1/trial-images/simple-pbt/pbt_test.py
```diff
@@ -0,0 +1,133 @@
...
+            self._accuracy -= (q_err - q_tolerance) * random.random()
+        self._accuracy += noise_level * np.random.normal()
+        self._accuracy = max(0, self._accuracy)
+
+        if self._step == 1 or self._step % self._log_interval == 0:
+            self.save_checkpoint()
+            if not self._writer:
+                self._writer = tf.summary.create_file_writer(self._log_dir)
+            with self._writer.as_default():
+                tf.summary.scalar(
```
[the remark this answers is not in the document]
a9p: Removed -- moved to using the pickle checkpoint.

pkg/new-ui/v1beta1/frontend/src/app/constants/algorithms-settings.const.ts
```diff
@@ -242,6 +242,29 @@ export const DartsSettings: AlgorithmSetting[] = [
   },
 ];
 
+export const PbtSettings: AlgorithmSetting[] = [
```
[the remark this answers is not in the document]
a9p: Separated UI changes are in #1862. It's currently branching from master, but will require the changes introduced in this PR to test E2E.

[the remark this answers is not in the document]
andreyvelich: Sounds good, thank you!

pkg/webhook/v1beta1/pod/inject_webhook.go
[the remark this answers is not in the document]
andreyvelich: Sounds good. @gaocegege @johnugeorge @tenzen-y What do you think about this Algorithm Settings name ?

pkg/suggestion/v1beta1/pbt/service.py
andreyvelich: I see, thank you for explanation.
Are we sure that Suggestion controller will reconcile it again ?
As I can see, when [`ReconcileSuggestion`](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L166-L177) failed, we don't requeue controller.
@gaocegege @johnugeorge @tenzen-y What do you think about this approach ?

Do we want to support "synchronous" algorithm service ? `GetSuggestions` (or any other API) is called only when all Trials have been completed in one generation ?

pkg/suggestion/v1beta1/pbt/service.py
andreyvelich: Let's see what others think about this feature.
cc @tenzen-y

pkg/apis/controller/suggestions/v1beta1/suggestion_types.go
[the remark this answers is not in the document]
tenzen-y: That makes sense @andreyvelich. 
In this use-case, it might be better to store it in annotions.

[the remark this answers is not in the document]
a9p: Thanks for the notes, it's a bit clearer now! I see how we could transmit the uid/gen via AlgorithmSettings, but once the experiment is updated, that information is not stored in a [Trial](https://github.com/kubeflow/katib/blob/280eaed8cc27e25241fdcd0517354ed5deb732d9/pkg/apis/manager/v1beta1/api.proto#L209). So if we have the `trial-1-uid: <Suggestion defined uid>` map in the Experiment, how might we tie that to the backend Trial uid that gets defined on spawn so we can link the two in the Suggestion service?

pkg/suggestion/v1beta1/pbt/service.py
tenzen-y: I am not sure of use-cases that would specify a special name for Trial.
So, I think it would be great to add uid annotation.
WDYT @andreyvelich @a9p ?

pkg/webhook/v1beta1/pod/inject_webhook.go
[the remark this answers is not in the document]
tenzen-y: SGTM.

pkg/apis/controller/suggestions/v1beta1/suggestion_types.go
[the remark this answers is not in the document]
andreyvelich: For example, since BO algorithm doesn't generate Trial name, we have to search for HyperParameters based on Trial assignments: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/skopt/base_service.py#L84-L98.

If Suggestion can generate Trial names, we can store these names in `recorded_trials_names` once SKOpt service [suggests new HPs](https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/skopt/base_service.py#L121).
Once `GetSuggestions` called again, we can easily update loss value in `recorded_trials_names` based on Trial name.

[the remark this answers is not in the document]
andreyvelich: I guess, if your Suggestion can define Trial name, you can link Trials and `generation`, `parent` data from `AlgorithmSettings`, because you are using Trial annotations only in PBT service: https://github.com/kubeflow/katib/blob/ff3da32328d1c119d98236943bfb79b07ead5207/pkg/suggestion/v1beta1/pbt/service.py#L295

But it looks tricky and it ties to my original question: https://github.com/kubeflow/katib/pull/1833#discussion_r865033485.
Do we want to set Trial name from Suggestion service ?

pkg/suggestion/v1beta1/pbt/service.py
andreyvelich: For example, since BO algorithm doesn't generate Trial name, we have to search for HyperParameters based on Trial assignments: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/skopt/base_service.py#L84-L98.

If Suggestion can generate Trial names, we can store these names in `recorded_trials_names` once SKOpt service [suggests new HPs](https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/skopt/base_service.py#L121).
Once `GetSuggestions` called again, we can easily update loss value in `recorded_trials_names` based on Trial name.

pkg/apis/controller/suggestions/v1beta1/suggestion_types.go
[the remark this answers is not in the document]
a9p: I agree that setting the name from the Suggestion service would be a cleaner flow to ensure we can track the Trial. Assuming we go with that, what is the preference for passing other metadata we want tied to the Trial (e.g. generation)? Is the Annotation field okay or should I change that to Label?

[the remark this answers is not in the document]
andreyvelich: I prefer `Labels`, since [Kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects) say that Lables are used to select objects. Also, I think that Labels should go to metadata so we can use [Kubernetes Label Selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) to select Trials for UI, etc.

Let's see what others say.
@gaocegege @johnugeorge @tenzen-y ?

pkg/suggestion/v1beta1/pbt/service.py
a9p: My observation was that suggestions are retrieved whenever running jobs < parallel trials. I changed the implementation to async so we are no longer relying on the controller to keep polling until it gets a response of the appropriate size.

pkg/suggestion/v1beta1/pbt/service.py
a9p: Moved to specifying trial name in the Reply rather than the Annotation per your linked issue.

tenzen-y: Thanks for clarifying, @andreyvelich!
That feature will help us.

pkg/apis/controller/suggestions/v1beta1/suggestion_types.go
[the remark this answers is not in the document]
tenzen-y: I have a question @andreyvelich @a9p.

Are these trial settings (`Labels` or `Annotations`) only referenced from Katib components (e.g., Katib UI, Suggestion Service)? Or are they expected end-users to use (e.g., select object)?

If used only in Katib components, I think it would be better `Annotations`.

Since the labels are propagated to the trial pod, it might reduce the visibility of `kubectl` results (e.g. `kubectl get pods -n kubeflow --show-labels`) to add many labels not to be used by end-users.

pkg/webhook/v1beta1/pod/utils.go
```diff
@@ -241,12 +254,15 @@ func mutateVolume(pod *v1.Pod, mountPath, sidecarContainerName, primaryContainer
 	}
 
 	for _, i := range indexList {
-		c := &pod.Spec.Containers[i]
-		if c.VolumeMounts == nil {
-			c.VolumeMounts = make([]v1.VolumeMount, 0)
+		if err := addContainerVolumeMount(&pod.Spec.Containers[i], &vm); err != nil {
+			return err
```
tenzen-y, line 258: I think `addContainerVolumeMount()` always returns `nil`. Do we need to check errors?

(Same with `pkg/webhook/v1beta1/pod/inject_webhook.go`)

pkg/webhook/v1beta1/pod/inject_webhook.go
```diff
@@ -316,3 +324,64 @@ func (s *SidecarInjector) getMetricsCollectorArgs(trial *trialsv1beta1.Trial, me
...
+	}
+
+	vm := v1.VolumeMount{
+		Name:      suggestionVolume.Name,
+		MountPath: checkpointPath,
+		SubPath:   checkpointFolder,
+	}
+
+	primaryContainerIndex := getPrimaryContainerIndex(pod.Spec.Containers, trial.Spec.PrimaryContainerName)
+	if err := addContainerVolumeMount(&pod.Spec.Containers[primaryContainerIndex], &vm); err != nil {
```
tenzen-y, line 374: I think addContainerVolumeMount() always returns nil. Do we need to check errors?

pkg/webhook/v1beta1/pod/inject_webhook.go
```diff
@@ -316,3 +324,64 @@ func (s *SidecarInjector) getMetricsCollectorArgs(trial *trialsv1beta1.Trial, me
...
+		return err
+	}
+	// c := &pod.Spec.Containers[primaryContainerIndex]
+	// if c.VolumeMounts == nil {
+	// 	c.VolumeMounts = make([]v1.VolumeMount, 0)
+	// }
+	// c.VolumeMounts = append(c.VolumeMounts, vm)
+	// pod.Spec.Containers[primaryContainerIndex] = *c
```
tenzen-y, line 382: Do we need these comments?
cc: @kubeflow/wg-automl-leads

pkg/webhook/v1beta1/pod/utils.go
```diff
@@ -241,12 +254,15 @@ func mutateVolume(pod *v1.Pod, mountPath, sidecarContainerName, primaryContainer
...
 		}
-		c.VolumeMounts = append(c.VolumeMounts, vm)
-		pod.Spec.Containers[i] = *c
+		// c := &pod.Spec.Containers[i]
+		// if c.VolumeMounts == nil {
+		// 	c.VolumeMounts = make([]v1.VolumeMount, 0)
+		// }
+		// c.VolumeMounts = append(c.VolumeMounts, vm)
+		// pod.Spec.Containers[i] = *c
```
tenzen-y, line 265: Do we need these comments?

(Same comment with pkg/webhook/v1beta1/pod/inject_webhook.go)

examples/v1beta1/hp-tuning/pbt.yaml
```diff
@@ -0,0 +1,60 @@
...
+        description: Learning rate for training the model
+        reference: lr
+    trialSpec:
+      apiVersion: batch/v1
+      kind: Job
+      spec:
+        template:
+          # metadata:
+          #   annotations:
+          #     sidecar.istio.io/inject: "false"
```
tenzen-y, line 48: Do we need these comments?

tenzen-y (commented):
Thanks for this great work, @a9p! Also, sorry for the late response.
I left a few comments.

@kubeflow/wg-automl-leads Could you approve workflows in Github Action?

pkg/apis/controller/suggestions/v1beta1/suggestion_types.go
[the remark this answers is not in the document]
a9p: The annotations are now only used to tie information to the Trial for the UI. I wasn't imagining users parsing through those fields for PBT at least.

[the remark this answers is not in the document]
andreyvelich: What if users want to compare all PBT Trials with `generation` 2 ? What do others prefer (`Labels` or `Annotations`) @gaocegege @johnugeorge ?
Do we have any examples from other CRDs when they store object metadata from various services ?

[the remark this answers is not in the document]
johnugeorge: As per docs, annotations contain non identifiable info which doesn't need selection or watch.  As @andreyvelich pointed out, won't they be used for selection/filter ?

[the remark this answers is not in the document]
tenzen-y: It might be better to use `Labels` in preference to Kubernetes docs rather than the visibility of kubectl results.
Sorry if I have confused you.

andreyvelich:
@a9p Do you have some time to check our other comments ? We really want to include this feature in our next release (Katib 0.14).
Also, it would be great if we could add this Experiment in our CI/CD.
I commented here: https://github.com/kubeflow/katib/pull/1833#pullrequestreview-961885456 how we can add Trial and Suggestion images to the AWS CI.

Thank you for your time!

tenzen-y:
Could you rebase this branch? @a9p 
We have fixed CI in https://github.com/kubeflow/katib/pull/1867 and https://github.com/kubeflow/katib/pull/1866.

tenzen-y:
@a9p Can you add [examples/v1beta1/trial-images/simple-pbt/Dockerfile](https://github.com/kubeflow/katib/blob/fe1b8e46bf836f154dafe1d5f0bb1a39285119bc/examples/v1beta1/trial-images/simple-pbt/Dockerfile) to [this section](https://github.com/kubeflow/katib/blob/fe1b8e46bf836f154dafe1d5f0bb1a39285119bc/test/e2e/v1beta1/argo_workflow.py#L43-L66) and [examples/v1beta1/hp-tuning/pbt.yaml](https://github.com/kubeflow/katib/blob/fe1b8e46bf836f154dafe1d5f0bb1a39285119bc/examples/v1beta1/hp-tuning/pbt.yaml) to [this section](https://github.com/kubeflow/katib/blob/fe1b8e46bf836f154dafe1d5f0bb1a39285119bc/test/e2e/v1beta1/argo_workflow.py#L70-L88)?

cmd/suggestion/pbt/v1beta1/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM python:3.9
```
tenzen-y, line 1: Can you add this Dockerfile to the [publish-algorithm-images.yaml](https://github.com/kubeflow/katib/blob/10f674a15512a2f1f2ebe52cab235d8702a2e1e4/.github/workflows/publish-algorithm-images.yaml)?

examples/v1beta1/trial-images/simple-pbt/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.9
```
tenzen-y, line 1: Can you add this Dockerfike to the [publish-trial-images.yaml](https://github.com/kubeflow/katib/blob/10f674a15512a2f1f2ebe52cab235d8702a2e1e4/.github/workflows/publish-trial-images.yaml)?

test/e2e/v1beta1/argo_workflow.py
```diff
@@ -63,6 +63,7 @@
     "trial-enas-cnn-cifar10-gpu":    "examples/v1beta1/trial-images/enas-cnn-cifar10/Dockerfile.gpu",
     "trial-enas-cnn-cifar10-cpu":    "examples/v1beta1/trial-images/enas-cnn-cifar10/Dockerfile.cpu",
     "trial-darts-cnn-cifar10":       "examples/v1beta1/trial-images/darts-cnn-cifar10/Dockerfile",
+    "trial-simple-pbt":              "examples/v1beta1/trial-images/simple-pbt/Dockerfile",
```
tenzen-y, line 66: Can you also add `suggestion-pbt` to `KATIB_IMAGES`?
Thanks for your contribution.

a9p: Sorry about that messy commit; added the `suggestion-pbt`!

tenzen-y:
@johnugeorge We must merge https://github.com/kubeflow/testing/pull/1002 before this PR is merged.

examples/v1beta1/hp-tuning/pbt.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: kubeflow.org/v1beta1
+kind: Experiment
+metadata:
+  name: random-experiment
+  namespace: ''
```
andreyvelich, line 5: Let's keep kubeflow namespace, to be consistent with other examples.
```suggestion
  namespace: kubeflow
```

examples/v1beta1/hp-tuning/pbt.yaml
```diff
@@ -0,0 +1,57 @@
...
+spec:
+  maxTrialCount: null
+  parallelTrialCount: 12
+  maxFailedTrialCount: 3
+  resumePolicy: FromVolume
+  objective:
+    type: maximize
+    goal: 0.99
+    objectiveMetricName: Validation-accuracy
+    additionalMetricNames: []
```
andreyvelich, line 15: You can remove it.
```suggestion
```

pkg/apis/manager/v1beta1/api.proto
```diff
@@ -302,10 +304,19 @@ message GetSuggestionsRequest {
...
+    message LabelAssignments {
+        map<string, string> labels = 1;
     }
+
     repeated ParameterAssignments parameter_assignments = 1;
     AlgorithmSpec algorithm = 2;
     repeated EarlyStoppingRule early_stopping_rules = 3;
+
+    // Optional field to add labels to the generated Trials
+    repeated LabelAssignments label_assignments = 5;
```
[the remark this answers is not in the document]
tenzen-y: SGTM

examples/v1beta1/hp-tuning/pbt.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: kubeflow.org/v1beta1
+kind: Experiment
+metadata:
+  name: random-experiment
+  namespace: ''
+spec:
+  maxTrialCount: null
+  parallelTrialCount: 12
+  maxFailedTrialCount: 3
+  resumePolicy: FromVolume
```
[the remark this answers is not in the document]
a9p: Yes, we require persistent storage to seed later generations.

johnugeorge:
@a9p Can you please update with all review comments so that we can get this in for the next Katib release? 

Thanks

examples/v1beta1/hp-tuning/pbt.yaml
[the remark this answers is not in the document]
andreyvelich: I see, that is why we reconcile Volume only if
[`resumePolicy` is `FromVolume`](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L195-L207).
I believe, that is not true for this use-case.
In this case, we want to create Suggestion volume for algorithm requirements regardless Experiment `resumePolicy`.

I think, we should give user ability to attach Suggestion volume without setting `resumePolicy`.
But if `resumePolicy` is `FromVolume`, Suggestion volume should be always reconciled.

What do you think @tenzen-y @johnugeorge ?

@a9p Please can you create a separate issue to track it.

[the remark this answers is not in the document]
tenzen-y: SGTM
I'd like to include this feature in the next release, so let's merge this PR leaving the `resumPolicy` field!

pkg/apis/manager/v1beta1/api.proto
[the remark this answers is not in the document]
gaocegege: SGTM

It is a breaking change, maybe we should document it in the release note.

tenzen-y:
@a9p 
It seems that this branch has a merge conflict. Could you rebase?
Thanks for your contribution!

examples/v1beta1/hp-tuning/pbt.yaml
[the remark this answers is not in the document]
a9p: New issue linked here: https://github.com/kubeflow/katib/issues/1893

a9p:
@johnugeorge I think all the action items from the review feedback are addressed now.

johnugeorge:
Thanks @a9p for this

As  we have limited resources per GHA job, can you also optimise Dockerfiles and examples as in  https://github.com/kubeflow/katib/pull/1882 ?  Can you also add it to GHA workflow so that it is tested during CI run

/cc @tenzen-y

.github/workflows/simple-pbt-e2e-test.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Detail: https://hub.docker.com/r/kindest/node
+        # TODO (tenzen-y): We need to consider running tests on more kubernetes versions.
+        # kubernetes-version: ["v1.20.15", "v1.21.12", "v1.22.9", "v1.23.6", "v1.24.1"]
+        kubernetes-version: ["v1.21.12", "v1.22.9", "v1.23.6"]
+        # Comma Delimited
+        experiments: ["simple-pbt"]
```
tenzen-y, line 33: We need to sync `experiments` name with the manifest name.
For example, If this experiments name is "simple-pbt", we need to modify [this](https://github.com/kubeflow/katib/blob/806354cfcf1cf169f00b575db67c867ed7dff3f9/examples/v1beta1/hp-tuning/pbt.yaml) manifest name from pbt.yaml to simple-pbt.yaml.

tenzen-y:
@a9p 
Can you run `make check` on your local machine to resolve [this](https://github.com/kubeflow/katib/runs/6826113088?check_suite_focus=true#step:5:1174) lint error?

a9p:
@tenzen-y I am not sure why that error is showing up in the build -- I don't see any unstaged changes locally when I run `make check` (even on a clean clone). `gofmt -l .` also doesn't  show any issues.

tenzen-y:
@a9p Thanks for this update!

>  I am not sure why that error is showing up in the build -- I don't see any unstaged changes locally when I run make check (even on a clean clone). gofmt -l . also doesn't show any issues. (ubuntu 20.04, go 1.17.10)

I will try to run `make check` on my local using your branch to confirm this error.

tenzen-y:
@a9p 
I see some changes when I ran `make check` using your branch, [`pbt`](https://github.com/a9p/katib/tree/4e672019711983c3772cbed869e12dfd6441b084) on my local (macOS v12.4, Go 1.17.10).

```shell
$ git diff --name-only 
pkg/apis/controller/suggestions/v1beta1/zz_generated.deepcopy.go
pkg/apis/controller/trials/v1beta1/zz_generated.deepcopy.go
pkg/apis/v1beta1/openapi_generated.go
pkg/apis/v1beta1/swagger.json
sdk/python/v1beta1/docs/V1beta1TrialAssignment.md
sdk/python/v1beta1/docs/V1beta1TrialSpec.md
sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_assignment.py
sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_spec.py
```

<details>
<summary>full git diff</summary>

```shell
$ git diff
diff --git a/pkg/apis/controller/suggestions/v1beta1/zz_generated.deepcopy.go b/pkg/apis/controller/suggestions/v1beta1/zz_generated.deepcopy.go
index 5bfa26b7..d2f37c0d 100644
--- a/pkg/apis/controller/suggestions/v1beta1/zz_generated.deepcopy.go
+++ b/pkg/apis/controller/suggestions/v1beta1/zz_generated.deepcopy.go
@@ -191,6 +191,13 @@ func (in *TrialAssignment) DeepCopyInto(out *TrialAssignment) {
                *out = make([]commonv1beta1.EarlyStoppingRule, len(*in))
                copy(*out, *in)
        }
+       if in.Labels != nil {
+               in, out := &in.Labels, &out.Labels
+               *out = make(map[string]string, len(*in))
+               for key, val := range *in {
+                       (*out)[key] = val
+               }
+       }
        return
 }
 
diff --git a/pkg/apis/controller/trials/v1beta1/zz_generated.deepcopy.go b/pkg/apis/controller/trials/v1beta1/zz_generated.deepcopy.go
index 8537398d..04010a68 100644
--- a/pkg/apis/controller/trials/v1beta1/zz_generated.deepcopy.go
+++ b/pkg/apis/controller/trials/v1beta1/zz_generated.deepcopy.go
@@ -135,6 +135,13 @@ func (in *TrialSpec) DeepCopyInto(out *TrialSpec) {
                        (*out)[key] = val
                }
        }
+       if in.Labels != nil {
+               in, out := &in.Labels, &out.Labels
+               *out = make(map[string]string, len(*in))
+               for key, val := range *in {
+                       (*out)[key] = val
+               }
+       }
        return
 }
 
diff --git a/pkg/apis/v1beta1/openapi_generated.go b/pkg/apis/v1beta1/openapi_generated.go
index 8eb788fe..0fe9de85 100644
--- a/pkg/apis/v1beta1/openapi_generated.go
+++ b/pkg/apis/v1beta1/openapi_generated.go
@@ -1755,6 +1755,22 @@ func schema_apis_controller_suggestions_v1beta1_TrialAssignment(ref common.Refer
                                                        },
                                                },
                                        },
+                                       "labels": {
+                                               SchemaProps: spec.SchemaProps{
+                                                       Description: "Suggestion label metadata to attach to Trial job",
+                                                       Type:        []string{"object"},
+                                                       AdditionalProperties: &spec.SchemaOrBool{
+                                                               Allows: true,
+                                                               Schema: &spec.Schema{
+                                                                       SchemaProps: spec.SchemaProps{
+                                                                               Default: "",
+                                                                               Type:    []string{"string"},
+                                                                               Format:  "",
+                                                                       },
+                                                               },
+                                                       },
+                                               },
+                                       },
                                },
                        },
                },
@@ -2017,6 +2033,22 @@ func schema_apis_controller_trials_v1beta1_TrialSpec(ref common.ReferenceCallbac
                                                        Format:      "",
                                                },
                                        },
+                                       "labels": {
+                                               SchemaProps: spec.SchemaProps{
+                                                       Description: "Labels that provide additional metadata for services (e.g. Suggestions tracking)",
+                                                       Type:        []string{"object"},
+                                                       AdditionalProperties: &spec.SchemaOrBool{
+                                                               Allows: true,
+                                                               Schema: &spec.Schema{
+                                                                       SchemaProps: spec.SchemaProps{
+                                                                               Default: "",
+                                                                               Type:    []string{"string"},
+                                                                               Format:  "",
+                                                                       },
+                                                               },
+                                                       },
+                                               },
+                                       },
                                },
                        },
                },
diff --git a/pkg/apis/v1beta1/swagger.json b/pkg/apis/v1beta1/swagger.json
index 82633465..b5a92761 100644
--- a/pkg/apis/v1beta1/swagger.json
+++ b/pkg/apis/v1beta1/swagger.json
@@ -207,6 +207,14 @@
             "$ref": "#/definitions/v1beta1.EarlyStoppingRule"
           }
         },
+        "labels": {
+          "description": "Suggestion label metadata to attach to Trial job",
+          "type": "object",
+          "additionalProperties": {
+            "type": "string",
+            "default": ""
+          }
+        },
         "name": {
           "description": "Name of the suggestion",
           "type": "string"
@@ -303,6 +311,14 @@
           "description": "Condition when trial custom resource is failed. Condition must be in GJSON format, ref https://github.com/tidwall/gjson. For example for BatchJob: status.conditions.#(type==\"Failed\")#|#(status==\"True\")#",
           "type": "string"
         },
+        "labels": {
+          "description": "Labels that provide additional metadata for services (e.g. Suggestions tracking)",
+          "type": "object",
+          "additionalProperties": {
+            "type": "string",
+            "default": ""
+          }
+        },
         "metricsCollector": {
           "description": "Describes how metrics will be collected",
           "default": {},
diff --git a/sdk/python/v1beta1/docs/V1beta1TrialAssignment.md b/sdk/python/v1beta1/docs/V1beta1TrialAssignment.md
index 1f7f2e4a..1e063a82 100644
--- a/sdk/python/v1beta1/docs/V1beta1TrialAssignment.md
+++ b/sdk/python/v1beta1/docs/V1beta1TrialAssignment.md
@@ -5,6 +5,7 @@ TrialAssignment is the assignment for one trial.
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
 **early_stopping_rules** | [**list[V1beta1EarlyStoppingRule]**](V1beta1EarlyStoppingRule.md) | Rules for early stopping techniques Contains rule name, value and comparison type | [optional] 
+**labels** | **dict(str, str)** | Suggestion label metadata to attach to Trial job | [optional] 
 **name** | **str** | Name of the suggestion | [optional] 
 **parameter_assignments** | [**list[V1beta1ParameterAssignment]**](V1beta1ParameterAssignment.md) | Suggestion results with Trial parameters | [optional] 
 
diff --git a/sdk/python/v1beta1/docs/V1beta1TrialSpec.md b/sdk/python/v1beta1/docs/V1beta1TrialSpec.md
index e93ca36b..580e4cfb 100644
--- a/sdk/python/v1beta1/docs/V1beta1TrialSpec.md
+++ b/sdk/python/v1beta1/docs/V1beta1TrialSpec.md
@@ -6,6 +6,7 @@ Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
 **early_stopping_rules** | [**list[V1beta1EarlyStoppingRule]**](V1beta1EarlyStoppingRule.md) | Rules for early stopping techniques. Each rule should be met to early stop Trial. | [optional] 
 **failure_condition** | **str** | Condition when trial custom resource is failed. Condition must be in GJSON format, ref https://github.com/tidwall/gjson. For example for BatchJob: status.conditions.#(type&#x3D;&#x3D;\&quot;Failed\&quot;)#|#(status&#x3D;&#x3D;\&quot;True\&quot;)# | [optional] 
+**labels** | **dict(str, str)** | Labels that provide additional metadata for services (e.g. Suggestions tracking) | [optional] 
 **metrics_collector** | [**V1beta1MetricsCollectorSpec**](V1beta1MetricsCollectorSpec.md) |  | [optional] 
 **objective** | [**V1beta1ObjectiveSpec**](V1beta1ObjectiveSpec.md) |  | [optional] 
 **parameter_assignments** | [**list[V1beta1ParameterAssignment]**](V1beta1ParameterAssignment.md) | Key-value pairs for hyperparameters and assignment values. | [optional] 
diff --git a/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_assignment.py b/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_assignment.py
index be3d7539..a106534e 100644
--- a/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_assignment.py
+++ b/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_assignment.py
@@ -34,29 +34,34 @@ class V1beta1TrialAssignment(object):
     """
     openapi_types = {
         'early_stopping_rules': 'list[V1beta1EarlyStoppingRule]',
+        'labels': 'dict(str, str)',
         'name': 'str',
         'parameter_assignments': 'list[V1beta1ParameterAssignment]'
     }
 
     attribute_map = {
         'early_stopping_rules': 'earlyStoppingRules',
+        'labels': 'labels',
         'name': 'name',
         'parameter_assignments': 'parameterAssignments'
     }
 
-    def __init__(self, early_stopping_rules=None, name=None, parameter_assignments=None, local_vars_configuration=None):  # noqa: E501
+    def __init__(self, early_stopping_rules=None, labels=None, name=None, parameter_assignments=None, local_vars_configuration=None):  # noqa: E501
         """V1beta1TrialAssignment - a model defined in OpenAPI"""  # noqa: E501
         if local_vars_configuration is None:
             local_vars_configuration = Configuration()
         self.local_vars_configuration = local_vars_configuration
 
         self._early_stopping_rules = None
+        self._labels = None
         self._name = None
         self._parameter_assignments = None
         self.discriminator = None
 
         if early_stopping_rules is not None:
             self.early_stopping_rules = early_stopping_rules
+        if labels is not None:
+            self.labels = labels
         if name is not None:
             self.name = name
         if parameter_assignments is not None:
@@ -85,6 +90,29 @@ class V1beta1TrialAssignment(object):
 
         self._early_stopping_rules = early_stopping_rules
 
+    @property
+    def labels(self):
+        """Gets the labels of this V1beta1TrialAssignment.  # noqa: E501
+
+        Suggestion label metadata to attach to Trial job  # noqa: E501
+
+        :return: The labels of this V1beta1TrialAssignment.  # noqa: E501
+        :rtype: dict(str, str)
+        """
+        return self._labels
+
+    @labels.setter
+    def labels(self, labels):
+        """Sets the labels of this V1beta1TrialAssignment.
+
+        Suggestion label metadata to attach to Trial job  # noqa: E501
+
+        :param labels: The labels of this V1beta1TrialAssignment.  # noqa: E501
+        :type: dict(str, str)
+        """
+
+        self._labels = labels
+
     @property
     def name(self):
         """Gets the name of this V1beta1TrialAssignment.  # noqa: E501
diff --git a/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_spec.py b/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_spec.py
index 71dd5296..ba9e6b8f 100644
--- a/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_spec.py
+++ b/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_trial_spec.py
@@ -35,6 +35,7 @@ class V1beta1TrialSpec(object):
     openapi_types = {
         'early_stopping_rules': 'list[V1beta1EarlyStoppingRule]',
         'failure_condition': 'str',
+        'labels': 'dict(str, str)',
         'metrics_collector': 'V1beta1MetricsCollectorSpec',
         'objective': 'V1beta1ObjectiveSpec',
         'parameter_assignments': 'list[V1beta1ParameterAssignment]',
@@ -48,6 +49,7 @@ class V1beta1TrialSpec(object):
     attribute_map = {
         'early_stopping_rules': 'earlyStoppingRules',
         'failure_condition': 'failureCondition',
+        'labels': 'labels',
         'metrics_collector': 'metricsCollector',
         'objective': 'objective',
         'parameter_assignments': 'parameterAssignments',
@@ -58,7 +60,7 @@ class V1beta1TrialSpec(object):
         'success_condition': 'successCondition'
     }
 
-    def __init__(self, early_stopping_rules=None, failure_condition=None, metrics_collector=None, objective=None, parameter_assignments=None, primary_container_name=None, primary_pod_labels=None, retain_run=None, run_spec=None, success_condition=None, local_vars_configuration=None):  # noqa: E501
+    def __init__(self, early_stopping_rules=None, failure_condition=None, labels=None, metrics_collector=None, objective=None, parameter_assignments=None, primary_container_name=None, primary_pod_labels=None, retain_run=None, run_spec=None, success_condition=None, local_vars_configuration=None):  # noqa: E501
         """V1beta1TrialSpec - a model defined in OpenAPI"""  # noqa: E501
         if local_vars_configuration is None:
             local_vars_configuration = Configuration()
@@ -66,6 +68,7 @@ class V1beta1TrialSpec(object):
 
         self._early_stopping_rules = None
         self._failure_condition = None
+        self._labels = None
         self._metrics_collector = None
         self._objective = None
         self._parameter_assignments = None
@@ -80,6 +83,8 @@ class V1beta1TrialSpec(object):
             self.early_stopping_rules = early_stopping_rules
         if failure_condition is not None:
             self.failure_condition = failure_condition
+        if labels is not None:
+            self.labels = labels
         if metrics_collector is not None:
             self.metrics_collector = metrics_collector
         if objective is not None:
@@ -143,6 +148,29 @@ class V1beta1TrialSpec(object):
 
         self._failure_condition = failure_condition
 
+    @property
+    def labels(self):
+        """Gets the labels of this V1beta1TrialSpec.  # noqa: E501
+
+        Labels that provide additional metadata for services (e.g. Suggestions tracking)  # noqa: E501
+
+        :return: The labels of this V1beta1TrialSpec.  # noqa: E501
+        :rtype: dict(str, str)
+        """
+        return self._labels
+
+    @labels.setter
+    def labels(self, labels):
+        """Sets the labels of this V1beta1TrialSpec.
+
+        Labels that provide additional metadata for services (e.g. Suggestions tracking)  # noqa: E501
+
+        :param labels: The labels of this V1beta1TrialSpec.  # noqa: E501
+        :type: dict(str, str)
+        """
+
+        self._labels = labels
+
     @property
     def metrics_collector(self):
         """Gets the metrics_collector of this V1beta1TrialSpec.  # noqa: E501
```

</details>

Can you show me logs from `make check` ?
Also, Can you check the required tools in [this developer guide](https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md#requirements)?

a9p:
@tenzen-y I had a bit more time to go glance through the logs and it seems like some new dependencies were missing (I needed to run `go mod download` manually to fix). Not sure why this wasn't causing an error originally, but seems like the file diff is now consistent with your last message!

tenzen-y:
> @tenzen-y I had a bit more time to go glance through the logs and it seems like some new dependencies were missing (I needed to run `go mod download` manually to fix). Not sure why this wasn't causing an error originally, but seems like the file diff is now consistent with your last message!

Thanks for updating this! @a9p 

@kubeflow/wg-automl-leads Can you start GH-Actions?

a9p:
Before kicking it off, there seems to be one more issue in the python tests that I'm tracking down right now!

johnugeorge:
@a9p PBT tests are failing

tenzen-y:
> @a9p PBT tests are failing

@a9p @johnugeorge 
Since we are overriding maxTrialCount and ParallelTrialCount in [this script](https://github.com/kubeflow/katib/blob/6a21af058ec8d02968538a0eb9a5dc2b84f83348/test/e2e/v1beta1/hack/gh-actions/run-e2e-experiment.go), maybe we need to specify maxTrialCount and ParallelTrialCount like the below.
https://github.com/kubeflow/katib/blob/6a21af058ec8d02968538a0eb9a5dc2b84f83348/test/e2e/v1beta1/hack/gh-actions/run-e2e-experiment.go#L70-L74

https://github.com/kubeflow/katib/runs/6854321927?check_suite_focus=true#step:4:9959

johnugeorge:
Waiting for Experiment simple-pbt to finish
    [10255](https://github.com/kubeflow/katib/runs/6892743686?check_suite_focus=true#step:4:10255)

    2022/06/15 04:34:41 Experiment is running: 0 Trials, 0 Pending Trials, 0 Running Trials, 0 Succeeded Trials, 0 Failed Trials,    0 EarlyStopped Trials, 0 MetricsUnavailable Trials

    [10256](https://github.com/kubeflow/katib/runs/6892743686?check_suite_focus=true#step:4:10256)
    2022/06/15 04:34:41 Current optimal Trial: { [] {[]}}

No trials are created. You should be able to reproduce it in kind cluster as well

tenzen-y:
@a9p 
Can you please update [katib/test/e2e/v1beta1/hack/gh-actions/run-e2e-experiment.go](https://github.com/kubeflow/katib/blob/6a21af058ec8d02968538a0eb9a5dc2b84f83348/test/e2e/v1beta1/hack/gh-actions/run-e2e-experiment.go) as described in https://github.com/kubeflow/katib/pull/1833#issuecomment-1153488409

manifests/v1beta1/components/controller/katib-config.yaml
```diff
@@ -57,6 +57,19 @@ data:
       },
       "darts": {
         "image": "docker.io/kubeflowkatib/suggestion-darts:latest"
+      },
+      "pbt": {
+        "image": "docker.io/kubeflowkatib/suggestion-pbt:latest",
+        "persistentVolumeClaimSpec": {
+          "accessModes": [
+            "ReadWriteMany"
```
tenzen-y, line 65:
```suggestion
            "ReadWriteOnce"
```

When I reproduced E2E errors on my local machine, suggestion-pbt is Pending since It seems that the KinD storage provisioner only supports "ReadWriteOnce".

ref: https://github.com/kubernetes-sigs/kind/issues/2371

```shell
$ kubectl get pvc --n kubeflow
NAMESPACE   NAME             STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
kubeflow    katib-mysql      Bound     pvc-ce5b364d-12b4-4681-9def-4b89d8a17004   2Gi        RWO            standard       7m19s
kubeflow    simple-pbt-pbt   Pending                                                                        standard       5m10s
```

I changed the mode of the PVC and PV was generated.

```shell
$ kubectl get pvc -n kubeflow
NAMESPACE   NAME             STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
kubeflow    katib-mysql      Bound    pvc-ce5b364d-12b4-4681-9def-4b89d8a17004   2Gi        RWO            standard       14m
kubeflow    simple-pbt-pbt   Bound    pvc-c3b71cdf-7b30-4f93-8631-67c6260752d6   5Gi        RWO            standard       7s
```

```shell
$ kubectl get cm -n kubeflow katib-config -ojsonpath='{.data}' | jq -r '.suggestion'
{
  "pbt": {
    "image": "docker.io/kubeflowkatib/suggestion-pbt:e2e-test",
    "persistentVolumeClaimSpec": {
      "accessModes": [
        "ReadWriteOnce"
      ],
      "resources": {
        "requests": {
          "storage": "5Gi"
        }
      }
    }
  }
}
```

Do we need "ReadWriteMany" PVC in suggestion-pbt?

a9p: I don't think we need to change the `run-e2e-experiment.go`, what you pointed out here is the main issue! The suggestion mount should have RWX access as it is a shared mount across the running trials + suggestion service. For CI testing, RWO should be fine though; is there any way to flag this mount mode for testing vs general use?

andreyvelich: @a9p What is the differences running PBT during CI and manually ?
Will your Experiment fail with RWO ?

a9p: Actually, I believe the experiment should always fail with RWO (CI or otherwise). Suggestion service should always have the claim to the volume so the spawned Trials will get stuck at pending while trying to get the subdirectory mount.

andreyvelich: I don't think Kind supports this out of the box as @tenzen-y mentioned above.
So we might consider add more unit/integration tests for PBT Suggestion and Trial service until Kind supports RWX.

andreyvelich: Thank you for investigating this @tenzen-y!
Do we have any limitations in our CI with Minikube ?
Do you know, is it possible to use Kind with different volume provisioner ( it seems they are using [Local Path Provisioner](https://github.com/kubernetes-sigs/kind/issues/2371#issuecomment-880633851) from Rancher) ?

tenzen-y: > Do we have any limitations in our CI with Minikube ?

Maybe, we do not have any limitations, but minikube cluster is a bit more heavy cluster rather than the KinD cluster.
We may face computing resource limitations of GH Actions.

> Do you know, is it possible to use Kind with different volume provisioner ( it seems they are using https://github.com/kubernetes-sigs/kind/issues/2371#issuecomment-880633851 from Rancher) ?

Yes, we can use other provisioners in the KinD cluster (ex. [topolvm](https://github.com/topolvm/topolvm)). 

However, it might be better to use RWO PVC in that use case since our CI system launch only one node.

What do you think? @andreyvelich @johnugeorge

tenzen-y: looks good

johnugeorge: @tenzen-y did you see if Minikube works?  Isn't that the only way forward?

tenzen-y: @johnugeorge 
I have not tried Minikube yet.
I will see it by today (UTC).

tenzen-y: FYI: https://github.com/kubeflow/katib/pull/1899

johnugeorge:
https://github.com/kubeflow/katib/pull/1899 is merged 

Can you do a rebase?

tenzen-y:
@kubeflow/wg-automl-leads Can you start GH Actions?

a9p:
It looks like no trials kicked off again, but this time due to a pull failure for the `simple-pbt` image. Did I miss something in the .github workflow?

tenzen-y:
> It looks like no trials kicked off again, but this time due to a pull failure for the `simple-pbt` image. Did I miss something in the .github workflow?

Thanks for checking CI. @a9p 
I guess that you need to add `run "suggestion-pbt" "$CMD_PREFIX/suggestion/pbt/$VERSION/Dockerfile"` to the following part.

https://github.com/kubeflow/katib/blob/3e0cfd112884e44e33f9406d97a78512d787c349/test/e2e/v1beta1/scripts/gh-actions/build-load.sh#L160-L167

a9p:
@johnugeorge I think the last approved run failed due to a pull policy that I had in simple-pbt. I checked against the other hp jobs and they should be in parity now (no additional fields)!

johnugeorge:
@a9p Tests didn't complete yet. Does it succeed when you tried locally?

a9p:
@johnugeorge it did work for me, but it seems like I had images that satisfied the docker image pull locally. I setup a new test vm -- the main issue that has been causing the ci failure was a missing entry for simple-pbt in update-images.sh which changes the image tag from latest to e2e-test for the tests.

johnugeorge:
Thanks @a9p for hard work and patience. We are ready to merge. 

One minor addition request. Can you update https://github.com/kubeflow/katib#search-algorithms to add PBT as well in the README table? 

In a separate PR, can you also add a new section https://www.kubeflow.org/docs/components/katib/experiment/#search-algorithms-in-detail  as well in https://github.com/kubeflow/website ? This will help drive users to try PBT.

andreyvelich (approved):
Thank you for this amazing contribution @a9p!
I am excited to see the outcome from our users.

/lgtm

a9p:
Thank you all for your guidance and patience with getting this merged in!

PR #1845, opened 2022-04-11
manifests: Increate the probes seconds

haoxins:

The current probe delay seconds are not suitable for the most environments.
We should increase these values.


anencore94:
I agree with you :) /lgtm

PR #1862, opened 2022-05-05
Population based training UI

a9p:

Support the discovery of modulated hyperparameters rather than attempting to find a fixed set over the entire training process within the Katib UI.

This PR provides some initial support for PBT within the Katib UI (#1382). Required core changes are introduced in #1833.

**Checklist:**


Arhell (commented):
/ok-to-test

a9p:
Changing this request back to WIP: following integration of feedback from #1833, this PR needs to be changed to parse Labels instead of Annotations for metadata.

a9p:
@johnugeorge this PR has two components:
1. pbt additions to the new experiment form
1. pbt tab on the experiment results view

The first one should be good to use and the changes are encompassed in [593881a](https://github.com/kubeflow/katib/pull/1862/commits/593881a2382889ad1d44a5c1c2071ccf1f5f9b58) and [6145922](https://github.com/kubeflow/katib/pull/1862/commits/6145922bc42d8df7d4bc216f047071b596846fb7) (only the prettier needs to be run again on this commit if we want to separate it).

The second one I think could use a bit more usability / code feedback and might be a better target for a subsequent cycle.

a9p:
Noting here if anyone has any thoughts -- 
A standard job for me runs ~1k trials. The UI data query (`fetch_hp_job_info`) seem to take a _significant_ amount of time to aggregate and transmit the data to the browser (order minutes). On cursory glance it might be a DB bottleneck...as we are likely introducing a second query (or augmenting the job info one) in this PR for PBT job information, I was curious if anyone else has seen this performance characteristic.

andreyvelich:
@a9p Yes, we also noticed this issue before.
Katib UI operates slowly when Experiment has many Trials.
We need to investigate if it is [backend or db](https://github.com/kubeflow/katib/blob/master/pkg/ui/v1beta1/hp.go#L82-L110) slowness. Also, UI rendering might take time.
@a9p It would be nice if you could create a separate issue to track this performance issue.

cc @kimwnasptd @elenzio9 @apo-ger

a9p:
Hi @andreyvelich, it doesn't look like I will be able to wrap up this PR before feature freeze. Could we push this to the next release target as well?

andreyvelich:
Sure, no problem @a9p!

PR #1874, opened 2022-05-25
additional metric names should not include objective metric name

### LINKED ISSUE #1873: additional metric names should not include objective metric name
/kind bug

Deploy an experiment which `spec.objective.additionalMetricNames` contains `spec.objective.objectiveMetricName`.

The experiment is created, but you can see nothing on the experiment UI, like below.

- You can see the log `GetLastConditionType failed: Experiment doesn't have any condition` in the Katib UI pod.
- Also in katib controller, you can see the log `"admission webhook \"validator.experiment.katib.kubeflow.org\" denied the request: only spec.parallelTrialCount, spec.maxTrialCount and spec.maxFailedTrialCount are editable",`.

After some investigation, we found that **the experiment is created, but failed to update** when the `additionalMetricNames` contains `objectiveMetricName`.

<details>
<summary>Detailed investigation below</summary>
<p>
This is because of the behavior of the mutating & validation webhook.

If `additionalMetricNames` contains `objectiveMetricName`, mutating webhook tries to add `objectiveMetricName` to `metricStrategies` like below.

- https://github.com/kubeflow/katib/blob/v0.13.0/pkg/apis/controller/experiments/v1beta1/experiment_defaults.go#L57-L59
- https://github.com/kubeflow/katib/blob/v0.13.0/pkg/apis/controller/experiments/v1beta1/experiment_defaults.go#L80-L90

So after the mutation, metric strategies look like below.

```yaml
    metricStrategies:
    - name: Validation-accuracy
      value: max
    - name: Train-accuracy
      value: max
    - name: Validation-accuracy
      value: max
```

Another mutation happens when the controller update finalizer
- https://github.com/kubeflow/katib/blob/v0.13.0/pkg/controller.v1beta1/experiment/experiment_controller.go#L187-L189

However, it is not allowed to update `metricsStrategies` when `oldInst` is not nil
- https://github.com/kubeflow/katib/blob/v0.13.0/pkg/webhook/v1beta1/experiment/validator/validator.go#L109-L111

So the updating experiment is failed and status of experiment will be never updated

</p>
</details>

The experiment should not be created when `additionalMetricNames` contains `objectiveMetricName`.

**Environment:**

- Katib version (check the Katib controller image version): v0.13

---

henrysecond1:

Please see https://github.com/kubeflow/katib/issues/1873 for detailed explanation

Fixes #1873 

**Checklist:**

### DIFF
--- a/pkg/webhook/v1beta1/experiment/validator/validator.go
+++ b/pkg/webhook/v1beta1/experiment/validator/validator.go
@@ -157,6 +157,9 @@ func (g *DefaultValidator) validateObjective(obj *commonapiv1beta1.ObjectiveSpec
 	if obj.ObjectiveMetricName == "" {
 		return fmt.Errorf("no spec.objective.objectiveMetricName specified")
 	}
+	if contains(obj.AdditionalMetricNames, obj.ObjectiveMetricName) {
+		return fmt.Errorf("spec.objective.additionalMetricNames should not contain spec.objective.objectiveMetricName")
+	}
 	return nil
 }
 
--- a/pkg/webhook/v1beta1/experiment/validator/validator_test.go
+++ b/pkg/webhook/v1beta1/experiment/validator/validator_test.go
@@ -108,6 +108,16 @@ func TestValidateExperiment(t *testing.T) {
 			Err:             true,
 			testDescription: "Objective metric name is empty",
 		},
+		{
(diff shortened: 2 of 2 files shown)


johnugeorge:
Thanks!

/lgtm

tenzen-y (commented):
Thanks for this improving @henrysecond1 !
/ok-to-test
/lgtm

PR #1877, opened 2022-05-28
Include MetricsUnavailable condition to Complete in Trial

### LINKED ISSUE #1343: Metrics unavailable Trial condition

tenzen-y:
It is not easy for users to find why Trial failed when training code output incorrect format logs since the trial-controller sets Succeeded condition with False to Trial if there are unavailable metrics in Katib DB as described in https://github.com/kubeflow/katib/issues/1343.
So I also included MetricsUnavailable condition to Complete in Trial.

Fixes #1343

**Checklist:**


johnugeorge:
Kubeflow-presubmit-tests have been removed from prow due to migration. We need to figure out how to enable tests again so that code is broken. 

https://github.com/kubeflow/testing/pull/1003 

We need to think of Github actions

tenzen-y:
Thanks for letting me know, @johnugeorge.
As commented in that PR, it seems that there are any plans for a new kubeflow test-infra from the Kubeflow AWS team.
Are the plans gone away?

johnugeorge:
Yes. There are some plans but completion time is not known yet.  Hence, thinking if we need to move to Github actions. 

Thoughts?

tenzen-y:
> Yes. There are some plans but completion time is not known yet. Hence, thinking if we need to move to Github actions.
> 
> Thoughts?

I see. Generally, I agree with your opinion.

However, if AWS test-infra is rebuilt, we should not go back to AWS test-infra since it is hard for us to go back to AWS test-infra from GitHub Actions.

What do you think, @johnugeorge?

johnugeorge:
Tricky issue is the amount of resources required for Katib tests. I am still not sure if we will be able to fit all tests in GH actions.

tenzen-y:
> Tricky issue is the amount of resources required for Katib tests. I am still not sure if we will be able to fit all tests in GH actions.

I see.
First of all, I will create an issue to track migrating to GitHub Actions.

tenzen-y:
I have rebased.

tenzen-y:
@kubeflow/wg-automl-leads Please take a look.

pkg/controller.v1beta1/experiment/util/status_util.go
```diff
@@ -192,7 +202,9 @@ func UpdateExperimentStatusCondition(collector *ExperimentsCollector, instance *
 	}
 
 	// First check if MaxFailedTrialCount is reached.
-	if (instance.Spec.MaxFailedTrialCount != nil) && (failedTrialsCount > *instance.Spec.MaxFailedTrialCount) {
+	if (instance.Spec.MaxFailedTrialCount != nil) &&
+		((*instance.Spec.MaxFailedTrialCount != *instance.Spec.MaxTrialCount && failedTrialsCount > *instance.Spec.MaxFailedTrialCount) ||
```
johnugeorge, line 206: Can you explain this condition from the current one?

tenzen-y: @johnugeorge  Thanks for your review!
Currently, if `*instance.Spec.MaxFailedTrialCount` is equal to `*instance.Spec.MaxTrialCount` and `failedTrialsCount` is equal to `*instance.Spec.MaxFailedTrialCount`, katib skip this condition and gives `Completed` to the Experiment status.

In other words, katib gives Experiment status `Completed` even though the number of failed Trials reaches `*instance.Spec.MaxFailedTrialCount`.

In this implementation, when the same condition as mentioned above, katib doesn't skip this condition and gives `Failed` to the experiment status.

sdk/python/v1beta1/kubeflow/katib/models/v1beta1_experiment_status.py
```diff
@@ -40,10 +40,12 @@ class V1beta1ExperimentStatus(object):
         'failed_trial_list': 'list[str]',
         'killed_trial_list': 'list[str]',
         'last_reconcile_time': 'datetime',
+        'metrics_unavailable_trial_list': 'list[str]',
```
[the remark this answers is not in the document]
tenzen-y: > You manually generated using gen-python-sdk/gen-sdk.sh ? 

Yes, I updated these files with `make generate`.

> This will help us to ensure that generated files are uptodate

It makes sense, thanks!

pkg/controller.v1beta1/experiment/util/status_util.go
johnugeorge: How does it matter with `*instance.Spec.MaxTrialCount`?  If I understand correctly, status is not set correctly when the number of failed Trials reaches `*instance.Spec.MaxFailedTrialCount`. Then, Isn't condition `failedTrialsCount >= *instance.Spec.MaxFailedTrialCount` enough ?

tenzen-y: In the current implementation, katib create `*instance.Spec.MaxFailedTrialCount + 1` times Trials if all Trials failed.
I implemented this part to keep this specification.
Should I change to `failedTrialsCount >= *instance.Spec.MaxFailedTrialCount`?

sdk/python/v1beta1/kubeflow/katib/models/v1beta1_experiment_status.py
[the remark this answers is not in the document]
tenzen-y: Done.

pkg/controller.v1beta1/experiment/util/status_util.go
johnugeorge: Isn't failedTrialsCount >= *instance.Spec.MaxFailedTrialCount the right one? Experiment should be completed when failed trials reach MaxFailedTrialCount.

pkg/controller.v1beta1/experiment/util/status_util.go
```diff
@@ -177,8 +186,9 @@ func getObjectiveMetricValue(trial trialsv1beta1.Trial) string {
 // UpdateExperimentStatusCondition updates the experiment status.
 func UpdateExperimentStatusCondition(collector *ExperimentsCollector, instance *experimentsv1beta1.Experiment, isObjectiveGoalReached bool, getSuggestionDone bool) {
 	logger := log.WithValues("Experiment", types.NamespacedName{Name: instance.GetName(), Namespace: instance.GetNamespace()})
-	completedTrialsCount := instance.Status.TrialsSucceeded + instance.Status.TrialsFailed + instance.Status.TrialsKilled + instance.Status.TrialsEarlyStopped
-	failedTrialsCount := instance.Status.TrialsFailed
+	completedTrialsCount :=
+		instance.Status.TrialsSucceeded + instance.Status.TrialsFailed + instance.Status.TrialsKilled + instance.Status.TrialsEarlyStopped + instance.Status.TrialMetricsUnavailable
+	failedTrialsCount := instance.Status.TrialsFailed + instance.Status.TrialMetricsUnavailable
```
johnugeorge, line 191: Discussion point:
Is TrialMetricsUnavailable state a failed state for a user?

pkg/controller.v1beta1/experiment/util/status_util.go
tenzen-y: > Isn't failedTrialsCount >= *instance.Spec.MaxFailedTrialCount the right one? Experiment should be completed when failed trials reach MaxFailedTrialCount.

Yes, I agree with @johnugeorge .
However, because as you say, there is an inconsistency between `maxFailedTrialCount` and `maxTrialCount`, we need to change definition of maxFailedTrialCount in order to change to `failedTrialsCount >= *instance.Spec.MaxFailedTrialCount` in this part.

pkg/controller.v1beta1/experiment/util/status_util.go
tenzen-y: I think that the TrialMetricsUnavailable state is a failed Trial for a user since katib can not use that Trial result to optimize hyperparameters even though the training is successful if the metrics-collector container fails to collect metrics.

What do you think? @johnugeorge

johnugeorge: I think so since it is a unrecoverable as well. We have to include this in docs as well to explain "Failed" trial"

tenzen-y: It makes sense.
As I can see katib docs, we don't have any sections to describe Experiment status, so It might be better to explain Experiment status in [this](https://www.kubeflow.org/docs/components/katib/overview/#experiment) paragraph describing Experiment.

Does it sound good to you? @johnugeorge

johnugeorge: Sure. We will wait for a day to get other reviews as well.

tenzen-y:
I will wait 2-3 days for comments from other reviewers.

pkg/controller.v1beta1/experiment/util/status_util.go
johnugeorge: @tenzen-y Please go ahead and make this change - `failedTrialsCount >= *instance.Spec.MaxFailedTrialCount`

tenzen-y: @johnugeorge Sure.

tenzen-y: Done.

johnugeorge:
Thanks @tenzen-y 

/lgtm
/approve

PR #1885, opened 2022-06-04
Add semicolon when using `command` command in Makefile

tenzen-y:
To avoid the `make: command: Command not found` error, I have added `-v` option when using the `command` command in https://github.com/kubeflow/katib/pull/1884, although we received the same error message again.

> Run go mod download
make: command: Command not found
make: command: Command not found
make: command: Command not found
go generate ./pkg/... ./cmd/...
Generate deepcopy, clientset, listers, informers for the APIs

https://github.com/kubeflow/katib/runs/6736557128?check_suite_focus=true#step:5:10

According to this StackOverflow comment, we need to add a semicolon when using the `command` command on the Ubuntu machine.

> you must have a version of GNU make pre 4.3, that doesn't realize command is a shell built-in. You can force it by adding a semicolon like: $(shell command -v python3;)

https://stackoverflow.com/questions/63286462/gnu-make-check-whether-a-python-module-needs-to-be-installed

After applying this patch, I confirmed that we don't receive the error message in [this](https://github.com/tenzen-y/katib/runs/6737111804?check_suite_focus=true#step:5:10) job.

> Run go mod download
go generate ./pkg/... ./cmd/...
Generate deepcopy, clientset, listers, informers for the APIs

/assign @johnugeorge 

Fixes #

**Checklist:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
-HAS_LINT := $(shell command -v golangci-lint)
-HAS_SHELLCHECK := $(shell command -v shellcheck)
-HAS_SETUP_ENVTEST := $(shell command -v setup-envtest)
+HAS_LINT := $(shell command -v golangci-lint;)
+HAS_SHELLCHECK := $(shell command -v shellcheck;)
+HAS_SETUP_ENVTEST := $(shell command -v setup-envtest;)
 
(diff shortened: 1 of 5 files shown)


johnugeorge:
Thanks
/lgtm
/approve

PR #1899, opened 2022-06-16
Change integration test sysytem from KinD Cluster to Minikube Cluster 

tenzen-y:
As discussed in our community meeting, we need to change the integration test system from KinD Cluster to Minikube Cluster to be able to use RWX PVC. 

/assign @johnugeorge @andreyvelich 

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/darts-cifar10-e2e-test.yaml
+++ b/.github/workflows/darts-cifar10-e2e-test.yaml
@@ -2,6 +2,9 @@ name: E2E Test with darts-cnn-cifar10
 on:
   - pull_request
 
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
 jobs:
   e2e:
     runs-on: ubuntu-20.04
@@ -25,9 +28,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Detail: https://hub.docker.com/r/kindest/node
         # TODO (tenzen-y): We need to consider running tests on more kubernetes versions.
-        # kubernetes-version: ["v1.20.15", "v1.21.12", "v1.22.9", "v1.23.6", "v1.24.1"]
-        kubernetes-version: ["v1.21.12", "v1.22.9", "v1.23.6"]
+        # kubernetes-version: ["v1.20.15", "v1.21.13", "v1.22.10", "v1.23.7", "v1.24.1"]
+        kubernetes-version: ["v1.21.13", "v1.22.10", "v1.23.7"]
         # Comma Delimited
         experiments: ["darts-cpu"]
--- a/.github/workflows/template-setup-e2e-test/action.yaml
+++ b/.github/workflows/template-setup-e2e-test/action.yaml
@@ -4,21 +4,17 @@ inputs:
   kubernetes-version:
     required: true
     type: string
-  cluster_name:
-    required: false
-    type: string
-    default: katib-e2e-cluster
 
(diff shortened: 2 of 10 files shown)


.github/workflows/darts-cifar10-e2e-test.yaml
```diff
@@ -2,6 +2,9 @@ name: E2E Test with darts-cnn-cifar10
 on:
   - pull_request
 
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
johnugeorge, line 6: Why do we need GITHUB_TOKEN? I don't think, it is exposed to pull_request type for forked repos anways.

johnugeorge, line 6: Why do we need GITHUB_TOKEN? I don't think, it is exposed to pull_request type for forked repos anways.

tenzen-y: @johnugeorge Thanks for your review.
According to [this](https://github.com/manusa/actions-setup-minikube#optional-input-parameters) document, it seems that [actions-setup-minikube](https://github.com/manusa/actions-setup-minikube) calls GitHub REST API.
So, I added GITHUB_TOKEN.

Would you like to remove GITHUB_TOKEN?

.github/workflows/template-setup-e2e-test/action.yaml
```diff
@@ -4,21 +4,17 @@ inputs:
...
-    type: string
-    default: katib-e2e-cluster
 
 runs:
   using: composite
   steps:
-    - name: Set Up KinD Cluster
-      uses: helm/kind-action@v1.2.0
+    - name: Set Up Minikube Cluster
+      uses: manusa/actions-setup-minikube@v2.6.0
```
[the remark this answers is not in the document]
tenzen-y: I used this action to change to Minikube Cluster with minimal changes.
Do you prefer this action or a manual installation?
If you prefer manual installation rather than this action, I will change the way to install Minkube.

.github/workflows/darts-cifar10-e2e-test.yaml
johnugeorge: For unauthenticated requests, it is limited to 60 per hour https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-personal-accounts Hence, we will need GitHub token if we use this action.

tenzen-y: Yes, it makes sense.
I don't have any concerns about the security issues of GITHUB_TOKEN. Do you have any concerns?

.github/workflows/template-setup-e2e-test/action.yaml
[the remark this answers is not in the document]
tenzen-y: WDYT? @johnugeorge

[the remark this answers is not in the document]
johnugeorge: Can we keep the manual way as kserve? Do you see any red flags?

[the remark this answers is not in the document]
tenzen-y: There is no difference in operation between this action and manual installation.
The only difference is whether we manage the scripts or the administrator of this action manages the scripts.

tenzen-y:
Thanks for the quick review! @johnugeorge

PR #1911, opened 2022-06-30
Fix push script to include new images

johnugeorge:
Push script was not updated with new image names


tenzen-y (approved):
Thanks! @johnugeorge 
/lgtm

PR #1934, opened 2022-08-20
[feat] health check for katib-controller

### LINKED ISSUE #1930: Implement Readiness/LivenessProbe endpoint for katib-controller
/kind feature

Even though katib-controller pod's status became ready. It does not mean that katib-controller is available for all feature.

For example, if user request `/validate-experiment` API to katib-controller, right after katib-controller was just created, It returns 200 response without actually processing the real logic.

First, we need to define how to determine if katib-controller is ready or live. and then implement it.

In usual cases, readiness endpoint checks whether app could process all requests (usually try connecting to the database).
Also, liveness endpoint checks just the container is alive or not.

Raised from https://github.com/kubeflow/katib/pull/1921#discussion_r944579336

---

anencore94:
- Need to check readiness/liveness when deploying katib-controller
  - The main things that we need to check is whether the webhook server is started well.

I've referenced the health check way from other components:
- kubeflow/pipelines's apiserver : https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml#L84-L122
- mysql : https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/components/mysql/mysql.yaml#L42-L67

So in my opinion, the liveness check with just `ping` is ok, but for readiness check, I'm not sure do we need more detailed way to check it or not.

Fixes #1930

**Checklist:**


manifests/v1beta1/components/controller/service.yaml
```diff
@@ -19,5 +19,8 @@ spec:
     - name: metrics
       port: 8080
       targetPort: 8080
+    - name: healthz
+      port: 18080
+      targetPort: 18080
```
tenzen-y, line 24:
```suggestion
```

I think that we do not have to add a port for probes since probes are executed by kubelet, not use kube-proxy.

tenzen-y (commented):
Thanks for your contribution! @anencore94 
I left a few comments.

manifests/v1beta1/components/controller/controller.yaml
```diff
@@ -40,6 +40,17 @@ spec:
...
+              name: healthz
+              protocol: TCP
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: healthz
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
```
tenzen-y, line 53: Should we also add [startupProbe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes)?

anencore94: Sure, with startup probe, we could setup more graceful period only when startup.

I didn't investigate deeply what values are the appropriate for `failureThreshold`, `periodSeconds`, `initialDelaySeconds` and other values, I just use the [default values](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) now.

But if we have some standards to determine readiness, liveness and startup, we will need to seperate those values in detail.

Do you think we need to test to find more appropriate values for each ? @tenzen-y

manifests/v1beta1/components/controller/service.yaml
anencore94: That makes sense! Thanks for finding this :)

manifests/v1beta1/components/controller/controller.yaml
tenzen-y: Sorry for my late response. @anencore94 
It might be better to set each parameter referred to [default values](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) since I think that it is hard to optimize these parameters.

anencore94: I also agree with you. Since currently there is no clue to specification

cmd/katib-controller/v1beta1/main.go
```diff
@@ -129,6 +135,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	log.Info("Setting up health checker.")
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		log.Error(err, "Unable to add healthz endpoint to the manager")
+		os.Exit(1)
+	}
+	// TODO (@anencore94) need to more detailed check whether is it possible to communicate with k8s-apiserver or db-manager at '/readyz' ?
```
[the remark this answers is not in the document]
anencore94: Is there any idea for this PR ? @kubeflow/wg-automl-leads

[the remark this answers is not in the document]
tenzen-y: ping @kubeflow/wg-automl-leads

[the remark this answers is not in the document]
johnugeorge: Sorry for late reply. We can create a issue and track it. For now, we can merge this

tenzen-y (commented):
Also, can you update [this doc](https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md#controller-flags)?

cmd/katib-controller/v1beta1/main.go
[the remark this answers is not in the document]
anencore94: Thanks, I raised this issue on #1973

manifests/v1beta1/components/controller/controller.yaml
anencore94: Also referenced at #1973

anencore94:
> Also, can you update [this doc](https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md#controller-flags)?

Thanks for checking carefully and sorry for the late reponse :) I've updated it!

tenzen-y (commented):
@anencore94 Thanks for the updates!
/lgtm

anencore94:
Could you re-run the failed action and re-review if you don't mind? @johnugeorge

Issue #1900, opened 2022-06-19
Support multi CPU architectures (amd64 and arm64) in all container images with one image tag


tenzen-y:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
As discussed In [this](https://github.com/kubeflow/katib/pull/1898#discussion_r897557325), we need to make all container images compatible with multi-CPU architectures (amd64 and arm64) with one image tag.

ref: https://docs.docker.com/desktop/multi-arch/

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

gaocegege:
The change is subtle and the benefit is huge!

tenzen-y:
I take this issue.
/assign

tenzen-y:
I have completed support for arm64 in all container images.
Next, I will build a CI to verify that the multi-platform image can be built in each PR.

PR #1935, opened 2022-08-20
Introduce Automatic platform ARGs

tenzen-y:
To resolve the complex build conditions for each CPU architecture, I refactored all Dockerfiles using [Automatic platform ARGs](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope).

Also, I switched builder to `docker buildx` since [Automatic platform ARGs](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope) are only available when using the BuildKit as described in the following documentation.

> This feature is only available when using the [BuildKit](https://docs.docker.com/engine/reference/builder/#buildkit) backend.

Part-of: #1900 

**Checklist:**

### DIFF
--- a/.github/workflows/template-publish-image/action.yaml
+++ b/.github/workflows/template-publish-image/action.yaml
@@ -30,7 +30,7 @@ runs:
           type=sha,prefix=v1beta1-
 
     - name: Build and Push
-      uses: docker/build-push-action@v2
+      uses: docker/build-push-action@v3
       with:
         context: .
         file: ${{ inputs.dockerfile }}
--- a/.github/workflows/template-publish-image/action.yaml
+++ b/.github/workflows/template-publish-image/action.yaml
@@ -12,7 +12,7 @@ runs:
   using: composite
   steps:
     - name: Set Up Docker Buildx
-      uses: docker/setup-buildx-action@v1
+      uses: docker/setup-buildx-action@v2
 
     - name: Docker Login
       uses: docker/login-action@v1
@@ -38,3 +38,4 @@ runs:
         tags: ${{ steps.meta.outputs.tags }}
         cache-from: type=gha
         cache-to: type=gha,mode=max
+        platforms: linux/amd64
--- a/.github/workflows/template-setup-e2e-test/action.yaml
+++ b/.github/workflows/template-setup-e2e-test/action.yaml
@@ -17,7 +17,7 @@ runs:
         github token: ${{ env.GITHUB_TOKEN }}
 
     - name: Set Up Docker Buildx
-      uses: docker/setup-buildx-action@v1
+      uses: docker/setup-buildx-action@v2
 
     - name: Set Up Go env
       uses: actions/setup-go@v2
--- a/.github/workflows/test-charmed-katib.yaml
+++ b/.github/workflows/test-charmed-katib.yaml
@@ -49,15 +49,19 @@ jobs:
           sudo snap install juju-wait --classic
           sudo pip3 install charmcraft==1.3.1
 
+      - name: Set Up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
       - name: Build Docker images
         run: |
           set -eux
           images=("katib-controller" "katib-ui" "katib-db-manager")
           folders=("katib-controller" "ui" "db-manager")
           for idx in {0..2}; do
-            docker build . \
+            docker buildx build . \
                 -t docker.io/kubeflowkatib/${images[$idx]}:latest \
-                -f cmd/${folders[$idx]}/v1beta1/Dockerfile
+                -f cmd/${folders[$idx]}/v1beta1/Dockerfile \
+                --load
             docker save docker.io/kubeflowkatib/${images[$idx]} > ${images[$idx]}.tar
             microk8s ctr image import ${images[$idx]}.tar
           done
--- a/cmd/cert-generator/v1beta1/Dockerfile
+++ b/cmd/cert-generator/v1beta1/Dockerfile
@@ -1,5 +1,7 @@
-# Build the Katib Cert Generatoe.
+# Build the Katib Cert Generator.
 FROM golang:alpine AS build-env
+
+ARG TARGETARCH
 
 WORKDIR /go/src/github.com/kubeflow/katib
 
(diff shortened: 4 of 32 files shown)


gaocegege (approved):
LGTM

anencore94 (commented):
/lgtm

Looks nice :) BTW, it's hard to test build all images for each cpu architecture now. Do you think we need to test docker bulid on gihtub action runner with arm-architecture ?

tenzen-y:
> /lgtm
> 
> Looks nice :) BTW, it's hard to test build all images for each cpu architecture now. Do you think we need to test docker bulid on gihtub action runner with arm-architecture ?

Thanks for your comments. @anencore94 
The goal of #1900 is to be able to build [a multi-architecture supported image](https://docs.docker.com/desktop/multi-arch/) (support arm64 and amd64 architecture with a single image tag) with gh-actions and a local machine.

anencore94:
> Thanks for your comments. @anencore94
The goal of https://github.com/kubeflow/katib/issues/1900 is to be able to build [a multi-architecture supported image](https://docs.docker.com/desktop/multi-arch/) (support arm64 and amd64 architecture with a single image tag) on gh-actions and a local machine.

I agree with your comment :), Though what I meant was that I think we may need to check arm-image is built "well".
For example, by contructing k8s cluster in arm-action-runner and setup katib with all arm-katib-image with same tag(supported feature in this PR), and see whether all e2e test passed. @tenzen-y 

What do you think of it ?

tenzen-y:
> > Thanks for your comments. @anencore94
> > The goal of #1900 is to be able to build [a multi-architecture supported image](https://docs.docker.com/desktop/multi-arch/) (support arm64 and amd64 architecture with a single image tag) on gh-actions and a local machine.
> 
> I agree with your comment :), Though what I meant was that I think we may need to check arm-image is built "well". For example, by contructing k8s cluster in arm-action-runner and setup katib with all arm-katib-image with same tag(supported feature in this PR), and see whether all e2e test passed. @tenzen-y
> 
> What do you think of it ?

Thanks for pointing out that. @anencore94
It would be great to verify katib components on  arm64 env by our e2e test. 
Although, it seems that [minikube qemu2 driver does not support emulation arm64 when the host machine's CPU architecture is amd64](https://github.com/kubernetes/minikube/blob/879c592e02668cd85334fce9b28c6ac6b6fb8780/pkg/minikube/registry/drvs/qemu2/qemu2.go#L129-L156). 
So, we need to consider other ways such as the below:

1. Run our e2e tests on Amazon EKS with AWS Graviton2 EC2 instance.
2. Run our e2e tests on Github Actions with a self-hosted runner using AWS Graviton2 EC2 instance.
3. And more...

In any case, we need to touch with the AWS team.
Do you have any other ideas?

anencore94:
> > > Thanks for your comments. @anencore94
> > > The goal of #1900 is to be able to build [a multi-architecture supported image](https://docs.docker.com/desktop/multi-arch/) (support arm64 and amd64 architecture with a single image tag) on gh-actions and a local machine.
> > 
> > 
> > I agree with your comment :), Though what I meant was that I think we may need to check arm-image is built "well". For example, by contructing k8s cluster in arm-action-runner and setup katib with all arm-katib-image with same tag(supported feature in this PR), and see whether all e2e test passed. @tenzen-y
> > What do you think of it ?
> 
> Thanks for pointing out that. @anencore94 It would be great to verify katib components on arm64 env by our e2e test. Although, it seems that [minikube qemu2 driver does not support emulating arm64 when the host machine's CPU architecture is amd64](https://github.com/kubernetes/minikube/blob/879c592e02668cd85334fce9b28c6ac6b6fb8780/pkg/minikube/registry/drvs/qemu2/qemu2.go#L129-L156). So, we need to consider other ways such as the below:
> 
> 1. Run our e2e tests on Amazon EKS with AWS Graviton2 EC2 instance.
> 2. Run our e2e tests on Github Actions with a self-hosted runner using AWS Graviton2 EC2 instance.
> 3. And more...
> 
> In any case, we need to touch with the AWS team. Do you have any other ideas?

Thanks for your detailed comments :+1: @tenzen-y 
For [minikube qemu2 driver does not support emulating arm64 when the host machine's CPU architecture is amd64](https://github.com/kubernetes/minikube/blob/879c592e02668cd85334fce9b28c6ac6b6fb8780/pkg/minikube/registry/drvs/qemu2/qemu2.go#L129-L156), I agree, so that's why I think we need to setup arm64-github-action runner. (it is second way of your suggestion)

tenzen-y:
> so that's why I think we need to setup arm64-github-action runner. (it is second way of your suggestion)

I think so too.

What do you think? @kubeflow/wg-automl-leads

johnugeorge:
My only concern is that our CI tests will take much longer to test both platforms.

tenzen-y:
> My only concern is that our CI tests will take much longer to test both platforms.

It makes sense.  @johnugeorge Deploying multiple VMs might be helpful for us.

Also, we need to build a CI system that deploys an ephemeral clean VM for each job, as described in the following document.

> GitHub-hosted runners execute code within ephemeral and clean isolated virtual machines, meaning there is no way to persistently compromise this environment, or otherwise gain access to more information than was placed in this environment during the bootstrap process.
>
> Self-hosted runners for GitHub do not have guarantees around running in ephemeral clean virtual machines, and can be persistently compromised by untrusted code in a workflow.
>
> As a result, self-hosted runners should almost [never be used for public repositories](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories) on GitHub, because any user can open pull requests against the repository and compromise the environment. Similarly, be cautious when using self-hosted runners on private or internal repositories, as anyone who can fork the repository and open a pull request (generally those with read access to the repository) are able to compromise the self-hosted runner environment, including gaining access to secrets and the GITHUB_TOKEN which, depending on its settings, can grant write access to the repository. Although workflows can control access to environment secrets by using environments and required reviews, these workflows are not run in an isolated environment and are still susceptible to the same risks when run on a self-hosted runner.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#hardening-for-self-hosted-runners

johnugeorge:
@tenzen-y Since we are planning to use AWS credits, we can create a new issue to track various possible solutions with their required CI changes for each. This will help to estimate efforts for each

tenzen-y:
> @tenzen-y Since we are planning to use AWS credits, we can create a new issue to track various possible solutions with their required CI changes for each. This will help to estimate efforts for each solution

Sounds good. @johnugeorge 
I will create an issue to keep tracking this.

anencore94:
> > @tenzen-y Since we are planning to use AWS credits, we can create a new issue to track various possible solutions with their required CI changes for each. This will help to estimate efforts for each solution
> 
> Sounds good. @johnugeorge I will create an issue to keep tracking this.

Thanks @johnugeorge , @tenzen-y . That would be nice

PR #1951, opened 2022-09-13
Create Tune API in the Katib SDK

### LINKED ISSUE #1585: Automate Container Build for Trial training containers
/kind feature

We discussed in our WG meetings and in the summit on 2021-07-16 about Trial training container's source code automation.

Many users request automation for container build instead of manually building images. In that case, users will have uniform experience to tune HP from the Jupyter Notebooks.
We might extend the Katib SDK to support it.

Let's continue discussion about it under this issue.

Resources from GCP: [using Buildpacks](https://cloud.google.com/blog/products/containers-kubernetes/google-cloud-now-supports-buildpacks?_lrsc=c87ebc65-7b47-4d40-bc5f-fce4a2f97f4b&utm_source=elevate&utm_medium=social&utm_campaign=elevate_TwG&utm_term=elevate_TwG_SEID:277467_BID:1111642_TID:6109_SHID:23512237).

cc @gaocegege @johnugeorge 
/priority p2

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1585.
Inspired by [KFP `create_component_from_func`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/component_factory.py).

This API will allow user to create Katib HP Tuning Experiment without building the Experiment YAML.
This is the first small step to simplify our Kubeflow SDKs and to avoid Kubernetes complexity.

Later we can extend this functionality (support more Katib features in a simple way), give more spec options via APIs.

Also, we can provide API (e.g. `katib.list_base_images`) to show list of supported images (e.g. `BASE_IMAGE_TENSORFLOW`, `BASE_IMAGE_MXNET`) with their description.

cc @kubeflow/wg-training-leads @tenzen-y @anencore94 Please give your feedback on the API design.


tenzen-y:
Thanks for implementing the new feature!
First, I'll review the PR submitted in training-operator repo.

README.md
```diff
@@ -189,10 +189,51 @@ katib-ui-5767cfccdc-pwg2x           1/1     Running     0          36s
...
+In the following example we are going to maximize a simple objective function:
+$F(a,b) = 4a - b^2$. The bigger $a$ and the lesser $b$ value, the bigger the function value $F$.
+
+```python
+import kubeflow.katib as katib
+
+# Step 1. Create an objective function.
+def objective(parameters):
+    # Import required packages.
+    import time
```
anencore94, line 205: I understand why this kind of logic is necesary, but we might find out more friendly way to specify packages by users.
However, as a minimal feature, this is ok if we could instruct users that they couldn't make a mistake to miss required packages in the objective function.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -430,7 +701,56 @@ def get_optimal_hyperparameters(self, name=None, namespace=None):
...
-        result["currentOptimalTrial"] = katibexp.get(
-            "status", {}).get("currentOptimalTrial")
+        result["currentOptimalTrial"] = katibexp.get("status", {}).get(
+            "currentOptimalTrial"
+        )
 
         return result
+
+
+def double(min: float, max: float, step: float = None):
```
anencore94, line 711: How about gather these hyperparameter helper function to another inner module?
For example, the users could call these functions by

```python

hp_search_space = {
    "a": katib.hp_search_space.int(min=10, max=20),
    "b": katib.hp_search_space.double(min=0.1, max=0.2)
}

katib_client.tune(
    name=name,
    objective=objective,
    hp_search_space=hp_search_space,
    objective_metric_name="result",
    max_trial_count=12
)
```

IMO, importing by `katib.int` makes me it's too board that we couldn't see it is for hyperparameter search space type immediately.

sdk/python/v1beta1/kubeflow/katib/constants/constants.py
```diff
@@ -15,13 +15,21 @@
...
 # Katib K8S constants
-KUBEFLOW_GROUP = 'kubeflow.org'
-EXPERIMENT_PLURAL = 'experiments'
-SUGGESTION_PLURAL = 'suggestions'
-TRIAL_PLURAL = 'trials'
+KUBEFLOW_GROUP = "kubeflow.org"
+EXPERIMENT_KIND = "Experiment"
+EXPERIMENT_PLURAL = "experiments"
+SUGGESTION_PLURAL = "suggestions"
+TRIAL_PLURAL = "trials"
```
anencore94, line 25: how about speficy these variables to type hint `Final[str]`?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -430,7 +701,56 @@ def get_optimal_hyperparameters(self, name=None, namespace=None):
...
+    parameter = models.V1beta1ParameterSpec(
+        parameter_type="int",
+        feasible_space=models.V1beta1FeasibleSpace(min=str(min), max=str(max)),
+    )
+    if step is not None:
+        parameter.feasible_space.step = str(step)
+    return parameter
+
+
+def categorical(list: List):
```
anencore94, line 747: do we need to check if list is `[]` or not ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -53,47 +60,253 @@ def _is_ipython(self):
...
+                objective function in addition to objective metric.
+            objective_type: Type for the Experiment optimization for the objective metric.
+                Must be one of `minimize` or `maximize`.
+            objective_goal: Objective goal that Experiment should reach to be Succeeded.
+            max_trial_count: Maximum number of Trials to run.
+            parallel_trial_count: Number of Trials that Experiment runs in parallel.
+            max_failed_trial_count: Maximum number of Trials allowed to fail.
+            retain_trials: Whether Trials' resources (e.g. pods) are deleted after Succeeded state.
+            packages_to_install: List of Python packages to install before
+                executing the objective function.
```
anencore94, line 162: how about saying that these field is for packages except already installed in the `base_image` ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -53,47 +60,253 @@ def _is_ipython(self):
...
+        parameters: Dict[str, Any],
+        base_image: str = constants.BASE_IMAGE_TENSORFLOW,
+        namespace: str = utils.get_default_target_namespace(),
+        algorithm_name: str = "random",
+        objective_metric_name: str = None,
+        additional_metric_names: List[str] = [],
+        objective_type: str = "maximize",
+        objective_goal: float = None,
+        max_trial_count: int = None,
+        parallel_trial_count: int = None,
```
anencore94, line 128: how about these fields' default values to `1` for who are not familiar with katib experiments?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -53,47 +60,253 @@ def _is_ipython(self):
...
+        packages_to_install: List[str] = None,
+        pip_index_url: str = "https://pypi.org/simple",
+    ):
+        """Create HyperParameter Tuning Katib Experiment from the objective function.
+
+        Args:
+            name: Name for the Experiment.
+            objective: Objective function that Katib uses to train the model.
+                This function must be Callable and it must have only one dict argument.
+                Katib uses this argument to send HyperParameters to the function.
```
anencore94, line 140: I think we need to strongly-specify that objective function should inner-import statements for all necesary packages

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -53,47 +60,253 @@ def _is_ipython(self):
...
+        # TODO (andreyvelich): How to be consistent with other APIs (name) ?
+        name: str,
+        objective: Callable,
+        parameters: Dict[str, Any],
+        base_image: str = constants.BASE_IMAGE_TENSORFLOW,
+        namespace: str = utils.get_default_target_namespace(),
+        algorithm_name: str = "random",
+        objective_metric_name: str = None,
+        additional_metric_names: List[str] = [],
+        objective_type: str = "maximize",
```
anencore94, line 125: maybe we could use "maximize" and "minimize" as an `Enum` class or `Literal`

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -53,47 +60,253 @@ def _is_ipython(self):
...
+        # Verify the objective function arguments.
+        objective_signature = inspect.signature(objective)
+        try:
+            objective_signature.bind({})
+        except Exception:
+            raise ValueError(
+                "Invalid args in the Objective function. "
+                "The function args must have only 'parameters' dictionary. "
+                f"Current Objective arguments: {objective_signature}"
+            ) from None
```
anencore94, line 212: How about extracting this part to a single function with naming something like `validate_objective_fucntion`?
I think this kinds of validation for user's wrong-use could become bigger and bigger.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -53,47 +60,253 @@ def _is_ipython(self):
...
+                )
+            else:
+                # Otherwise, add value to the function input.
+                input_params[p_name] = p_value
+
+        # Wrap objective function to execute it from the file. For example
+        # def objective(parameters):
+        #     print(f'Parameters are {parameters}')
+        # objective({'lr': '${trialParameters.lr}', 'epochs': '${trialParameters.epochs}', 'is_dist': False})
+        objective_code = f"{objective_code}\n{objective.__name__}({input_params})\n"
```
anencore94, line 247: I'm not sure do we need the last `\n` or not

anencore94 (commented):
Thanks for awesome feature !!! users could generate katib experiment much easier now :)
I left some personal opinions.

sdk/python/v1beta1/kubeflow/katib/constants/constants.py
```diff
@@ -15,13 +15,21 @@
...
 
 # How long to wait in seconds for requests to the ApiServer
 APISERVER_TIMEOUT = 120
+
+DEFAULT_PRIMARY_CONTAINER_NAME = "training-container"
+
+# Supported base images for the Katib Trials.
+# TODO (andreyvelich): Implement list_base_images function to get each image description.
+BASE_IMAGE_TENSORFLOW = "docker.io/tensorflow/tensorflow:2.9.1"
+BASE_IMAGE_MXNET = "docker.io/mxnet/python:1.9.1_native_py3"
```
terrytangyuan, line 35: Can these be modified by the user? Otherwise, the versions here are tied to Katib Python SDK releases.

README.md
```diff
@@ -189,10 +189,51 @@ katib-ui-5767cfccdc-pwg2x           1/1     Running     0          36s
...
+    time.sleep(5)
+    # Calculate objective function.
+    result = 4 * int(parameters["a"]) - float(parameters["b"]) ** 2
+    # Katib parses metrics in this format: <metric-name>=<metric-value>.
+    print(f"result={result}")
+
+# Step 2. Create HyperParameter search space.
+parameters = {
+    "a": katib.int(min=10, max=20),
+    "b": katib.double(min=0.1, max=0.2)
```
terrytangyuan, line 215: This doesn't seem Pythonic to me. Is it easy to support user specified list ranges, e.g. `range(10, 20)`?

README.md
andreyvelich: I agree, but I think Kubeflow Pipelines follows the same way: https://www.kubeflow.org/docs/components/pipelines/sdk/python-function-components/#building-python-function-based-components
@zijianjoy Do you know if Pipelines SDK verifies that all function imports are defined properly ?

@anencore94 Any ideas how to verify this?
cc @tenzen-y @gaocegege @johnugeorge

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: I was even thinking to change this API in the future to support various distributions: `uniform`, `loguniform`, `categorical`, etc.

FYI, hyperopt uses [`hp` module](https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/hyperopt/base_service.py#L59), skop uses [`space` module](https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/skopt/base_service.py#L54), optuna uses [`distributions` module](https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/optuna/base_service.py#L94), Ray uses [tune module](https://docs.ray.io/en/latest/tune/api_docs/suggestion.html?highlight=search#ax-tune-search-ax-axsearch). I think we can learn something from it. 

`search` module looks good to me, but I am not sure what is the best UX here.

What others think @terrytangyuan @gaocegege @johnugeorge @c-bata @a9p @g-votte @tenzen-y ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Should we introduce this check in [Validation Webhook](https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/experiment/validator/validator.go#L244-L246), so we can have single place for spec validation ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Sure. @anencore94 What do you think about this:
```
List of Python packages to install in addition to base image packages. These packages are installed before executing the objective function.
        ```

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 6 above)
Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:128]: We set default Trial Threshold here: https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/experiments/v1beta1/experiment_defaults.go#L35-L39
I am not sure, if we want to introduce another default value in the `tune` API.
WDYT @anencore94 @tenzen-y ?

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 7 above)
Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:140]: Let's discuss about it in this comment: https://github.com/kubeflow/katib/pull/1951#discussion_r972199141.
At the minimum, we can just extend this comment as you mentioned.

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 8 above)
Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:125]: I was also thinking about it. 
Is there a way to automatically generate enums for our [Python Models](https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/kubeflow/katib/models/v1beta1_objective_spec.py#L40) ?
In our APIs `type` param is enum: https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/common/v1beta1/common_types.go#L96

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 10 above)
Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:247]: I think that is required for the newline in `ephemeral_objective.py` file. It's good practice to have it.

File: sdk/python/v1beta1/kubeflow/katib/constants/constants.py (hunk 11 above)
Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/constants/constants.py:35]: @terrytangyuan Sure, users can set [the custom base image](https://github.com/kubeflow/katib/blob/86d3a686f3c5d4e3e98d68465aade78d74c96b4e/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L120).

File: README.md (hunk 12 above)
Comment (andreyvelich) [reply README.md:215]: @terrytangyuan Let's continue discussion around it here: https://github.com/kubeflow/katib/pull/1951#discussion_r970890433.

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 2 above)
Comment (terrytangyuan) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:697]: Ref https://github.com/kubeflow/katib/pull/1951#discussion_r971944632. 

IMO it feels more natural if we could support built-in lists or common numpy arrays, etc. Is there a particular motivation for bringing a separate module in the Katib SDK just to do the same thing?

Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:697]: @terrytangyuan List and numpy arrays might not be enough. 
How we can identify search space for `uniform` or `loguniform` in the future?

e.g. Hyperopt utilise some numpy APIs, but it still has it's own functions for the search space: https://github.com/hyperopt/hyperopt/blob/master/hyperopt/pyll/stochastic.py#L41-L43

Comment (terrytangyuan) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:697]: The functions in that module seem to be very thin wrappers. If we could completely avoid creating additional wrappers/layers, it would be great.

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 4 above)
Comment (anencore94) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:733]: Of course but I'm not sure if we need some fast-validation for the user who are not familiar with katib. Also representing the validation rule in python's stdout would be better for them to find out a mistake.

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 5 above)
Comment (anencore94) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py]: Looks nice :) Thanks

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 6 above)
Comment (anencore94) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:128]: I introduce this because of the same reason with this comment: https://github.com/kubeflow/katib/pull/1951#discussion_r972651507.

I'm not sure if we have to implement such fast-validation (which is duplicated) or not.

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 8 above)
Comment (anencore94) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:125]: By "automatically" means that automatically generate python Enum class from the golang file or openapi.json ?

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 2 above)
Comment (tenzen-y) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:697]: I agree with @andreyvelich.
In this PR, we can start in the minimal feature. So, we can support other distributions in other PRs.

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 6 above)
Comment (tenzen-y) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:128]: Since introducing duplicated values will cause the bug, it would be good not to set the default values in this.

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 2 above)
Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:697]: @terrytangyuan Do you have any ideas how we can improve UX on this ?

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 4 above)
Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:733]: I think if Validation Webhook fails, user also will see an error in the Python's stdout before creating Katib Experiment.
What do you think @anencore94 ?

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client.py (hunk 8 above)
Comment (andreyvelich) [reply sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:125]: > By "automatically" means that automatically generate python Enum class from the golang file or openapi.json ?

Exactly. I am not sure if `kube-openapi` supports enum generation. Check this: https://github.com/kubernetes/enhancements/issues/2887.
Maybe we should update `kube-openapi` version for this feature.

File: sdk/python/v1beta1/setup.py [hunk 13]
```diff
@@ -14,40 +14,44 @@
...
 import setuptools
 
-with open('requirements.txt') as f:
-    REQUIRES = f.readlines()
+REQUIRES = [
+    "certifi>=14.05.14",
+    "six>=1.10",
+    "setuptools>=21.0.0",
+    "urllib3>=1.15.1",
+    "kubernetes>=12.0.0",
```
tenzen-y, line 22: I think that `12.0.0` is so old.
Should we change the required kubernetes client version?

https://github.com/kubernetes-client/python#compatibility-matrix-of-supported-client-versions

sdk/python/v1beta1/kubeflow/katib/constants/constants.py
```diff
@@ -13,15 +13,28 @@
...
+TRIAL_PLURAL: Final[str] = "trials"
 
-# How long to wait in seconds for requests to the ApiServer
-APISERVER_TIMEOUT = 120
+
+DEFAULT_PRIMARY_CONTAINER_NAME: Final[str] = "training-container"
+
+# Supported base images for the Katib Trials.
+# TODO (andreyvelich): Implement list_base_images function to get each image description.
+BASE_IMAGE_TENSORFLOW: Final[str] = "docker.io/tensorflow/tensorflow:2.9.1"
```
tenzen-y, line 36: Should we add the base image for tensorflow with GPU support same as the tfjob client?

tenzen-y (commented):
@andreyvelich Thanks for your contributions!
I left a few comments.

sdk/python/v1beta1/setup.py
andreyvelich: Sure, let's be in sync with Training Operator SDK.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: If the validadation webhook's failure message is detailed enough, your suggestion would be much better than implementing the same validation check in both!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: I agree with you, but then I think we should write docstring and tutorial page more detailed and kind!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
anencore94: I have no many experience with it sadly. How about make an issue with this feature, and merge this PR with hardcoded for now ?

README.md
anencore94: We may need to automatically validating this after this PR. How about making an issue for tracking this feature now ?
If we could make an idea for this validation in general, we may apply this for other kf components like pipelines and training-operators and so on.

andreyvelich: Sure, that sound good @anencore94.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: @anencore94 @tenzen-y Maybe we update docstring with link to [this page](https://www.kubeflow.org/docs/components/katib/experiment/#configuration-spec) for Trial budget info?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Sounds good @anencore94 !

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
tenzen-y: That sounds good.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
terrytangyuan: I am ok as is if other projects are following the same convention.

andreyvelich:
Thanks everyone for review!
I addressed the remaining comments and [create separate search module for HPs](https://github.com/kubeflow/katib/pull/1951/commits/316d4a4fbf9f50f839b9141908da3198f0884f72).
Please take a look @anencore94 @tenzen-y @johnugeorge @terrytangyuan

sdk/python/v1beta1/kubeflow/katib/api/search.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2021 The Kubeflow Authors.
```
tenzen-y, line 1: Probably, this is `2022`, not `2021`.

tenzen-y (commented):
@andreyvelich Thanks for updating this!

README.md
```diff
@@ -189,10 +189,51 @@ katib-ui-5767cfccdc-pwg2x           1/1     Running     0          36s
...
+# Step 4. Get the best HyperParameters.
+print(katib_client.get_optimal_hyperparameters(name))
+```
+
 # Documentation
 
-- Run your first Katib Experiment in the
-  [getting started guide](https://www.kubeflow.org/docs/components/katib/hyperparameter/#example-using-random-algorithm).
+- Check
+  [the Katib getting started guide](https://www.kubeflow.org/docs/components/katib/hyperparameter/#example-using-random-algorithm).
```
tenzen-y, line 236:
```suggestion
  [the Katib getting started guide](https://www.kubeflow.org/docs/components/katib/hyperparameter/#example-using-random-search-algorithm).
```

I found this link had been broken for some time.

tenzen-y:
Thanks for addressing my comments!
/lgtm

johnugeorge:
Thanks @andreyvelich for this great feature

andreyvelich:
Thanks everyone for your feedback and review!
I am looking forward for the new Katib usage.
/hold cancel

terrytangyuan (commented):
This is great! 

/lgtm

PR #1980, opened 2022-10-24
[SDK] Remove Final Keyword from constants

andreyvelich:
I think, we should remove Final keyword from our constants in Python SDK.
Final is available only in >= Python 3.8 version: https://peps.python.org/pep-0591/
Currently, we say that [Katib SDK is supported in Python 3 and 3.7](https://github.pie.apple.com/kubeflow-oss/katib/blob/master/sdk/python/v1beta1/setup.py#L42-L47).
Also, it's better to be compatible with other Kubeflow components SDKs, [e.g. with Kubeflow Pipelines](https://github.com/kubeflow/pipelines/blob/master/sdk/python/setup.py#L85-L87).

We can think about Python upgrade in the future releases/versions across various Kubeflow components.

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/constants/constants.py
+++ b/sdk/python/v1beta1/kubeflow/katib/constants/constants.py
@@ -13,29 +13,26 @@
 # limitations under the License.
 
 import os
-from typing import Final
 
 # How long to wait in seconds for requests to the ApiServer
-APISERVER_TIMEOUT: Final[int] = 120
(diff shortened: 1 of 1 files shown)


johnugeorge:
@andreyvelich Can you do a rebase?

tenzen-y (commented):
@andreyvelich  Thanks for pointing this out!
/lgtm

I think that we could upgrade the minimal support python version to v3.8 still releasing katib v0.15.

PR #1981, opened 2022-10-24
[SDK] Fix namespace parameter in tune API

andreyvelich:
We should send `namespace` parameter to `create_experiment` API, so user can set it using `tune` API.

cc @tenzen-y @johnugeorge @anencore94


tenzen-y (commented):
@andreyvelich Thanks fro fixing this!
/lgtm

Issue #1440, opened 2021-02-18
[UI Feature] Page to manage Trial templates


andreyvelich:
/kind feature

The new Katib UI should have a page to manage Trial templates.

/cc @kimwnasptd

PR #1983, opened 2022-10-25
Add authorization mechanisms in new Katib UI backend

### LINKED ISSUE #1547: [ui-authz-subjectaccessreviews] Use SubjectAccessReviews to ensure the web app respects K8s RBAC
/kind feature

Right now the backend for both the new and old UIs does not perform any authorization checks and only relies on the permissions of the backend's ServiceAccount. 

The backend should follow the architecture of the other backends https://github.com/kubeflow/kubeflow/issues/5566 and use SubjectAccessReviews before making an action, to ensure it doesn't violate K8s RBAC permissions.

We should enforce these authz checks at least for actions on Experiments and Trials. The tricky part will be the checks for ConfigMaps.

The app handles ConfigMaps for the TrialTemplates which, right now, only reside in the `kubeflow` namespace. IMO we shouldn't give permissions on users to edit/delete ConfigMaps in the `kubeflow` namespace. We had already started a discussion about having TrialTemplates reside in each namespace, and not in `kubeflow` https://github.com/kubeflow/katib/issues/1546. 

I think we have two options:
1. Postpone this effort until we've changed our UX to have TrialTemplates per namespace and not always in `kubeflow` namespace
2. Move forward but don't perform any authnz checks for ConfigMaps in the `kubeflow` namespace that have the TrialTemplate specific label

@andreyvelich @johnugeorge @gaocegege WDYT?

apo-ger:
The current backend for the new UI does not perform any authorization checks. It should comply with the authz guidelines (https://github.com/kubeflow/kubeflow/issues/5566) and use `SubjectAccessReviews` (SAR) to check if a user has permissions to perform an action.

This PR:
- introduces authz checks for actions on:
  - `Trials`
  - `Experiments`
  - `TrialTemplates`

- introduces three ENV variables:
  - `APP_DISABLE_AUTH`: Skip authorization if an admin is explicitly requested it.
  - `BACKEND_MODE`: dev or prod mode. Skip authorization when in dev mode.
  - `USERID_HEADER`: to check and ensure that a username is provided into the header of every http request towards the api-server.

Before any request proceed to K8s api-server:
* check if authorization must be skipped (BACKEND_MODE, APP_DISBLE_AUTH)
* check if a username is proviced in request Header (authentication)
* query the K8s api-server with SAR to ensure that the user has appropriate access privileges

Other changes:
- Introduce a client-go client to construct `SubjectAccessReview` objects.
- Replace the `/katib/fetch_experiment/` route with `/katib/fetch_namespaced_experiments`. This route expects a namespace as a query parameter from which all experiments will be fetched.
- Update the `README` of the web app to expose that devs should set `APP_DISABLE_AUTH=true` when running locally

Fixes #1547

### DIFF
--- a/cmd/new-ui/v1beta1/main.go
+++ b/cmd/new-ui/v1beta1/main.go
@@ -48,7 +48,7 @@ func main() {
 	http.HandleFunc("/katib/", kuh.ServeIndex(*buildDir))
 	http.Handle("/katib/static/", http.StripPrefix("/katib/", frontend))
 
-	http.HandleFunc("/katib/fetch_experiments/", kuh.FetchAllExperiments)
+	http.HandleFunc("/katib/fetch_namespaced_experiments/", kuh.FetchNamespacedExperiments)
 
 	http.HandleFunc("/katib/create_experiment/", kuh.CreateExperiment)
 
--- a/cmd/new-ui/v1beta1/main.go
+++ b/cmd/new-ui/v1beta1/main.go
@@ -48,7 +48,7 @@ func main() {
 	http.HandleFunc("/katib/", kuh.ServeIndex(*buildDir))
 	http.Handle("/katib/static/", http.StripPrefix("/katib/", frontend))
 
-	http.HandleFunc("/katib/fetch_namespaced_experiments/", kuh.FetchNamespacedExperiments)
+	http.HandleFunc("/katib/fetch_experiments/", kuh.FetchExperiments)
 
 	http.HandleFunc("/katib/create_experiment/", kuh.CreateExperiment)
 
--- a/pkg/new-ui/v1beta1/authzn.go
+++ b/pkg/new-ui/v1beta1/authzn.go
@@ -2,13 +2,13 @@ package v1beta1
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
 	"strings"
 
 	"github.com/kubeflow/katib/pkg/util/v1beta1/env"
-	"github.com/pkg/errors"
 	v1 "k8s.io/api/authorization/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -30,7 +30,7 @@ func GetUsername(r *http.Request) (string, error) {
 	}
 
 	if r.Header.Get(USER_HEADER) == "" {
-		return "", errors.New("User header not present!")
+		return "", errors.New("user header not present")
 	}
 
 	user := r.Header.Get(USER_HEADER)
--- a/pkg/new-ui/v1beta1/authzn.go
+++ b/pkg/new-ui/v1beta1/authzn.go
@@ -10,8 +10,8 @@ import (
 
 	"github.com/kubeflow/katib/pkg/util/v1beta1/env"
 	v1 "k8s.io/api/authorization/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ENV variables
@@ -40,8 +40,7 @@ func GetUsername(r *http.Request) (string, error) {
 }
 
 // Function for constructing SubjectAccessReviews (SAR) objects
-func CreateSAR(user, verb, namespace, group,
-	version, resource, subresource, name string) *v1.SubjectAccessReview {
+func CreateSAR(user, verb, namespace, resource, subresource, name string, schema schema.GroupVersion) *v1.SubjectAccessReview {
 
 	sar := &v1.SubjectAccessReview{
 		Spec: v1.SubjectAccessReviewSpec{
@@ -49,8 +48,8 @@ func CreateSAR(user, verb, namespace, group,
 			ResourceAttributes: &v1.ResourceAttributes{
 				Namespace:   namespace,
 				Verb:        verb,
-				Group:       group,
-				Version:     version,
+				Group:       schema.Group,
+				Version:     schema.Version,
 				Resource:    resource,
 				Subresource: subresource,
 				Name:        name,
@@ -60,8 +59,7 @@ func CreateSAR(user, verb, namespace, group,
 	return sar
 }
 
-func IsAuthorized(user, verb, namespace, group,
-	version, resource, subresource, name string, client *kubernetes.Clientset) error {
+func IsAuthorized(user, verb, namespace, resource, subresource, name string, schema schema.GroupVersion, client client.Client) error {
 
 	// Skip authz when in dev_mode
 	if BACKEND_MODE == "dev" || BACKEND_MODE == "development" {
@@ -74,32 +72,30 @@ func IsAuthorized(user, verb, namespace, group,
 		return nil
 	}
 
-	sar := CreateSAR(user, verb, namespace, group, version, resource, subresource, name)
+	sar := CreateSAR(user, verb, namespace, resource, subresource, name, schema)
 
-	res, err := client.AuthorizationV1().SubjectAccessReviews().Create(context.TODO(), sar, metav1.CreateOptions{})
+	err := client.Create(context.TODO(), sar)
 	if err != nil {
 		log.Printf("Error submitting SubjectAccessReview: %v, %s", sar, err.Error())
 		return err
 	}
 
-	if res.Status.Allowed {
+	if sar.Status.Allowed {
 		return nil
 	}
 
-	msg := generateUnauthorizedMessage(user, verb, namespace, group, version, resource, subresource, res)
-	err = errors.New(msg)
-	return err
+	msg := generateUnauthorizedMessage(user, verb, namespace, resource, subresource, schema, sar)
+	return errors.New(msg)
 }
 
-func generateUnauthorizedMessage(user, verb, namespace, group,
-	version, resource, subresource string, sar *v1.SubjectAccessReview) string {
+func generateUnauthorizedMessage(user, verb, namespace, resource, subresource string, schema schema.GroupVersion, sar *v1.SubjectAccessReview) string {
 
 	msg := fmt.Sprintf("User: %s is not authorized to %s", user, verb)
 
-	if group == "" {
-		msg += fmt.Sprintf(" %s/%s", version, resource)
+	if schema.Group == "" {
+		msg += fmt.Sprintf(" %s/%s", schema.Version, resource)
 	} else {
-		msg += fmt.Sprintf(" %s/%s/%s", group, version, resource)
+		msg += fmt.Sprintf(" %s/%s/%s", schema.Group, schema.Version, resource)
 	}
 
 	if subresource != "" {
--- a/pkg/new-ui/v1beta1/authzn.go
+++ b/pkg/new-ui/v1beta1/authzn.go
@@ -18,14 +18,13 @@ import (
 var (
 	USER_HEADER  = env.GetEnvOrDefault("USERID_HEADER", "kubeflow-userid")
 	USER_PREFIX  = env.GetEnvOrDefault("USERID_PREFIX", ":")
-	DISABLE_AUTH = env.GetEnvOrDefault("APP_DISABLE_AUTH", "false") == "true"
-	BACKEND_MODE = env.GetEnvOrDefault("BACKEND_MODE", "prod")
+	DISABLE_AUTH = env.GetEnvOrDefault("APP_DISABLE_AUTH", "false")
 )
 
 func GetUsername(r *http.Request) (string, error) {
 	var username string
-	if DISABLE_AUTH {
-		log.Printf("APP_DISABLE_AUTH set to True. Skipping authorization check")
+	if DISABLE_AUTH == "true" {
+		log.Printf("APP_DISABLE_AUTH set to True. Skipping authentication check")
 		return "", nil
 	}
 
@@ -61,13 +60,8 @@ func CreateSAR(user, verb, namespace, resource, subresource, name string, schema
 
 func IsAuthorized(user, verb, namespace, resource, subresource, name string, schema schema.GroupVersion, client client.Client) error {
 
-	// Skip authz when in dev_mode
-	if BACKEND_MODE == "dev" || BACKEND_MODE == "development" {
-		log.Printf("Skipping authorization check in development mode")
-		return nil
-	}
 	// Skip authz when admin is explicity requested it
-	if DISABLE_AUTH {
+	if DISABLE_AUTH == "true" {
 		log.Printf("APP_DISABLE_AUTH set to True. Skipping authorization check")
 		return nil
 	}
--- a/pkg/new-ui/v1beta1/authzn.go
+++ b/pkg/new-ui/v1beta1/authzn.go
@@ -1,0 +1,116 @@
+package v1beta1
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+	"strings"
+
+	"github.com/kubeflow/katib/pkg/util/v1beta1/env"
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/authorization/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// ENV variables
+var (
+	USER_HEADER  = env.GetEnvOrDefault("USERID_HEADER", "kubeflow-userid")
+	USER_PREFIX  = env.GetEnvOrDefault("USERID_PREFIX", ":")
+	DISABLE_AUTH = env.GetEnvOrDefault("APP_DISABLE_AUTH", "false") == "true"
+	BACKEND_MODE = env.GetEnvOrDefault("BACKEND_MODE", "prod")
+)
+
+func GetUsername(r *http.Request) (string, error) {
+	var username string
+	if DISABLE_AUTH {
+		log.Printf("APP_DISABLE_AUTH set to True. Skipping authorization check")
+		return "", nil
+	}
+
+	if r.Header.Get(USER_HEADER) == "" {
+		return "", errors.New("User header not present!")
+	}
+
+	user := r.Header.Get(USER_HEADER)
+	username = strings.Replace(user, USER_PREFIX, "", 1)
+
+	return username, nil
+}
+
+// Function for constructing SubjectAccessReviews (SAR) objects
+func CreateSAR(user string, verb string, namespace string, group string,
+	version string, resource string, subresource string, name string) *v1.SubjectAccessReview {
+
+	sar := &v1.SubjectAccessReview{
+		Spec: v1.SubjectAccessReviewSpec{
+			User: user,
+			ResourceAttributes: &v1.ResourceAttributes{
+				Namespace:   namespace,
+				Verb:        verb,
+				Group:       group,
+				Version:     version,
+				Resource:    resource,
+				Subresource: subresource,
+				Name:        name,
+			},
+		},
+	}
+	return sar
+}
+
+func IsAuthorized(user string, verb string, namespace string, group string,
+	version string, resource string, subresource string, name string, client *kubernetes.Clientset) error {
+
+	// Skip authz when in dev_mode
+	if BACKEND_MODE == "dev" || BACKEND_MODE == "development" {
+		log.Printf("Skipping authorization check in development mode")
+		return nil
+	}
+	// Skip authz when admin is explicity requested it
+	if DISABLE_AUTH {
+		log.Printf("APP_DISABLE_AUTH set to True. Skipping authorization check")
+		return nil
+	}
+
+	sar := CreateSAR(user, verb, namespace, group, version, resource, subresource, name)
+
+	res, err := client.AuthorizationV1().SubjectAccessReviews().Create(context.TODO(), sar, metav1.CreateOptions{})
+	if err != nil {
+		log.Printf("Error submitting SubjectAccessReview: %v, %s", sar, err.Error())
+		return err
+	}
+
+	if res.Status.Allowed {
+		return nil
+	}
+
+	msg := generateUnauthorizedMessage(user, verb, namespace, group, version, resource, subresource, res)
+	err = errors.New(msg)
+	return err
+}
+
+func generateUnauthorizedMessage(user string, verb string, namespace string, group string,
+	version string, resource string, subresource string, sar *v1.SubjectAccessReview) string {
+
+	msg := fmt.Sprintf("User: %s is not authorized to %s", user, verb)
+
+	if group == "" {
+		msg += fmt.Sprintf(" %s/%s", version, resource)
+	} else {
+		msg += fmt.Sprintf(" %s/%s/%s", group, version, resource)
+	}
+
+	if subresource != "" {
+		msg += fmt.Sprintf("/%s", subresource)
+	}
+
+	if namespace != "" {
+		msg += fmt.Sprintf(" in namespace: %s", namespace)
+	}
+	if sar.Status.Reason != "" {
+		msg += fmt.Sprintf(" ,reason: %s", sar.Status.Reason)
+	}
+	return msg
+}
--- a/pkg/new-ui/v1beta1/authzn.go
+++ b/pkg/new-ui/v1beta1/authzn.go
@@ -18,25 +18,8 @@ import (
 var (
 	USER_HEADER  = env.GetEnvOrDefault("USERID_HEADER", "kubeflow-userid")
 	USER_PREFIX  = env.GetEnvOrDefault("USERID_PREFIX", ":")
-	DISABLE_AUTH = env.GetEnvOrDefault("APP_DISABLE_AUTH", "false")
+	DISABLE_AUTH = env.GetEnvOrDefault("APP_DISABLE_AUTH", "true")
 )
-
-func GetUsername(r *http.Request) (string, error) {
-	var username string
-	if DISABLE_AUTH == "true" {
-		log.Printf("APP_DISABLE_AUTH set to True. Skipping authentication check")
-		return "", nil
-	}
-
-	if r.Header.Get(USER_HEADER) == "" {
-		return "", errors.New("user header not present")
-	}
-
-	user := r.Header.Get(USER_HEADER)
-	username = strings.Replace(user, USER_PREFIX, "", 1)
-
-	return username, nil
-}
 
 // Function for constructing SubjectAccessReviews (SAR) objects
 func CreateSAR(user, verb, namespace, resource, subresource, name string, schema schema.GroupVersion) *v1.SubjectAccessReview {
@@ -58,28 +41,34 @@ func CreateSAR(user, verb, namespace, resource, subresource, name string, schema
 	return sar
 }
 
-func IsAuthorized(user, verb, namespace, resource, subresource, name string, schema schema.GroupVersion, client client.Client) error {
+func IsAuthorized(verb, namespace, resource, subresource, name string, schema schema.GroupVersion, client client.Client, r *http.Request) (string, error) {
 
-	// Skip authz when admin is explicity requested it
+	// We disable authn/authz checks when in standalone mode.
 	if DISABLE_AUTH == "true" {
-		log.Printf("APP_DISABLE_AUTH set to True. Skipping authorization check")
-		return nil
+		log.Printf("APP_DISABLE_AUTH set to True. Skipping authentication/authorization checks")
+		return "", nil
 	}
+	// Check if an incoming request is from an authenticated user (kubeflow mode: kubeflow-userid header)
+	if r.Header.Get(USER_HEADER) == "" {
+		return "", errors.New("user header not present")
+	}
+	user := r.Header.Get(USER_HEADER)
+	user = strings.Replace(user, USER_PREFIX, "", 1)
 
+	// Check if the user is authorized to perform a given action on katib/k8s resources.
 	sar := CreateSAR(user, verb, namespace, resource, subresource, name, schema)
-
 	err := client.Create(context.TODO(), sar)
 	if err != nil {
 		log.Printf("Error submitting SubjectAccessReview: %v, %s", sar, err.Error())
-		return err
+		return user, err
 	}
 
 	if sar.Status.Allowed {
-		return nil
+		return user, nil
 	}
 
 	msg := generateUnauthorizedMessage(user, verb, namespace, resource, subresource, schema, sar)
-	return errors.New(msg)
+	return user, errors.New(msg)
 }
 
 func generateUnauthorizedMessage(user, verb, namespace, resource, subresource string, schema schema.GroupVersion, sar *v1.SubjectAccessReview) string {
--- a/pkg/new-ui/v1beta1/authzn.go
+++ b/pkg/new-ui/v1beta1/authzn.go
@@ -40,8 +40,8 @@ func GetUsername(r *http.Request) (string, error) {
 }
 
 // Function for constructing SubjectAccessReviews (SAR) objects
-func CreateSAR(user string, verb string, namespace string, group string,
-	version string, resource string, subresource string, name string) *v1.SubjectAccessReview {
+func CreateSAR(user, verb, namespace, group,
+	version, resource, subresource, name string) *v1.SubjectAccessReview {
 
 	sar := &v1.SubjectAccessReview{
 		Spec: v1.SubjectAccessReviewSpec{
@@ -60,8 +60,8 @@ func CreateSAR(user string, verb string, namespace string, group string,
 	return sar
 }
 
-func IsAuthorized(user string, verb string, namespace string, group string,
-	version string, resource string, subresource string, name string, client *kubernetes.Clientset) error {
+func IsAuthorized(user, verb, namespace, group,
+	version, resource, subresource, name string, client *kubernetes.Clientset) error {
 
 	// Skip authz when in dev_mode
 	if BACKEND_MODE == "dev" || BACKEND_MODE == "development" {
@@ -91,8 +91,8 @@ func IsAuthorized(user string, verb string, namespace string, group string,
 	return err
 }
 
-func generateUnauthorizedMessage(user string, verb string, namespace string, group string,
-	version string, resource string, subresource string, sar *v1.SubjectAccessReview) string {
+func generateUnauthorizedMessage(user, verb, namespace, group,
+	version, resource, subresource string, sar *v1.SubjectAccessReview) string {
 
 	msg := fmt.Sprintf("User: %s is not authorized to %s", user, verb)
 
--- a/pkg/new-ui/v1beta1/backend.go
+++ b/pkg/new-ui/v1beta1/backend.go
@@ -129,7 +129,7 @@ func (k *KatibUIHandler) CreateExperiment(w http.ResponseWriter, r *http.Request
 	}
 }
 
-func (k *KatibUIHandler) FetchNamespacedExperiments(w http.ResponseWriter, r *http.Request) {
+func (k *KatibUIHandler) FetchExperiments(w http.ResponseWriter, r *http.Request) {
 
 	// check if user's username is provided in request Header
 	user, err := GetUsername(r)
--- a/pkg/new-ui/v1beta1/backend.go
+++ b/pkg/new-ui/v1beta1/backend.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	"encoding/json"
+	"errors"
 	"log"
 	"net/http"
 	"path/filepath"
@@ -31,7 +32,6 @@ import (
 	experimentv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/experiments/v1beta1"
 	api_pb_v1beta1 "github.com/kubeflow/katib/pkg/apis/manager/v1beta1"
 	"github.com/kubeflow/katib/pkg/util/v1beta1/katibclient"
-	"github.com/pkg/errors"
 )
 
 func NewKatibUIHandler(dbManagerAddr string) *KatibUIHandler {
@@ -40,8 +40,12 @@ func NewKatibUIHandler(dbManagerAddr string) *KatibUIHandler {
 		log.Printf("NewClient for Katib failed: %v", err)
 		panic(err)
 	}
-	// create a new client-go client for sending SAR objects in the API-SERVER
+	// create a new client for manipulating SAR objects.
 	conf, err := config.GetConfig()
+	if err != nil {
+		log.Printf("Failed to create k8s rest config: %v", err)
+		panic(err)
+	}
 	sarclient, err := kubernetes.NewForConfig(conf)
 	if err != nil {
 		log.Printf("SarClient for Katib failes: %v", err)
@@ -147,7 +151,7 @@ func (k *KatibUIHandler) FetchNamespacedExperiments(w http.ResponseWriter, r *ht
 	namespaces, ok := r.URL.Query()["namespace"]
 	if !ok {
 		log.Printf("No 'namespace' query parameter was provided.")
-		err := errors.New("No 'namespace' query parameter was provided.")
+		err := errors.New("no 'namespace' query parameter was provided")
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
@@ -218,14 +222,14 @@ func (k *KatibUIHandler) DeleteExperiment(w http.ResponseWriter, r *http.Request
 	namespaces, ok := r.URL.Query()["namespace"]
 	if !ok {
 		log.Printf("No 'namespace' query parameter was provided.")
-		err := errors.New("No 'namespace' query parameter was provided.")
+		err := errors.New("no 'namespace' query parameter was provided")
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 	experimentNames, ok := r.URL.Query()["experimentName"]
 	if !ok {
 		log.Printf("No experimentName provided in Query parameteres! Provide an 'experimentName' param")
-		err := errors.New("No experimentName provided!")
+		err := errors.New("no experimentName provided")
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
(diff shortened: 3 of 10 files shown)


johnugeorge:
/assign @kimwnasptd 

Can you review this PR for both standalone and Kubeflow mode of deployment?

andreyvelich:
Thank you so much for these improvements @apo-ger!
Before merging this please can we have discussion at [AutoML WG Community Meeting](https://docs.google.com/document/d/1MChKfzrKAeFRtYqypFbMXL6ZIc_OgijjkvbqmwRV-64/edit#heading=h.o8oe6e5kry87) on **November 16th** ?
In the meantime, @kimwnasptd you can start reviewing this.
 
 cc @akshaychitneni

apo-ger:
@andreyvelich 
Sure, we'll be more than happy to join and go over the details!

apo-ger:
During the review, we took another look at the kubeflow-katib installation manifests and it appears that Istio sidecar injection is disabled for the katib-ui component ([ui.yaml](https://github.com/kubeflow/katib/blob/cdd0b909955350158ac75c87458c2509129e8643/manifests/v1beta1/components/ui/ui.yaml#L19)). The same applies to the other katib components. In addition, we lack the proper AuthorizationPolicies for this PR to work properly, as currently any workloda/pod can set the `kubeflow-userid` header and communicate with the backend of the Katib UI. 

I can prepare a following PR with the minimum required changes in manifests:
- introduce a patch in kubeflow-katib installation manifests to inject an istio sidecar in `katib-ui` Pod.
- introduce an `AuthorizationPolicy` to secure `katib-ui` and allow traffic only from `istio-ingressgateway` workload. (authenticated requests from the UI with the kubeflow-userid header set)
 
  
I could also prepare another public consultation later, which would allow the injection of istio sidecar for the other components of katib and introduce appropriate AuthorizationPolicies to secure them. However, this is orthogonal to this PR.

cc. @kimwnasptd, @andreyvelich, @johnugeorge.

johnugeorge:
Looks good @apo-ger

pkg/new-ui/v1beta1/backend.go
```diff
@@ -225,7 +349,14 @@ func (k *KatibUIHandler) AddTemplate(w http.ResponseWriter, r *http.Request) {
 	updatedConfigMapPath := data["updatedConfigMapPath"].(string)
 	updatedTemplateYaml := data["updatedTemplateYaml"].(string)
 
-	newTemplates, err := k.updateTrialTemplates(updatedConfigMapNamespace, updatedConfigMapName, "", updatedConfigMapPath, updatedTemplateYaml, ActionTypeAdd)
+	err = IsAuthorized(user, "add", updatedConfigMapNamespace, "", "v1", "configmaps", "", updatedConfigMapName, &k.sarClient)
```
apo-ger, line 352: fix verb

pkg/new-ui/v1beta1/backend.go
```diff
@@ -265,7 +404,14 @@ func (k *KatibUIHandler) EditTemplate(w http.ResponseWriter, r *http.Request) {
 	updatedConfigMapPath := data["updatedConfigMapPath"].(string)
 	updatedTemplateYaml := data["updatedTemplateYaml"].(string)
 
-	newTemplates, err := k.updateTrialTemplates(updatedConfigMapNamespace, updatedConfigMapName, configMapPath, updatedConfigMapPath, updatedTemplateYaml, ActionTypeEdit)
+	err = IsAuthorized(user, "edit", updatedConfigMapNamespace, "", "v1", "configmaps", "", updatedConfigMapName, &k.sarClient)
```
apo-ger, line 407: same, fix verb

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+
+	user := r.Header.Get(USER_HEADER)
+	username = strings.Replace(user, USER_PREFIX, "", 1)
+
+	return username, nil
+}
+
+// Function for constructing SubjectAccessReviews (SAR) objects
+func CreateSAR(user string, verb string, namespace string, group string,
+	version string, resource string, subresource string, name string) *v1.SubjectAccessReview {
```
tenzen-y, line 44: Should we remove duplicated variable types like `func CreateSAR(foo, bar string)`?

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+				Subresource: subresource,
+				Name:        name,
+			},
+		},
+	}
+	return sar
+}
+
+func IsAuthorized(user string, verb string, namespace string, group string,
+	version string, resource string, subresource string, name string, client *kubernetes.Clientset) error {
```
tenzen-y, line 64: Should we remove duplicated variable types like `func CreateSAR(foo, bar string)`?

pkg/new-ui/v1beta1/backend.go
```diff
@@ -302,7 +456,14 @@ func (k *KatibUIHandler) DeleteTemplate(w http.ResponseWriter, r *http.Request)
...
 
-	newTemplates, err := k.updateTrialTemplates(updatedConfigMapNamespace, updatedConfigMapName, "", updatedConfigMapPath, "", ActionTypeDelete)
+	err = IsAuthorized(user, "delete", updatedConfigMapNamespace, "", "v1", "configmaps", "", updatedConfigMapName, &k.sarClient)
+	if err != nil {
+		log.Printf("The user: %s is not authorized to delete configmap: %s from namespace: %s \n", user, updatedConfigMapName, updatedConfigMapNamespace)
+		http.Error(w, err.Error(), http.StatusForbidden)
+		return
+	}
+
+	newTemplates, err := k.updateTrialTemplates(updatedConfigMapNamespace, updatedConfigMapName, "", updatedConfigMapPath, "", ActionTypeDelete, user)
```
apo-ger, line 466: @kimwnasptd The updateTrialTemplate() function will first try to retrieve the TrialTemplate ConfigMap and then perform update/delete/create depending on the route (Edit/Add/DeleteTemplate) from which it is called. 

The current implementation will either delete the TrialTemplate ConfigMap if its empty or update the ConfigMap.
Currently we check only if a user has permissions to delete ConfigMaps. Should we move the authorization checks  inside updateTrialTemplate()?

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+	"strings"
+
+	"github.com/kubeflow/katib/pkg/util/v1beta1/env"
+	"github.com/pkg/errors"
```
tenzen-y, line 11: Should we stop using this library since this has been archived?

https://github.com/pkg/errors

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil
+	}
+
+	msg := generateUnauthorizedMessage(user, verb, namespace, group, version, resource, subresource, res)
+	err = errors.New(msg)
+	return err
+}
+
+func generateUnauthorizedMessage(user string, verb string, namespace string, group string,
+	version string, resource string, subresource string, sar *v1.SubjectAccessReview) string {
```
tenzen-y, line 95: Should we remove duplicated variable types like `func CreateSAR(foo, bar string)`?

pkg/new-ui/v1beta1/hp.go
```diff
@@ -27,14 +27,45 @@ import (
 
 	commonv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/common/v1beta1"
 	api_pb_v1beta1 "github.com/kubeflow/katib/pkg/apis/manager/v1beta1"
+	"github.com/pkg/errors"
```
tenzen-y, line 30: Should we stop using this library since this has been archived?

https://github.com/pkg/errors

pkg/new-ui/v1beta1/backend.go
```diff
@@ -37,8 +40,16 @@ func NewKatibUIHandler(dbManagerAddr string) *KatibUIHandler {
 		log.Printf("NewClient for Katib failed: %v", err)
 		panic(err)
 	}
+	// create a new client-go client for sending SAR objects in the API-SERVER
+	conf, err := config.GetConfig()
```
tenzen-y, line 44: Should we handle this error?

tenzen-y (commented):
@apo-ger Thanks for driving this PR!
I left a few comments.

pkg/new-ui/v1beta1/authzn.go
apo-ger: Sure, I'll switch to Go’s built-in errors package

pkg/new-ui/v1beta1/authzn.go
apo-ger: done

pkg/new-ui/v1beta1/authzn.go
apo-ger: fixed

pkg/new-ui/v1beta1/authzn.go
apo-ger: done!

pkg/new-ui/v1beta1/backend.go
apo-ger: fixed!

pkg/new-ui/v1beta1/hp.go
apo-ger: switched to Go's build-in errors package

apo-ger:
> @apo-ger Thanks for driving this PR! I left a few comments.

Thank you for the comments! Fixes are now included

pkg/new-ui/v1beta1/authzn.go
tenzen-y: @apo-ger 
Probably, you forget to change to the standard errors library in this section.

andreyvelich:
> During the review, we took another look at the kubeflow-katib installation manifests and it appears that Istio sidecar injection is currently disabled for the katib-ui component ([ui.yaml](https://github.com/kubeflow/katib/blob/cdd0b909955350158ac75c87458c2509129e8643/manifests/v1beta1/components/ui/ui.yaml#L19)).

@apo-ger Should we remove this annotation from our manifests ? I guess, Katib UI should work with istio sidecar injected, right ?
What about our DB Manager: https://github.com/kubeflow/katib/blob/cdd0b909955350158ac75c87458c2509129e8643/manifests/v1beta1/components/db-manager/db-manager.yaml#L19 ? Will gRPC communication works with injected sidecar ?

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+	version, resource, subresource, name string) *v1.SubjectAccessReview {
+
+	sar := &v1.SubjectAccessReview{
+		Spec: v1.SubjectAccessReviewSpec{
+			User: user,
+			ResourceAttributes: &v1.ResourceAttributes{
+				Namespace:   namespace,
+				Verb:        verb,
+				Group:       group,
+				Version:     version,
```
andreyvelich, line 53: Similar comment for `group` and `version`, we can just use constants here.

pkg/new-ui/v1beta1/backend.go
```diff
@@ -37,8 +40,20 @@ func NewKatibUIHandler(dbManagerAddr string) *KatibUIHandler {
...
+	if err != nil {
+		log.Printf("Failed to create k8s rest config: %v", err)
+		panic(err)
+	}
+	sarclient, err := kubernetes.NewForConfig(conf)
+	if err != nil {
+		log.Printf("SarClient for Katib failes: %v", err)
+		panic(err)
+	}
```
andreyvelich, line 53: Do you need to create new K8s client here ?
You can get controller runtime client from the Katib client using [`GetClient`](https://github.com/kubeflow/katib/blob/0377ed3932502e658e22a1a553f82db19280bab5/pkg/util/v1beta1/katibclient/katib_client.go#L36) API. You can use this client to create SAR object.

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+
+	if res.Status.Allowed {
+		return nil
+	}
+
+	msg := generateUnauthorizedMessage(user, verb, namespace, group, version, resource, subresource, res)
+	err = errors.New(msg)
+	return err
```
andreyvelich, line 91: I would simplify this to just this:
```suggestion
	return errors.New(msg)
```

pkg/new-ui/v1beta1/backend.go
```diff
@@ -96,6 +120,16 @@ func (k *KatibUIHandler) CreateExperiment(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	namespace := job.ObjectMeta.Namespace
+	expName := job.ObjectMeta.Name
+
+	err = IsAuthorized(user, "create", namespace, "kubeflow.org", "v1beta1", "experiments", "", "", &k.sarClient)
+	if err != nil {
+		log.Printf("The user: %s is not authorized to create experiment: %s from namespace: %s \n", user, expName, namespace)
```
andreyvelich, line 128:
```suggestion
		log.Printf("The user: %s is not authorized to create experiment: %s in namespace: %s \n", user, expName, namespace)
```

andreyvelich (commented):
I left few comments.

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		log.Printf("Error submitting SubjectAccessReview: %v, %s", sar, err.Error())
+		return err
+	}
+
+	if res.Status.Allowed {
+		return nil
+	}
+
+	msg := generateUnauthorizedMessage(user, verb, namespace, group, version, resource, subresource, res)
```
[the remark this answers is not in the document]
apo-ger: The `SubjectAccessReview.reason` field returns a message when a request is allowed. for example:
>Reason: RBAC: allowed by RoleBinding "namespaceAdmin/kubeflow-user-example-com" of ClusterRole "kubeflow-admin" to User "user@example.com"

However its empty when the request is denied (`SubjectAccessReview.Status.allowed=false`).

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+				Subresource: subresource,
+				Name:        name,
+			},
+		},
+	}
+	return sar
+}
+
+func IsAuthorized(user, verb, namespace, group,
+	version, resource, subresource, name string, client *kubernetes.Clientset) error {
```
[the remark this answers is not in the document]
apo-ger: Well, we also perform authorization checks for ConfigMap resources (group: "", version: "v1"). Thus we need a way to set these values accordingly.

apo-ger:
@andreyvelich 
 > @apo-ger Should we remove this annotation from our manifests ? I guess, Katib UI should work with istio sidecar injected, right ? 

Yes, removing the annotation will do. 


>What about our DB Manager:
> 
> https://github.com/kubeflow/katib/blob/cdd0b909955350158ac75c87458c2509129e8643/manifests/v1beta1/components/db-manager/db-manager.yaml#L19
> 
> ? Will gRPC communication works with injected sidecar ?

Yes, gRPC communication will work with injected sidecar as long as we provide the appropriate ALLOW AuthorizationPolicy when in kubeflow mode because by default we deny all traffic in the mesh:
https://github.com/kubeflow/manifests/blob/master/common/istio-1-14/istio-install/base/deny_all_authorizationpolicy.yaml.

pkg/new-ui/v1beta1/authzn.go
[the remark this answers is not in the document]
tenzen-y: Maybe, you can get group and version using the type `schema.GroupVersion` in the following:

```go

func Foo() {
  err := IsAuthorized(user, "create", namespace, experimentv1beta1.SchemeGroupVersion, "experiments", "", "", &k.sarClient)
  err := IsAuthorized(user, "create", namespace, corev1.SchemeGroupVersion, "configmap", "", "", &k.sarClient)
  ...
}

func IsAuthorized(user, verb, namespace, schema schema.GroupVersion, resource, subresource, name string, client *kubernetes.Clientset) error {

  sar := CreateSAR(user, verb, namespace, schema schema.GroupVersion, resource, subresource, name)
...
}

func CreateSAR(user, verb, namespace, schema schema.GroupVersion, resource, subresource, name string) *v1.SubjectAccessReview {

	sar := &v1.SubjectAccessReview{
		Spec: v1.SubjectAccessReviewSpec{
			User: user,
			ResourceAttributes: &v1.ResourceAttributes{
				Namespace:   namespace,
				Verb:        verb,
				Group:       schema.Group(),
				Version:     schema.Version(),
...
}
```

pkg/new-ui/v1beta1/backend.go
apo-ger: Fixed!

pkg/new-ui/v1beta1/backend.go
apo-ger: fixed

pkg/new-ui/v1beta1/authzn.go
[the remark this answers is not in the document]
andreyvelich: Got it, thanks @apo-ger.
I like @tenzen-y solution, let's utilise constants as much as we can.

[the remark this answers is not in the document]
apo-ger: Done!

pkg/new-ui/v1beta1/authzn.go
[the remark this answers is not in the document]
andreyvelich: Make sense, thank you!

pkg/new-ui/v1beta1/authzn.go
apo-ger: Fixed!

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// ENV variables
+var (
+	USER_HEADER  = env.GetEnvOrDefault("USERID_HEADER", "kubeflow-userid")
+	USER_PREFIX  = env.GetEnvOrDefault("USERID_PREFIX", ":")
+	DISABLE_AUTH = env.GetEnvOrDefault("APP_DISABLE_AUTH", "false") == "true"
+	BACKEND_MODE = env.GetEnvOrDefault("BACKEND_MODE", "prod")
```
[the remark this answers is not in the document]
apo-ger: Not really, I'll drop the BACKEND_MODE ENV var

cmd/new-ui/v1beta1/main.go
```diff
@@ -48,7 +48,7 @@ func main() {
 	http.HandleFunc("/katib/", kuh.ServeIndex(*buildDir))
 	http.Handle("/katib/static/", http.StripPrefix("/katib/", frontend))
 
-	http.HandleFunc("/katib/fetch_experiments/", kuh.FetchAllExperiments)
+	http.HandleFunc("/katib/fetch_namespaced_experiments/", kuh.FetchNamespacedExperiments)
```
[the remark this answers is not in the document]
apo-ger: done :)

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,116 @@
...
+		log.Printf("APP_DISABLE_AUTH set to True. Skipping authorization check")
+		return "", nil
+	}
+
+	if r.Header.Get(USER_HEADER) == "" {
+		return "", errors.New("user header not present")
+	}
+
+	user := r.Header.Get(USER_HEADER)
+	username = strings.Replace(user, USER_PREFIX, "", 1)
```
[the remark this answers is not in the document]
apo-ger: This is related to https://github.com/kubeflow/kubeflow/issues/3553.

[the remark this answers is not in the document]
andreyvelich: @apo-ger @kimwnasptd Does it mean that for SAR we always should remove prefix from `kubeflow-userid` HEADER ?
I guess, prefix is set by default in GCP, AWS environmnets, right ?

pkg/new-ui/v1beta1/backend.go
```diff
@@ -104,6 +138,53 @@ func (k *KatibUIHandler) CreateExperiment(w http.ResponseWriter, r *http.Request
...
+	if _, err = w.Write(response); err != nil {
+		log.Printf("Write experiments failed: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+}
+
 // FetchAllExperiments gets HP and NAS experiments in all namespaces.
 func (k *KatibUIHandler) FetchAllExperiments(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
kimwnasptd: @andreyvelich right now the Katib web app does not have an `All namespaces` option, but we'll soon include this functionality as part of https://github.com/kubeflow/kubeflow/blob/master/components/proposals/20220621-all-namespaces.md.

But even once we'll introduce this the frontend will be making requests to each namespace, when the `All namespaces` option will be selected. So I believe we can remove this API endpoint.

apo-ger:
@andreyvelich Thank you for your comments. Let me know if everything is okay

Now, regarding the manifests changes: 

1. IIUC I should enable istio sidecar injection for both **katib-ui**  and **DB manager** pods, right?
    - by removing the annotation: https://github.com/kubeflow/katib/blob/cdd0b909955350158ac75c87458c2509129e8643/manifests/v1beta1/components/db-manager/db-manager.yaml#L19
2. I'll create a follow-up PR as soon as this PR is merged.

pkg/new-ui/v1beta1/backend.go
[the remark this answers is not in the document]
andreyvelich: Sounds good, we can remove it then.

andreyvelich:
@apo-ger I am not sure if we should enable istio for DB Manager. Will it block gRPC traffic that call APIs:
 Experiment Controller <-> Suggestion Service, Trial Metrics Collector <-> Katib DB Manager ?

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,106 @@
...
+	v1 "k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ENV variables
+var (
+	USER_HEADER  = env.GetEnvOrDefault("USERID_HEADER", "kubeflow-userid")
+	USER_PREFIX  = env.GetEnvOrDefault("USERID_PREFIX", ":")
+	DISABLE_AUTH = env.GetEnvOrDefault("APP_DISABLE_AUTH", "false")
```
andreyvelich, line 21: Let's keep `DISABLE_AUTH` by default as `true`.
Since currently Katib doesn't support it in standalone mode, and we should update all of our [Installs](https://github.com/kubeflow/katib/tree/master/manifests/v1beta1/installs).
For [Katib with Kubeflow install](https://github.com/kubeflow/katib/tree/master/manifests/v1beta1/installs/katib-with-kubeflow) we can make appropriate patch for the Katib UI manifests.

WDYT @kimwnasptd @apo-ger ?

pkg/new-ui/v1beta1/backend.go
```diff
@@ -37,8 +40,11 @@ func NewKatibUIHandler(dbManagerAddr string) *KatibUIHandler {
 		log.Printf("NewClient for Katib failed: %v", err)
 		panic(err)
 	}
+	sarclient := kclient.GetClient()
+
 	return &KatibUIHandler{
 		katibClient:   kclient,
+		sarClient:     sarclient,
 		dbManagerAddr: dbManagerAddr,
 	}
```
andreyvelich, line 49: Do you need `sarClient` parameter in `KatibUIHandler` ?
I think you can pass K8s client to `IsAuthorized` as follows:
```go
err=  IsAuthorized(user, "create", namespace, "kubeflow.org", "", "", experimentv1beta1.SchemeGroupVersion, k.GetClient())
```

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+func GetUsername(r *http.Request) (string, error) {
+	var username string
+	if DISABLE_AUTH == "true" {
+		log.Printf("APP_DISABLE_AUTH set to True. Skipping authentication check")
+		return "", nil
+	}
```
andreyvelich, line 29: Can we avoid duplicate check for [`if DISABLE_AUTH == "true"`](https://github.com/kubeflow/katib/blob/9e283983cb26a4bb318d2d86d7795f3b2b4bca2e/pkg/new-ui/v1beta1/authzn.go#L64-L68) by adding `GetUsername` logic under `IsAuthorized` call ?
Then, we have 1 place where we verify if auth is required.
WDYT @apo-ger ?

pkg/new-ui/v1beta1/backend.go
```diff
@@ -96,6 +111,16 @@ func (k *KatibUIHandler) CreateExperiment(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	namespace := job.ObjectMeta.Namespace
+	expName := job.ObjectMeta.Name
+
+	err = IsAuthorized(user, "create", namespace, "kubeflow.org", "", "", experimentv1beta1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 117: I guess, here you should have `experiments` instead of `kubeflow.org`, right ?

pkg/new-ui/v1beta1/backend.go
```diff
@@ -104,6 +129,53 @@ func (k *KatibUIHandler) CreateExperiment(w http.ResponseWriter, r *http.Request
...
+	if !ok {
+		log.Printf("No 'namespace' query parameter was provided.")
+		err := errors.New("no 'namespace' query parameter was provided")
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	namespace := namespaces[0]
+
+	err = IsAuthorized(user, "list", namespace, "experiments", "", "", experimentv1beta1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 152: @apo-ger Please can we add plural for `experiments`, `trials`, and `suggestions` to [constants](https://github.com/kubeflow/katib/blob/9e283983cb26a4bb318d2d86d7795f3b2b4bca2e/pkg/controller.v1beta1/consts/const.go#L29-L30) and use it here ?

pkg/new-ui/v1beta1/backend.go
```diff
@@ -225,7 +344,14 @@ func (k *KatibUIHandler) AddTemplate(w http.ResponseWriter, r *http.Request) {
 	updatedConfigMapPath := data["updatedConfigMapPath"].(string)
 	updatedTemplateYaml := data["updatedTemplateYaml"].(string)
 
-	newTemplates, err := k.updateTrialTemplates(updatedConfigMapNamespace, updatedConfigMapName, "", updatedConfigMapPath, updatedTemplateYaml, ActionTypeAdd)
+	err = IsAuthorized(user, "create", updatedConfigMapNamespace, "configmaps", "", updatedConfigMapName, corev1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 347: `configmaps` constant [can be used from K8s API](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L6087).

pkg/new-ui/v1beta1/backend.go
andreyvelich, line 152: Also, please if K8s APIs doesn't have constants for CRUD actions (e.g. `list`). Let's add them to our [constants](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/consts/const.go).

pkg/new-ui/v1beta1/backend.go
```diff
@@ -130,15 +202,45 @@ func (k *KatibUIHandler) FetchAllExperiments(w http.ResponseWriter, r *http.Requ
...
+	if !ok {
+		log.Printf("No experimentName provided in Query parameteres! Provide an 'experimentName' param")
+		err := errors.New("no experimentName provided")
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	experimentName := experimentNames[0]
+	namespace := namespaces[0]
+
+	err = IsAuthorized(user, "delete", namespace, "experiments", "", "", experimentv1beta1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 230: Similar comment about constants.

pkg/new-ui/v1beta1/backend.go
```diff
@@ -265,7 +399,14 @@ func (k *KatibUIHandler) EditTemplate(w http.ResponseWriter, r *http.Request) {
 	updatedConfigMapPath := data["updatedConfigMapPath"].(string)
 	updatedTemplateYaml := data["updatedTemplateYaml"].(string)
 
-	newTemplates, err := k.updateTrialTemplates(updatedConfigMapNamespace, updatedConfigMapName, configMapPath, updatedConfigMapPath, updatedTemplateYaml, ActionTypeEdit)
+	err = IsAuthorized(user, "update", updatedConfigMapNamespace, "configmaps", "", updatedConfigMapName, corev1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 402: Similar comment about constants.

pkg/new-ui/v1beta1/backend.go
```diff
@@ -302,7 +451,14 @@ func (k *KatibUIHandler) DeleteTemplate(w http.ResponseWriter, r *http.Request)
 	updatedConfigMapName := data["updatedConfigMapName"].(string)
 	updatedConfigMapPath := data["updatedConfigMapPath"].(string)
 
-	newTemplates, err := k.updateTrialTemplates(updatedConfigMapNamespace, updatedConfigMapName, "", updatedConfigMapPath, "", ActionTypeDelete)
+	err = IsAuthorized(user, "delete", updatedConfigMapNamespace, "configmaps", "", updatedConfigMapName, corev1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 454: Similar comment about constants.

pkg/new-ui/v1beta1/backend.go
```diff
@@ -351,8 +507,37 @@ func (k *KatibUIHandler) FetchNamespaces(w http.ResponseWriter, r *http.Request)
...
+		log.Printf("No experimentName provided in Query parameteres! Provide an 'experimentName' param")
+		err := errors.New("no experimentName provided")
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	experimentName := experimentNames[0]
+	namespace := namespaces[0]
+
+	err = IsAuthorized(user, "get", namespace, "experiments", "", experimentName, experimentv1beta1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 535: Similar comment about constants.

pkg/new-ui/v1beta1/backend.go
```diff
@@ -375,8 +560,37 @@ func (k *KatibUIHandler) FetchExperiment(w http.ResponseWriter, r *http.Request)
...
+		log.Printf("No experimentName provided in Query parameteres! Provide an 'experimentName' param")
+		err := errors.New("no experimentName provided")
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	suggestionName := suggestionNames[0]
+	namespace := namespaces[0]
+
+	err = IsAuthorized(user, "get", namespace, "suggestions", "", suggestionName, suggestionv1beta1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 588: Similar comment about constants.

pkg/new-ui/v1beta1/hp.go
```diff
@@ -19,22 +19,55 @@ package v1beta1
...
+	if !ok {
+		log.Printf("No experimentName provided in Query parameteres! Provide an 'experimentName' param")
+		err := errors.New("no experimentName provided")
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	experimentName := experimentNames[0]
+	namespace := namespaces[0]
+
+	err = IsAuthorized(user, "get", namespace, "experiments", "", experimentName, experimentv1beta1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 64: Similar comment about constants.

pkg/new-ui/v1beta1/hp.go
```diff
@@ -65,6 +98,14 @@ func (k *KatibUIHandler) FetchHPJobInfo(w http.ResponseWriter, r *http.Request)
 	}
 	log.Printf("Got Parameters names")
 
+	err = IsAuthorized(user, "list", namespace, "trials", "", "", trialv1beta1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 101: Similar comment about constants.

pkg/new-ui/v1beta1/hp.go
```diff
@@ -151,16 +192,50 @@ func (k *KatibUIHandler) FetchHPJobInfo(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+
+	trialName := trialNames[0]
+	namespace := namespaces[0]
+
 	conn, c := k.connectManager()
 	defer conn.Close()
 
+	err = IsAuthorized(user, "list", namespace, "trials", "", trialName, trialv1beta1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 225: Similar comment about constants.

pkg/new-ui/v1beta1/util.go
```diff
@@ -94,8 +94,30 @@ func (k *KatibUIHandler) getTrialTemplatesViewList() ([]TrialTemplatesDataView,
...
+		// configmaps that belong to namespaces other than kubeflow and kubeflow-user.
+		var newTrialTemplatesConfigMapList []apiv1.ConfigMap
+		for _, cmap := range trialTemplatesConfigMapList.Items {
+			if ns == "kubeflow" {
+				// Add the configmaps from kubeflow namespace (no user can have access to kubeflow ns, so we hardcode it)
+				newTrialTemplatesConfigMapList = append(newTrialTemplatesConfigMapList, cmap)
+			} else {
+				// for all other namespaces check authorization rbac
+				configmapName := cmap.ObjectMeta.Name
+				err = IsAuthorized(user, "get", ns, "", "", configmapName, apiv1.SchemeGroupVersion, k.sarClient)
```
andreyvelich, line 107: Similar comment about constants.

johnugeorge:
> @apo-ger I am not sure if we should enable istio for DB Manager. Will it block gRPC traffic that call APIs: Experiment Controller <-> Suggestion Service, Trial Metrics Collector <-> Katib DB Manager ?
> 
> cc @johnugeorge @kimwnasptd @tenzen-y

Note: We have to test it carefully before any side car changes.  The reason is that we don't have any tests in Kubeflow mode. So, manual testing is the only option

pkg/new-ui/v1beta1/backend.go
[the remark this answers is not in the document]
apo-ger: Removed this API endpoint!

pkg/new-ui/v1beta1/authzn.go
apo-ger: Makes sense! Fixed!

pkg/new-ui/v1beta1/backend.go
apo-ger: Valid point. Fixed!

pkg/new-ui/v1beta1/backend.go
apo-ger: Oh, I missed this. Fixed!

pkg/new-ui/v1beta1/backend.go
apo-ger: Done!

pkg/new-ui/v1beta1/backend.go
apo-ger: Done!

pkg/new-ui/v1beta1/backend.go
apo-ger: Done!

pkg/new-ui/v1beta1/backend.go
apo-ger: Done!

pkg/new-ui/v1beta1/backend.go
apo-ger: fixed!

pkg/new-ui/v1beta1/backend.go
apo-ger: fixed!

pkg/new-ui/v1beta1/backend.go
apo-ger: Fixed!

pkg/new-ui/v1beta1/backend.go
apo-ger: fixed!

pkg/new-ui/v1beta1/hp.go
apo-ger: fixed!

pkg/new-ui/v1beta1/hp.go
apo-ger: done!

pkg/new-ui/v1beta1/util.go
apo-ger: fixed!

pkg/new-ui/v1beta1/authzn.go
apo-ger: Sure, I moved the `GetUsername` logic under `IsAuthorized` and introduced the appropriate checks to handle the various errors properly!

pkg/new-ui/v1beta1/util.go
```diff
@@ -94,8 +95,31 @@ func (k *KatibUIHandler) getTrialTemplatesViewList() ([]TrialTemplatesDataView,
...
+				// for all other namespaces check authorization rbac
+				configmapName := cmap.ObjectMeta.Name
+				user, err := IsAuthorized(consts.ActionTypeGet, ns, apiv1.ResourceConfigMaps.String(), "", configmapName, apiv1.SchemeGroupVersion, k.katibClient.GetClient(), r)
+				if err != nil {
+					log.Printf("The user: %s is not authorized to view configMap: %s in namespace: %s \n", user, configmapName, ns)
+					return nil, err
+				} else {
+					log.Printf("The user: %s is authorized to view configMap: %s in namespace: %s", user, configmapName, ns)
+					newTrialTemplatesConfigMapList = append(newTrialTemplatesConfigMapList, cmap)
+				}
```
andreyvelich, line 116: @apo-ger Maybe we should make the `IsAuthorized` check before we are [getting Trial templates](https://github.com/kubeflow/katib/blob/ae941c4cc66747a7e0db363ec7c9e1f877a1b9c9/pkg/new-ui/v1beta1/util.go#L92) ?
Then, you don't need to have additional condition for `kubeflow` namespace.

apo-ger: IIUC the current implementation of the UI only allows users to view the available `TrialTemplates` ([/katib/fetch_trial_templates/](https://github.com/apo-ger/katib/blob/ae941c4cc66747a7e0db363ec7c9e1f877a1b9c9/pkg/new-ui/v1beta1/backend.go#L256)), right?

andreyvelich: I see, makes sense @apo-ger.
> IIUC the current implementation of the UI only allows users to view the available TrialTemplates ([/katib/fetch_trial_templates/](https://github.com/apo-ger/katib/blob/ae941c4cc66747a7e0db363ec7c9e1f877a1b9c9/pkg/new-ui/v1beta1/backend.go#L256)), right?

We have API to create/delete template. For example: https://github.com/apo-ger/katib/blob/ae941c4cc66747a7e0db363ec7c9e1f877a1b9c9/pkg/new-ui/v1beta1/backend.go#L280.
The old UI has feature to CRUD Trial templates, but the new UI still don't support it, tracking here: https://github.com/kubeflow/katib/issues/1440. cc @kimwnasptd 
We might need to have discussion around this. Should we show Trial templates in `kubeflow` namespace or just avoid it.

pkg/new-ui/v1beta1/util.go
```diff
@@ -94,8 +95,31 @@ func (k *KatibUIHandler) getTrialTemplatesViewList() ([]TrialTemplatesDataView,
 			return nil, err
 		}
 
+		// Iterate over the trialTemplatesConfigMapList from all namespaces and filter out the
+		// configmaps that belong to namespaces other than kubeflow and the ones that the user has
+		// access privileges.
+		var newTrialTemplatesConfigMapList []apiv1.ConfigMap
+		for _, cmap := range trialTemplatesConfigMapList.Items {
+			if ns == "kubeflow" {
```
andreyvelich, line 103: Can you use this `const` variable please for `kubeflow` namespace: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/consts/const.go#L175 ?

andreyvelich:
Thanks a lot for implementing this @apo-ger!
/lgtm
/assign @kimwnasptd @johnugeorge @tenzen-y

pkg/new-ui/v1beta1/util.go
kimwnasptd: > We might need to have discussion around this. Should we show Trial templates in kubeflow namespace or just avoid it.

+1 on having a discussion around this. Although I think the current approach makes sense:
1. An admin adds TrialTemplate (ConfigMap) in the `kubeflow` namespace. These are *visible* from all users
    * But the backend won't allow users to edit/delete ConfigMaps in the `kubeflow` namespace. For these actions there are SubjectAccessReview checks in place. 
    * If an admin wants to allow a user to manipulate the global (`kubeflow` namespace) ConfigMaps then they just need to give them RBAC permissions
2. Users will then see/interact-with ConfigMaps in other namespaces that they have permissions for

kimwnasptd:
The current code looks good from my side. @andreyvelich thank you for the thorough review!

/lgtm

tenzen-y (commented):
@apo-ger Thanks a lot!

/lgtm

kimwnasptd:
🎉 🎉 

/approve

andreyvelich:
/approve
@apo-ger Feel free to unhold the PR when you are ready.

apo-ger:
:tada:
/unhold

PR #1990, opened 2022-11-01
[kwa-show-status-first] Show the trials table's status column first

elenzio9:
In this PR, we move the status column to the first position of the trials table as it is in the other tables.

Here's a screenshot:
image

Signed-off-by: Elena Zioga <elena@arrikto.com>


johnugeorge:
Can you rebase?

PR #1992, opened 2022-11-01
Suggestion volume resumePolicy override

### LINKED ISSUE #1893: Separate suggestion volume reconcile from resumePolicy
/kind feature

Volumes used by Suggestion services (e.g. PBT) currently rely on `resumePolicy` to be `FromVolume` to run properly. As the Suggestion volumes always need to be mounted, it should be independent of the `resumePolicy` field.

Discussion around requested change: https://github.com/kubeflow/katib/pull/1833#discussion_r890123411

---

a9p:

PBT currently relies on `resumePolicy` to be `FromVolume` to run properly. The proposed changes follow up on [previous discussion](https://github.com/kubeflow/katib/pull/1833#discussion_r890123411) on removing the requirement for users specifying this field and instead handling the logic automatically for suggestions that require it.

Fixes #1893

**Checklist:**


a9p:
cc/ @andreyvelich was this closer to what you were looking for with your original comment?

pkg/controller.v1beta1/suggestion/suggestion_controller.go
```diff
@@ -47,6 +47,7 @@ import (
 const (
 	ControllerName = "suggestion-controller"
 )
+var ReconcileSuggestionVolumeOverrides = [...] string {"pbt"}
```
tenzen-y, line 50: I think that we can read `ReconcileSuggestionVolumeOverrides` from katib-config so that the users can use this feature for custom suggestion services.
But, in this PR, it would be good to use hardcoded suggestions as the first step for this new feature.

cc: @andreyvelich @johnugeorge @gaocegege

tenzen-y (commented):
@a9p Thanks for driving this!
I left a few comments.

pkg/controller.v1beta1/suggestion/suggestion_controller.go
a9p: Are you thinking we mount the volumes when `persistentVolumeClaimSpec` exists (and is not `{}`) in the katib-config?

tenzen-y: I think we can add a new field to katib-config, not `persistentVolumeClaimSpec` field.

tenzen-y:
@a9p please get in touch with me when this PR is ready for review.

a9p:
@tenzen-y thanks for the suggestion! I think this PR is ready for your review now -- let me know if you have any other comments!

pkg/controller.v1beta1/suggestion/suggestion_controller.go
```diff
@@ -191,8 +192,14 @@ func (r *ReconcileSuggestion) ReconcileSuggestion(instance *suggestionsv1beta1.S
...
-	// If ResumePolicy = FromVolume volume is reconciled for suggestion
-	if instance.Spec.ResumePolicy == experimentsv1beta1.FromVolume {
+	// Check if algorithm overrides volume reconcile policy
+	suggestionConfigData, err := katibconfig.GetSuggestionConfigData(instance.Spec.Algorithm.AlgorithmName, r.Client)
+	if err != nil {
+		return err
+	}
+
+	// If ResumePolicy = FromVolume (or overriden), volume is reconciled for suggestion
+	if suggestionConfigData.VolumeForceMount || instance.Spec.ResumePolicy == experimentsv1beta1.FromVolume {
```
tenzen-y, line 202: Can you add the test for this logic?

tenzen-y (commented):
@a9p Thanks for updating this PR! This feature looks good to me.

pkg/util/v1beta1/katibconfig/config.go
```diff
@@ -42,6 +42,7 @@ type SuggestionConfig struct {
 	PersistentVolumeClaimSpec corev1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
 	PersistentVolumeSpec      corev1.PersistentVolumeSpec      `json:"persistentVolumeSpec,omitempty"`
 	PersistentVolumeLabels    map[string]string                `json:"persistentVolumeLabels,omitempty"`
+	VolumeForceMount          bool                             `json:"volumeForceMount,omitempty"`
```
andreyvelich, line 45: Can we avoid additional `VolumeForceMount` parameter in config ?

One suggestion that I have: Always [get  desired volume](https://github.com/kubeflow/katib/blob/3f1eba01d6f6b1520d81b7c71dbfb57c143b62d9/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L202) if Suggestion Config Data contains Persistent Volume Spec or if Resume Policy = From Volume. 
The default values for PVC, we can set under [the Desired Volume](https://github.com/kubeflow/katib/blob/3f1eba01d6f6b1520d81b7c71dbfb57c143b62d9/pkg/controller.v1beta1/suggestion/composer/composer.go#L260) function.

I understand that it doesn't solve the problem when user wants to use the default values for Suggestion volume and still attach the volume, but we can find the better solution.

Any other ideas how to avoid more parameters in Katib Config @a9p @tenzen-y @johnugeorge ?

pkg/controller.v1beta1/suggestion/suggestion_controller.go
andreyvelich, line 202: I think we need to identify more places where we check if Suggestion has volume and we need to do something.
e.g. Controller deletes volume when Experiment is finished: https://github.com/kubeflow/katib/blob/3f1eba01d6f6b1520d81b7c71dbfb57c143b62d9/pkg/controller.v1beta1/experiment/experiment_controller_util.go#L168


Maybe we can update [the Suggestion CR's API with `volumeAttached: bool` in addition to `resumePolicy`](https://github.com/kubeflow/katib/blob/3f1eba01d6f6b1520d81b7c71dbfb57c143b62d9/pkg/apis/controller/suggestions/v1beta1/suggestion_types.go#L41).
So we can do appropriate changes based on this spec setting.
We [are creating Suggestion CR](https://github.com/kubeflow/katib/blob/3f1eba01d6f6b1520d81b7c71dbfb57c143b62d9/pkg/controller.v1beta1/experiment/suggestion/suggestion.go#L73) before reconcile Suggestion, so [you don't need to get Katib Config](https://github.com/kubeflow/katib/blob/3f1eba01d6f6b1520d81b7c71dbfb57c143b62d9/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L196) before reconcile Suggestion volume.

What do you think about this API change @tenzen-y @johnugeorge @a9p ?

andreyvelich (commented):
Thanks for this @a9p.
I left few suggestions.

pkg/util/v1beta1/katibconfig/config.go
tenzen-y: > Always [get desired volume](https://github.com/kubeflow/katib/blob/3f1eba01d6f6b1520d81b7c71dbfb57c143b62d9/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L202) if Suggestion Config Data contains Persistent Volume Spec or if Resume Policy = From Volume.
The default values for PVC, we can set under [the Desired Volume](https://github.com/kubeflow/katib/blob/3f1eba01d6f6b1520d81b7c71dbfb57c143b62d9/pkg/controller.v1beta1/suggestion/composer/composer.go#L260) function.

Maybe, it's idea is better to avoid more parameters in katib-config.

> Any other ideas how to avoid more parameters in Katib Config

I have no good ideas.

pkg/controller.v1beta1/suggestion/suggestion_controller.go
tenzen-y: What are the conditions under which katib-controller sets `volumeAttached` to true? Do you plan to do only when `Spec.ResumePolicy == experimentsv1beta1.FromVolume`? Or also check whether `status.phase == Bound` and so on in PVC?

pkg/util/v1beta1/katibconfig/config.go
a9p: I agree with removing the new parameter 👍🏽

I'm not sure what a good default is for the mounts or if there is one -- it seems like that would be setup dependent. One thought would be to prevent the suggestion service from starting if the mount doesn't exist in the config (making the config issue apparent up front) for services that require one? Currently, PBT for example, basically does a random search at the initial seed until run limits are hit which is a bit opaque unless you read the documentation for the config needed to operate correctly.

pkg/controller.v1beta1/suggestion/suggestion_controller.go
a9p: One more case where it would be true is where the service needs a volume independent of ResumePolicy (this is what I think the original issue was pointing at separating)

pkg/util/v1beta1/katibconfig/config.go
andreyvelich: >One thought would be to prevent the suggestion service from starting if the mount doesn't exist in the config (making the config issue apparent up front) for services that require one?

@a9p How we are going to understand for which Suggestion service the Volume is mandatory ?

a9p: I was originally thinking something like [this](https://github.com/kubeflow/katib/blob/1e7e3c16fed97c3d3bb1c4e5fd48bae5ebd9401f/pkg/controller.v1beta1/suggestion/suggestion_controller.go#L50), but hardcoding this in the suggestion controller isn't ideal. Possibly linking the required resources to the suggestion service and reading that in at the controller (e.g. an api change to the rpc endpoint implemented by the suggestion defaulting to no additional requirements)?

andreyvelich: @a9p Or maybe we can use [`ValidateAlgorithmSettings` gRPC API](https://github.com/kubeflow/katib/blob/1e7e3c16fed97c3d3bb1c4e5fd48bae5ebd9401f/pkg/apis/manager/v1beta1/api.proto#L36) ?
In addition to the Experiment's Algorithm Setting we can pass information (with appropriate algorithm name) from the Katib config to the Suggestion service ?
Then, service can verify if settings are correct and user will see the results on the Experiment status.

What do you think @a9p @johnugeorge @tenzen-y @gaocegege ?

a9p: Does something like the following match what you were thinking:
```
message AlgorithmSpec {
    string algorithm_name = 1;
    repeated AlgorithmSetting algorithm_settings = 2;
    repeated <ServiceSetting> <service>_settings = 3; // ServiceSetting, SuggestionSetting, etc. 
}
```

tenzen-y: @a9p @andreyvelich Looks good about expanding gRPC API.

> Any ideas how we should define SuggestionSetting API in our gRPC since we might want to send all [SuggestionConfig parameters to the Suggestion settings (which is Deployment spec)](https://github.com/kubeflow/katib/blob/master/pkg/util/v1beta1/katibconfig/config.go#L36)?

Maybe, to use all Deployment fields, we need to hold the [kubernetes/api](https://github.com/kubernetes/api) repository as git submodules and import deployment spec gRPC API to our gRPC API.

And then, we will need to manage to support K8s versions more carefully since the above means that our gRPC API intensely depends on K8s API.

@andreyvelich @a9p wdyt?

andreyvelich: Sounds great @tenzen-y , so should we [import the Deployment spec from the proto file](https://github.com/kubernetes/api/blob/master/apps/v1/generated.proto#L268) and make the following API structure:
```proto
message AlgorithmSpec {
    string algorithm_name = 1;
    repeated AlgorithmSetting algorithm_settings = 2;
    DeploymentSpec suggestion_spec = 3;
}
```
> And then, we will need to more carefully manage K8s versions supported by katib since the above means that our gRPC API intensely depends on K8s API.

I think we already use the Kubernetes API for our CRD API, so we already have to be compatible with the version (https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/common/v1beta1/common_types.go#L233).
cc @johnugeorge

tenzen-y: sgtm

a9p: Across most of the repo where the k8s spec is used, it is used in go via imports -- is git submoduling in the manager/v1beta1 directory the preferred approach? I put a temporary shim in the build.sh tying it to the minor k8s version just to verify static checks, but the code should still fail at runtime since it's deleting the directory.

a9p: Hi @andreyvelich / @andreyvelich! Curious if either of you have any thoughts on the above?

andreyvelich: @a9p Apologies for the late reply, please can we discuss it again after the Kubeflow release (e.g. after 2 weeks) ?

a9p: @andreyvelich / @tenzen-y -- just following up if either of you have any thoughts on the best way to include the k8s spec in this repo?

tenzen-y: Sorry for the late response. I'm missing the notifications. I came back to this thread right now.
I'll check your changes.

tenzen-y: > is git submoduling in the manager/v1beta1 directory the preferred approach? I put a temporary shim in the build.sh tying it to the minor k8s version just to verify static checks, but the code should still fail at runtime since it's deleting the directory.

@a9p Thanks for your investigating. 
Can we store the proto definition for the k8s.io/apps/v1 only when generating codes? I guess we can store the k8s proto file locally like the following.

https://github.com/kubeflow/mpi-operator/blob/c034b8caf613b2951aa2550c32f1333c58f56a39/Makefile#L170-L172

I guess that we can do that with `GOPATH=/tmp go install k8s.io/api`.
I appreciate your effort!

andreyvelich (commented):
Hi @a9p, do you have any free time to complete this PR so we can merge it as part of our next release (Katib release-0.16)?

a9p:
Hi @andreyvelich, it doesn't look like I will be able to wrap up this PR before feature freeze. Could we push this to the next release target? I should have a fair amount more time to work on this closer to October.

andreyvelich:
Sure, no problem @a9p!
Thank you for your time

tenzen-y:
@a9p Hi, are you still here?

tenzen-y:
/remove-lifecycle stale

Still, we should move this forward.

PR #1991, opened 2022-11-01
[kwa-kfp-component] Introduce KWA's frontend component for kfp links

elenzio9:
In this PR:
- Introduce the `kfp-run` component as a distinct component.
- Make the pipeline button a link.

Here's a screenshot:
image

Signed-off-by: Elena Zioga <elena@arrikto.com>

PR #1977, opened 2022-10-21
UI: Fix unit tests

elenzio9:
This PR fixes the Katib unit tests. In order to run the unit tests, try:
- `npm i` to install the updated packages and after
- `npm run test` to run the tests

Related PRs: https://github.com/kubeflow/kubeflow/pull/6676, https://github.com/kubeflow/katib/pull/1974#pullrequestreview-1150544918

Signed-off-by: Elena Zioga <elena@arrikto.com>

PR #1998, opened 2022-11-03
gh-actions: Extend action to run Frontend Unit tests

orfeas-k:
This is a follow up PR to this https://github.com/kubeflow/katib/pull/1977. Here, we extend the GH action [Frontend Test](https://github.com/kubeflow/katib/blob/master/.github/workflows/test-node.yaml) to include a job that runs KWA frontend unit tests as well. This also is needed before this PR https://github.com/kubeflow/katib/pull/1991#issuecomment-1301793351 is merged to ensure that nothing breaks.

### DIFF
--- a/.github/workflows/publish-core-images.yaml
+++ b/.github/workflows/publish-core-images.yaml
@@ -1,8 +1,12 @@
 name: Publish Katib Core Images
 
 on:
-  - push
-  - pull_request
+  push:
+    paths-ignore:
+      - 'pkg/new-ui/v1beta1/frontend/**'
+  pull_request:
+    paths-ignore:
+      - 'pkg/new-ui/v1beta1/frontend/**'
 
 jobs:
   core:
@@ -15,7 +19,6 @@ jobs:
     secrets:
       DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
       DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
-
 
     strategy:
       fail-fast: false
--- a/.github/workflows/test-node.yaml
+++ b/.github/workflows/test-node.yaml
@@ -14,7 +14,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Setup Node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: 12.18.1
 
@@ -22,3 +22,37 @@ jobs:
         run: |
           npm install prettier --prefix ./pkg/new-ui/v1beta1/frontend
           make prettier-check
+
+  frontend-unit-tests:
+    name: Frontend Unit Tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 12.18.1
+
+      - name: Fetch Kubeflow and install common code dependencies
+        run: |
+          cd /tmp && git clone https://github.com/kubeflow/kubeflow.git
+          cd kubeflow
+          git checkout c4ca7a9
+          cd components/crud-web-apps/common/frontend/kubeflow-common-lib
+          npm i
+          npm run build
+          npm link ./dist/kubeflow
+
+      - name: Install KWA dependencies
+        run: |
+          cd pkg/new-ui/v1beta1/frontend
+          npm i
+          npm link kubeflow
+
+      - name: Run unit tests
+        run: |
+          cd pkg/new-ui/v1beta1/frontend
+          npm run test:prod
--- a/.github/workflows/darts-cifar10-e2e-test.yaml
+++ b/.github/workflows/darts-cifar10-e2e-test.yaml
@@ -1,6 +1,8 @@
 name: E2E Test with darts-cnn-cifar10
 on:
-  - pull_request
+  pull_request:
+    paths-ignore:
+      - 'pkg/new-ui/v1beta1/frontend/**'
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/enas-cifar10-e2e-test.yaml
+++ b/.github/workflows/enas-cifar10-e2e-test.yaml
@@ -1,6 +1,8 @@
 name: E2E Test with enas-cnn-cifar10
 on:
-  - pull_request
+  pull_request:
+    paths-ignore:
+      - 'pkg/new-ui/v1beta1/frontend/**'
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/mxnet-mnist-e2e-test.yaml
+++ b/.github/workflows/mxnet-mnist-e2e-test.yaml
@@ -1,6 +1,8 @@
 name: E2E Test with mxnet-mnist
 on:
-  - pull_request
+  pull_request:
+    paths-ignore:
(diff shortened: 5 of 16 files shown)


.github/workflows/test-node.yaml
```diff
@@ -22,3 +22,37 @@ jobs:
...
           npm install prettier --prefix ./pkg/new-ui/v1beta1/frontend
           make prettier-check
+
+  frontend-unit-tests:
+    name: Frontend Unit Tests
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
```
tenzen-y, line 32: Can you rebase this PR and upgrade this `actions/checkout` version to v3 since we upgrade the `actions/checkout` version in #1996 to resolve the warning?

.github/workflows/test-node.yaml
```diff
@@ -22,3 +22,37 @@ jobs:
...
+        - name: Install KWA dependencies
+          run: |
+            cd pkg/new-ui/v1beta1/frontend
+            npm i
+            npm link kubeflow
+
+        - name: Run unit tests
+          run: |
+            cd katib/pkg/new-ui/v1beta1/frontend
+            npm run test
```
tenzen-y, line 58: Can you remove one of the top tabs since these items maybe are wrong place?

tenzen-y (commented):
@orfeas-k Thanks for creating this!
I left a few comments.

/assign @kimwnasptd

.github/workflows/test-node.yaml
orfeas-k: Sure, I just force-pushed the branch with both of your comments fixed. Thanks for noticing @tenzen-y

.github/workflows/test-node.yaml
orfeas-k: Fixed this! Thank you @tenzen-y

.github/workflows/test-node.yaml
tenzen-y: I'm ok with upgrading the` actions/setup-node` version. Thanks!

orfeas-k:
While working on this PR, I noticed that all jobs were queued and ran and thus, had to wait for a long time in order to verify if something breaks. After some discussion with @kimwnasptd , we thought it would be preferable if we could eliminate and run the jobs that are specific to each PR's changes. For that, we could utilise https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths and exclude the frontend path in jobs that are not related to it. 

WDYT? Which workflows do you think we should run when there's a change in the frontend?   cc @johnugeorge @andreyvelich @tenzen-y

johnugeorge:
I think, for UI related changes, we need to run only  https://github.com/kubeflow/katib/blob/master/.github/workflows/katib-ui-e2e-test.yaml and can skip all e2e tests.

@tenzen-y Any comments?

kimwnasptd:
@johnugeorge @tenzen-y  in this PR we also changed the [.github/workflows/test-node.yaml](https://github.com/kubeflow/katib/pull/1998/files#diff-950cb7be033a3cd5ed7606e4923ff5a038fccc3d8bb32edc5f5358ad5143a290) for running the unit tests, and would extend it in the future to also run the integration tests.

Should we include this action as well for running unit/integration tests or should we migrate everything for the UI into the action you mentioned above?
https://github.com/kubeflow/katib/blob/master/.github/workflows/katib-ui-e2e-test.yaml

johnugeorge:
@kimwnasptd If the integration tests requires images to be built, it is best to be added in https://github.com/kubeflow/katib/blob/master/.github/workflows/katib-ui-e2e-test.yaml  as the workflow already does it for you.
If not, you can keep all tests  in [github/workflows/test-node.yaml](https://github.com/kubeflow/katib/pull/1998/files#diff-950cb7be033a3cd5ed7606e4923ff5a038fccc3d8bb32edc5f5358ad5143a290) itself.

tenzen-y:
> I think, for UI related changes, we need to run only https://github.com/kubeflow/katib/blob/master/.github/workflows/katib-ui-e2e-test.yaml and can skip all e2e tests.
> 
> @tenzen-y Any comments?

@johnugeorge 
Thanks for reminding me.

@kimwnasptd 
I also think that we can skip any other e2e tests except katib-ui-e2e-test for only UI code changes.

orfeas-k:
@kimwnasptd @johnugeorge @tenzen-y I pushed another commit in order to exclude workflows from running when there are changes only in the frontend, just as we discussed here. I'm also removing the `WIP -` also from this PR.

.github/workflows/publish-core-images.yaml
```diff
@@ -1,8 +1,12 @@
...
 
 on:
-  - push
-  - pull_request
+  push:
+    paths-ignore:
+      - 'pkg/new-ui/v1beta1/frontend/**'
+  pull_request:
+    paths-ignore:
+      - 'pkg/new-ui/v1beta1/frontend/**'
```
tenzen-y, line 9: We need to check whether building the UI image supports multi-platform.
it might be better to separate the job to build the UI image.

This means creating the following new actions file.

```yaml
name: Publish Katib UI Image

on:
  push:
  pull_request:

jobs:
  ui:
    name: Publish Image
    uses: ./.github/workflows/build-and-publish-images.yaml
    with:
      component-name: katib-ui
      platforms: linux/amd64,linux/arm64
      dockerfile: cmd/new-ui/v1beta1/Dockerfile
    secrets:
      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
```

.github/workflows/lint.yaml
```diff
@@ -1,8 +1,12 @@
...
 
 on:
-  - push
-  - pull_request
+  push:
+    paths-ignore:
+      - 'pkg/new-ui/v1beta1/frontend/**'
+  pull_request:
+    paths-ignore:
+      - 'pkg/new-ui/v1beta1/frontend/**'
```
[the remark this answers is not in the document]
orfeas-k: Ok I 'll remove the change in the `lint.yaml` file completely then.

.github/workflows/publish-core-images.yaml
orfeas-k: Just to be sure, we should keep the `paths-ignore` changes made to `publish-core-images.yaml`, right?

orfeas-k: Also, a small correction on the new workflow file
```
on:
  - push
  - pull_request
```

tenzen-y: > Just to be sure, we should keep the paths-ignore changes made to publish-core-images.yaml, right?

Yes, that's right.

> Also, I 'll add a small correction on the new workflow file

I'm ok.

orfeas-k: I just pushed the branch with the changes included.

kimwnasptd:
Thanks @orfeas-k. LGTM from my side. I think we'll only need to re-run one action.

/lgtm

.github/workflows/public-katib-ui-image.yaml
```diff
@@ -0,0 +1,17 @@
+name: Publish Katib UI Image
```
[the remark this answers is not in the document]
orfeas-k: Thank you for this @tenzen-y as this was not intentional. Just pushed the branch with the file renamed.

tenzen-y (commented):
@orfeas-k Thanks for updating!
/lgtm

/assign @kubeflow/wg-automl-leads

PR #1999, opened 2022-11-03
Add scripts to generate Go mock files to Makefile

### LINKED ISSUE #59: [test] Add verify-mockgen in CI
Now we do not check if generated code is changed, thus we should do it. Refer to https://github.com/kubeflow/tf-operator/blob/master/hack/verify-codegen.sh

tenzen-y:
Signed-off-by: tenzen-y <yuki.iwai.tz@gmail.com>

I added scripts to generate Go mock files to Makefile.

Fixes #59

**Checklist:**


Makefile
```diff
@@ -21,9 +23,9 @@ test: envtest
 envtest:
 ifndef HAS_SETUP_ENVTEST
 	go install sigs.k8s.io/controller-runtime/tools/setup-envtest@2c3a6fa2996c026b284c7fe2b055274cd9a556bc #v0.13.0
-	@echo "setup-envtest has been installed"
+	$(info "setup-envtest has been installed")
```
andreyvelich, line 26: Why do we want to change `echo` to `info` ?

andreyvelich (commented):
Thanks for taking this @tenzen-y!

I think to cover this issue: https://github.com/kubeflow/katib/issues/59 we also want to add unit test to verify that contributor runs `make generate` before creating PR.

Should we update our unit tests to verify that source code doesn't have any changes after running `make generate` ?

Makefile
tenzen-y: @andreyvelich  I changed them since I think that it might be better to use the Make function instead of shell scripts in Makefile.

Would you like to keep using `@echo`?

tenzen-y:
> Thanks for taking this @tenzen-y!
> 
> I think to cover this issue: #59 we also want to add unit test to verify that contributor runs `make generate` before creating PR.
> 
> Should we update our unit tests to verify that source code doesn't have any changes after running `make generate` ?

Thanks for the review!

Does that mean we add scripts to Makefile or create the shellscript file as shown the below:

https://github.com/kubeflow/katib/blob/0848e0303d44f24951374720dd2ff3c09b9d8c43/.github/workflows/test-go.yaml#L38-L43

Makefile
andreyvelich: Sounds good, as I can see [K8s also uses `info`](https://github.com/kubernetes/kubernetes/blob/master/build/root/Makefile#L50)

andreyvelich:
@tenzen-y It's a good idea to re-use this unit test to check if contributor forgot to run `make check` or `go mod tidy`.
Should we add more files in addition to (`git add pkg/apis hack/gen-python-sdk`) to check if `make generate` needs to be ran ?

tenzen-y:
> @tenzen-y It's a good idea to re-use this unit test to check if contributor forgot to run `make check` or `go mod tidy`. Should we add more files in addition to (`git add pkg/apis hack/gen-python-sdk`) to check if `make generate` needs to be ran ?

It makes sense.

hack/verify-generated-codes.sh
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+cd "$(dirname "$0")/.."
+
+git diff --exit-code -- 'pkg/apis' 'hack/gen-python-sdk' 'sdk/python' 'pkg/mock' > /dev/null || \
```
tenzen-y, line 23: @andreyvelich Are there any other directories that need to be checked?

hack/verify-go-mod.sh
```diff
@@ -0,0 +1,26 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+cd "$(dirname "$0")/.."
+
+GO_VERSION="$(grep '^go' go.mod | cut -d ' ' -f 2)"
+go mod tidy -go "${GO_VERSION}"
+git diff --exit-code -- 'go.*' > /dev/null || \
+  (echo "Please run \"go mod tidy -go ${GO_VERSION}\" to sync Go modules" && exit 1)
```
andreyvelich, line 26: Should we keep code generation check in 1 hack script ?
Then, we can reduce our number of scripts.

hack/verify-generated-codes.sh
andreyvelich: @tenzen-y What do you think about verifying the whole code base on the changes
(e.g. `git diff --exit-code -- . `)?
For example, tomorrow we are going to introduce script check to automatically sync `operator` manifests with our manifests and we ask users to run this script before submitting PR.

tenzen-y: It makes sense.
I will change so that developer can check all files.

hack/verify-go-mod.sh
tenzen-y: Sure, thanks!

tenzen-y:
@andreyvelich I addressed your suggestions. PTAL

andreyvelich:
Thanks for the updates @tenzen-y!
/lgtm
/assign @johnugeorge

PR #2000, opened 2022-11-05
More container fields for SuggestionConfig

### LINKED ISSUE #1737: Support users to set more fields in suggetion deployment
/kind feature

Is it possible to allow users to set more fields in suggetion deployment like we set the trailSpec
such as imagePullSecrets 、args、command and so on

fischor:

This PR adds more fields to the `SuggestionConfig` that is used to populate the main container of the Suggestion deployment. It adds fields `command`, `args`, `envs` that can be used to pass arguments to the executable. This allows the user to easily parametrize the suggestion algorithm/executable itself by passing values using these new fields.   

Fixes #1737 (at least partly)

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/components/katib/): the docs will need to mention the new fields

### DIFF
--- a/pkg/controller.v1beta1/suggestion/composer/composer.go
+++ b/pkg/controller.v1beta1/suggestion/composer/composer.go
@@ -191,13 +191,12 @@ func (g *General) desiredContainers(s *suggestionsv1beta1.Suggestion,
 		suggestionContainer.Name = consts.ContainerSuggestion
 	}
 
-	if len(suggestionContainer.Ports) == 0 {
-		suggestionContainer.Ports = []corev1.ContainerPort{
-			{
-				Name:          consts.DefaultSuggestionPortName,
-				ContainerPort: consts.DefaultSuggestionPort,
-			},
+	if !containsContainerPort(suggestionContainer.Ports, consts.DefaultSuggestionPortName) {
+		suggestionPort := corev1.ContainerPort{
+			Name:          consts.DefaultSuggestionPortName,
+			ContainerPort: consts.DefaultSuggestionPort,
 		}
+		suggestionContainer.Ports = append(suggestionContainer.Ports, suggestionPort)
 	}
 
 	if viper.GetBool(consts.ConfigEnableGRPCProbeInSuggestion) && suggestionContainer.ReadinessProbe == nil {
@@ -233,15 +232,14 @@ func (g *General) desiredContainers(s *suggestionsv1beta1.Suggestion,
 		}
 	}
 
-	// Attach volume mounts to the suggestion container if ResumePolicy = FromVolume
-	if s.Spec.ResumePolicy == experimentsv1beta1.FromVolume && len(suggestionContainer.VolumeMounts) == 0 {
-		suggestionContainer.VolumeMounts = []corev1.VolumeMount{
-			{
-				Name:      consts.ContainerSuggestionVolumeName,
-				MountPath: suggestionConfigData.VolumeMountPath,
-			},
+	if s.Spec.ResumePolicy == experimentsv1beta1.FromVolume && !containsVolumeMount(suggestionContainer.VolumeMounts, consts.ContainerSuggestionVolumeName) {
+		suggestionVolume := corev1.VolumeMount{
+			Name:      consts.ContainerSuggestionVolumeName,
+			MountPath: suggestionConfigData.VolumeMountPath,
 		}
+		suggestionContainer.VolumeMounts = append(suggestionContainer.VolumeMounts, suggestionVolume)
 	}
+
 	containers = append(containers, suggestionContainer)
 
 	if s.Spec.EarlyStopping != nil && s.Spec.EarlyStopping.AlgorithmName != "" {
@@ -260,6 +258,26 @@ func (g *General) desiredContainers(s *suggestionsv1beta1.Suggestion,
 		containers = append(containers, earlyStoppingContainer)
 	}
 	return containers
+}
+
+func containsVolumeMount(volumeMounts []corev1.VolumeMount, name string) bool {
+	for i := range volumeMounts {
+		if volumeMounts[i].Name == name {
+			return true
+		}
+	}
+
+	return false
+}
+
+func containsContainerPort(ports []corev1.ContainerPort, name string) bool {
+	for i := range ports {
+		if ports[i].Name == name {
+			return true
+		}
+	}
+
+	return false
 }
 
 // DesiredVolume returns desired PVC and PV for Suggestion.
--- a/pkg/controller.v1beta1/suggestion/composer/composer.go
+++ b/pkg/controller.v1beta1/suggestion/composer/composer.go
@@ -186,6 +186,11 @@ func (g *General) desiredContainers(s *suggestionsv1beta1.Suggestion,
 		Name:            consts.ContainerSuggestion,
 		Image:           suggestionConfigData.Image,
 		ImagePullPolicy: suggestionConfigData.ImagePullPolicy,
+		Command:         suggestionConfigData.Command,
+		Args:            suggestionConfigData.Args,
+		WorkingDir:      suggestionConfigData.WorkingDir,
+		EnvFrom:         suggestionConfigData.EnvFrom,
+		Env:             suggestionConfigData.Env,
 		Ports: []corev1.ContainerPort{
 			{
 				Name:          consts.DefaultSuggestionPortName,
--- a/pkg/controller.v1beta1/suggestion/composer/composer.go
+++ b/pkg/controller.v1beta1/suggestion/composer/composer.go
@@ -76,6 +76,11 @@ func (g *General) DesiredDeployment(s *suggestionsv1beta1.Suggestion) (*appsv1.D
 	if err != nil {
 		return nil, err
 	}
+	if containsContainerPortWithName(suggestionConfigData.Ports, consts.DefaultSuggestionPortName) ||
+		containsContainerPort(suggestionConfigData.Ports, consts.DefaultSuggestionPort) {
+		return nil, fmt.Errorf("invalid suggestion config: a port with name %q or number %d must not be specified",
+			consts.DefaultSuggestionPortName, consts.DefaultSuggestionPort)
+	}
 
 	// If early stopping is used, get the config data.
 	earlyStoppingConfigData := katibconfig.EarlyStoppingConfig{}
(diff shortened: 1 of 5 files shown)


pkg/util/v1beta1/katibconfig/config.go
```diff
@@ -36,6 +36,11 @@ import (
 type SuggestionConfig struct {
 	Image                     string                           `json:"image"`
 	ImagePullPolicy           corev1.PullPolicy                `json:"imagePullPolicy,omitempty"`
+	Command                   []string                         `json:"command,omitempty"`
+	Args                      []string                         `json:"args,omitempty"`
+	WorkingDir                string                           `json:"workingDir,omitempty"`
+	EnvFrom                   []corev1.EnvFromSource           `json:"envFrom,omitempty"`
+	Env                       []corev1.EnvVar                  `json:"env,omitempty"`
```
andreyvelich, line 43: Maybe we could start adding more fields to SuggestionConfig with `corev1.Container{}` field, similar to [`PodSpec` API in K8s](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3105) ? I think you can use inline settings to avoid modification in existing Katib config YAML:
```go
 v1.Container `json:",inline"`
 ```

I guess, at the end we want to give user ability to modify the whole Suggestion's Deployment spec via Katib config, but it requires much more changes.

WDYT @tenzen-y @johnugeorge @fischor ?

**Review by andreyvelich — COMMENTED** [review]
Thank you for taking this @fischor!

File: pkg/util/v1beta1/katibconfig/config.go (hunk 1 above)
Comment (tenzen-y) [reply pkg/util/v1beta1/katibconfig/config.go:43]: > Maybe we could start adding more fields to SuggestionConfig with corev1.Container{} field, similar to [PodSpec API in K8s](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3105) ? I think you can use inline param to avoid modification in existing Katib config YAML

sgtm

> I guess, at the end we want to give user ability to modify the whole Suggestion's Deployment spec via Katib config, but it requires much more changes.

I think so too. Maybe, to support customizing Tolerations, imagePullSecret and so on, we want to support customizing PodSpec for suggestion service deployment.

Comment (fischor) [reply pkg/util/v1beta1/katibconfig/config.go:43]: Yes, inlining `corev1.Container` should also be doable.  
Certain fields of the container will be set by the katib-Controller, like some ports, the liveness and readiness probes and possibly the "suggestion-volume". How to handle these?

> I think so too. Maybe, to support customizing Tolerations, imagePullSecret and so on, we want to support customizing PodSpec for suggestion service deployment.

To support his, how about specifying a default deployment in the suggestion config (instead of just the container).

Comment (johnugeorge) [reply pkg/util/v1beta1/katibconfig/config.go:43]: For namespace specific specific suggestion deployments, we can allow to have a configmap in a user namespace(where experiment is run). If suggestion spec is provided in the configmap of the user namespace, it can override the global katib config.  
Ref: https://github.com/kubeflow/katib/issues/1976

Comment (andreyvelich) [reply pkg/util/v1beta1/katibconfig/config.go:43]: > Certain fields of the container will be set by the katib-Controller, like some ports, the liveness and readiness probes and possibly the "suggestion-volume". How to handle these? Overwrite or not?

@fischor I think if parameters are set by the user we should use them, otherwise use the default values.

> To support his, how about specifying a default deployment in the suggestion config (instead of just the container)?.

We might want to set `spec.template.spec` for our Suggestion config, as I mentioned here: https://github.com/kubeflow/katib/issues/1737#issuecomment-974245081, but that requires much more changes since we need to modify manifests.

>  If suggestion spec is provided in the configmap of the user namespace, it can override the global katib config.

@johnugeorge That might be the next step. We need to have discussion what is the best way to give user ability to set experiment-scoped Katib Config.

**Comment by johnugeorge** [comment]
@fischor Can you rebase and fix?

**Comment by fischor** [comment]
The name of the suggestion algorithm container was not defaulted. Creating the suggestion deployment failed because of a missing container  name.

@johnugeorge I have rebased and fixed.

File: pkg/util/v1beta1/katibconfig/config.go [hunk 2]
```diff
@@ -34,9 +34,7 @@ import (
 
 // SuggestionConfig is the JSON suggestion structure in Katib config.
 type SuggestionConfig struct {
-	Image                     string                           `json:"image"`
-	ImagePullPolicy           corev1.PullPolicy                `json:"imagePullPolicy,omitempty"`
-	Resource                  corev1.ResourceRequirements      `json:"resources,omitempty"`
+	corev1.Container          `json:",inline"`
 	ServiceAccountName        string                           `json:"serviceAccountName,omitempty"`
 	VolumeMountPath           string                           `json:"volumeMountPath,omitempty"`
```
andreyvelich, line 39: @fischor Do we need `VolumeMountPath` if that is part of Container parameters: https://github.com/kubeflow/katib/blob/7f1b2842f87158c25c54f577da1b82ddc7a958e3/pkg/controller.v1beta1/suggestion/composer/composer.go#L237 ?

fischor: The current Katib `SuggestionConfig` has a JSON field `volumeMountPath` thats just a filepath, the `corev1.Container` has a json field `volumeMounts` thats an array of volume mounts. So `volumeMountPath` is not part if the container fields.

The referenced snippet kind of follows the logic:
> ... if parameters are set by the user we should use them, otherwise use the default values.

In case the `volumeMounts` are set we prefer that, otherwise mount the default suggestion volume under the specificed `volumeMountPath`.

andreyvelich: I see, that is because we set [the default value name for Suggestion Volume](https://github.com/kubeflow/katib/blob/7f1b2842f87158c25c54f577da1b82ddc7a958e3/pkg/controller.v1beta1/suggestion/composer/composer.go#L121). Please can you leave comment that we should refactor `VolumeMountPath` parameter once we allow to set Deployment spec in our Suggestion config ?

Also, I think we should modify [`len(suggestionContainer.VolumeMounts) == 0` check](https://github.com/kubeflow/katib/blob/7f1b2842f87158c25c54f577da1b82ddc7a958e3/pkg/controller.v1beta1/suggestion/composer/composer.go#L237).
We always should append `VolumeMount` if ResumePolicy is `FromVolume`.

```go
if s.Spec.ResumePolicy == experimentsv1beta1.FromVolume {
volumeMount := corev1.VolumeMount{
	Name:      consts.ContainerSuggestionVolumeName,
	MountPath: suggestionConfigData.VolumeMountPath,
}
suggestionContainer.VolumeMounts = append(suggestionContainer.VolumeMounts, volumeMount)
}
```
WDYT @fischor ?

fischor: @andreyvelich Have you seen that the consts.DefaultSuggestionPortName port is also just set if the Container specified in the suggestion has no ports set? See https://github.com/fischor/katib/blob/7f1b2842f87158c25c54f577da1b82ddc7a958e3/pkg/controller.v1beta1/suggestion/composer/composer.go#L194-L201
Would you also change that to always append the default port?

I think both ways of doing this have pros/cons. If we always append the default volume/port then the user has no option to overwrite it. If we never append the default volume/port, then the user needs to be aware that it has to specify it when setting any volumes/ports.

A third option would be to check if the default volume/port is already set in the container and only append if its not. I think I would prefer that.

andreyvelich: > A third option would be to check if the default volume/port is already set in the container and only append if its not. I think I would prefer that.

I think that should work, but we need to make sure that volume and port name is correct (`DefaultSuggestionPortName` and `ContainerSuggestionVolumeName`).

pkg/controller.v1beta1/suggestion/composer/composer.go
```diff
@@ -181,21 +181,30 @@ func (g *General) desiredContainers(s *suggestionsv1beta1.Suggestion,
...
+	// Assign default values for suggestionContainer fields that are not set via
+	// the suggestion config.
+
+	if suggestionContainer.Name == "" {
+		suggestionContainer.Name = consts.ContainerSuggestion
 	}
 
-	if viper.GetBool(consts.ConfigEnableGRPCProbeInSuggestion) {
+	if !containsContainerPortWithName(suggestionContainer.Ports, consts.DefaultSuggestionPortName) &&
+		!containsContainerPort(suggestionConfigData.Ports, consts.DefaultSuggestionPort) {
```
andreyvelich, line 199: Should we verify that port `name` and `containerPort` is set together in the Katib Config ?
[`DefaultSuggestionPortName` and `DefaultSuggestionPort` is always used for the Suggestion's service](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/suggestion/composer/composer.go#L147-L151).
So if user set `DefaultSuggestionPortName` port with different address, the Suggestion service failed.

cc @tenzen-y

andreyvelich: cc @johnugeorge

tenzen-y: It might be better to prohibit setting containerPort in katibConfig.

andreyvelich: @fischor Do you have time to check my latest commit ? We are getting closer for the Katib 0.15 release, and it would be great to include your feature in that release.

fischor: @andreyvelich I should have some spare time tomorrow. Do we do it as you suggested: verify that name and port are correct, otherwise return an error in `DesiredDeployment` method? Or prohibit the DefaultSuggestionPort to be set at all?

andreyvelich: @fischor Thank you! Let's prohibit to set `DefaultSuggestionPort` for now as @tenzen-y suggested. So we can avoid errors from the user side.

johnugeorge: @fischor We are planning to cut a release in a day. Do you have any update?

fischor: @johnugeorge Thanks for the reminder...you can check the latest commit now.

andreyvelich:
Thanks you for updates @fischor.
Excited to see a new usage of Katib config!
/lgtm
/approve

tenzen-y:
@fischor Great work!
Thanks for your effort!

Issue #1967, opened 2022-09-30
Katib experiment detail page 500 error after refresh


shaowei-su:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
Katib experiment detail page can be accessed after a given namespace (non default one) is chosen, however, upon page refresh or direct access page URL e.g `https://service-name/katib/experiment/exp-name`, UI will throw out 500 error.
<img width="783" alt="Screen Shot 2022-09-29 at 20 02 01" src="https://user-images.githubusercontent.com/7682657/193181398-047075ac-f4d9-4305-8156-cea8bfd8185c.png">

By checking the requested URL, it seems that the actual API request sent is `https://service-name/katib/fetch_hp_job_info/?experimentName=exp-name&namespace=kubeflow-user` where the namespace is default to `namespace=kubeflow-user`.

**What did you expect to happen:**

The experiment detail page should able to refresh and directly accessible. 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Katib version (check the Katib controller image version): v0.13.0
- Kubernetes version: (`kubectl version`): N/A
- OS (`uname -a`): N/A

---

Impacted by this bug? Give it a 👍 We prioritize the issues with the most 👍

tenzen-y:
@shaowei-su Thanks for reporting this!
Did you get the same error with katib v0.14.0?

shaowei-su:
Hi @tenzen-y yeah I just tried `v0.14.0` UI component and it failed with the same error. Are you able to reproduce this on your side?

tenzen-y:
@shaowei-su Sorry for the late response.
In my local kind cluster, I faced that issue.

<img width="1909" alt="2022-11-03 4 52 36" src="https://user-images.githubusercontent.com/68272500/199589534-8944f43a-7402-497c-bd81-a49a44a84f11.png">

tenzen-y:
@kimwnasptd Can you help with fixing this bug?

johnugeorge:
/cc @elenzio9 

Can you please take a look at this? This seems to be a blocker for new UI 

Ref: https://github.com/kubeflow/katib/pull/1982#issuecomment-1300813767

kimwnasptd:
Thanks for the ping! 

Yes we'll take a look at this one within the next week.

johnugeorge:
Thanks @kimwnasptd 

[...]

PR #2001, opened 2022-11-07
frontend: Fix 500 error after detail page refresh (#1967)

elenzio9:
Fix 500 error when refreshing KWA's detail page by also adding the namespace variable as a query parameter to the route.

Related issue: https://github.com/kubeflow/katib/issues/1967

Signed-off-by: Elena Zioga <elena@arrikto.com>

### DIFF
--- a/pkg/new-ui/v1beta1/frontend/src/app/app-routing.module.ts
+++ b/pkg/new-ui/v1beta1/frontend/src/app/app-routing.module.ts
@@ -7,7 +7,10 @@ import { TrialModalComponent } from './pages/experiment-details/trials-table/tri
 
 const routes: Routes = [
   { path: '', component: ExperimentsComponent },
-  { path: 'experiment/:experimentName', component: ExperimentDetailsComponent },
+  {
+    path: 'experiment/:namespace/:experimentName',
+    component: ExperimentDetailsComponent,
+  },
   { path: 'new', component: ExperimentCreationComponent },
   {
     path: 'experiment/:experimentName/trial/:trialName',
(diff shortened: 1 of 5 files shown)


johnugeorge:
Thanks @elenzio9 

@shaowei-su Can you check this out in your env if it works ?

kimwnasptd:
@elenzio9 thanks for the PR!

With a quick glance I see that you only changed the files of the page that handles an Experiment. Should we also change the main page with the table that redirects users to the details page of an Experiment? To ensure the correct link is used there as well

elenzio9:
> @elenzio9 thanks for the PR!
> 
> With a quick glance I see that you only changed the files of the page that handles an Experiment. Should we also change the main page with the table that redirects users to the details page of an Experiment? To ensure the correct link is used there as well

I've also updated this page as you can see here: https://github.com/kubeflow/katib/pull/2001/files#diff-4e45400d0d644ac998f731d708df6edd6667d51b19c7be069e2522dfcef3f137R89

shaowei-su:
Thanks @elenzio9 @johnugeorge , just tested this branch offline and the 500 error is fixed. Huge thanks for your timely fix 👍

shaowei-su:
Actually, I found an issue with loading the visualized graph if the experiment is in "running" state:
Screen Shot 2022-11-07 at 15 37 50

elenzio9:
> Actually, I found an issue with loading the visualized graph if the experiment is in "running" state (6 out of 10 trials are already succeeded but failed to load)

Since the above issue is related to the experiment graph, it would be helpful to create another issue and describe everything there. Thank you!

johnugeorge:
@

> Actually, I found an issue with loading the visualized graph if the experiment is in "running" state (6 out of 10 trials are already succeeded but failed to load) :
> Screen Shot 2022-11-07 at 15 37 50

@shaowei-su Can you create a separate issue to track this?  This PR is good to go

/assign @kimwnasptd

kimwnasptd:
@elenzio9 can you do a rebase here as well to ensure the unittests are passing as well?

Issue #2002, opened 2022-11-08
Duplicated suggestions generated when early stopping is enabled


shaowei-su:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

When early stopping is enabled with `bayesianoptimization`, duplicated suggestions/trials are generated. 
See suggestion service log:
```
INFO:pkg.suggestion.v1beta1.skopt.base_service:Optimizer tell method takes 0 seconds
INFO:pkg.suggestion.v1beta1.skopt.base_service:List of recorded Trials names: ['exp-6f14cce0-5ac6-11ed-bd7e-acde48001122-6wcpxc58', 'exp-6f14cce0-5ac6-11ed-bd7e-acde48001122-ss88lpl9', 'exp-6f14cce0-5ac6-11ed-bd7e-acde48001122-8t8ftjlj', 'exp-6f14cce0-5ac6-11ed-bd7e-acde48001122-szmjrvjc']

INFO:pkg.suggestion.v1beta1.skopt.base_service:Running Optimizer ask to query new parameters for Trials

INFO:pkg.suggestion.v1beta1.skopt.base_service:New suggested parameters for Trial: [0.0008577340571058496, 0.23212435323286384, 'True', 13]
INFO:pkg.suggestion.v1beta1.skopt.base_service:GetSuggestions returns 1 new Trials


INFO:pkg.suggestion.v1beta1.skopt.base_service:----------------------------------------------------------------------------------------------------

INFO:pkg.suggestion.v1beta1.skopt.base_service:New GetSuggestions call

INFO:pkg.suggestion.v1beta1.skopt.base_service:Succeeded Trials didn't change: 4

INFO:pkg.suggestion.v1beta1.skopt.base_service:Running Optimizer ask to query new parameters for Trials

INFO:pkg.suggestion.v1beta1.skopt.base_service:New suggested parameters for Trial: [0.0008577340571058496, 0.23212435323286384, 'True', 13]
INFO:pkg.suggestion.v1beta1.skopt.base_service:GetSuggestions returns 1 new Trials
```

Here the highlight is **Succeeded Trials didn't change: 4** 

**What did you expect to happen:**

Suggestion service should see early stopped trials in the service request and apply updates to `skopt_suggested` and `loss_for_skopt`: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/skopt/base_service.py#L105

**Anything else you would like to add:**
[...]

Issue #577, opened 2019-05-24
[Enhancement Request] Metrics Collector Push-based Implementation


gaocegege:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Now the design of metrics collector is based on pull. We have a metrics collector cron job for one trial. And it collects logs according to the pods log. Then it parses the log and persist the logs in MySQL.

The design has some problems (https://github.com/kubeflow/tf-operator/issues/722#issuecomment-405669269) @johnugeorge proposed a push-based model to avoid the problems caused by the current design. And I also have some ideas about it.

In my design, we need a push-based implementation to push the metrics to prometheus. Then we can use custom-metrics-server to expose the trial or job level metrics. Then katib could get all periodical metrics from k8s master API. The early stopping services can use the API to determine if we should kill the trial. And UI can use the API to show the metrics.

And, tfjob and pytorchjob can also benefit from the metrics collector. Because we can use it collect periodical metrics for them, too. And the metrics will be exposed by a kubernetes native way: [K8s metrics API](https://github.com/kubernetes/metrics#custom-metrics-api)

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

gaocegege:
I am glad to discuss it after v1alpha2 is released.

johnugeorge:
Yes. This is a long standing need :) Lets take this up during next api design phase

johnugeorge:
This is fixed with new metric collector design in v1alpha3.

johnugeorge:
Closing the issue

Electronic-Waste:
Hi everyone!

I'm Electronic-Waste, a.k.a. Shao Wang, a senior student from Shanghai Jiao Tong Univeristy. My major interests lies in cloud infra concerned with k8s, and also, ai infra. I have two professional experience concerd with k8s and go:
1. For the first experience, I wrote a [naive kubernetes integrated with knative](https://github.com/Electronic-Waste/Minik8s) along with my classmates.
2. In the second experience, I implement a [heterogenous network emulation system](https://github.com/dtn-dslab/Satellite-SDN) based on kubernetes in @sjtu-sail, in which I'll pursue my master degree.

And also, I previous made some [contributions](https://github.com/tensorchord/envd/commits?author=Electronic-Waste) to tensorchord/envd, an open source software designed to solve AI/ML development environment

[...]

PR #2004, opened 2022-11-08
[Fix] add early stopped trials in converter

shaowei-su:
Python suggestion services will convert the trials from request payload before passing in actual search algos. However, early stopped trials are filtered out in this process and thus no updates are provided to search algo and there are duplicated trials created.

Fixes #

https://github.com/kubeflow/katib/issues/2002

**Checklist:**

### DIFF
--- a/pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go
+++ b/pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go
@@ -343,7 +343,7 @@ func (g *General) ConvertTrials(ts []trialsv1beta1.Trial) []*suggestionapi.Trial
 		if t.IsMetricsUnavailable() {
 			continue
 		}
-		if !t.IsObservationAvailable() {
+		if !t.IsObservationAvailable() && t.IsEarlyStopped() {
 			continue
 		}
 		trial := &suggestionapi.Trial{
--- a/pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go
+++ b/pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go
@@ -343,6 +343,9 @@ func (g *General) ConvertTrials(ts []trialsv1beta1.Trial) []*suggestionapi.Trial
 		if t.IsMetricsUnavailable() {
 			continue
 		}
+		if !t.IsObservationAvailable() {
+			continue
+		}
 		trial := &suggestionapi.Trial{
 			Name: t.Name,
 			Spec: &suggestionapi.TrialSpec{
--- a/pkg/suggestion/v1beta1/skopt/base_service.py
+++ b/pkg/suggestion/v1beta1/skopt/base_service.py
@@ -73,7 +73,7 @@ class BaseSkoptService(object):
         Get the new suggested trials with skopt algorithm.
         """
         logger.info("-" * 100 + "\n")
-        logger.info("New GetSuggestions call\n")
+        logger.info("New GetSuggestions call with current request number: {}\n".format(current_request_number))
         skopt_suggested = []
         loss_for_skopt = []
         if len(trials) > self.succeeded_trials or self.succeeded_trials == 0:
@@ -113,8 +113,6 @@ class BaseSkoptService(object):
 
         else:
             logger.error("Succeeded Trials didn't change: {}\n".format(self.succeeded_trials))
-            logger.error("No new suggestions could be generated, return early..\n")
-            return []
 
         logger.info("Running Optimizer ask to query new parameters for Trials\n")
 
--- a/pkg/suggestion/v1beta1/skopt/base_service.py
+++ b/pkg/suggestion/v1beta1/skopt/base_service.py
@@ -112,7 +112,9 @@ class BaseSkoptService(object):
                 logger.info("List of recorded Trials names: {}\n".format(self.recorded_trials_names))
 
         else:
-            logger.info("Succeeded Trials didn't change: {}\n".format(self.succeeded_trials))
+            logger.error("Succeeded Trials didn't change: {}\n".format(self.succeeded_trials))
+            logger.error("No new suggestions could be generated, return early..\n")
+            return []
 
         logger.info("Running Optimizer ask to query new parameters for Trials\n")
 
--- a/pkg/apis/controller/trials/v1beta1/util.go
+++ b/pkg/apis/controller/trials/v1beta1/util.go
@@ -92,6 +92,9 @@ func (trial *Trial) IsMetricsUnavailable() bool {
 
 // IsObservationAvailable return ture if the Trial has valid observations updated
 func (trial *Trial) IsObservationAvailable() bool {
+	if trial.Spec.Objective == nil {
+		return false
+	}
 	objectiveMetricName := trial.Spec.Objective.ObjectiveMetricName
 	if trial.Status.Observation != nil && trial.Status.Observation.Metrics != nil {
 		for _, metric := range trial.Status.Observation.Metrics {
--- a/pkg/apis/controller/trials/v1beta1/util.go
+++ b/pkg/apis/controller/trials/v1beta1/util.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	"errors"
+	"github.com/kubeflow/katib/pkg/controller.v1beta1/consts"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -87,6 +88,19 @@ func (trial *Trial) IsKilled() bool {
 // IsMetricsUnavailable returns true if Trial metrics are not available
 func (trial *Trial) IsMetricsUnavailable() bool {
 	return hasCondition(trial, TrialMetricsUnavailable)
+}
+
+// IsObservationAvailable return ture if the Trial has valid observations updated
+func (trial *Trial) IsObservationAvailable() bool {
+	objectiveMetricName := trial.Spec.Objective.ObjectiveMetricName
+	if trial.Status.Observation != nil && trial.Status.Observation.Metrics != nil {
+		for _, metric := range trial.Status.Observation.Metrics {
+			if metric.Name == objectiveMetricName && metric.Latest != consts.UnavailableMetricValue {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 func (trial *Trial) IsCompleted() bool {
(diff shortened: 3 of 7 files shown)


andreyvelich (commented):
Thank you @shaowei-su!
/lgtm

tenzen-y (commented):
@shaowei-su Thanks for fixing this!
/lgtm

@kubeflow/wg-automl-leads Can you restart failure CI?

shaowei-su:
Due to the aysnc nature of the trials' updates, the suggestion service actually might be invoked without all the observed metrics available. I updated this logic too so that suggestion service return empty list if no new updates been made. 
PTAL! @andreyvelich @tenzen-y Thanks!

tenzen-y:
@shaowei-su Thanks for updating!
/lgtm

pkg/suggestion/v1beta1/skopt/base_service.py
```diff
@@ -112,7 +112,9 @@ def getSuggestions(self, trials, current_request_number):
                 logger.info("List of recorded Trials names: {}\n".format(self.recorded_trials_names))
 
         else:
-            logger.info("Succeeded Trials didn't change: {}\n".format(self.succeeded_trials))
+            logger.error("Succeeded Trials didn't change: {}\n".format(self.succeeded_trials))
+            logger.error("No new suggestions could be generated, return early..\n")
+            return []
```
andreyvelich, line 117: @shaowei-su What if Trial was failed for some reason (e.g. networking) and we ask for the new Trials ?
In that case `succeeded_trials` list will stay the same.

shaowei-su: That's a great question @andreyvelich , I can think of two possible routes:

1. Add more conditional branches in the suggestions service, e.g is the service invoked as a result of failed trials? if so, suggestion service will return duplicated suggestions as before this PR change;

2. Retry for trials should be kept as a logic within trial templates themself i.e Job, TFJob, MPIJob etc.. HPO search experiment only make progress when there are succeeded (including early stopped ones) trials complete. In the case of multiple concurrent trials, suggestion service will not return new trials until next succeeded trial's ready.

andreyvelich: > Add more conditional branches in the suggestions service, e.g is the service invoked as a result of failed trials?

But do we need conditional branch to the suggestion service ?
If `succeeded_trials` = `Succeeded Trials` + `Early Stopped Trials`, then if Trial was failed the Suggestion service will just ask for the new trials (e.g. with duplicated HPs) using `current_request_number`. 

> Retry for trials should be kept as a logic within trial templates themself i.e Job, TFJob, MPIJob etc.. 

That's valid assumption. In that case, what should be the nature of [MaxFailedTrialCount](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/experiments/v1beta1/experiment_types.go#L51) API ?

cc @johnugeorge @tenzen-y

shaowei-su: > But do we need conditional branch to the suggestion service ?

The reason is that suggestion service can't tell whether the `current_request_number` is requested as a result of failed trails (in this case service should return duplicated HPs), or as a result of delay in trial status/observations updates (in this case service should wait until trials are fully updated).

The latter can be reproduced by this [example](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/early-stopping/median-stop.yaml#L17) and change the algo to `bayesianoptimization`.

andreyvelich: @shaowei-su Should we maintain this logic in our skopt suggestion ? E.g. maintain list of created Trials in Suggestion service (similar to our [Chocolate Service](https://github.com/kubeflow/katib/blob/1613191656368e0d9d84a7de1615153584814a99/pkg/suggestion/v1beta1/chocolate/base_service.py#L54)) ?  Then, if Suggestion was invoked, but no new Trials were Succeeded we just return HPs from that list.
WDYT ?

shaowei-su: @andreyvelich I checked the logic from Chocolate Service and not quite sure if that can solve the issue here, e.g when trial early stopped without observation updated in time, if suggestion service consider there is no new succeeded trials, and returning HPs from created list will generate duplicated trials.

Another solution is to root cause why suggestion service is called with `current_request_number > 0` before trial observation is fully updated for early stopping. In theory, [trial_controller](https://github.com/kubeflow/katib/blob/v0.14.0/pkg/controller.v1beta1/trial/trial_controller.go#L255) should guarantee that metric is fully up to date.

shaowei-su: hmm I think the reason might be that for early stopped trials, the status is updated by a separate service call (https://github.com/kubeflow/katib/blob/master/pkg/earlystopping/v1beta1/medianstop/service.py#L184) in async, whereas for succeeded trials update metrics and change status happen in the same trial controller loop (https://github.com/kubeflow/katib/blob/v0.14.0/pkg/controller.v1beta1/trial/trial_controller.go#L264).

To fix this, we could try to add the early stopped but non-observation ready trials into the active trials count as well. wdyt?

andreyvelich: > when trial early stopped without observation updated in time

Should we avoid sending Early Stopped Trials without observation to the Suggestion service ? e.g. we don't send Trials with [Metrics Unavailable condition](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go#L343-L344). I think, Trials logically shouldn't be called "EarlyStopped" until Observation is ready for them. WDYT @shaowei-su @johnugeorge @tenzen-y ?

I know it is hacky that we set EarlyStopping trial condition from the Early Stopping service and only after that we add Observation Trial results. We might want to reconsider some early stopping design when we introduce push-based Metrics Collector: https://github.com/kubeflow/katib/issues/577

andreyvelich: > To fix this, we could try to add the early stopped but non-observation ready trials into the active trials [count](https://github.com/kubeflow/katib/blob/v0.14.0/pkg/controller.v1beta1/experiment/experiment_controller.go#L293) as well. wdyt?

Since Suggestion and Trial controller works in async manner, I think we should introduce another check to Suggestion Client as I mentioned above.

shaowei-su: Actually those trials are already [filtered out](https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/internal/trial.py#L61) on the suggestion service side.

What's missing here is the correct [currentRequestNum](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go#L88), which should not be incremented before the early stopping metrics are fully updated.

andreyvelich: @shaowei-su But is that correct approach to filter them on the Suggestion side instead of Katib Controller side ?

Some of the suggestion services might not use `internal` implementation and just forget about it.
I think the Katib controller should just invoke suggestion service with the correct statuses for the Trials. 
E.g. [these Trial statuses](https://github.com/kubeflow/katib/blob/master/pkg/apis/manager/v1beta1/api.proto#L232-L242).
Which means if EarlyStopped Trial doesn't have observation, Controller should not send it to Suggestion Service.
The main purpose of the Suggestion service is just produce new HPs based on Trial results.

I doubt that Suggestion service should do any additional work with Trials from Katib controller.

What do other think about it ? @johnugeorge @gaocegege @tenzen-y

pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go
```diff
@@ -343,6 +343,9 @@ func (g *General) ConvertTrials(ts []trialsv1beta1.Trial) []*suggestionapi.Trial
 		if t.IsMetricsUnavailable() {
 			continue
 		}
+		if !t.IsObservationAvailable() {
+			continue
+		}
```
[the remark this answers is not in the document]
shaowei-su: I'd suggest to keep this filtering logic since others might run different version of `controller` vs `suggestion service` docker images.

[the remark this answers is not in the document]
shaowei-su: I think it make more sense to filter the failed trails out as those won't provide any updates to the suggestion service.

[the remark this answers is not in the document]
andreyvelich: Usually, we ask our users to run the same version of Katib Control Plane.
Should add TODO in this section to remove it in the future versions ?
What others think @gaocegege @johnugeorge @tenzen-y ?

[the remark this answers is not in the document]
shaowei-su: Good call, I didn't realize that failed trials are tracked by other suggestion services. Updated in the latest commit to filter on `early stopped & without observations` trials only, PTAL.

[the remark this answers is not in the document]
shaowei-su: Same, changed the filter logic to early stopped trials only and let's keep the service side check as it is.

andreyvelich:
Thank you for these updates @shaowei-su!
/lgtm
/assign @johnugeorge @gaocegege @tenzen-y

pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go
```diff
@@ -343,6 +343,9 @@ func (g *General) ConvertTrials(ts []trialsv1beta1.Trial) []*suggestionapi.Trial
 		if t.IsMetricsUnavailable() {
 			continue
 		}
+		if !t.IsObservationAvailable() && t.IsEarlyStopped() {
+			continue
+		}
```
tenzen-y, line 348: @shaowei-su Can we add a test for this condition to the [`TestSyncAssignments`](https://github.com/kubeflow/katib/blob/8190f9c69c6ae2763257e57a67206b725a363e1f/pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient_test.go#L76) func or to a new test function?

shaowei-su: Thanks @tenzen-y ! I added few more trails in the mock so this branching logic is also validated.

Btw, could help restart the failed unit tests? Looks like a flaky test failed.

tenzen-y (commented):
@shaowei-su Thanks for the updates! I appreciate your work in fixing this bug.
/lgtm

/cc @johnugeorge

PR #2008, opened 2022-11-10
added yaml-python hybrid example

### LINKED ISSUE #2006: yaml to python sdk submission example needed
/kind feature

There is a need for an example that submits an Experiment YAML via the Katib SDK. I have written an example, and I plan to submit it as a PR.

My reasoning is this:

The Katib SDK exposes the power of Python programming to Katib users; however, the code structure is significantly more complex than in the case of direct YAML submission using kubectl.

In contrast, direct Katib Experiment submission via YAML offers an easy-to-read (and to maintain) structure, but it is very basic, and it lacks advantages that the SDK route provides.

There is a way to submit a YAML file by converting it to a Python dictionary and then passing it to the Katib SDK, so an example is in order.

In the long run, it would be nice to have this functionality in the SDK (that is, load a YAML file, then update chosen parameters, all in an object that can be passed to create_experiment). For now, this example can help users engage easier with the SDK, especially in the case where they already have experiments defined in YAML, but want to submit via the SDK.

---

dandawg:
https://github.com/kubeflow/katib/issues/2006

Fixes #2006
**Checklist:**

### DIFF
--- a/examples/v1beta1/sdk/yaml-python-submit.ipynb
+++ b/examples/v1beta1/sdk/yaml-python-submit.ipynb
@@ -1,0 +1,529 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# YAML-Python Katib Example Submission"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


andreyvelich:
Sorry for the late reply @dandawg!
Do you want to still work on this example ? I think, that might be useful for users who still want to use YAML and submit Experiment via Katib SDK.

PR #2009, opened 2022-11-10
Add More Katib Presentations 2022

andreyvelich:
I've added more Katib Presentations from 2022.
Also, I've added "Event" column to the table, so we can identify the talk source easier.

Please take a look @terrytangyuan @johnugeorge @tenzen-y @kubeflow/wg-training-leads @kubeflow/wg-automl-leads


terrytangyuan (approved):
Looking great!

/lgtm

PR #2014, opened 2022-11-16
kwa(build): Introduce COMMIT file for building KWA

orfeas-k:
This PR introduces a new COMMIT file that contains the commit where Katib needs to checkout inside Kubeflow's common code in order to be built. The file was integrated in the following places as well:
 - Dockerfile
 - GH actions
 - README.md
 
This way, a developer may only update one file each time we need to checkout to a newer commit.


kimwnasptd:
Thanks @orfeas-k! This will make it easier to update the common code in one place without having to touch many different parts.

Plus, without this we would need to update the GH Actions every time which means triggering everything. Now we'll only need to update the `frontend/COMMIT` file.

From my side
/lgtm

Issue #1764, opened 2021-12-22
Dedicated logs tab for Trials


kimwnasptd:
/kind feature

**Describe the solution you'd like**
As part of https://github.com/kubeflow/katib/issues/1763, and https://github.com/kubeflow/katib/issues/1745, lets us this issue to discuss how to expose the logs for a Trial.

Looking at the docs there are many different types of workers for a Trial https://www.kubeflow.org/docs/components/katib/trial-template/#custom-resource. The K8s clients can allow the backend to fetch logs from a Pod. With this , the main question I have is how can the backend find the **Pod** for a specific Trial worker type? 
* Will there always be an annotation/label that reflects the Trial that the Pod belongs to? 
* Is this the case for all worker types?
* What could we show when we have an Argo Workflow?

I suggest that we handle each worker type separately. We can start with K8s Jobs and/or TFJobs and start adding more later on.

@d-gol @andreyvelich @johnugeorge @gaocegege 

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

d-gol:
@kimwnasptd thank you for creating this issue and organizing the effort!

Let me try to answer your questions to the best of my knowledge:

> Will there always be an annotation/label that reflects the Trial that the Pod belongs to?

Yes. There is a label `job-name` in every pod which belongs to a specific trial - https://github.com/kubeflow/common/blob/v0.4.1/pkg/apis/common/v1/interface.go#L30

`job-name` is equal to the `trial-name` - https://github.com/kubeflow/katib/blob/release-0.13/pkg/webhook/v1beta1/pod/inject_webhook.go#L120

This allows us to obtain all pods belonging to various Trial types (`Job`, `TFJob`, `PyTorchJob`...)

> Is this the case for all worker types?

Yes for `Job`, `TFJob`, `PyTorchJob`, `MXJob`, `XGBoostJob` and `MPIJob`.
For Pipelines, I would expect the same, but I would first focus on getting the logs from the other CRDs since Pipeline logs can already be accessed on the Run page.

> What could we show when we have an Argo Workflow?

We could show a link (button) that would route to the Pipelines Run page with logs, as in trials table - https://github.com/kubeflow/katib/blob/release-0.13/pkg/new-ui/v1beta1/frontend/src/app/pages/experiment-details/trials-table/trials-table.component.html#L34
Later, we could technically parse responses from the pipeline Run page and show them in Katib UI as well.
[...]

PR #2039, opened 2022-11-25
Backend for getting logs of a trial

d-gol:
Implementation of the backend for fetching logs, as a part of https://github.com/kubeflow/katib/issues/1764.  

Providing a route `/katib/fetch_trial_logs/` to obtain logs for a specific trial.  
Logs are obtained from a master pod.

### DIFF
--- a/manifests/v1beta1/components/ui/rbac.yaml
+++ b/manifests/v1beta1/components/ui/rbac.yaml
@@ -19,6 +19,14 @@ rules:
       - suggestions
     verbs:
       - "*"
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+      - pods/log
+    verbs:
+      - list
+      - get
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/manifests/v1beta1/components/ui/rbac.yaml
+++ b/manifests/v1beta1/components/ui/rbac.yaml
@@ -23,9 +23,13 @@ rules:
       - ""
     resources:
       - pods
-      - pods/log
     verbs:
       - list
+  - apiGroups:
+      - ""
+    resources:
+      - pods/log
+    verbs:
       - get
 ---
 apiVersion: v1
--- a/pkg/new-ui/v1beta1/backend.go
+++ b/pkg/new-ui/v1beta1/backend.go
@@ -604,7 +604,11 @@ func (k *KatibUIHandler) FetchTrialLogs(w http.ResponseWriter, r *http.Request) 
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
-	w.Write(response)
+	if _, err = w.Write(response); err != nil {
+		log.Printf("Write logs failed: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
 }
 
 // GetTrialLogs returns logs of a master Pod for the given job name and namespace
@@ -624,9 +628,10 @@ func getTrialLogs(k *KatibUIHandler, trialName string, namespace string) (string
 		return "", err
 	}
 
-	selectionLabel := "training.kubeflow.org/job-name=" + trialName + ",training.kubeflow.org/job-role=master"
-	if trial.Spec.RunSpec.GetKind() == "Job" {
-		selectionLabel = "job-name=" + trialName
+	selectionLabel := "training.kubeflow.org/job-name=" + trialName
+
+	for primaryKey, primaryValue := range trial.Spec.PrimaryPodLabels {
+		selectionLabel = selectionLabel + "," + primaryKey + "=" + primaryValue
 	}
 
 	podList, err := clientset.Pods(namespace).List(context.Background(), metav1.ListOptions{LabelSelector: selectionLabel})
--- a/pkg/new-ui/v1beta1/backend.go
+++ b/pkg/new-ui/v1beta1/backend.go
@@ -45,6 +45,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 func NewKatibUIHandler(dbManagerAddr string) *KatibUIHandler {
@@ -603,7 +605,57 @@ func (k *KatibUIHandler) FetchTrialLogs(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	logs, err := getTrialLogs(k, trialName, namespace)
+	user, err = IsAuthorized(consts.ActionTypeList, namespace, "pods", "", "", schema.GroupVersion{Group: "apps", Version: "v1"}, k.katibClient.GetClient(), r)
+	if user == "" && err != nil {
+		log.Printf("No user provided in kubeflow-userid header.")
+		http.Error(w, err.Error(), http.StatusUnauthorized)
+		return
+	} else if err != nil {
+		log.Printf("The user: %s is not authorized to list pods in namespace: %s \n", user, namespace)
+		http.Error(w, err.Error(), http.StatusForbidden)
+		return
+	}
+
+	trial := &trialsv1beta1.Trial{}
+	if err := k.katibClient.GetClient().Get(context.TODO(), types.NamespacedName{Name: trialName, Namespace: namespace}, trial); err != nil {
+		log.Printf("GetLogs failed: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// TODO: Use controller-runtime client instead of kubernetes client to get logs, once this is available
+	clientset, err := createKubernetesClientset()
+	if err != nil {
+		log.Printf("GetLogs failed: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	podName, err := fetchMasterPodName(clientset, trial)
+	if err != nil {
+		log.Printf("GetLogs failed: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	user, err = IsAuthorized(consts.ActionTypeGet, namespace, "pods", "log", podName, schema.GroupVersion{Group: "apps", Version: "v1"}, k.katibClient.GetClient(), r)
+	if user == "" && err != nil {
+		log.Printf("No user provided in kubeflow-userid header.")
+		http.Error(w, err.Error(), http.StatusUnauthorized)
+		return
+	} else if err != nil {
+		log.Printf("The user: %s is not authorized to list pod logs: %s in namespace: %s \n", user, podName, namespace)
+		http.Error(w, err.Error(), http.StatusForbidden)
+		return
+	}
+
+	podLogOpts := apiv1.PodLogOptions{}
+	podLogOpts.Container = trial.Spec.PrimaryContainerName
+	if trial.Spec.MetricsCollector.Collector.Kind == common.StdOutCollector {
+		podLogOpts.Container = mccommon.MetricLoggerCollectorContainerName
+	}
+
+	logs, err := fetchPodLogs(clientset, namespace, podName, podLogOpts)
 	if err != nil {
 		log.Printf("GetLogs failed: %v", err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
@@ -622,51 +674,46 @@ func (k *KatibUIHandler) FetchTrialLogs(w http.ResponseWriter, r *http.Request) 
 	}
 }
 
-// GetTrialLogs returns logs of a master Pod for the given job name and namespace
-func getTrialLogs(k *KatibUIHandler, trialName string, namespace string) (string, error) {
-	trial := &trialsv1beta1.Trial{}
-	if err := k.katibClient.GetClient().Get(context.TODO(), types.NamespacedName{Name: trialName, Namespace: namespace}, trial); err != nil {
-		return "", err
+// createKubernetesClientset returns kubernetes clientset
+func createKubernetesClientset() (*kubernetes.Clientset, error) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return nil, err
 	}
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return clientset, nil
+}
 
-	selectionLabel := consts.LabelTrialName + "=" + trialName
+// fetchMasterPodName returns name of the master pod for a trial
+func fetchMasterPodName(clientset *kubernetes.Clientset, trial *trialsv1beta1.Trial) (string, error) {
+	selectionLabel := consts.LabelTrialName + "=" + trial.ObjectMeta.Name
 	for primaryKey, primaryValue := range trial.Spec.PrimaryPodLabels {
 		selectionLabel = selectionLabel + "," + primaryKey + "=" + primaryValue
 	}
 
-	// TODO: Use controller-runtime client instead of kubernetes client to get logs, once this is available
-	cfg, err := config.GetConfig()
-	if err != nil {
-		return "", err
-	}
-	clientset, err := kubernetes.NewForConfig(cfg)
-	if err != nil {
-		return "", err
-	}
-
-	podList, err := clientset.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{LabelSelector: selectionLabel})
+	podList, err := clientset.CoreV1().Pods(trial.ObjectMeta.Namespace).List(context.Background(), metav1.ListOptions{LabelSelector: selectionLabel})
 	if err != nil {
 		return "", err
 	}
 
 	if len(podList.Items) == 0 {
-		message := `Logs for the trial could not be found.
+		return "", errors.New(`Logs for the trial could not be found.
 Was 'retain: true' specified in the Experiment definition?
-An example can be found here: https://github.com/kubeflow/katib/blob/7bf39225f7235ee4ba6cf285ecc2c455c6471234/examples/v1beta1/argo/argo-workflow.yaml#L33`
-
-		return message, nil
+An example can be found here: https://github.com/kubeflow/katib/blob/7bf39225f7235ee4ba6cf285ecc2c455c6471234/examples/v1beta1/argo/argo-workflow.yaml#L33`)
 	}
 	if len(podList.Items) > 1 {
 		return "", errors.New("More than one master replica found")
 	}
 
-	podLogOpts := apiv1.PodLogOptions{}
-	podLogOpts.Container = trial.Spec.PrimaryContainerName
-	if trial.Spec.MetricsCollector.Collector.Kind == common.StdOutCollector {
-		podLogOpts.Container = mccommon.MetricLoggerCollectorContainerName
-	}
+	return podList.Items[0].Name, nil
+}
 
-	req := clientset.CoreV1().Pods(namespace).GetLogs(podList.Items[0].Name, &podLogOpts)
+// fetchPodLogs returns logs of a pod for the given job name and namespace
+func fetchPodLogs(clientset *kubernetes.Clientset, namespace string, podName string, podLogOpts apiv1.PodLogOptions) (string, error) {
+	req := clientset.CoreV1().Pods(namespace).GetLogs(podName, &podLogOpts)
 	podLogs, err := req.Stream(context.Background())
 	if err != nil {
 		return "", err
--- a/pkg/new-ui/v1beta1/backend.go
+++ b/pkg/new-ui/v1beta1/backend.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1beta1
 
 import (
+	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"log"
@@ -28,11 +30,18 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	experimentv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/experiments/v1beta1"
+	trialsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/trials/v1beta1"
 	suggestionv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/suggestions/v1beta1"
 	api_pb_v1beta1 "github.com/kubeflow/katib/pkg/apis/manager/v1beta1"
 	consts "github.com/kubeflow/katib/pkg/controller.v1beta1/consts"
 	"github.com/kubeflow/katib/pkg/util/v1beta1/katibclient"
 	corev1 "k8s.io/api/core/v1"
+
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 func NewKatibUIHandler(dbManagerAddr string) *KatibUIHandler {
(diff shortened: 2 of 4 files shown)


pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,87 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		return "", err
+	}
+
+	trial := &trialsv1beta1.Trial{}
+	if err := k.katibClient.GetClient().Get(context.TODO(), types.NamespacedName{Name: trialName, Namespace: namespace}, trial); err != nil {
+		return "", err
+	}
+
+	selectionLabel := "training.kubeflow.org/job-name=" + trialName + ",training.kubeflow.org/job-role=master"
```
johnugeorge, line 475: How about using Trial Spec Primary Pod Labels?

https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/pod/inject_webhook.go#L129 

/cc @andreyvelich

d-gol: Agree, I can change it. Did you mean something like this?
```
selectionLabel := "training.kubeflow.org/job-name=" + trialName

for primaryKey, primaryValue := range trial.Spec.PrimaryPodLabels {
	selectionLabel = selectionLabel + "," + primaryKey + "=" + primaryValue
}
```

d-gol: Edited here: https://github.com/kubeflow/katib/pull/2039/commits/4d979d9199c0c6470aec5573c4611926b045510c

pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,92 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		return "", err
+	}
+
+	trial := &trialsv1beta1.Trial{}
+	if err := k.katibClient.GetClient().Get(context.TODO(), types.NamespacedName{Name: trialName, Namespace: namespace}, trial); err != nil {
+		return "", err
+	}
+
+	selectionLabel := "training.kubeflow.org/job-name=" + trialName
```
andreyvelich, line 479: I think you can remove `selectionLabel` variable.
Primary Pod Labels already has necessary labels to get the pods (e.g. [the default for Training Operator](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/experiments/v1beta1/constants.go#L41))
Or [the required for Argo Workflow example](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/argo/argo-workflow.yaml#L35).

pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,92 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
 	}
 }
+
+// FetchTrialLogs fetches logs for a trial in specific namespace.
+func (k *KatibUIHandler) FetchTrialLogs(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Requesting logs")
+
+	trialName := r.URL.Query()["trialName"][0]
+	namespace := r.URL.Query()["namespace"][0]
+	log.Printf("Requesting logs")
```
andreyvelich, line 441: Should you remove this ?
```suggestion
```

pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,92 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 }
+
+// FetchTrialLogs fetches logs for a trial in specific namespace.
+func (k *KatibUIHandler) FetchTrialLogs(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Requesting logs")
```
andreyvelich, line 437:
```suggestion
	log.Println("Requesting logs")
```

pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,92 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+}
+
+// GetTrialLogs returns logs of a master Pod for the given job name and namespace
+func getTrialLogs(k *KatibUIHandler, trialName string, namespace string) (string, error) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return "", err
+	}
+
+	clientset, err := corev1.NewForConfig(cfg)
```
andreyvelich, line 469: Can we use [the Controller Runtime client from the `KatibUIHandler`](https://github.com/kubeflow/katib/blob/4d979d9199c0c6470aec5573c4611926b045510c/pkg/util/v1beta1/katibclient/katib_client.go#L60-L81) to get the Pod logs ?

pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,92 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		return "", err
+	}
+
+	trial := &trialsv1beta1.Trial{}
+	if err := k.katibClient.GetClient().Get(context.TODO(), types.NamespacedName{Name: trialName, Namespace: namespace}, trial); err != nil {
+		return "", err
+	}
```
andreyvelich, line 477: Should we verify the auth here to check if user can get the Pod logs, similar to [other API calls](https://github.com/kubeflow/katib/blob/master/pkg/new-ui/v1beta1/backend.go#L108) ?
cc @apo-ger @kimwnasptd

pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,92 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+
+	podList, err := clientset.Pods(namespace).List(context.Background(), metav1.ListOptions{LabelSelector: selectionLabel})
+	if err != nil {
+		return "", err
+	}
+
+	if len(podList.Items) == 0 {
+		message := `Logs for the trial could not be found.
+Was 'retain: true' specified in the Experiment definition?
+An example can be found here: https://github.com/kubeflow/katib/blob/master/examples/v1beta1/argo/argo-workflow.yaml#L33`
```
andreyvelich, line 493: Let's use the example with commit so it will never be changed ?
```suggestion
An example can be found here: https://github.com/kubeflow/katib/blob/7bf39225f7235ee4ba6cf285ecc2c455c6471234/examples/v1beta1/argo/argo-workflow.yaml#L33`
```

pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,92 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+
+		return message, nil
+	}
+
+	podLogOpts := apiv1.PodLogOptions{}
+	podLogOpts.Container = trial.Spec.PrimaryContainerName
+	for container := range podList.Items[0].Spec.Containers {
+		if podList.Items[0].Spec.Containers[container].Name == "metrics-logger-and-collector" {
+			podLogOpts.Container = "metrics-logger-and-collector"
```
andreyvelich, line 502: You can get the constant from here: https://github.com/kubeflow/katib/blob/4d979d9199c0c6470aec5573c4611926b045510c/pkg/metricscollector/v1beta1/common/const.go#L53.

pkg/new-ui/v1beta1/backend.go
```diff
@@ -421,3 +431,92 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+		message := `Logs for the trial could not be found.
+Was 'retain: true' specified in the Experiment definition?
+An example can be found here: https://github.com/kubeflow/katib/blob/master/examples/v1beta1/argo/argo-workflow.yaml#L33`
+
+		return message, nil
+	}
+
+	podLogOpts := apiv1.PodLogOptions{}
+	podLogOpts.Container = trial.Spec.PrimaryContainerName
+	for container := range podList.Items[0].Spec.Containers {
```
andreyvelich, line 500: I think, you can skip this check.
Trial instance has [Metrics Collector spec](https://github.com/kubeflow/katib/blob/4d979d9199c0c6470aec5573c4611926b045510c/pkg/apis/controller/trials/v1beta1/trial_types.go#L47).
You can just do this: 
```go
if trial.spec.metricsCollector.collector.kind == apis.StdOut {
 podLogOpts.Container = const. MetricLoggerCollectorContainerName
}
```

andreyvelich (commented):
Thank you for implementing this @d-gol!
I left few comments.

pkg/new-ui/v1beta1/backend.go
apo-ger: @andreyvelich I think we should. This is the required authorization check before fetching the trial:

```go
user, err := IsAuthorized(consts.ActionTypeGet, namespace, consts.PluralTrial, "", trialName, trialv1beta1.SchemeGroupVersion, k.katibClient.GetClient(), r)
if user == "" && err != nil {
	log.Printf("No user provided in kubeflow-userid header.")
	http.Error(w, err.Error(), http.StatusUnauthorized)
	return
} else if err != nil {
	log.Printf("The user: %s is not authorized to delete tial: %s in namespace: %s \n", user, trialName, namespace)
	http.Error(w, err.Error(), http.StatusForbidden)
	return
}

```
Similar to what we do in [new-ui/v1beta1/hp.go](https://github.com/kubeflow/katib/blob/bd91301d36051d4cf04fe88bb3751591df4141b9/pkg/new-ui/v1beta1/hp.go#L98-L107)

pkg/new-ui/v1beta1/backend.go
d-gol: Don't we also need a selection based on the name of the trial? Without the `training.kubeflow.org/job-name` label, we would get all the master replica pods in the namespace.

pkg/new-ui/v1beta1/backend.go
d-gol: From my understanding based on this issue https://github.com/kubernetes-sigs/controller-runtime/issues/452, it is not possible to obtain logs using the controller runtime client, we need a Kubernetes client. If I'm missing something, please let me know.

tenzen-y:
@d-gol Thanks for your effort.
We merged #2047 with the master branch to mutate the trial name label to pods. So, can you rebase this PR?

pkg/new-ui/v1beta1/backend.go
d-gol: Thanks @apo-ger for the code snippet!

pkg/new-ui/v1beta1/backend.go
d-gol: Edited the label https://github.com/kubeflow/katib/pull/2047 was merged.

d-gol:
Thank you @tenzen-y and @andreyvelich, rebased it now.

pkg/new-ui/v1beta1/backend.go
d-gol: Thanks @tenzen-y, added the comment.

johnugeorge:
Thanks Dejan

/lgtm

tenzen-y (commented):
@d-gol Thanks for driving this!

/lgtm
/assign @andreyvelich

pkg/new-ui/v1beta1/backend.go
```diff
@@ -574,3 +586,99 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 }
+
+// FetchTrialLogs fetches logs for a trial in specific namespace.
+func (k *KatibUIHandler) FetchTrialLogs(w http.ResponseWriter, r *http.Request) {
+	trialName := r.URL.Query()["trialName"][0]
+	namespace := r.URL.Query()["namespace"][0]
+
+	user, err := IsAuthorized(consts.ActionTypeGet, namespace, consts.PluralTrial, "", trialName, trialsv1beta1.SchemeGroupVersion, k.katibClient.GetClient(), r)
```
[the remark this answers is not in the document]
d-gol: Yes, I agree. Added authorization checks for listing the pods and getting the logs. Had to reorganize the code a bit to fit in the additional checks, but the logic is the same. Adding as a separate commit, in the end we can squash the commits.

pkg/new-ui/v1beta1/backend.go
```diff
@@ -574,3 +588,144 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+		log.Printf("No user provided in kubeflow-userid header.")
+		http.Error(w, err.Error(), http.StatusUnauthorized)
+		return
+	} else if err != nil {
+		log.Printf("The user: %s is not authorized to get trial: %s in namespace: %s \n", user, trialName, namespace)
+		http.Error(w, err.Error(), http.StatusForbidden)
+		return
+	}
+
+	user, err = IsAuthorized(consts.ActionTypeList, namespace, "pods", "", "", schema.GroupVersion{Group: "apps", Version: "v1"}, k.katibClient.GetClient(), r)
```
andreyvelich, line 608: Should we use group version for pods similar to ConfigMaps (https://github.com/kubeflow/katib/blob/365154b2ebfcdc5fdacdbd60c4dc813fc9c60c39/pkg/new-ui/v1beta1/backend.go#L359)  ?
`corev1.SchemeGroupVersion` ?

manifests/v1beta1/components/ui/rbac.yaml
```diff
@@ -19,6 +19,18 @@ rules:
...
+    resources:
+      - pods
+    verbs:
+      - list
+  - apiGroups:
+      - ""
+    resources:
+      - pods/log
+    verbs:
+      - get
```
andreyvelich, line 33: I think, we should also update the `kubeflow-user-edit` role: https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-with-kubeflow/kubeflow-katib-roles.yaml#L22-L36 with this permission.

cc @kimwnasptd

pkg/new-ui/v1beta1/backend.go
```diff
@@ -574,3 +588,144 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	podName, err := fetchMasterPodName(clientset, trial)
+	if err != nil {
+		log.Printf("GetLogs failed: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	user, err = IsAuthorized(consts.ActionTypeGet, namespace, "pods", "log", podName, schema.GroupVersion{Group: "apps", Version: "v1"}, k.katibClient.GetClient(), r)
```
andreyvelich, line 641: @kimwnasptd @apo-ger Please review this change. Does it sound good for you ?

pkg/new-ui/v1beta1/backend.go
d-gol: Thank you, edited!

manifests/v1beta1/components/ui/rbac.yaml
d-gol: Edited this as well.

pkg/new-ui/v1beta1/backend.go
d-gol: Let's wait for the review on this one.

apo-ger: Hey @andreyvelich @d-gol! The changes look good.

d-gol: Thanks @apo-ger!

pkg/new-ui/v1beta1/backend.go
```diff
@@ -574,3 +586,144 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 }
+
+// FetchTrialLogs fetches logs for a trial in specific namespace.
+func (k *KatibUIHandler) FetchTrialLogs(w http.ResponseWriter, r *http.Request) {
+	trialName := r.URL.Query()["trialName"][0]
+	namespace := r.URL.Query()["namespace"][0]
```
apo-ger, line 593: @d-gol Just a minor comment here. Maybe add checks here similar to what we do [here](https://github.com/d-gol/katib/blob/4afe97603ef2ffbe868c42371cf6716e431754bd/pkg/new-ui/v1beta1/hp.go#L196-L213)

pkg/new-ui/v1beta1/backend.go
```diff
@@ -574,3 +586,160 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+		log.Printf("No trialName provided in Query parameters! Provide a 'trialName' param")
+		err := errors.New("no 'trialName' provided")
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	trialName := trialNames[0]
+	namespace := namespaces[0]
+
+	user, err := IsAuthorized(consts.ActionTypeGet, namespace, consts.PluralTrial, "", trialName, trialsv1beta1.SchemeGroupVersion, k.katibClient.GetClient(), r)
```
kimwnasptd, line 611: Slightly orthogonal to the PR, but I think the function signature `user, err = IsAuthorized(...)` is not ideal. We end up getting the same information and do duplicate checks on errors depending on the user value.

Why not have a distinct function for getting the current user and do this check once? Then `IsAuthorized(user, ...)` will only be responsible for the SubjectAccessReviews check.

Or at least for this PR, we could only check if the returned user is not "" only once, the first time we call this function

pkg/new-ui/v1beta1/backend.go
```diff
@@ -574,3 +586,160 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+	user, err = IsAuthorized(consts.ActionTypeList, namespace, corev1.ResourcePods.String(), "", "", corev1.SchemeGroupVersion, k.katibClient.GetClient(), r)
+	if user == "" && err != nil {
+		log.Printf("No user provided in kubeflow-userid header.")
+		http.Error(w, err.Error(), http.StatusUnauthorized)
+		return
+	} else if err != nil {
+		log.Printf("The user: %s is not authorized to list pods in namespace: %s \n", user, namespace)
+		http.Error(w, err.Error(), http.StatusForbidden)
+		return
+	}
```
kimwnasptd, line 631: I'd be OK even if we'd remove the check for whether the user can list Pods. IMO just checking if the User has permissions to:
1. GET the specific Trial
2. GET the logs of each specific underlying Pod

would be enough. Even in plain K8s a user could have permissions to GET the logs of a Pod, but without having permissions to list all the Pods in the namespace

pkg/new-ui/v1beta1/backend.go
```diff
@@ -574,3 +586,160 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		log.Printf("Marshal logs failed: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	if _, err = w.Write(response); err != nil {
+		log.Printf("Write logs failed: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
kimwnasptd, line 688: I understand that the backend will end up returning a string value, that is the logs of the "master" Pod of the Trial?

Question 1: Is there always a master pod or do we expect this to fail for some cases?

kimwnasptd, line 688: Question 2: Would we expect in the future to return logs from other worker pods?

If that's the case I'd propose that the backend actually returns a JSON type response like
```
logs: {
    master: "..."
}
```

kimwnasptd (commented):
@d-gol @andreyvelich apologies for the late reply. Mostly have some questions around the return type and the structure of the response data.

pkg/new-ui/v1beta1/backend.go
d-gol: Agree, we are checking twice the user. I think we can proceed fixing the auth after we merge this PR? Then we can have a separate PR to improve the authentication in the entire file. Or we can do the other way around?

pkg/new-ui/v1beta1/backend.go
d-gol: Thanks @kimwnasptd , I can change this.

pkg/new-ui/v1beta1/backend.go
d-gol: I think the consensus is there will be a master pod. Please correct me if I'm wrong @johnugeorge @andreyvelich

d-gol: This is a good idea, if we want to change in the future. @johnugeorge @andreyvelich what do you think?

tenzen-y:
@d-gol Can you rebase this since we have merged #2064 to fix CI into the master branch?

pkg/new-ui/v1beta1/backend.go
andreyvelich: @kimwnasptd We are using [PrimaryPodLabel to identify the pod for getting the logs](https://github.com/kubeflow/katib/pull/2039/files#diff-368f67643a9976088bc884f3f659f60407978206a4142bd75b0902411f2d4e72R707)

andreyvelich: I like @kimwnasptd idea, let's add Primary Pod Label to the JSON response.

d-gol: Hey @andreyvelich did you mean the result to be in the form: 
``` 
logs: {
    master: "..."
}
```
or something else?
We can have multiple primary pod labels, did you mean to also add them as key value pairs?

johnugeorge: master is bit confusing term. Eg: There can be job with just workers where worker0 acts as the master. If we really need to add pod info, pod name might be better.

andreyvelich: @d-gol It's not always "master" label for the pod that we get labels.
For example, for Argo we get pod with [`katib.kubeflow.org/model-training: true` label](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/argo/argo-workflow.yaml#L35).
Maybe pod name to include in the response make sense as @johnugeorge mentioned.

andreyvelich: I am ok with creating separate issue to track improvements for this API response (e.g. add trial name to the response).
So we can merge this PR and unblock UI team to start working on the UI changes to have this feature in the next release.
What do you think @d-gol @kimwnasptd @johnugeorge ?

johnugeorge: I agree. +1 to merge

d-gol: I agree to merge it, and later we can improve the API response with more information if needed. So again, to clarify, we want to merge this PR with a simple string response (current implementation)? Or in the form of json, like below?
```
{
    "pod_name": logs
}
```

johnugeorge:
@d-gol We can merge this. Can you create an issue to track the Json response discussion ? Also, please do a rebase

d-gol:
@johnugeorge great, I will rebase it. Also, I can remove the check for listing pods, as Kimonas suggested - https://github.com/kubeflow/katib/pull/2039#discussion_r1048278596

d-gol:
Rebased @johnugeorge  @andreyvelich should I squash the commits, or will you do it on merge? There were many commits during the discussions, but I think they can all be squashed.

andreyvelich:
> There were many commits during the discussions, but I think they can all be squashed.

@d-gol They will be automatically squashed by tide before the merge.
Thank you for the updates!
/lgtm

/assign @johnugeorge @tenzen-y @kimwnasptd

d-gol:
@andreyvelich great, thank you!

johnugeorge:
@d-gol Can you try a failing e2e test locally?

d-gol:
@johnugeorge sure, checking it.

pkg/new-ui/v1beta1/backend.go
```diff
@@ -574,3 +586,149 @@ func (k *KatibUIHandler) FetchTrial(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, err.Error(), http.StatusUnauthorized)
+		return
+	} else if err != nil {
+		log.Printf("The user: %s is not authorized to get trial: %s in namespace: %s \n", user, trialName, namespace)
+		http.Error(w, err.Error(), http.StatusForbidden)
+		return
+	}
+
+	trial := &trialsv1beta1.Trial{}
+	if err := k.katibClient.GetClient().Get(context.TODO(), types.NamespacedName{Name: trialName, Namespace: namespace}, trial); err != nil {
```
tenzen-y, line 623: nit: 
Do you have any reason for using `context.TODO()`?
It might be better to use `context.Background()` or to take context out of `r *http.Request` (r.Context()).

tenzen-y (commented):
@d-gol Thanks for implementing this powerful feature!

LGTM
Although I wonder why our E2E failed.

pkg/new-ui/v1beta1/backend.go
d-gol: Thanks a lot @tenzen-y, and thank you for all the help!

Actually no particular reason, I will change now to `context.Background()`.

For E2E, strange, they were passing on each previous commit, this one today should not affect them.
Checking locally as well.

tenzen-y:
@johnugeorge @andreyvelich @d-gol Same errors seem to occur for `E2E Test with Katib UI, random search, and postgres / e2e` and `E2E Test with mxnet-mnist / e2e` in #2060 and #2067.

tenzen-y:
It seems that errors are caused by mxnet-mnist image. You can reproduce by `ytenzen/mxnet-mnist:debug-error`.

tenzen-y:
ASAP, I will create a PR to fix this issue.

tenzen-y:
Blocked by: #2070

tenzen-y:
@d-gol Can you rebase since we fixed CI?

d-gol:
@tenzen-y done, thank you for fixing the CI!

tenzen-y (approved):
@d-gol Thank you for the update!
/lgtm
/approve

d-gol:
@tenzen-y great, thanks a lot! 
And thanks everyone for all the help with this @andreyvelich @johnugeorge @kimwnasptd @apo-ger

Issue #1638, opened 2021-08-24
Support Istio-sidecar handling


anencore94:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

I'd like to run katib-experiment with istio-sidecar injection.
I guess the reason why katib induce user to run katib-experiment with istio-sidecar injection "false" was something like this:
- https://github.com/kubeflow/katib/issues/955
- https://github.com/kubernetes/kubernetes/issues/65502
- https://github.com/istio/istio/issues/6324
- https://medium.com/redbox-techblog/handling-istio-sidecars-in-kubernetes-jobs-c392661c4af7

Thus, the following problem is maybe the main issue.
A) katib's metrics-logger-and-collector sidecar container should know when the istio-proxy sidecar container started
B) metrics-logger-and-collector sidecar container should kill when the main container finished well or not.

---
By the way I think changing initial-logic and end-logic in metrics-collector image, katib could handle this.

Maybe something like this;
1) Handling issue B above from metrics-collector (https://github.com/kubeflow/katib/blob/eea6ada974ee11daf13fc66627696c458b4bad77/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L339)
```go
func killIstioProxySidecar() {
	klog.Infof("Let's kill istio-proxy sidecar")

	// curl -fsI -X POST http://localhost:15020/quitquitquit
	resp, err := http.Post("http://localhost:15020/quitquitquit", "text/plain", bytes.NewBufferString(""))
	if err != nil {
		// TODO 
		klog.Infof("there is no istio-proxy sidecar")
	}

	defer resp.Body.Close()
	klog.Infof("istio-proxy sidecar killed")
}

func main() {
	defer killIstioProxySidecar()

	flag.Var(&stopRules, "stop-rule", "The list of early stopping stop rules")
	flag.Parse()
	klog.Infof("Trial Name: %s", *trialName)
.....
```

2) Also, pass the process from istio-proxy sidecar nearby [here](https://github.com/kubeflow/katib/blob/eea6ada974ee11daf13fc66627696c458b4bad77/pkg/metricscollector/v1beta1/common/pns.go#L94)

```go
		// Ignore istio-sidecar processes
                // TODO this must be more sophisticated to catch istio-proxy's process
		if strings.Contains(cmdline, "envoy") || strings.Contains(cmdline, "sidecar") {
			fmt.Printf("Ignore istio-sidecar container's processes")
			continue
		}
```

[...]

PR #2041, opened 2022-11-29
Update manifests to enable authorization check mechanisms for katib-UI in kubeflow mode

apo-ger:
This is a follow-up PR from: https://github.com/kubeflow/katib/pull/1983

Currently, in kubeflow mode, any workload/pod can set the `kubeflow-userid` header and communicate with the backend of the Katib UI. Since the code for securing the backend is merged we need to provide a way to enable this feature.

Changes to `install-with-kubeflow` manifests:

- Enable Istio sidecar injection for katib-ui component
- Introduce AuthorizationPolicies: 
  - `katib-ui`: Allow traffic only from the `istio-ingressgateway` towards the UI backend. We can trust `IGW` as we are sure it's setting the user header correctly.
- Enable authz checks for `katib-ui` backend (set `APP_DISABLE_AUTH` ENV var to false)
- Update `ClusterRole` to provide RBAC persmissions for the `katib-ui` so it can create `SubjectAccessReview` objects

Refs: https://github.com/kubeflow/katib/issues/1547

### DIFF
--- a/manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
+++ b/manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
@@ -7,6 +7,7 @@ resources:
   # Kubeflow Katib components.
   - kubeflow-katib-roles.yaml
   - ui-virtual-service.yaml
+  - istio-authorizationpolicy.yaml
 images:
   - name: docker.io/kubeflowkatib/katib-controller
     newName: docker.io/kubeflowkatib/katib-controller
@@ -20,6 +21,32 @@ images:
 
 patchesStrategicMerge:
   - patches/remove-namespace.yaml
+
+patches:
+  # Extend RBAC permission list of katib-ui so it can
+  # create SubjectAccessReview resources.
+  - target:
+      kind: ClusterRole
+      name: katib-ui
+      group: rbac.authorization.k8s.io
+      version: v1
+    path: patches/ui-rbac.yaml
+  # Enable RBAC authz checks in UI's backend.
+  - target:
+      version: v1
+      kind: Deployment
+      name: katib-ui
+    patch: |-
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
+        value:
+          name: APP_DISABLE_AUTH
+          value: "false"
+  # Allow istio sidecar injection in katib-UI Pod.
+  - target:
+      kind: Deployment
+      name: katib-ui
+    path: patches/istio-sidecar-injection.yaml
 
 vars:
   - fieldref:
--- a/manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
+++ b/manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
@@ -36,12 +36,7 @@ patches:
       version: v1
       kind: Deployment
       name: katib-ui
-    patch: |-
-      - op: add
-        path: /spec/template/spec/containers/0/env/-
-        value:
-          name: APP_DISABLE_AUTH
-          value: "false"
+    path: patches/enable-ui-authz-checks.yaml
   # Allow istio sidecar injection in katib-UI Pod.
   - target:
       kind: Deployment
--- a/pkg/new-ui/v1beta1/backend.go
+++ b/pkg/new-ui/v1beta1/backend.go
@@ -33,6 +33,7 @@ import (
 	experimentv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/experiments/v1beta1"
 	suggestionv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/suggestions/v1beta1"
 	trialsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/trials/v1beta1"
+	trialv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/trials/v1beta1"
 	api_pb_v1beta1 "github.com/kubeflow/katib/pkg/apis/manager/v1beta1"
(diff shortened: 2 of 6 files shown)


apo-ger:
@kimwnasptd  @andreyvelich @johnugeorge In this PR I introduce manifests changes only for the `katib-ui` component. However we could go one step further and secure the rest of components as well. 

This would require at least an `AuthorizationPolicy` for each katib component to allow traffic without the `kubeflow-userid` header set (to avoid impersonations). 

Furthermore, in some cases we could apply narrower policies for hardening the security even more (for example allow only  `katib-db-manager` to talk to `katib-mysql` for storing/fetching metric logs).

As mentioned in https://github.com/kubeflow/katib/pull/1983#issuecomment-1320183507 and https://github.com/kubeflow/katib/pull/1983#issuecomment-1320236489 we need to test sidecar changes carefully. I can help on this!

johnugeorge:
@apo-ger We don't have a mechanism to test these changes. Did you test all workflows from Kubeflow dashboard?

apo-ger:
Hi @johnugeorge, apologies for the late reply.

>We don't have a mechanism to test these changes. Did you test all workflows from Kubeflow dashboard?

Enabling Istio sicecar injection for all Katib components would require us to introduce appropriate `AuthorizationPolices` to ensure nothing breaks. More specifically:

- `katib-controller`: Allow traffic to the webhook port in order to let the K8s api server send traffic to this endpoint.
  - When an `Experiment` CR is submitted, the Katib `Experiment` mutating and validating webhook is called to set the default values for the `Experiment` CR and validate the CR separately.
- `katib-ui`: Allow traffic only from the `istio-ingressgateway` towards the UI backend
  - https://github.com/kubeflow/katib/pull/1983
  - We can trust `IGW` as we are sure it's setting the user header correctly.
- `katib-mysql`: The db manager's persistence layer. Allow traffic only from db-manager. 
- `katib-db-manager`: Allow `ALL` traffic as Katib jobs are currently not supporting Istio sidecar injection.
   - This is because trial pods are part of a Job, existing issue for this: https://github.com/kubeflow/katib/issues/1638
   
I have tested that the above `AuthorizationPolicies` are sufficient

johnugeorge:
Can you fix lint? 

Also, which rule is responsible for `katib-mysql`?  `The db manager's persistence layer. Allow traffic only from db-manager.`

manifests/v1beta1/components/controller/controller.yaml
```diff
@@ -18,7 +18,6 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/port: "8080"
-        sidecar.istio.io/inject: "false"
```
johnugeorge, line 21: We need this flag added for standalone mode

apo-ger: I've updated the PR!

andreyvelich:
@apo-ger Please can we post-pone changes for other components (`katib-controller`, `katib-db-manager`, `katib-mysql`) to the next release ? 
I think we should be very careful with enabling istio proxy for Katib Controller, etc.
E.g. When Early Stopping is enabled, early stopping container queries data from the Katib DB: https://github.com/kubeflow/katib/blob/master/pkg/earlystopping/v1beta1/medianstop/service.py#L151.
/hold for the review

johnugeorge:
@apo-ger Can you update with Katib-UI changes in this PR and take other changes in a separate PR?

apo-ger:
I updated this PR to keep only the changes related to the katib-ui component as mentioned in https://github.com/kubeflow/katib/pull/2041#issuecomment-1398812885

pkg/new-ui/v1beta1/backend.go
```diff
@@ -33,6 +33,7 @@ import (
 	experimentv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/experiments/v1beta1"
 	suggestionv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/suggestions/v1beta1"
 	trialsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/trials/v1beta1"
+	trialv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/trials/v1beta1"
```
tenzen-y, line 36: Do we need this dependency?
Probably, we can use `trialsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/trials/v1beta1"` above.

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
```diff
@@ -21,6 +22,32 @@ images:
...
+  - target:
+      version: v1
+      kind: Deployment
+      name: katib-ui
+    patch: |-
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
+        value:
+          name: APP_DISABLE_AUTH
+          value: "false"
```
tenzen-y, line 44: Should we move this patch to a separate file, the same as other patches?

https://github.com/kubeflow/katib/tree/master/manifests/v1beta1/installs/katib-leader-election/patches

tenzen-y:
@apo-ger Thanks for submitting this PR.
Can you update the PR description? The correct description helps to track changes.

manifests/v1beta1/installs/katib-with-kubeflow/kustomization.yaml
apo-ger: Fixed!

tenzen-y:
LGTM
/assign @johnugeorge @andreyvelich

tenzen-y:
@andreyvelich PTAL.

andreyvelich (approved):
Thanks a lot for these updates @apo-ger!
/lgtm
/approve

PR #2045, opened 2022-11-30
Remove Certificate Chain from Cert Generator

andreyvelich:
I removed Certificate Chain from our Cert generator.
I think having chain is not necessary for Kubernetes Webhooks. I was able to create Katib Experiments in my environment with these changes.
Also, we can leave only Full Service Domain in our DNS names.

Do you have any objections @tenzen-y @johnugeorge @shaowei-su ?

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -58,7 +58,7 @@ make generate
 Below is a list of command-line flags accepted by Katib controller:
 
 | Name                            | Type                      | Default                       | Description                                                                                                            |
-|---------------------------------|---------------------------|-------------------------------|------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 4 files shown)


tenzen-y (commented):
@andreyvelich Thanks for refactoring the katib-cert-generator!

> Do you have any objections @tenzen-y @johnugeorge @shaowei-su ?

I agree with you since, according to the x509 Go library documentation, we can create the self-signed certificate by specifying the same certificates in the `x509.CreateCertificate` function.

> The certificate is signed by parent. If parent is equal to template then the certificate is self-signed. 

https://pkg.go.dev/crypto/x509#CreateCertificate

Also, can you update the docs for the katib-cert-generator?

https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md#katib-cert-generator

tenzen-y (commented):
@andreyvelich Looks great! Thanks for the update!

/lgtm

PR #2047, opened 2022-12-01
Add Trial Labels During Pod Mutation

andreyvelich:
As we discussed on the recent AutoML WG meeting, we should add Trial labels to the pod during mutation.
We need it to get Trial logs from the Katib UI backend: https://github.com/kubeflow/katib/pull/2039.

cc @kimwnasptd @johnugeorge @tenzen-y @d-gol


tenzen-y (commented):
@andreyvelich  Thanks for this work!  This feature makes helpful users!

/lgtm

Issue #2022, opened 2022-11-18
[SDK] Create API to get Trial metrics from Katib DB


andreyvelich:
/kind feature
/area sdk

Our Katib Python SDK doesn't have an API to get Trial metrics from Katib DB.
Currently, user can see the Trial metrics only using Katib UI.
We should give an ability to query metrics using [`GetObservationLog` gRPC API](https://github.com/kubeflow/katib/blob/4a2db414d85f29f17bc8ec6ff3462beef29585da/pkg/apis/manager/v1beta1/api.proto#L23) via Katib SDK.

From the security perspective user can run this gRPC API from any namespace and any experiment since our DB Manager doesn't have any auth checks, right ?
Should we investigate how to improve user isolation for Katib ("multi-user mode feature") ?
One solution could be to use Istio to allow traffic only from the appropriate user, as @apo-ger mentioned here: https://github.com/kubeflow/katib/pull/1983#issuecomment-1319674570.

What do you think @johnugeorge @gaocegege @tenzen-y @anencore94 @kimwnasptd @apo-ger ?

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

kimwnasptd:
@andreyvelich that's a great feature! 

Regarding the authnz part, I think this discussion will revolve around having programmatic client support for the DB Manager API Server. This is the same with how KFP allows Pods from other namespaces to use its API Server to perform CRUD tasks https://github.com/kubeflow/pipelines/issues/5138. 

And this is done by:
1. Allowing everyone to talk to the DB Manager, but **without** setting the `kubeflow-userid` header (to avoid impersonations).
    * The `istio-ingressgateway` should still be able to set this header though, since we trust that it's setting the user header correctly
    * This is the AuthorizationPolicy in KFP for this https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/istio-authorization-config.yaml#L36-L38
3. The DB Manager will drop any requests that are not authenticated 
4. In-cluster pods that will need to talk to the DB Manager will need to provide an audience scoped ServiceAccount token
5. The DB Manager will need to validate the token
6. The DB Manager will then extract the identity (ServiceAccount name) from that token and perform a SubjectAccessReview

Then there's also the discussion on how to use the ServiceAccount tokens from outside the cluster. But this is a next step once we have the above in-cluster behavior working

andreyvelich:
[...]

PR #2050, opened 2022-12-05
[SDK] Get Trial Metrics from Katib DB

andreyvelich:
Related: https://github.com/kubeflow/katib/issues/2022.

I've added API to get Trial metrics from the Katib DB using Katib DB Manager gRPC API.
Usage:
```python
cl = KatibClient()
trial_name = cl.get_experiment(name="test-experiment-1")["status"]["succeededTrialList"][0]
cl.get_trial_metrics(name=trial_name)[0]

"
time_stamp: "2022-12-05T14:42:08Z"
metric {
  name: "Train-accuracy"
  value: "0.112390"
}
"
```

Since we don't publish our gRPC as Python package, we can't modify [the requirements for our Katib SDK](https://github.com/andreyvelich/katib/blob/b6755bbdc4d73eee18f8025c0bc68b5d90c8747b/sdk/python/v1beta1/setup.py#L19-L26).
In the `setup.py` script [I copied the Katib gRPC APIs file to the appropriate location](https://github.com/kubeflow/katib/blob/b6755bbdc4d73eee18f8025c0bc68b5d90c8747b/sdk/python/v1beta1/setup.py#L30-L36), so we can use it in the Katib client.

What do you think about it @johnugeorge @gaocegege @tenzen-y @kimwnasptd @anencore94  ?

In the long-term (if we are going to create Katib API Server) we might want to consider publishing the Katib API Server Python APIs to PyPI, so other users can make integration on top of it (what @anencore94 mentioned here: https://github.com/kubeflow/katib/issues/2022#issuecomment-1336863064).

Kubeflow Pipelines team also publish their [python `kfp-server-api ` to the public registry](https://github.com/kubeflow/pipelines/tree/6e364d35b1e66e6b7a7985b2364d29e52ba316ce/backend/api/v2beta1/python_http_client).

/assign @johnugeorge @kimwnasptd @tenzen-y @anencore94 

/hold for review

### DIFF
--- a/sdk/python/v1beta1/setup.py
+++ b/sdk/python/v1beta1/setup.py
@@ -25,13 +25,13 @@ REQUIRES = [
     "grpcio==1.41.1",
 ]
 
-katib_grpc_api_file = "kubeflow/katib/katib_api_pb2.py"
+katib_grpc_api_file = "../../../pkg/apis/manager/v1beta1/python/api_pb2.py"
 
 # Copy Katib gRPC Python APIs to use it in the Katib SDK Client.
-# We need to copy this file only on the SDK building stage, not on installation stage.
-if not os.path.exists(katib_grpc_api_file):
+# We need to always copy this file only on the SDK building stage, not on SDK installation stage.
+if os.path.exists(katib_grpc_api_file):
     shutil.copy(
-        "../../../pkg/apis/manager/v1beta1/python/api_pb2.py", katib_grpc_api_file,
+        katib_grpc_api_file, "kubeflow/katib/katib_api_pb2.py",
     )
 
 setuptools.setup(
--- a/sdk/python/v1beta1/setup.py
+++ b/sdk/python/v1beta1/setup.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+import shutil
 import setuptools
 
 REQUIRES = [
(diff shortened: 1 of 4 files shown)


sdk/python/v1beta1/setup.py
```diff
@@ -20,8 +22,18 @@
...
     "urllib3>=1.15.1",
     "kubernetes>=23.6.0",
+    "grpcio==1.41.1",
 ]
 
+katib_grpc_api_file = "kubeflow/katib/katib_api_pb2.py"
+
+# Copy Katib gRPC Python APIs to use it in the Katib SDK Client.
+# We need to copy this file only on the SDK building stage, not on installation stage.
+if not os.path.exists(katib_grpc_api_file):
```
johnugeorge, line 32: How do we ensure files are uptodate and in sync?

andreyvelich: @johnugeorge It's a good point.

Maybe instead of checking that file `kubeflow/katib/katib_api_pb2.py` exists, we should check if `../../../pkg/apis/manager/v1beta1/python/api_pb2.py` file exists and always replace it with the newest ?

```python
katib_grpc_api_file = "../../../pkg/apis/manager/v1beta1/python/api_pb2.py"

# Copy Katib gRPC Python APIs to use it in the Katib SDK Client.
# We need to copy this file only on the SDK building stage, not on installation stage.
if os.path.exists(katib_grpc_api_file):
    shutil.copy(
        katib_grpc_api_file, "kubeflow/katib/katib_api_pb2.py",
    )
```

In that case, on the SDK building stage we always copy the newest gRPC APIs file ?

Any other suggestions @johnugeorge @tenzen-y ?

johnugeorge: Yeah. I think, "always copy" is the safer solution

andreyvelich:
@johnugeorge @tenzen-y I've made the change. It would be great if you could take a look.

sdk/python/v1beta1/setup.py
tenzen-y: > In that case, on the SDK building stage we always copy the newest gRPC APIs file?

sgtm

tenzen-y (commented):
@andreyvelich Sorry for the late response.
Thanks for implementing this! This feature looks great!
/lgtm
/hold cancel

PR #2065, opened 2022-12-15
frontend: Enable actions in experiment graph

elenzio9:
This PR enables actions in the experiment graph and specifically:
- Restore action which resets the graph
- Save image action

image


kimwnasptd:
Nice!

/lgtm
/approve

Issue #2062, opened 2022-12-14
delete ns label validation


zhixian82:
/kind bug

the metrics collector sidecar injection is in job level for now，so the ns label validation should be removed？

https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/experiment/validation_webhook.go#L83

PR #1303, opened 2020-08-18
Fix Pod's ownership to inject metrics collector

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1271.

I refactored `getKatibJob` to correctly inject metrics collector.
`getKatibJob` is looking for Trial ownership in all pod's nested objects recursively.
These objects should be created ones Pod is mutated.

Also, I moved functions that are not belong to `sidecarInjector` struct to `utils` to be consistent with [Validation webhook](https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/experiment/validation_webhook.go).

/assign @sperlingxx @gaocegege @johnugeorge

PR #2066, opened 2022-12-20
fix: remove unneccessary ns validation

zhixian82:
https://github.com/kubeflow/katib/issues/2062


pkg/webhook/v1beta1/experiment/validation_webhook.go
```diff
@@ -79,25 +78,6 @@ func (v *ExperimentValidator) Handle(ctx context.Context, req admission.Request)
 		}
 	}
 
-	// After metrics collector sidecar injection in Job level done, delete validation for namespace labels
```
tenzen-y, line 82: I'm not sure what `sidecar injection in Job level` means. This comment seemed to add in #828 3 years ago.
After that, @andreyvelich refactored the `getKatibJob` function (maybe, for job level injection?) in #1303.

@andreyvelich @johnugeorge Do you remember what this comment means?

johnugeorge: I don't think, this comment is still valid.

 In Job level, we look at type of Job(TFJob, PytorchJob etc) and decide the job pod which needs metrics collector injection. In Pod level, we check pod level labels and decide if we need to inject metrics collector or not.  Right now, we are doing pod level where inject metric collector in the pods that  contain PrimaryPodLabels or not.

johnugeorge:
The reason behind namespace selector was that we need to identify namespaces that can run katib workloads.  In order to reduce the performance effect(due to webhooks) for the whole cluster which can have non Kubeflow/Katib workloads as well, we were specifically looking for namespaces that has `katib.kubeflow.org/metrics-collector-injection` label. This change will have  performance issues in a shared cluster

pkg/webhook/v1beta1/experiment/validation_webhook.go
tenzen-y: I see. Thanks for clarifying! @johnugeorge

andreyvelich: @tenzen-y I guess the idea behind Job level mutation was described here: https://github.com/kubeflow/katib/blob/master/docs/proposals/metrics-collector.md#mutating-webhook.
But since we dynamically support any kind of Kubernetes resources, we mutate only in the Pod level.

operators/katib-controller/src/webhooks.yaml
```diff
@@ -63,9 +63,6 @@ webhooks:
         name: katib-controller
         namespace: kubeflow
         path: /mutate-pod
-    namespaceSelector:
-      matchLabels:
-        katib.kubeflow.org/metrics-collector-injection: enabled
```
andreyvelich, line 68: We shouldn't remove this `namespaceSelector`.
Otherwise, Katib Mutation Webhook will be executing in every Kubernetes namespace.

pkg/webhook/v1beta1/experiment/validation_webhook.go
tenzen-y: Yes, I agree. I thought we cannot inject sidecars in Job level since we support any K8s resources, as you say.
Maybe, it would be good to remove the comment to avoid confusion.

zhixian82: @johnugeorge @andreyvelich @tenzen-y Thanks for the explanation! I will close this pr

Issue #2061, opened 2022-12-14
stop trials when experiment completed


zhixian82:
/kind bug

**What steps did you take and what happened:**
create an experiment with: 
```
MaxTrialCount: 12
MaxFailedTrialCount: 3
```
when 3 trials failed, the experiment status was set to failed, bug some trials still in running status

**What did you expect to happen:**

when experiment completed(successed/failed), stop running trials

- Katib version (check the Katib controller image version): v0.14.0

johnugeorge:
To be more clear, running trials are the ones which got started before experiment is in completed state?

zhixian82:
> To be more clear, running trials are the ones which got started before experiment is in completed state?

yes

PR #2068, opened 2022-12-21
fix: cleanup trial resources when experiment completed

zhixian82:
Fixes #https://github.com/kubeflow/katib/issues/2061

### DIFF
--- a/pkg/controller.v1beta1/experiment/experiment_controller.go
+++ b/pkg/controller.v1beta1/experiment/experiment_controller.go
@@ -46,6 +46,7 @@ import (
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/experiment/manifest"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/experiment/suggestion"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/experiment/util"
+	"github.com/kubeflow/katib/pkg/controller.v1beta1/trial"
 )
 
 const (
@@ -189,6 +190,11 @@ func (r *ReconcileExperiment) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	if instance.IsCompleted() {
+		if err := r.cleanupTrialResources(instance); err != nil {
+			logger.Error(err, "cleanup Trial Resources error")
+			return reconcile.Result{}, err
+		}
+
 		// Cleanup suggestion after Experiment is finished.
 		// If ResumePolicy = Never or ResumePolicy = FromVolume, delete suggestion deployment, service and mark suggestion status succeeded.
 		if instance.Spec.ResumePolicy == experimentsv1beta1.NeverResume || instance.Spec.ResumePolicy == experimentsv1beta1.FromVolume {
@@ -276,8 +282,7 @@ func (r *ReconcileExperiment) ReconcileExperiment(instance *experimentsv1beta1.E
 			return err
 		}
 	}
-	reconcileRequired := !instance.IsCompleted()
-	if reconcileRequired {
+	if !instance.IsCompleted() {
 		return r.ReconcileTrials(instance, trials.Items)
 	}
 
@@ -504,4 +509,36 @@ func (r *ReconcileExperiment) ReconcileSuggestions(instance *experimentsv1beta1.
 		}
 	}
 	return assignments, nil
+}
+
+func (r *ReconcileExperiment) cleanupTrialResources(instance *experimentsv1beta1.Experiment) error {
+
+	logger := log.WithValues("Experiment", types.NamespacedName{Name: instance.GetName(), Namespace: instance.GetNamespace()})
+	trials := &trialsv1beta1.TrialList{}
+	labels := map[string]string{consts.LabelExperimentName: instance.Name}
+
+	if err := r.List(context.TODO(), trials, client.InNamespace(instance.Namespace), client.MatchingLabels(labels)); err != nil {
+		logger.Error(err, "Trial List error")
+		return err
+	}
+
+	timeNow := metav1.Now()
+	for i := range trials.Items {
+		if trials.Items[i].IsCompleted() {
+			continue
+		}
+		trials.Items[i].MarkTrialStatusKilled(trial.TrialKilledReason, "experiment already completed")
+		instance.Status.CompletionTime = &timeNow
+		eventMsg := fmt.Sprintf("Job %v has killed", trials.Items[i].Spec.RunSpec.GetName())
+		r.recorder.Eventf(instance, corev1.EventTypeNormal, trial.TrialKilledReason, eventMsg)
+		logger.Info("Trial status changed to Killed", "trial name", trials.Items[i].Name)
+
+		if err := r.Status().Update(context.TODO(), &trials.Items[i]); err != nil {
+			logger.Error(err, "Trials status updated failed", "trial name", trials.Items[i].Name)
+			return err
+		}
+	}
+
+	return nil
+
 }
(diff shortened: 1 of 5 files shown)


pkg/controller.v1beta1/experiment/experiment_controller.go
```diff
@@ -505,3 +511,35 @@ func (r *ReconcileExperiment) ReconcileSuggestions(instance *experimentsv1beta1.
...
+		logger.Error(err, "Trial List error")
+		return err
+	}
+
+	timeNow := metav1.Now()
+	for i := range trials.Items {
+		if trials.Items[i].IsCompleted() {
+			continue
+		}
+		trials.Items[i].MarkTrialStatusKilled(trialsv1beta1.TrialKilledReason, "experiment already completed")
```
johnugeorge, line 531: What do you mean by Trial Killed status?  Underlying job will be still running ?

zhixian82: the trial status will turn to killed， and whether the underlying  job still running or not was decided by the filed `trialSpec.RetainRun`.

code: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller.go#L303

johnugeorge: Isn't it misleading saying that Trial is killed but underlying job is still running?

zhixian82: > Isn't it misleading saying that Trial is killed but underlying job is still running?

I think this is an another question.  Maybe we can create an issue to discuss this case.

my pr was trying to change  trial status when experiment completed.

johnugeorge: /cc @tenzen-y

pkg/controller.v1beta1/trial/trial_controller_status.go
```diff
@@ -23,20 +23,14 @@ import (
 )
 
 const (
-	// For Trials
-	TrialCreatedReason            = "TrialCreated"
-	TrialRunningReason            = "TrialRunning"
-	TrialSucceededReason          = "TrialSucceeded"
-	TrialMetricsUnavailableReason = "MetricsUnavailable"
-	TrialFailedReason             = "TrialFailed"
-
```
andreyvelich, line 32: I think we want to keep these constants in `trial_controller_status`, similar to Suggestion Controller: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/suggestion/suggestion_controller_status.go#L26

pkg/controller.v1beta1/experiment/experiment_controller.go
andreyvelich: Hmm, if we want to have Killed status for the Trials, we need to decide how to handle this condition and if it is even useful for the user. Maybe we should just delete Trial in such cases to reduce status complexity.
We don't handle Killed Trial status: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_util.go#L42.

johnugeorge:
To me, it doesn't add any value in providing TrialKilled Status if underlying running resources are untouched.

pkg/controller.v1beta1/experiment/experiment_controller.go
tenzen-y: I agree with @andreyvelich.
The killed condition looks curious to me. When the number of failed Jobs reaches `MaxFailedCount`, it might be better to remove running Trial excepted failed Trial.

zhixian82: If we remove running Trial in such cases directly, how to update status in experiment? 

in such case, the experiment may be like:
```
status:
  runningTrialList:
  - task-tjgndlgd
  - task-gz78qbcm
  - task-xm6c2lqn
  trials: 6
  trialsRunning: 3
```
how to update the runningTrialList and trialsRunning field?

andreyvelich: @zhixian82 Maybe we should run [`updateTrialSummary` when Trial is deleted](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/experiment/util/status_util.go#L56) ?
WDYT @tenzen-y @johnugeorge ?

johnugeorge: TrialsKilled status is bit confusing term. Are we really Killing the trials or underlying pods? Deletion of running trials can be done ways IMO.  What extra info does it provide to the user?

andreyvelich: @johnugeorge I don't think we should introduce TrialKilled status.
However, we can delete Trial if that is required from the Experiment controller.
Underlying Trial resources should be deleted also, since they have an `OwnerReference`.

tenzen-y: > Maybe we should run [updateTrialSummary when Trials are deleted](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/experiment/util/status_util.go#L56) ?

I agree with that idea. We can update the Experiment status using that helper after unneeded running Trials are deleted.

zhixian82: I updated the pr，correct me if i understood wrong。
@andreyvelich @tenzen-y @johnugeorge 

But if we delete running trials directly, the status maybe changed from 
```
status:
  runningTrialList:
  - task-tjgndlgd
  - task-gz78qbcm
  - task-xm6c2lqn
  trials: 6
  trialsRunning: 3
```

to 
```
status:
  ...
  trials: 3
  trialsSucceeded: 3
```
the total trials decreased，this seems unreasonable。

tenzen-y: That behavior is expected. I think we can remove Job resources, not Trial resources.
@andreyvelich @johnugeorge WDYT?

tenzen-y:
@zhixian82 Can you sign the CLA with `<zh****n​@cambricon.com>`?

pkg/controller.v1beta1/experiment/experiment_controller.go
tenzen-y: Although, the trial-controller may create new Job resources once we remove Job resources, forcefully.
So, maybe we want to consider the feature of ignoring Trial resources that we don't want to reconcile.

tenzen-y: > So, maybe we want to consider the feature of ignoring Trial resources that we don't want to reconcile.

Ah, we can mark target Trial resources `Completed`. So we don't need to implement that feature.

andreyvelich: > But if we delete running trials directly, the status maybe changed from

@zhixian82 @johnugeorge @tenzen-y Any concerns of modifying underlying Experiment status in case of Trial cleanup?

johnugeorge: I was thinking - We delete underlying jobs and then change Trial status  to TrialKilled status. There is no further reconcile for TrialKilled status and we record the same in experiment status. Any red flags here?

tenzen-y: @johnugeorge That idea sounds better. @andreyvelich WDYT?

tenzen-y: Good point. Maybe, we can deal with KIlled Trials similarly to EarlyStopping Trials.

Issue #1639, opened 2021-08-25
Narrow and explicit RBAC


maanur:
/kind feature

I am trying to deploy Katib in an enterprise environment and have some hard time explaining Katib's requested RBAC
 rules. It would be much appreciated if ClusterRole explicitly declared only necessary verbs for each individual resource.

Look at this:
``` yaml
rules:
  - apiGroups:
      - ""
    resources:
      - configmaps
      - serviceaccounts
      - services
      - events
      - namespaces
      - persistentvolumes
      - persistentvolumeclaims
      - pods
      - pods/log
      - pods/status
    verbs:
      - "*"
```
Full access for Namespaces, Roles and RoleBindings effectively gives Katib unconstrained privileges to do anything it wants. This is plainly unacceptable in my case.

I suggest that ClusterRoles would be narrow and explicit like that:
```
rules:
  - apiGroups:
      - ""
    resources:
      - namespaces
    verbs:
      - get
      - list
      - watch
 - apiGroups:
      - ""
    resources:
      - roles
    verbs:
      - get
      - list
    ... e.t.c
```

E.g. let's get rid of those stars in rules.verbs and olso remove unnecessary verbs from there

maanur:
I'll look into the issue, but I may need help with testing because of lack of free time.
Firstly, I'm going to look at kubebuilder markers and try making them work, thus getting rid of "stars" in ClusterRoles.

After that I'd be willing to switch the operator's scope to multiple namespaces instead of a whole cluster, as the current level of privileges is not necessary in fact (remember `katib-metricscollector-injection` label on tracked namespaces). That would be a breaking change, so I'll make a separate proposal for thorough consideration.

andreyvelich:
That would be great, thank you @maanur!

> After that I'd be willing to switch the operator's scope to multiple namespaces instead of a whole cluster

Are you referring to this issue: https://github.com/kubeflow/katib/issues/1295 ?
Which runs Katib controller without cluster role access ?

maanur:
> Are you referring to this issue: #1295 ?\nWhich runs Katib controller without cluster role access ?

Yes, partially. I suppose the particular problem of deploying multiple controllers can be solved by OLM and its OperatorGroups, but that can be considered a separate feature.

andreyvelich:
@maanur I think that also requires additional proposal on how we want to implement this.
Currently, we don't have solution on how to enable our Webhooks on namespace mode.

[...]

PR #2091, opened 2023-01-20
Narrow down RBAC rules

johnugeorge:
This PR narrows down the RBAC rules for Katib controller installation

Fixes: #1639


tenzen-y (commented):
@johnugeorge Thanks for this!
I confirmed this RBAC on my local using the following operation:

- create and delete random.yaml by kubectl and UI.
- create and delete from-volume-resume.yaml by kubectl and UI.

That works well.

Also, maybe we want to generate RBAC manifests by `sigs.k8s.io/controller-tools/cmd/controller-gen` in the future.

> Note: Currently, cert-generator requires full access to secrets and UI needs full access to configmaps. Can we reduce it in the future?

I guess the cert-generator has limited access to secrets, now.

https://github.com/kubeflow/katib/blob/0749265a22f5fad30ac148ffc35885f47dbb5b1a/manifests/v1beta1/components/cert-generator/rbac.yaml#L12-L15

/lgtm

johnugeorge:
> @johnugeorge Thanks for this! I confirmed this RBAC on my local using the following operation:
> 
> * create and delete random.yaml by kubectl and UI.
> * create and delete from-volume-resume.yaml by kubectl and UI.
> 
> That works well.
> 
> Also, maybe we want to generate RBAC manifests by `sigs.k8s.io/controller-tools/cmd/controller-gen` in the future.
> 
> > Note: Currently, cert-generator requires full access to secrets and UI needs full access to configmaps. Can we reduce it in the future?
> 
> I guess the cert-generator has limited access to secrets, now.
> 
> https://github.com/kubeflow/katib/blob/0749265a22f5fad30ac148ffc35885f47dbb5b1a/manifests/v1beta1/components/cert-generator/rbac.yaml#L12-L15
> 
> /lgtm

Fixed configmaps and secrets as well

tenzen-y:
I have confirmed new changes using the same way as https://github.com/kubeflow/katib/pull/2091#pullrequestreview-1263745731, again.

/lgtm

andreyvelich:
Thank you for this @johnugeorge!
I will check it on Monday.

manifests/v1beta1/components/controller/rbac.yaml
```diff
@@ -7,19 +7,38 @@ rules:
   - apiGroups:
       - ""
     resources:
-      - configmaps
-      - serviceaccounts
       - services
       - events
-      - namespaces
+    verbs:
+      - "*"
```
andreyvelich, line 13: Do we need `*` for services ?

manifests/v1beta1/components/controller/rbac.yaml
```diff
@@ -7,19 +7,38 @@ rules:
...
       - persistentvolumeclaims
+    verbs:
+      - "get"
+      - "list"
+      - "watch"
+      - "create"
+  - apiGroups:
+      - ""
+    resources:
+      - namespaces
```
andreyvelich, line 28: Do we need permission for namespace in the Controller @johnugeorge @tenzen-y ?

manifests/v1beta1/components/controller/rbac.yaml
```diff
@@ -32,7 +51,10 @@ rules:
...
       - rolebindings
     verbs:
-      - "*"
+      - "get"
+      - "create"
+      - "list"
+      - "watch"
   - apiGroups:
       - batch
     resources:
```
andreyvelich, line 60: @johnugeorge Should we limit scope for `deployment`, `batch`, and `kubeflow.org jobs` ?

manifests/v1beta1/components/controller/rbac.yaml
johnugeorge: Yes. We do it in  https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/suggestion/suggestion_controller_util.go

manifests/v1beta1/components/controller/rbac.yaml
johnugeorge: Yes. It is needed only in UI.

manifests/v1beta1/components/controller/rbac.yaml
johnugeorge: We do get/create/delete/watch

manifests/v1beta1/components/controller/rbac.yaml
johnugeorge: I retested it. Currently, Katib needs get/list/watch for namespaces as well. Need to investigate separately

andreyvelich (commented):
/lgtm

PR #2057, opened 2022-12-11
Upgrade Python version to 3.10

tenzen-y:
Signed-off-by: tenzen-y <yuki.iwai.tz@gmail.com>

**What this PR does / why we need it**:
I upgraded Python version to 3.10 since Python 3.9 no longer fixes bugs, according to [this](https://devguide.python.org/versions/#supported-versions).

> 3.9 | PEP 596 | security | 2020-10-05 | 2025-10 | Łukasz Langa

Also, I upgraded pytest version to avoid the error, `E   TypeError: required field "lineno" missing from alias`.

ref: https://github.com/pytest-dev/pytest/pull/8540

**Which issue(s) this PR fixes** _(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_:
Fixes #

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/components/katib/) included if any changes are user facing

PR #2092, opened 2023-01-20
Run e2e with various Python versions to verify Python SDK

tenzen-y:
Signed-off-by: Yuki Iwai <yuki.iwai.tz@gmail.com>

As discussed in https://github.com/kubeflow/katib/pull/2057#issuecomment-1346966020, run e2e with various Python versions to verify Python SDK.

Fixes #

**Checklist:**


andreyvelich (commented):
@tenzen-y @johnugeorge Any thoughts how to keep Python versions and this version consistent: https://github.com/kubeflow/katib/blob/989d3da6cae3224acb2e5b3e9e6e1badb1cd2ca0/sdk/python/v1beta1/setup.py#L55-L60 ?
Then, we can say that we test our SDK on every version that we suggested.

johnugeorge:
I can't think of easy way to use the same versions in sync. (other than some kind of parse)

tenzen-y:
I don't have any good idea. As @johnugeorge says, we might need to parse setup.py and workflow files.

andreyvelich:
@johnugeorge @tenzen-y I see, let's keep the sync for minor version (3.7, 3.8, etc.) manually for now.

PR #2098, opened 2023-01-24
[SDK] Use Katib Client without Kube Config

andreyvelich:
Use the same logic as for the SDK Training Client: https://github.com/kubeflow/training-operator/pull/1740.

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -30,34 +30,33 @@ import kubeflow.katib.katib_api_pb2 as katib_api_pb2
 class KatibClient(object):
     def __init__(
         self,
-        config_file=None,
-        context=None,
-        client_configuration=None,
-        persist_config=True,
+        config_file: str = None,
+        context: str = None,
+        client_configuration: client.Configuration = None,
     ):
         """KatibClient constructor.
 
         Args:
-            config_file: Name of the kube-config file. Defaults to ~/.kube/config.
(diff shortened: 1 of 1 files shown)


sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -30,34 +30,33 @@
...
-                config_file=config_file,
-                context=context,
-                client_configuration=client_configuration,
-                persist_config=persist_config,
-            )
-            self.in_cluster = False
-        else:
-            config.load_incluster_config()
-            self.in_cluster = True
+        self.in_cluster = False
```
tenzen-y, line 48: Do we need a var, `in_cluster`?

andreyvelich: @tenzen-y Currently, we use it when Katib is deployed as part of Kubeflow Central Dashboard: https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L112-L124.
Although, it can generate incorrect links when user relies on `load_incluster_config()` and not run it from Kubeflow Notebooks.
We will discuss about it in the future.

tenzen-y: I see.
Thanks for clarifying! Could you create an issue to keep tracking this?

andreyvelich: Sure, let's track this.

tenzen-y:
Thanks!
/lgtm
/hold for other reviewers.

Issue #2055, opened 2022-12-09
Use Never Resume Policy by Default 


andreyvelich:
/kind discussion

Currently, we are using [`LongRunning` resume policy as the default value](https://github.com/kubeflow/katib/blob/ab2f59621efcf3c65c934a557b7f14671e654785/pkg/apis/controller/experiments/v1beta1/constants.go#L23-L24).
From my point of view, [Resume Experiment feature](https://www.kubeflow.org/docs/components/katib/resume-experiment/) is not widely used and it's usually used by experience users (at least who are familiar with the Katib CRD APIs).

Maybe we should change the default value to `Never` ?
In that case, new users won't see long-running Suggestion's deployments and they don't waste compute resources unexpectedly. 

What do you think @johnugeorge @tenzen-y @gaocegege @anencore94 ?

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

tenzen-y:
@andreyvelich  SGTM
It's helpful to release computing resources across the cluster.
Although we would be good to announce this as breaking changes since the Suggestion deployment will be automatically removed by changing the default policy to `Never`.

wdyt?

tenzen-y:
> @tenzen-y Yes, I think we should highlight it in the next Katib release and update the docs: https://www.kubeflow.org/docs/components/katib/resume-experiment/#resume-policy-longrunning

SGTM

tenzen-y:
@andreyvelich Would you like to include this feature in the next katib release (v0.15.0)?

johnugeorge:
Since it is straightforward, we can take up in this release

tenzen-y:
Sounds good.

andreyvelich:
Sure, let me work on that today or tomorrow before the release.

PR #2102, opened 2023-01-25
Use Never Resume Policy as Default

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/2055.

We are going to use the `Never` resume policy as default.
I will submit PR to update the website doc: https://www.kubeflow.org/docs/components/katib/resume-experiment/#resume-policy-longrunning

**Checklist:**


tenzen-y (commented):
@andreyvelich Thanks!
/lgtm

Issue #2019, opened 2022-11-17
A metrics collector for Kubeflow Pipeline Metrics artifacts


votti:
/kind feature

**Describe the solution you'd like**
Currently a aim is to do parameter tuning over pipelines in katib (#1914, #1993).

Kubeflow pipelines allow for dedicated metrics artifacts:
https://kubeflow-pipelines.readthedocs.io/en/master/source/dsl.html?h=metrics#kfp.dsl.Metrics
https://www.kubeflow.org/docs/components/pipelines/v1/sdk/pipelines-metrics/

Having a dedicated[ Katib sidecar metrics collector](https://www.kubeflow.org/docs/components/katib/experiment/#metrics-collector) that collects the metrics from this artifacts, would make pipelines and katib work together quite nicely.

The current workaround is to use the stdout collector, but this causes issues with the complex commands in pipeline components (#1914, will add dedicated issue soon).

**Anything else you would like to add:**
-

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

votti:
I think I may give this a go - I would try to build this in Python analogous to the [tfevent-metricscollector](https://github.com/kubeflow/katib/tree/master/pkg/metricscollector/v1beta1/tfevent-metricscollector).
Does this sound like a reasonable approach? I am also happy for any other suggestion.

AlexandreBrown:
@votti Hello, did you have the chance to give it a try?

votti:
Small update: 
I have now a metrics collector for kubeflow v1 pipelines that I think should work (modeled after [tfevent-metricscollector](https://github.com/kubeflow/katib/tree/master/pkg/metricscollector/v1beta1/tfevent-metricscollector)).

What I am failing is to pass the trial name to the custom connector.
Essentially I am using the a very similar configuration as in the custom connector example here:
[tfevent-metricscollector](https://github.com/kubeflow/katib/tree/master/pkg/metricscollector/v1beta1/tfevent-metricscollector).

My cli metricscollector takes an argument "-t" or "--trial_name" with the trial name to use for reporting (exactly as the tfevent-metricscollector).
Would maybe someone know a hint how to configure this such that the current trial-name would be passed as arg?

votti:
I am now really a bit confused:
Reading the source code of the metrics collector sidecar injection `inject_webhook`, it looks to me as if the trial name should be actually added to the args: 
[...]

PR #2118, opened 2023-02-15
Feature/Example for training KFP v1

### LINKED ISSUE #1914: Add support for entire kubeflow pipelines as trial target (in addition to containers)
/kind feature

Hyper parameters not only affect the training step but also upstream pipeline components like feature transformation for example (e.g. parameters of a normalization transformation). In addition, transformation and training steps should be able to make use of kfp's parallel components (e.g. SparkJob, TFJob, ...). It would be helpful to not only allow to specify containers as trial targets but also complete kubeflow pipelines. As the latter also expose parameters they can be either set directly (non-hyperparameters) or added to the hyper parameter space.

I've started to create simple container image which can be used as trial target which acts as a proxy and downstream triggers parameterized Kubeflow pipeline executions with the respective hyper parameters. A Kubernetes Custom Resource can be created as well down the line.

---

votti:

Currently there is no example how to do parameter tuning over whole Kubeflow pipelines (#1914).
This example shows a how parameter tuning with Katib can be done on a Kubeflow pipeline based on the pipeline v1 SDK.
It adds and uses a new Metrics Collector (`kfpv1-metricscollector`, #2019) that is used as a custom metrics collector in the  pipeline.
The `kfpv1-metricscollector` image is also published to the `katib` docker environment.

Fixes #1914, #2019

**Checklist:**


cmd/metricscollector/v1beta1/kfpv1-metricscollector/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM python:3.9-slim
```
tenzen-y, line 1:
```suggestion
FROM python:3.10-slim
```

Should we use Python v3.10 similar to other components?

cmd/metricscollector/v1beta1/kfpv1-metricscollector/main.py
```diff
@@ -0,0 +1,113 @@
+# Copyright 2022 The Kubeflow Authors.
```
tenzen-y, line 1:
```suggestion
# Copyright 2023 The Kubeflow Authors.
```

pkg/metricscollector/v1beta1/common/const.py
```diff
@@ -20,6 +20,9 @@
 DEFAULT_WAIT_ALL_PROCESSES = "True"
 # Default value for directory where TF event metrics are reported
 DEFAULT_METRICS_FILE_DIR = "/log"
+# Default value for directory where TF event metrics are reported
+DEFAULT_METRICS_FILE_KFPV1_DIR = "/tmp/outputs/mlpipeline_metrics"
```
tenzen-y, line 24: Can we reuse `DEFAULT_METRICS_FILE_DIR`? Does this define in KFP v1 as a default value?

pkg/metricscollector/v1beta1/kfpv1-metricscollector/metrics_loader.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2022 The Kubeflow Authors.
```
tenzen-y, line 1:
```suggestion
# Copyright 2023 The Kubeflow Authors.
```

tenzen-y (commented):
@votti Thanks for creating this PR, and sorry for the late response.
I left a few review comments for the first pass.

> currently the docker image with the custom metrics collector lives in my personal docker. Maybe it would be better to add it to the kubeflow organisation? What would the steps be to achieve this?

Yes, we should build and push the image to our container registry.
You can add the image name and dockerfile path to the following:

https://github.com/kubeflow/katib/blob/b6afce7d8903610d861ca02d04f66c4a3774e7b3/.github/workflows/publish-core-images.yaml#L19-L34

Moreover, can you add a building script to https://github.com/kubeflow/katib/blob/master/scripts/v1beta1/build.sh and https://github.com/kubeflow/katib/blob/master/scripts/v1beta1/push.sh?

pkg/metricscollector/v1beta1/common/const.py
votti: In KFP v1 this is a default path where output metrics are saved in Python operators.

votti:
@tenzen-y  Thanks a lot for your comments!
 I have now weaved the `kfpv1-metricscollector` docker image into the build process and addressed the minor comments.

pkg/metricscollector/v1beta1/common/const.py
tenzen-y: sgtm.

pkg/metricscollector/v1beta1/common/const.py
```diff
@@ -20,6 +20,9 @@
 DEFAULT_WAIT_ALL_PROCESSES = "True"
 # Default value for directory where TF event metrics are reported
 DEFAULT_METRICS_FILE_DIR = "/log"
+# Default value for directory where TF event metrics are reported
```
tenzen-y, line 23: Please update this comment.

pkg/metricscollector/v1beta1/kfpv1-metricscollector/metrics_loader.py
```diff
@@ -0,0 +1,94 @@
...
+# limitations under the License.
+
+# TFEventFileParser parses tfevent files and returns an ObservationLog of the metrics specified.
+# When the event file is under a directory(e.g. test dir), please specify "{{dirname}}/{{metrics name}}"
+# For example, in the Tensorflow MNIST Classification With Summaries:
+# https://github.com/kubeflow/katib/blob/master/examples/v1beta1/trial-images/tf-mnist-with-summaries/mnist.py.
+# The "accuracy" and "loss" metric is saved under "train" and "test" directories.
+# So in the Metrics Collector specification, please specify name of "train" or "test" directory.
+# Check TFJob example for more information:
+# https://github.com/kubeflow/katib/blob/master/examples/v1beta1/kubeflow-training-operator/tfjob-mnist-with-summaries.yaml#L16-L22
```
tenzen-y, line 22: Please update this comment.

pkg/metricscollector/v1beta1/kfpv1-metricscollector/metrics_loader.py
```diff
@@ -0,0 +1,94 @@
...
+        handler = StreamHandler()
+        handler.setLevel(INFO)
+        self.logger.setLevel(INFO)
+        self.logger.addHandler(handler)
+        self.logger.propagate = False
+        self.metrics = metric_names
+
+    def parse_file(self, filename):
+        self.logger.info(filename + " will be parsed.")
+        mls = parse_metrics(filename)
```
tenzen-y, line 69: Why this metrics collector doesn't search all files in  `--metrics_file_dir`?
We should search all files and collect metrics if we don't have limitations related to KFP.

cmd/metricscollector/v1beta1/kfpv1-metricscollector/main.py
tenzen-y, line 1: Should we create this file under the `cmd/metricscollector/v1beta1/kfp-metricscollector/v1`?

I believe that when we add the metrics collector for KFP v2, we can commonize some codes with between kfpv1 and kfpv2.

pkg/metricscollector/v1beta1/kfpv1-metricscollector/metrics_loader.py
tenzen-y, line 1: > Should we create this file under the cmd/metricscollector/v1beta1/kfp-metricscollector/v1?
>
> I believe that when we add the metrics collector for KFP v2, we can commonize some codes with between kfpv1 and kfpv2.

Same comment.

tenzen-y:
@votti Thanks for your diligent work.

We must add an E2E test for this metrics collector. However, I think we can work on it on follow-up PRs.
Which do you prefer to add E2E in this PR or other PRs?

@kubeflow/wg-automl-leads Can you approve CI?

examples/v1beta1/kubeflow-pipelines/README.md
```diff
@@ -25,10 +31,12 @@ to give an access Kubeflow Notebook to run Kubeflow Pipelines.
 
 The following Pipelines are deployed from Kubeflow Notebook:
 
-- [Kubeflow E2E MNIST](kubeflow-e2e-mnist.ipynb)
+1) [Kubeflow E2E MNIST](kubeflow-e2e-mnist.ipynb)
+
+2) [Katib Experiment with Early Stopping](early-stopping.ipynb)
 
-- [Katib Experiment with Early Stopping](early-stopping.ipynb)
+3) [Tune parameters of a `MNIST` kubeflow pipeline with Katib](pipeline-parameters.ipynb)
```
axel7083, line 38:
```suggestion
3) [Tune parameters of a `MNIST` kubeflow pipeline with Katib](kubeflow-kfpv1-opt-mnist.ipynb)
```

.github/workflows/publish-core-images.yaml
```diff
@@ -32,3 +32,5 @@ jobs:
             dockerfile: cmd/metricscollector/v1beta1/file-metricscollector/Dockerfile
           - component-name: tfevent-metrics-collector
             dockerfile: cmd/metricscollector/v1beta1/tfevent-metricscollector/Dockerfile
+          - component-name: kfpv1-metrics-collector
+            dockerfile: cmd/metricscollector/v1beta1/kvpv1-metricscollector/Dockerfile
```
axel7083, line 36:
```suggestion
            dockerfile: cmd/metricscollector/v1beta1/kfpv1-metricscollector/Dockerfile
```

pkg/metricscollector/v1beta1/kfpv1-metricscollector/metrics_loader.py
votti: The comment is now updated.

tenzen-y:
@votti Thanks for the updates! I'll check this PR again later.

> As far as I know nobody is actively using this (yet), so I think there would be time for adding an E2E together with this PR making it easier to maintain long-term.
Would you have some pointers (eg existing tests) how this could be achieved?

Maybe, we need to update [`setup-katib.sh`](https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/test/e2e/v1beta1/scripts/gh-actions/setup-katib.sh) and [`run-e2e-experiment.py`](https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py).

- `setup-katib.sh`: Install Kubeflow Pipeline.
- `run-e2e-experiment.py`: Verify kubeflow integration.

cmd/metricscollector/v1beta1/kfp-metricscollector/v1/main.py
```diff
@@ -0,0 +1,101 @@
...
+    if len(db_manager_server) != 2:
+        raise Exception(
+            "Invalid Katib DB manager service address: %s" % opt.db_manager_server_addr
+        )
+
+    WaitMainProcesses(
+        pool_interval=opt.poll_interval,
+        timout=opt.timeout,
+        wait_all=wait_all_processes,
+        completed_marked_dir=None,
```
tenzen-y, line 78: Why do we set `None` to `completed/marked_dir`? Can we set `opt.metrics_file_dir` instead?

pkg/metricscollector/v1beta1/kfp-metricscollector/v1/metrics_loader.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self, metric_names):
+        self.metric_names = metric_names
+
+    @staticmethod
+    def find_all_files(directory):
+        for root, dirs, files in os.walk(directory):
+            for f in files:
+                yield os.path.join(root, f)
+
+    def parse_metrics(self, fn: str) -> List[api_pb2.MetricLog]:
```
tenzen-y, line 40:
```suggestion
    def parse_metrics(self, metric_file_path: str) -> List[api_pb2.MetricLog]:
```

Should we use `metric_file_path` as a name for more clarification?

tenzen-y (commented):
Thanks for the updates! I left a few comments.
Also, once adding tests are done, can you send a ping to me?

cmd/metricscollector/v1beta1/kfp-metricscollector/v1/main.py
votti: The documentation on this is a bit sparse, but if I understand the code right, this would require the Kubeflow pipeline to write a file `<pid>.pid`  with some `TRAINING_COMPLETED` text into this directory which it does not do:
https://github.com/kubeflow/katib/blob/f740889569afd93ca6be709a337ba0cd8bdd8920/pkg/metricscollector/v1beta1/common/pns.py#L95-L104

So I think `None` is correct here.

scripts/v1beta1/build.sh
```diff
@@ -71,6 +71,9 @@ docker buildx build --platform "linux/${ARCH}" -t "${REGISTRY}/cert-generator:${
 echo -e "\nBuilding file metrics collector image...\n"
 docker buildx build --platform "linux/${ARCH}" -t "${REGISTRY}/file-metrics-collector:${TAG}" -f ${CMD_PREFIX}/metricscollector/${VERSION}/file-metricscollector/Dockerfile .
 
+echo -e "\nBuilding kfpv1 metrics collector image...\n"
+docker buildx build --platform "linux/${ARCH}" self, -t "${REGISTRY}/kfpv1-metrics-collector:${TAG}" -f ${CMD_PREFIX}/metricscollector/${VERSION}/kfp-metricscollector/v1/Dockerfile .
```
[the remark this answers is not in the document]
votti: An embarrassing mistake it is :o 
Fixed now.

andreyvelich (commented):
Thank you for this great contribution @votti!

Please can you let me know why do we need separate Metrics Collector for KFP ?
Why we can't use our default Metrics Collector in a File mode [when you can specify file from which we should parse the metrics](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/metrics-collector/file-metrics-collector.yaml#L19-L21) ?
In that case, Metrics Collector doesn't parse Stdout, and read data from the file.

cmd/metricscollector/v1beta1/kfp-metricscollector/v1/main.py
tenzen-y: Thanks for the explanation. Let me check.

cmd/metricscollector/v1beta1/kfp-metricscollector/v1/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM python:3.10-slim
```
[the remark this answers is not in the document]
tenzen-y: @andreyvelich My concern is where we will put Dockerfile for KFP v2. So I would suggest we put Dockerfile for the KFP v1 on here.
wdyt?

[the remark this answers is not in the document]
andreyvelich: Oh, I see. Do we really need to support KFP v1 if, eventually, every Kubeflow users should migrate to KFP v2 ?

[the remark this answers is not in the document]
tenzen-y: Because KFP v1 and KFP v2 aren't compatible, I think migrating v1 to v2 is harder in production.
So I guess users need a lot of time to update the version.

Hence, supporting KFP v1 in Katib would be useful. WDYT?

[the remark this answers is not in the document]
andreyvelich: I see, in any case I still have a question ( https://github.com/kubeflow/katib/pull/2118#pullrequestreview-1536917369) why do we need separate Metrics Collector for KFP if we need to just read the logs from the metrics file ?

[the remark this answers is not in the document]
votti: Great point, I think if we could get away without a custom metrics collector this would be even better.

I am pretty sure I tested this in the beginning but failed.
As far as I remember the File metrics collector was not compatible with the metrics as required by KFP.

So the format is:
```
{
        "metrics": [
            {
                "name": "accuracy",  # The name of the metric. Visualized as the column name in the runs table.
                "numberValue": accuracy,  # The value of the metric. Must be a numeric value.
                "format": "PERCENTAGE",  # The optional format of the metric. Supported values are "RAW" (displayed in raw format) and "PERCENTAGE" (displayed in percentage format).
            },
            {
                "name": "val-accuracy",  # The name of the metric. Visualized as the column name in the runs table.
                "numberValue": val_accuracy,  # The value of the metric. Must be a numeric value.
                "format": "PERCENTAGE",  # The optional format of the metric. Supported values are "RAW" (displayed in raw format) and "PERCENTAGE" (displayed in percentage format).
            },
        ]
    }
```
usually dumped as a JSON.

Do I read the code of the file-metricscollector correctly, that this would require the JSON to be written on a single line:
https://github.com/kubeflow/katib/blame/89bd21f710fb4cd153a33c94e8892ec079cf63c8/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L224-L228

as `logText` is a line: 
https://github.com/kubeflow/katib/blame/89bd21f710fb4cd153a33c94e8892ec079cf63c8/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L176-L177

So maybe this could work if I make sure that the KFP metrics JSON file is dumped in a form where each metrics is written on a single line?

I now  added a more minimal example pipeline  that could be used to try this a bit more in-depth.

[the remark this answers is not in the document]
andreyvelich: Yeah, that's right @votti, JSON metric name and value should be on the same line. Is there any way to configure KFP keep to keep this metrics JSON file in the one line ?
Otherwise, we can introduce another [format to our file metrics collector](https://github.com/kubeflow/katib/blob/89bd21f710fb4cd153a33c94e8892ec079cf63c8/pkg/apis/controller/common/v1beta1/common_types.go#L195) to parse KFP metrics file.
Any other suggestions @tenzen-y ?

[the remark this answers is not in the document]
tenzen-y: @votti It's a great explanation! Yes, the JSON logs must be written on a single line.

> Is there any way to configure KFP keep to keep this metrics JSON file in the one line ?

If it's possible, we should take it this way.

> Otherwise, we can introduce another [format to our file metrics collector](https://github.com/kubeflow/katib/blob/89bd21f710fb4cd153a33c94e8892ec079cf63c8/pkg/apis/controller/common/v1beta1/common_types.go#L195) to parse KFP metrics file.

If the metrics format from KFP is always the following (it means users can not modify the format), we can add a new file format, `KubeflowPiplinesv1`.

> ```
>             {
>                 "name": "val-accuracy",  # The name of the metric. Visualized as the column name in the runs table.
>                 "numberValue": val_accuracy,  # The value of the metric. Must be a numeric value.
>                 "format": "PERCENTAGE",  # The optional format of the metric. Supported values are "RAW" (displayed in raw format) and "PERCENTAGE" (displayed in percentage format).
>             },
> ```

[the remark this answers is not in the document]
tenzen-y: Maybe, this is the metric form for the KFP v1?

https://github.com/kubeflow/pipelines/blob/1619f2c0b605f202c1cbb8488c64805ccff759c8/backend/api/v1beta1/run.proto#L283C11-L310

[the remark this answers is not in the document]
tenzen-y: For KFP v2, maybe this?

https://www.kubeflow.org/docs/components/pipelines/v1/sdk-v2/run-comparison/#scalar-metrics

https://github.com/kubeflow/pipelines/blob/55a2fb5c20011b01945c9867ddff0d39e9db1964/sdk/python/kfp/v2/components/types/artifact_types.py#L113-L131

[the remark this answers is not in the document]
andreyvelich: @james-jwu @zijianjoy Would you mind help us here ? We are trying to use KFP within Katib Trial and see if we can modify the output format for metrics file. Basically, save metrics in file in one line.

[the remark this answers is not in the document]
votti: Writing out the json on one line was not a big deal (actually it is already in this form. All that needed to be changed for the fileCollector (I think) is the metrics format regexp: `'"name": ([\w|-]+), "numberValue": ((-?\d+)(\.\d+)?)'`
Using this I tested the `fileCollector` approach again and run into an issue that this seems to break the argo pipeline with the error:
```
time="2023-07-25T20:53:32.488Z" level=info msg="Copying /tmp/outputs/mlpipeline_metrics/data from container base image layer to /tmp/argo/outputs/artifacts/mlpipeline-metrics.tgz"
time="2023-07-25T20:53:32.488Z" level=info msg="/var/run/argo/outputs/artifacts/tmp/outputs/mlpipeline_metrics/data.tgz -> /tmp/argo/outputs/artifacts/mlpipeline-metrics.tgz"
time="2023-07-25T20:53:32.488Z" level=error msg="executor error: open /var/run/argo/outputs/artifacts/tmp/outputs/mlpipeline_metrics/data.tgz: no such file or directory\n
```

Not sure why. I attached a failing experiment.
[katib-kfp-example-e2e-v1_filecollector.txt](https://github.com/kubeflow/katib/files/12165569/katib-kfp-example-e2e-v1_filecollector.txt)


Btw: If you wondering why I dont just use the Stdout collector and in addition print the metrics to the log: this is because this also [ broke the argo command: ](https://github.com/kubeflow/katib/issues/1914#issuecomment-1272648383) and needed a[ quite ugly workaround to still work](https://github.com/kubeflow/katib/issues/1914#issuecomment-1281480140).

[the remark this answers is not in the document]
andreyvelich: > Is there any reason for restricting metrics file configuration in one line?
@zijianjoy Katib metrics collector parses metrics file line by line and expects metrics name and value to be located in a single line.

@votti from the log line I can see that metrics are written to `/tmp/argo/outputs/artifacts/mlpipeline-metrics.tgz` file, isn't it ?

>Btw: If you wondering why I dont just use the Stdout collector and in addition print the metrics to the log: this is because this also broke the argo command: 

@votti Yeah, this could be an issue since we override the start command to make sure we redirect StdOut to `/var/log/katib/metrics.log` file, so Katib Metrics Collector can parse this file. Otherwise, Metrics Collector can't parse the StdOut. The main differences between StdOut and File metrics collector is that StdOut tails `/var/log/katib/metrics.log` file and prints logs.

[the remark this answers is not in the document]
votti: @metrics file:
One of the complexities kubeflow `pipeline` manages is to handle output artifacts (usually compressing them and storing them and saving them to an s3 storage). This is what seems to be broken when using the `filecollector`, as something while compressing and copying the file to `/tmp/argo/outputs/artifacts/mlpipeline-metrics.tgz` seems to go wrong.

After finding some time to look into it, I think the reason is very similar to the `stdout` collector:
The collector modifies the argo CMD/ARG in a way that I think causes these issues:

From the pod definition: Unmodified (eg when using the kubeflow custom metrics collector):
```
...
      _outputs = train_e2e(**_parsed_args)
      
    Args:

      --input-nr
      /tmp/inputs/input_nr/data
      --lr
      0.0005293023468535503
      --optimizer
      Adam
      --loss
      categorical_crossentropy
      --epochs
      3
      --batch-size
      36
      --mlpipeline-metrics
      /tmp/outputs/mlpipeline_metrics/data
```

When using the `filecollector` as metrics collector:

```
...
      _outputs = train_e2e(**_parsed_args)
       --input-nr /tmp/inputs/input_nr/data --lr 0.00021802007326291811 --optimizer Adam --loss categorical_crossentropy --epochs 3 --batch-size 53 --mlpipeline-metrics /tmp/outputs/mlpipeline_metrics/data && echo completed > /tmp/outputs/mlpipeline_metrics/$$$$.pid

```

I think this could be solved by following this proposal: https://github.com/kubeflow/katib/issues/2181
Until this is fixed, I think having a custom metrics collector that does not modify the command is a necessary workaround.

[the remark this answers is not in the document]
andreyvelich: @votti I think this also could solve this issue, isn't: https://github.com/kubeflow/katib/issues/577 ?
Basically, we can use Katib SDK to implement API for pushing metrics for Katib DB instead of using pull-based metrics collectors which require to change entrypoint.

For example:
```python

client = katib.KatibClient()
client.report(metrics={"accuracy": 0.9, "loss": 0.01"})
```
We might need to do additional changes to Katib controller to verify that metrics were reported by user.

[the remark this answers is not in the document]
votti: @push based metrics collection: that sounds like a good potential solution!
So the `KatibClient` can infer automatically which `trial` these `metrics` are associated with?

votti:
I tried now to implement an `e2e` test and checked that it runs locally (on a local `k3s` cluster).
What I dont know is how to trigger the `ci` to run the test. Could someone help me and check if the config looks right?
Thanks already!

cmd/metricscollector/v1beta1/kfp-metricscollector/v1/Dockerfile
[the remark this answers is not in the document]
andreyvelich: @votti Currently, user can get the Trial name using [`${trialSpec.Name}` template](https://www.kubeflow.org/docs/components/katib/trial-template/#use-trial-metadata-in-template) in their Trial's Pod environment vars. Then, user can run `KatibClient` API with appropriate Trial Name to insert metrics to the Katib DB.
I think, we should always add `TRIAL_NAME` env to the Trial pod since it is useful for many use-cases (e.g. for exporting trained model to S3, saving Trial metrics to DB, etc.)
WDYT @tenzen-y @johnugeorge ?

PR #2138, opened 2023-03-29
Namespace and trial pod annotations as CLI argument

nagar-ajay:
* For the [conformance test](https://docs.google.com/document/d/1TRUKUY1zCCMdgF-nJ7QtzRwifsoQop0V8UnRo-GWlpI/edit?pli=1#), we want to run all e2e tests in `kf-conformance` namespace which is Istio enabled (created by Kubeflow Profile CRD)
* Currently, the namespace is hardcoded in all example manifests files and we don't have the option to configure the namespace.
* Updated `run-e2e-experiment.py` to support namespace as a command line argument. Also, add `katib.kubeflow.org/metrics-collector-injection` label to the test namespace, if missing.
* Changed test namespace to `default` from `kubeflow`.
* In Istio enabled namespace, Pods of example manifests get stuck in the `NotReady` state. To fix this, added an option to pass `trial-pod-annotations`. Users can pass `'{"sidecar.istio.io/inject": "false"}'` annotation to disable istio sidecar injection.
* Since, we're loading kube config as part of `KatibClient` initialization, we don't need to do it again. Also `config.load_kube_config()` statement breaks if we run the test inside k8s cluster.
* Updated `random.yaml` to put resource limits on trail spec pod container. This is required as the namespace created for the conformance test, has resource quota specified. https://github.com/kubeflow/kubeflow/blob/master/conformance/1.5/setup.yaml#L24-L28

* **Testing**: Tested modified manifests locally. Also, all these manifests run in e2e test as part of Katib CI.

### DIFF
--- a/examples/v1beta1/early-stopping/median-stop-with-json-format.yaml
+++ b/examples/v1beta1/early-stopping/median-stop-with-json-format.yaml
@@ -59,9 +59,6 @@ spec:
       kind: Job
       spec:
         template:
-          metadata:
-            annotations:
-              sidecar.istio.io/inject: "false"
           spec:
             containers:
               - name: training-container
--- a/examples/v1beta1/early-stopping/median-stop-with-json-format.yaml
+++ b/examples/v1beta1/early-stopping/median-stop-with-json-format.yaml
@@ -59,6 +59,9 @@ spec:
       kind: Job
       spec:
         template:
+          metadata:
+            annotations:
+              sidecar.istio.io/inject: "false"
           spec:
             containers:
               - name: training-container
--- a/examples/v1beta1/hp-tuning/random.yaml
+++ b/examples/v1beta1/hp-tuning/random.yaml
@@ -62,4 +62,8 @@ spec:
                   - "--lr=${trialParameters.learningRate}"
                   - "--num-layers=${trialParameters.numberLayers}"
                   - "--optimizer=${trialParameters.optimizer}"
+                resources:
+                  limits:
+                    memory: "1Gi"
+                    cpu: "0.5"
             restartPolicy: Never
--- a/examples/v1beta1/hp-tuning/random.yaml
+++ b/examples/v1beta1/hp-tuning/random.yaml
@@ -51,6 +51,9 @@ spec:
       kind: Job
       spec:
         template:
+          metadata:
+            annotations:
+              sidecar.istio.io/inject: "false"
           spec:
             containers:
               - name: training-container
--- a/examples/v1beta1/hp-tuning/random.yaml
+++ b/examples/v1beta1/hp-tuning/random.yaml
@@ -51,9 +51,6 @@ spec:
       kind: Job
       spec:
         template:
-          metadata:
-            annotations:
-              sidecar.istio.io/inject: "false"
           spec:
             containers:
               - name: training-container
--- a/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
+++ b/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
@@ -220,6 +220,9 @@ if __name__ == "__main__":
         "--namespace", type=str, required=True, help="Namespace for the Katib E2E test",
     )
     parser.add_argument(
+        "--trial-pod-annotations", type=str, help="Annotation for the pod created by trial",
+    )
+    parser.add_argument(
         "--verbose", action="store_true", help="Verbose output for the Katib E2E test",
     )
     args = parser.parse_args()
@@ -253,6 +256,11 @@ if __name__ == "__main__":
     if experiment.metadata.name == "random":
         MAX_TRIAL_COUNT += 1
         PARALLEL_TRIAL_COUNT += 1
+        if args.trial_pod_annotations:
+            trial_spec = experiment.spec.trial_template.trial_spec
+            trial_spec_metadata = trial_spec['spec']['template'].get('metadata', {})
+            trial_spec_metadata['annotations'] = eval(args.trial_pod_annotations)
+            trial_spec['spec']['template']['metadata'] = trial_spec_metadata
 
     # Hyperband will validate the parallel trial count, thus we should not change it.
     # We don't need to test parallel Trials for Darts.
--- a/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
+++ b/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
@@ -256,10 +256,11 @@ if __name__ == "__main__":
         MAX_TRIAL_COUNT += 1
         PARALLEL_TRIAL_COUNT += 1
         if args.trial_pod_annotations:
-            trial_spec = experiment.spec.trial_template.trial_spec
-            trial_spec_metadata = trial_spec['spec']['template'].get('metadata', {})
-            trial_spec_metadata['annotations'] = eval(args.trial_pod_annotations)
-            trial_spec['spec']['template']['metadata'] = trial_spec_metadata
+            trial_spec_metadata = experiment.spec.trial_template.trial_spec['spec']['template'].get('metadata', {})
+            trial_spec_pod_annotations = trial_spec_metadata.get('annotations', {})
+            trial_spec_pod_annotations.update(eval(args.trial_pod_annotations))
+            trial_spec_metadata['annotations'] = trial_spec_pod_annotations
+            experiment.spec.trial_template.trial_spec['spec']['template']['metadata'] = trial_spec_metadata
 
     # Hyperband will validate the parallel trial count, thus we should not change it.
     # We don't need to test parallel Trials for Darts.
--- a/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
+++ b/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
@@ -256,6 +256,10 @@ if __name__ == "__main__":
         MAX_TRIAL_COUNT += 1
         PARALLEL_TRIAL_COUNT += 1
         if args.trial_pod_annotations:
+            kind = experiment.spec.trial_template.trial_spec['kind']
+            if kind != "Job":
+                raise NotImplementedError(f'Trail pod annotations not implemented for {kind}!')
+
             trial_spec_metadata = experiment.spec.trial_template.trial_spec['spec']['template'].get('metadata', {})
             trial_spec_pod_annotations = trial_spec_metadata.get('annotations', {})
             trial_spec_pod_annotations.update(eval(args.trial_pod_annotations))
--- a/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
+++ b/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
@@ -2,7 +2,7 @@ import argparse
 import yaml
 import time
 import logging
-from kubernetes import client, config
+from kubernetes import client
 from kubeflow.katib import ApiClient, KatibClient, models
 from kubeflow.katib.utils.utils import FakeResponse
 from kubeflow.katib.constants import constants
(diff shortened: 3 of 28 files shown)


andreyvelich (commented):
Thank you for updating this @nagar-ajay!

examples/v1beta1/early-stopping/median-stop-with-json-format.yaml
```diff
@@ -59,6 +59,9 @@ spec:
       kind: Job
       spec:
         template:
+          metadata:
+            annotations:
+              sidecar.istio.io/inject: "false"
```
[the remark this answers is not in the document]
tenzen-y: I agree with @andreyvelich.
It might be better to run the `kubectl patch` command in `run-e2e-experiment.sh`.

johnugeorge:
For conformance, we don't need to run all tests. Is [random example](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/hp-tuning/random.yaml) enough? Having distributed training operators in examples might not be a good idea for users who are not using it

examples/v1beta1/early-stopping/median-stop-with-json-format.yaml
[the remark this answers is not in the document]
nagar-ajay: Actually, as per [conformance test design ](https://docs.google.com/document/d/1TRUKUY1zCCMdgF-nJ7QtzRwifsoQop0V8UnRo-GWlpI/edit?pli=1#heading=h.tubso4j7bqvw), we want to run these tests inside a pod. I don't think we will have access to `kubectl` inside the pod. Correct me if I'm wrong.

We will need to add support for annotations in `run-e2e-experiment.py`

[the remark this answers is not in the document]
nagar-ajay: I think programmatically patching the resources will not help here as the resources are already created with Istio sidecar.

We will need to update the definition somehow before creating the experiment. This is tricky. If we provided an option to pass annotation, this can create confusion as the passed annotation can be applied in more than one place (experiment metadata, pod metadata, etc).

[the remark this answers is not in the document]
andreyvelich: I think, @tenzen-y meant not use `kubectl patch`, but change the run-e2e-experiment script to modify Trial template before creating Experiment. Similar as [here for Trial spec](https://github.com/kubeflow/katib/blob/master/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py#L259-L261).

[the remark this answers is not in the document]
tenzen-y: > I think, @tenzen-y meant not use kubectl patch, but change the run-e2e-experiment script to modify Trial template before creating Experiment. Similar as [here for Trial spec](https://github.com/kubeflow/katib/blob/master/test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py#L259-L261).

Yes, that's right. Sorry for your confusion.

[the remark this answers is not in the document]
johnugeorge: @andreyvelich It is bit tricky. Trial Spec can be different. Say distributed job vs non distributed job.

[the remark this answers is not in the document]
andreyvelich: @johnugeorge That's right, but since we do it only for Random experiment which uses Kubernetes Job, should we try to `deserialize` TrialSpec to `V1Job` ?

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
```diff
@@ -249,6 +255,11 @@ def run_e2e_experiment(
     if experiment.metadata.name == "random":
         MAX_TRIAL_COUNT += 1
         PARALLEL_TRIAL_COUNT += 1
+        if args.trial_pod_annotations:
+            trial_spec = experiment.spec.trial_template.trial_spec
+            trial_spec_metadata = trial_spec['spec']['template'].get('metadata', {})
+            trial_spec_metadata['annotations'] = eval(args.trial_pod_annotations)
```
johnugeorge, line 261: What format do you plan to pass? Also, Consider a case where annotations already present

examples/v1beta1/early-stopping/median-stop-with-json-format.yaml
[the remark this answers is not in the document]
tenzen-y: I'm ok.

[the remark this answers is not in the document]
andreyvelich: @johnugeorge @tenzen-y Other solution might be to add another experiment just for conformance test here: https://github.com/kubeflow/katib/tree/master/test/e2e/v1beta1/testdata with required annotation and other specification.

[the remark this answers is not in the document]
tenzen-y: If we add another experiment, I think we should run the experiment on CI every PR similar to other examples.
Since we need to verify whether the experiment can be runnable.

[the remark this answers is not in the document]
johnugeorge: @tenzen-y  I see the same exact problem. Conformance tests should be tested regularly. Else, it will diverge from the core examples.  I would suggest to go with current **random** experiment. For annotations, we have to handle 3 cases- non distributed trials, distributed trials and custom trials. We are handling just non distributed trials now. Others can be handled later.

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
nagar-ajay: I'm planning to pass string representation of a python dictionary (annotations's key-valye pairs) e.g. I'll pass `'{"sidecar.istio.io/inject": "false"}'` to disable sidecar injection.

Will update the code to consider already present annotations.

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
```diff
@@ -249,6 +255,12 @@ def run_e2e_experiment(
     if experiment.metadata.name == "random":
         MAX_TRIAL_COUNT += 1
         PARALLEL_TRIAL_COUNT += 1
+        if args.trial_pod_annotations:
```
andreyvelich, line 258: Please can we also throw an error if `experiment.spec.trial_template.trial_spec['kind'] != "Job"`.

That will help users of conformance to not define incorrect Trial Spec (e.g. TFJob, PyTorchJob), since we only support `Job` for now.

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
```diff
@@ -216,6 +215,12 @@ def run_e2e_experiment(
         required=True,
         help="Path to the Katib Experiment.",
     )
+    parser.add_argument(
+        "--namespace", type=str, required=True, help="Namespace for the Katib E2E test",
+    )
+    parser.add_argument(
+        "--trial-pod-annotations", type=str, help="Annotation for the pod created by trial",
```
andreyvelich, line 222: In which formant we are going to pass annotations ?

examples/v1beta1/hp-tuning/random.yaml
```diff
@@ -62,4 +62,8 @@ spec:
                   - "--lr=${trialParameters.learningRate}"
                   - "--num-layers=${trialParameters.numberLayers}"
                   - "--optimizer=${trialParameters.optimizer}"
+                resources:
+                  limits:
+                    memory: "1Gi"
+                    cpu: "0.5"
```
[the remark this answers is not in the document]
andreyvelich: We just discussed this with @johnugeorge during the AutoML WG sync. Let's keep this resources since we have `ResourceQuota` in our Conformance profile.

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
andreyvelich: I saw your comment above: https://github.com/kubeflow/katib/pull/2138#discussion_r1155243034.
@nagar-ajay Did you test it if you can pass multiple annotation in that format ? E.g.:
```
'{"sidecar.istio.io/inject": "false", "custom-key": "custom-value"}' 
```

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.sh
```diff
@@ -44,7 +44,8 @@ fi
 for exp_name in "${EXPERIMENT_FILE_ARRAY[@]}"; do
   echo "Running Experiment from $exp_name file"
   exp_path=$(find ../../../../../examples/v1beta1 -name "${exp_name}.yaml")
-  python run-e2e-experiment.py --experiment-path "${exp_path}" --verbose || (kubectl get pods -n kubeflow && exit 1)
+  python run-e2e-experiment.py --experiment-path "${exp_path}" --namespace kubeflow \
```
tenzen-y, line 47: Do we need to run the test in the kubeflow namespace? I'd like to verify whether the katib controller can operate Experiments deployed out of the namespace in which the katib controller is deployed.

andreyvelich: It's a good point, maybe we could run our E2Es on `default` namespace.
WDYT @johnugeorge @nagar-ajay ?

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
nagar-ajay: yes

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.sh
nagar-ajay: @andreyvelich  @tenzen-y I tried running random experiment test in the default namespace, but it failed. As per the error logs, I think other namespaces require `katib.kubeflow.org/metrics-collector-injection: enabled` label.
Error log:
```
kubernetes.client.exceptions.ApiException: (400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Audit-Id': '689dd995-d83a-4915-9e69-130043993f14', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': '66727794-bef2-4ca7-a384-6a1a7b8a571d', 'X-Kubernetes-Pf-Prioritylevel-Uid': 'c6d64b7c-508c-4954-b20b-0b391dcc2e85', 'Date': 'Thu, 06 Apr 2023 12:40:50 GMT', 'Content-Length': '326'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"admission webhook \"validator.experiment.katib.kubeflow.org\" denied the request: Cannot create the Experiment \"random\" in namespace \"default\": the namespace lacks label \"katib.kubeflow.org/metrics-collector-injection: enabled\"","code":400}
```

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.py
nagar-ajay: Added `NotImplementedError`

test/e2e/v1beta1/scripts/gh-actions/run-e2e-experiment.sh
tenzen-y: Yes, we must add the label to the namespace.

Can you add processes to update the namespace label in `run-e2e-experiment.py` if the namespace passed by `--namespace` doesn't hold the label?

nagar-ajay: sure, will add.

nagar-ajay: Done. Also updated test namespace to `default`.

andreyvelich (commented):
Thank you for updating this @nagar-ajay!
/lgtm

PR #2140, opened 2023-03-31
Relax dependencies restriction for the gRPC libraries

### LINKED ISSUE #2139: katib python sdk incompatible pinned version
/kind bug

pip install latest kserve and katib, recieve 

```
ERROR: Cannot install -r /tmp/requirements.txt (line 16), -r /tmp/requirements.txt (line 18) and ray[serve]==2.0.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    kserve 0.10.0 depends on grpcio>=1.34.0
    kubeflow-katib 0.15.0 depends on grpcio==1.41.1
    ray[serve] 2.0.0 depends on grpcio<=1.43.0 and >=1.42.0; python_version >= "3.10"
```

is it possible to change the require grpcio version in katib?
Able to install both to use with kubeflow

tenzen-y:
I relaxed the version restrictions on dependencies in the Python SDK.

Fixes #2139

**Checklist:**

### DIFF
--- a/sdk/python/v1beta1/setup.py
+++ b/sdk/python/v1beta1/setup.py
@@ -22,8 +22,8 @@ REQUIRES = [
     "setuptools>=21.0.0",
     "urllib3>=1.15.1",
     "kubernetes>=23.6.0",
-    "grpcio==1.41.1",
-    "protobuf==3.19.5",
+    "grpcio>=1.41.1",
+    "protobuf>=3.19.5, <=3.20.3",
 ]
 
 katib_grpc_api_file = "../../../pkg/apis/manager/v1beta1/python/api_pb2.py"
--- a/cmd/earlystopping/medianstop/v1beta1/requirements.txt
+++ b/cmd/earlystopping/medianstop/v1beta1/requirements.txt
@@ -1,5 +1,5 @@
-grpcio==1.41.1
-protobuf==3.19.5
+grpcio>=1.41.1
+protobuf>=3.19.5, <=3.20.3
 googleapis-common-protos==1.6.0
 kubernetes==22.6.0
 cython>=0.29.24
--- a/cmd/metricscollector/v1beta1/tfevent-metricscollector/requirements.txt
+++ b/cmd/metricscollector/v1beta1/tfevent-metricscollector/requirements.txt
@@ -1,5 +1,5 @@
 psutil==5.9.4
 rfc3339>=6.2
-grpcio==1.41.1
+grpcio>=1.41.1
 googleapis-common-protos==1.6.0
 tensorflow==2.11.0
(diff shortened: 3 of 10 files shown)


tenzen-y:

cc: @ryandawsonuk

sdk/python/v1beta1/setup.py
```diff
@@ -22,8 +22,8 @@
     "setuptools>=21.0.0",
     "urllib3>=1.15.1",
     "kubernetes>=23.6.0",
-    "grpcio==1.41.1",
-    "protobuf==3.19.5",
+    "grpcio>=1.41.1",
+    "protobuf>=3.19.5",
```
andreyvelich, line 26: Thanks for updating this @tenzen-y!
Since we don't have proper tests in SDK to verify the GRPC lib functionality yet, can we update the versions in other Katib services as well. For example, [tf event collector](https://github.com/kubeflow/katib/blob/master/cmd/metricscollector/v1beta1/tfevent-metricscollector/requirements.txt#L3) and [other Suggestion services.](https://github.com/kubeflow/katib/blob/master/cmd/suggestion/hyperopt/v1beta1/requirements.txt#L7)

tenzen-y: You're right.
I will change the versions in this PR.

tenzen-y: Done.

tenzen-y:
It seems that we must use the `protobuf` version lower or equal to v3.20.

> Traceback (most recent call last):
  File "run-e2e-experiment.py", line 6, in <module>
    from kubeflow.katib import ApiClient, KatibClient, models
  File "/home/runner/work/katib/katib/sdk/python/v1beta1/kubeflow/katib/__init__.py", line 73, in <module>
    from kubeflow.katib.api.katib_client import KatibClient
  File "/home/runner/work/katib/katib/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py", line 27, in <module>
    import kubeflow.katib.katib_api_pb2 as katib_api_pb2
  File "/home/runner/work/katib/katib/sdk/python/v1beta1/kubeflow/katib/katib_api_pb2.py", line 35, in <module>
    type=None),
  File "/opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/site-packages/google/protobuf/descriptor.py", line 796, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

https://github.com/kubeflow/katib/actions/runs/4577679497/jobs/8083373344?pr=2140#step:4:10112

tenzen-y:
Maybe, we must regenerate codes using a newer protoc in follow-up PRs.

tenzen-y:
> Should we consider alternative ?

@andreyvelich I agree. However, I think we can do that in another PR. wdyt?

PR #2145, opened 2023-04-07
Adding groups support from request

axel7083:
## Description

The [oidc-authservice](https://github.com/arrikto/oidc-authservice) support the `GROUPS_HEADER`. In the current implementation the `Groups` field in the `SubjectAccessReviewSpec` is not used. Kubernetes have a build-in support for them.

The format for the value of `GROUPS_HEADER` is a string containing the groups separated with commas. 

By default `Kubeflow` and specifically the `authn-filter` envoyFilter will not propagate the `kubeflow-groups` header only the `kubeflow-userid`, but this is only a configuration files, the changes in this PR will allow to get the groups the user is associated with, and create a subject review with its user-id (email) and groups.

### DIFF
--- a/pkg/new-ui/v1beta1/authzn.go
+++ b/pkg/new-ui/v1beta1/authzn.go
@@ -16,17 +16,19 @@ import (
 
 // ENV variables
 var (
-	USER_HEADER  = env.GetEnvOrDefault("USERID_HEADER", "kubeflow-userid")
-	USER_PREFIX  = env.GetEnvOrDefault("USERID_PREFIX", ":")
-	DISABLE_AUTH = env.GetEnvOrDefault("APP_DISABLE_AUTH", "true")
+	USER_HEADER   = env.GetEnvOrDefault("USERID_HEADER", "kubeflow-userid")
+	GROUPS_HEADER = env.GetEnvOrDefault("GROUPS_HEADER", "kubeflow-groups")
+	USER_PREFIX   = env.GetEnvOrDefault("USERID_PREFIX", ":")
+	DISABLE_AUTH  = env.GetEnvOrDefault("APP_DISABLE_AUTH", "true")
 )
 
 // Function for constructing SubjectAccessReviews (SAR) objects
-func CreateSAR(user, verb, namespace, resource, subresource, name string, schema schema.GroupVersion) *v1.SubjectAccessReview {
+func CreateSAR(user string, userGroups []string, verb, namespace, resource, subresource, name string, schema schema.GroupVersion) *v1.SubjectAccessReview {
 
 	sar := &v1.SubjectAccessReview{
 		Spec: v1.SubjectAccessReviewSpec{
-			User: user,
+			User:   user,
+			Groups: userGroups,
 			ResourceAttributes: &v1.ResourceAttributes{
 				Namespace:   namespace,
 				Verb:        verb,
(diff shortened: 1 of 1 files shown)


tenzen-y:
@axel7083 Thank you for creating this PR.

@apo-ger Do you remember the reason why we didn't support the feature for the user groups in katib-ui?

apo-ger:
hey @tenzen-y, IIRC KF did not support user groups back then.

tenzen-y:
> hey @tenzen-y, IIRC KF did not support user groups back then.

Ah, I see. Thank you for reminding me!
@apo-ger I'd be happy if you help us with reviewing this PR :)

axel7083:
The [Go Test / Unit Test (1.25.0) (pull_request)](https://github.com/kubeflow/katib/actions/runs/4636908132/jobs/8447124243?pr=2145) which failed surprise me a bit, I tried to reproduce it in local but got the following for the failing test
````bash
$ go mod download && make test ENVTEST_K8S_VERSION=1.25.0
[...]
ok      github.com/kubeflow/katib/pkg/controller.v1beta1/experiment     6.842s  coverage: 76.7% of statements
[...]
```

andreyvelich:
@axel7083 Yeah, that tests are flaky. I will restart it

pkg/new-ui/v1beta1/authzn.go
```diff
@@ -52,11 +54,17 @@ func IsAuthorized(verb, namespace, resource, subresource, name string, schema sc
 	if r.Header.Get(USER_HEADER) == "" {
 		return "", errors.New("user header not present")
 	}
+
 	user := r.Header.Get(USER_HEADER)
 	user = strings.Replace(user, USER_PREFIX, "", 1)
 
+	var groups []string
+	if header := r.Header.Get(GROUPS_HEADER); header != "" {
+		groups = strings.Split(header, ",")
```
tenzen-y, line 63: Uhm. Probably, we should mention using `,` as a separator in the documentation.

@andreyvelich @johnugeorge Where should we notify users about this?

tenzen-y (commented):
Overall looks good to me :)

pkg/new-ui/v1beta1/authzn.go
andreyvelich: I don't think we have any documentation regarding the Katib UI auth. I think, we should create separate page for Katib UI, and move all relative docs there. For example, move this one: https://www.kubeflow.org/docs/components/katib/hyperparameter/#accessing-the-katib-ui.

@apo-ger @kimwnasptd @elenzio9 Do we have any place in the docs where we explain users how to use Kubeflow UIs ?

tenzen-y: Sounds great.
Also, adding more documentation for the Katib UI auth there would be better. Through I see the section, we don't mention the auth.

axel7083: @tenzen-y do you want me to include in this PR a section in the [accessing-the-katib-ui](https://www.kubeflow.org/docs/components/katib/hyperparameter/#accessing-the-katib-ui) about the auth system ?

andreyvelich: @axel7083 Let's wait until @apo-ger, @elenzio9 , or @kimwnasptd comment.
They were the main contributors to support authzn in Katib UI.

tenzen-y: I meant we'd like to work on that at https://github.com/kubeflow/website.

andreyvelich:
Hi @axel7083, do we have enough time to finish this PR before our feature freeze date Aug 2nd to include it in the next Katib release ?

axel7083:
> Hi @axel7083, do we have enough time to finish this PR before our feature freeze date Aug 2nd to include it in the next Katib release ?

As discussed with @kimwnasptd and @thesuperzapper  in [kubeflow#7085 Enabling RBAC Based on User Groups](https://github.com/kubeflow/kubeflow/issues/7085) we aim for another approach than making the change of adding the support for another header.

The goal would probably to migrate out of [oidc-authservice](https://github.com/arrikto/oidc-authservice) in favor of [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy), making a JWT token propagate instead of a plain text header.

Closing this as will not be completed in this form.

Issue #2110, opened 2023-02-10
KatibClient().list_experiments/list_trials fails with RuntimeError


votti:
/kind bug

**What steps did you take and what happened:**
I tried to list the experiments/trials using the Katib Client:
```
from kubeflow.katib import KatibClient

client=KatibClient()

my_namespace="my-namespace"
client.list_experiments(namespace=my_namespace)
client.list_trials(namespace=my_namespace)
```
raises the error

```
~/miniconda3/envs/katib-exp/lib/python3.11/site-packages/kubeflow/katib/api_client.py in __deserialize_model(self, data, klass)
    657                     value = data[klass.attribute_map[attr]]
--> 658                     kwargs[attr] = self.__deserialize(value, attr_type)
...
--> 285             raise RuntimeError(
    286                 "There was a problem to get experiments in namespace {0}. Exception: \
    287           {1} ".format(namespace, e))

RuntimeError: There was a problem to get experiments in namespace my-namespace. Exception:           module 'kubeflow.katib.models' has no attribute 'V1ResourceRequirements' 
```

**What did you expect to happen:**
No error.

**Anything else you would like to add:**
I was able to temporarily fix this by patching the `kubeflow.katib.models` namespace:

```
import kubeflow.katib.models as models
from kubernetes import client as kclient
models.V1ResourceRequirements = kclient.V1ResourceRequirements
```

**Environment:**

-     Katib version (check the Katib controller image version): 
       python sdk: katib==0.14.0
       docker.io/kubeflowkatib/katib-controller:v0.14.0"
-     Kubernetes version: (kubectl version):
-     Client Version: v1.26.0
-     Kustomize Version: v4.5.7
-     Server Version: v1.23.14-gke.1800
-     OS (uname -a): Linux LAPTOP-SS9UALAT 4.19.128-microsoft-standard SMP Tue Jun 23 12:58:10 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

---

Impacted by this bug? Give it a 👍 We prioritize the issues with the most 👍

votti:
I guess the issue is that I have a `V1ResourceRequirements` in one of my experiments/trials.
This causes the `deserialization` from str to class to fail.

Not sure if this is common enough to need to be supported.
This fix would work: https://github.com/kubeflow/katib/commit/815b34db23228f323f4d0684f480207e89440b68

andreyvelich:
[...]

PR #2148, opened 2023-04-20
SDK: Import all Kubernetes Models

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/2110
Currently, we need to import all Kubernetes models to Katib SDK, so SDK can do proper deserialization.

We need to track better solution in: https://github.com/kubeflow/training-operator/issues/1723

/assign @johnugeorge @tenzen-y @votti


tenzen-y (commented):
Thanks @andreyvelich!

/lgtm
/approve

Issue #2072, opened 2022-12-24
The `operators` directory is a bit out of date


tenzen-y:
/kind discussion

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
The [`operators`](https://github.com/kubeflow/katib/tree/master/operators) directory corresponds to katib v0.12.0, which is a bit out of date.

Also, it looks like the latest Charmed katib-operator exists at https://github.com/canonical/katib-operators. Those Charmed katib-operators don't seem to sync. This situation is likely to be confusing for users.

@DomFleischmann @DnPlas @ca-scribner @knkski Would you like to keep maintaining both `kubeflow/katib/operators` and `canonical/katib-operators`? Or would you like to remove Charmed katib-operator from this repository (katib repo)?

/cc @kubeflow/wg-automl-leads

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

tenzen-y:
ping @DomFleischmann @DnPlas @ca-scribner @knkski

ca-scribner:
Thanks @tenzen-y, I'll raise this in the team

tenzen-y:
> Thanks @tenzen-y, I'll raise this in the team

Thanks!

PR #2161, opened 2023-06-07
Remove Charmed Operators for Katib

ca-scribner:
This PR removes the Charmed Operators for Katib, as well as the associated tests.  

In the past kubeflow/katib was the source of truth for these operators, but they have since been maintained [here](https://github.com/canonical/katib-operators/) and we've done a poor job of keeping the repos in sync.

Reviewers, I think I've found everything but please let me know if I missed something that should also be removed. 

Fixes kubeflow/katib#2072 

**Checklist:**

  - afaict, there are no docs for this.


andreyvelich (commented):
Thank you for the update @ca-scribner!
I think, it makes sense to add Katib Charmed GitHub link to the installation docs: https://www.kubeflow.org/docs/components/katib/hyperparameter/#installing-katib

PR #2164, opened 2023-06-19
[UI] Fix Trial Logs when Kubernetes Job Fails

andreyvelich:
I fixed UI backend that should return proper value when Kubernetes Job Fails. When Job Fails, it spawns multiple Pods and our UI shows this: `More than one master replica found`.

To fix this, I follow this:

1.  If one Succeeded or Running Pod exists, we print the logs.
1. Otherwise, print logs from one of Failed Pods.

/assign @tenzen-y @kimwnasptd @johnugeorge @apo-ger @d-gol

### DIFF
--- a/pkg/new-ui/v1beta1/backend.go
+++ b/pkg/new-ui/v1beta1/backend.go
@@ -733,10 +733,15 @@ func fetchMasterPodName(clientset *kubernetes.Clientset, trial *trialsv1beta1.Tr
 		field to "true" in the Experiment definition. If this error persists then the Pod's logs are not currently
 		persisted in the cluster.`)
 	}
-	if len(podList.Items) > 1 {
-		return "", errors.New("More than one master replica found")
+
+	// If Pod is Running or Succeeded Pod, return it.
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodRunning {
+			return pod.Name, nil
+		}
 	}
 
+	// Otherwise, return the first Failed Pod.
 	return podList.Items[0].Name, nil
 }
 
--- a/pkg/new-ui/v1beta1/backend.go
+++ b/pkg/new-ui/v1beta1/backend.go
@@ -742,7 +742,14 @@ func fetchMasterPodName(clientset *kubernetes.Clientset, trial *trialsv1beta1.Tr
 	}
 
 	// Otherwise, return the first Failed Pod.
-	return podList.Items[0].Name, nil
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodFailed {
+			return pod.Name, nil
+		}
+	}
+
+	// Otherwise, return error since Pod is in the Pending state.
+	return "", errors.New("Failed to get logs for this Trial. Pod is in the Pending or Unknown state.")
 }
 
 // fetchPodLogs returns logs of a pod for the given job name and namespace


pkg/new-ui/v1beta1/backend.go
```diff
@@ -733,10 +733,15 @@ func fetchMasterPodName(clientset *kubernetes.Clientset, trial *trialsv1beta1.Tr
...
-		return "", errors.New("More than one master replica found")
+
+	// If Pod is Running or Succeeded Pod, return it.
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodRunning {
+			return pod.Name, nil
+		}
 	}
 
+	// Otherwise, return the first Failed Pod.
```
tenzen-y, line 744:
```suggestion
	// Otherwise, return the first Failed or Pending Pod.
}
```

Is this included in Pending Pods too, right?
So, should we return a failed pod if there are pending pods and failed pods in podList?

tenzen-y (commented):
@andreyvelich Thank you for fixing this bug!

pkg/new-ui/v1beta1/backend.go
andreyvelich: >Is this included in Pending Pods too, right?

Yes, that's right. In case of Pending pod UI shows the similar error with my change:
```
One of the Pod containers is is ContainerCreating stage
```
@tenzen-y Do you think we should process only Failed pod after we processed Succeeded and Running ?

And then, if there are no more Failed pods, just return `podList.Items[0].Name` (Pending Pod) and UI should show the above error ?

tenzen-y: > Do you think we should process only Failed pod after we processed Succeeded and Running ?

Yes, I do. I think users may prefer that show logs for a failed pod rather than a pending pod. wdyt?

Also in the current implementation, If there are pending pods and failed pods, maybe, the selected pods will differ each time we run `clientset.CoreV1().Pods(trial.ObjectMeta.Namespace).List...`.

Hence, for either a pending or failed pod, we should guarantee which phase this function returns.

> And then, if there are no more Failed pods, just return podList.Items[0].Name (Pending Pod) and UI should show the above error ?

Uhm. I'm not sure the above message is appropriate. However, I don't think that pending means `ContainerCreating`. For example, when using a custom scheduler to enforce pending pods based on priority or requested resources, it will take a bit much time until is scheduled into a Node and starts.

andreyvelich: >  logs for a failed pod rather than a pending pod. wdyt

That makes sense.

> The selected pods will differ each time we run clientset.CoreV1().Pods(trial.ObjectMeta.Namespace).

From my testing however, the [0] element is always the last created Pod (despite that list in `client-go` doesn't guarantee ordering).
I guess, that is how caching works behind list operation.

> However, I don't think that pending means ContainerCreating

That is right, it can be different message I just added this as an example. We can log that message to the user always (when Pod is a pending state): `Failed to get logs for this Trial. Pod is in the Pending state`. WDYT @tenzen-y ?

tenzen-y: > From my testing however, the [0] element is always the last created Pod (despite that list in client-go doesn't guarantee ordering).
I guess, that is how caching works behind list operation.

It's interesting. Thanks for sharing the result of the investigation!

> Failed to get logs for this Trial. Pod is in the Pending state.

Looks great :)

andreyvelich: @tenzen-y @johnugeorge I've changed it.

tenzen-y (commented):
Thanks!
/lgtm
/approve

PR #2165, opened 2023-06-30
feature: Adding trial name as env for metrics collectors

axel7083:

This will add the `KATIB_TRIAL_NAME `  env containing the trial name to all the containers.

This would simply many interactions, the trial name is injected for every metrics collectors, which made it more difficult to create custom collectors.

The typical example is the following in the https://github.com/kubeflow/katib/pull/2118

https://github.com/kubeflow/katib/blob/25ac27f644fc5fb2439ab375366fa3d3c66b3aba/cmd/metricscollector/v1beta1/kfp-metricscollector/v1/main.py#L68

Parsing the pod_name is not really a good practice. Instead injecting with a proper env is probably more suitable.
Injecting in every pods make it also easier for trial without metrics collector to push themself metrics to katib.

Fixes #

**Checklist:**

### DIFF
--- a/pkg/webhook/v1beta1/pod/inject_webhook.go
+++ b/pkg/webhook/v1beta1/pod/inject_webhook.go
@@ -234,6 +234,10 @@ func (s *SidecarInjector) getMetricsCollectorContainer(trial *trialsv1beta1.Tria
 		Args:            args,
 		ImagePullPolicy: metricsCollectorConfigData.ImagePullPolicy,
 		Resources:       metricsCollectorConfigData.Resource,
+		Env: []v1.EnvVar{{
+			Name:  consts.KatibTrialEnvName,
+			Value: trial.Name,
+		}},
 	}
 
 	// Inject the security context when the flag is enabled.
--- a/pkg/webhook/v1beta1/pod/inject_webhook.go
+++ b/pkg/webhook/v1beta1/pod/inject_webhook.go
@@ -144,6 +144,9 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 		return nil, err
 	}
 
+	// Add KATIB_TRIAL_NAME env name to all containers
+	mutatePodContainersEnv(mutatedPod, trial)
+
 	// Add Katib Trial labels to the Pod metadata.
 	mutatePodMetadata(mutatedPod, trial)
 
--- a/pkg/webhook/v1beta1/pod/utils.go
+++ b/pkg/webhook/v1beta1/pod/utils.go
@@ -265,7 +265,7 @@ func mutatePodContainersEnv(pod *v1.Pod, trial *trialsv1beta1.Trial) {
 	for i := range pod.Spec.Containers {
 		container := &pod.Spec.Containers[i]
 		container.Env = append(container.Env, v1.EnvVar{
-			Name:  "KATIB_TRIAL_NAME",
+			Name:  consts.KatibTrialEnvName,
 			Value: trial.Name,
 		})
 	}
--- a/pkg/webhook/v1beta1/pod/utils.go
+++ b/pkg/webhook/v1beta1/pod/utils.go
@@ -261,6 +261,16 @@ func mutateMetricsCollectorVolume(pod *v1.Pod, mountPath, sidecarContainerName, 
 	return nil
 }
 
+func mutatePodContainersEnv(pod *v1.Pod, trial *trialsv1beta1.Trial) {
+	for i := range pod.Spec.Containers {
+		container := &pod.Spec.Containers[i]
+		container.Env = append(container.Env, v1.EnvVar{
+			Name:  "KATIB_TRIAL_NAME",
+			Value: trial.Name,
+		})
+	}
+}
+
 func mutatePodMetadata(pod *v1.Pod, trial *trialsv1beta1.Trial) {
 	podLabels := map[string]string{}
 
--- a/pkg/controller.v1beta1/consts/const.go
+++ b/pkg/controller.v1beta1/consts/const.go
@@ -103,6 +103,8 @@ const (
 	// DefaultKatibComposerEnvName is the default env name of katib suggestion composer
 	DefaultKatibComposerEnvName = "KATIB_SUGGESTION_COMPOSER"
 
+	KatibTrialEnvName = "KATIB_TRIAL_NAME"
+
 	// DefaultKatibDBManagerServiceNamespaceEnvName is the env name of Katib DB Manager namespace
 	DefaultKatibDBManagerServiceNamespaceEnvName = "KATIB_DB_MANAGER_SERVICE_NAMESPACE"
 	// DefaultKatibDBManagerServiceIPEnvName is the env name of Katib DB Manager IP
--- a/pkg/webhook/v1beta1/pod/inject_webhook_test.go
+++ b/pkg/webhook/v1beta1/pod/inject_webhook_test.go
@@ -1065,3 +1065,68 @@ func TestMutatePodMetadata(t *testing.T) {
 		}
 	}
 }
+
+func TestMutatePodContainersEnv(t *testing.T) {
+	testTrialName := "hello-trial"
+
+	testCases := []struct {
+		pod             *v1.Pod
+		trial           *trialsv1beta1.Trial
+		mutatedPod      *v1.Pod
+		testDescription string
+	}{
+		{
+			pod: &v1.Pod{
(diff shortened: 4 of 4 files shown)


pkg/webhook/v1beta1/pod/inject_webhook.go
```diff
@@ -144,6 +144,9 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 		return nil, err
 	}
 
+	// Add KATIB_TRIAL_NAME env name to all containers
```
tenzen-y, line 147:
```suggestion
	// Add KATIB_TRIAL_NAME env name to all containers.
```

pkg/webhook/v1beta1/pod/inject_webhook.go
```diff
@@ -144,6 +144,9 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
...
+	mutatePodContainersEnv(mutatedPod, trial)
```
tenzen-y, line 148: If we mutate env vars to containers here, we can not mutate env vars to a metrics-collector container since we mutate a metrics-collector container to Pod in L173.

So should we move this process to place after all other mutations are done?

pkg/webhook/v1beta1/pod/utils.go
```diff
@@ -261,6 +261,16 @@ func mutateMetricsCollectorVolume(pod *v1.Pod, mountPath, sidecarContainerName,
 	return nil
 }
 
+func mutatePodContainersEnv(pod *v1.Pod, trial *trialsv1beta1.Trial) {
+	for i := range pod.Spec.Containers {
+		container := &pod.Spec.Containers[i]
+		container.Env = append(container.Env, v1.EnvVar{
+			Name:  "KATIB_TRIAL_NAME",
```
tenzen-y, line 268: Should we make `KATIB_TRIAL_NAME` constants here: https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/pkg/controller.v1beta1/consts/const.go#L101-L111?

tenzen-y (commented):
@axel7083 Thank you for the contribution! I left a few comments.

@kubeflow/wg-automl-leads Could you approve CI?

pkg/webhook/v1beta1/pod/inject_webhook.go
axel7083: To avoid refactoring all the Mutate function, I added manually the env when creating the metric collector container in getMetricsCollectorContainer.

tenzen-y: @axel7083 Should we put `mutatePodContainersEnv(mutatedPod, trial)` on L191?

tenzen-y: If doing so, we can remove L237-L240 maybe.

axel7083: I put the mutatePodContainersEnv at line [L148](https://github.com/kubeflow/katib/blob/efb18b376de261cd3f6336d506d5c16083041c33/pkg/webhook/v1beta1/pod/inject_webhook.go#L148) because the metric collector could be None, when it is none, the mutate return early:

https://github.com/kubeflow/katib/blob/efb18b376de261cd3f6336d506d5c16083041c33/pkg/webhook/v1beta1/pod/inject_webhook.go#L161-L163

And the goal of adding the mutatePodContainersEnv  and adding in getMetricsCollectorContainer is to avoid removing the early return, and limit the refractoring.
If desired, I can change to avoid making early return and place the mutatePodContainersEnv  at the end.

axel7083:
> Thank you for this contribution @axel7083! Currently, user can get the Trial name inside the training container using `TrialSpec`: https://www.kubeflow.org/docs/components/katib/trial-template/#use-trial-metadata-in-template. Here is an example: https://github.com/kubeflow/katib/blob/master/examples/v1beta1/trial-template/trial-metadata-substitution.yaml. With this change, do we want to always add the Trial name to the training containers ?
> 
> Also, I don't understand why we need to add Trial name to the metrics collector container ?

Hummm, I did not know that, I tried to made a custom collectors myself. Base on what I saw in the code

https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/pkg/webhook/v1beta1/pod/inject_webhook.go#L301-L303

I thought it was not possible to get the trial name without doing some weird parsing of the hostname. Like I saw in https://github.com/kubeflow/katib/pull/2118.

Then I do not really understand the following example inside `spec.metricsCollectorSpec`:

https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/examples/v1beta1/metrics-collector/custom-metrics-collector.yaml#L14-L36

TrialNamePrefix will be equal to the trial name ? Why the word `prefix` is used here ?

pkg/webhook/v1beta1/pod/inject_webhook.go
tenzen-y: Right. Thank you for the clarification.

andreyvelich:
>Then I do not really understand the following example inside spec.metricsCollectorSpec:

I think, that is required to get Trial name inside Metrics Collector container since it is not part of Trial Spec. 

>Other question then, can we use ${trialParameters.trialName} in custom metrics collector ?

No, that is not possible today.

I think, we have a few options:

1. Always inject Trial name to Metrics Collector and Training containers. Then, we should refactor [Trial Metadata substitution](https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/examples/v1beta1/trial-template/trial-metadata-substitution.yaml#L26).
2. Allow to set `${trialParameters.x}` in the Metrics Collector container. And do the appropriate substitution during [Trial creation](https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/pkg/controller.v1beta1/experiment/experiment_controller_util.go#L78-L79).
3. Get the Trial name from Pod labels. Since we always inject [Trial name to Trial pods](https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/pkg/webhook/v1beta1/pod/utils.go#L278), you can get this value in the metrics collector container as follows: `metadata.labels[katib.kubeflow.org/trial]`

What do you think @tenzen-y @johnugeorge @axel7083 ?

tenzen-y:
> > Then I do not really understand the following example inside spec.metricsCollectorSpec:
> 
> I think, that is required to get Trial name inside Metrics Collector container since it is not part of Trial Spec.
> 
> > Other question then, can we use ${trialParameters.trialName} in custom metrics collector ?
> 
> No, that is not possible today.
> 
> I think, we have a few options:
> 
> 1. Always inject Trial name to Metrics Collector and Training containers. Then, we should refactor [Trial Metadata substitution](https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/examples/v1beta1/trial-template/trial-metadata-substitution.yaml#L26).
> 2. Allow to set `${trialParameters.x}` in the Metrics Collector container. And do the appropriate substitution during [Trial creation](https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/pkg/controller.v1beta1/experiment/experiment_controller_util.go#L78-L79).
> 3. Get the Trial name from Pod labels. Since we always inject [Trial name to Trial pods](https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/pkg/webhook/v1beta1/pod/utils.go#L278), you can get this value in the metrics collector container as follows: `metadata.labels[katib.kubeflow.org/trial]`
> 
> What do you think @tenzen-y @johnugeorge @axel7083 ?

@andreyvelich Thanks for the summary.
+1 on the last option (get the trial name from pod labels) since I think we should use an existing feature rather than expanding the feature if we can inject TrialName into metrics-collector logs.

@axel7083 You can refer to the following how to get trialName from labels:

https://github.com/kubeflow/katib/blob/ede6e7410c92df359cb79eb01debfeac3507c59d/manifests/v1beta1/components/controller/controller.yaml#L54-L58

axel7083:
Alright the following solution as @andreyvelich suggested is very suitable for the requirements I had in mind.

```yaml
...
- name: KATIB_TRIAL_NAME
   valueFrom:
      fieldRef:
         fieldPath: metadata.labels['katib.kubeflow.org/trial']
...
```

Closing as resolved without modification

Issue #2155, opened 2023-05-17
Kubernetes versions for the next katib version


tenzen-y:
/kind discussion

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
Currently, we support 3 K8s versions, such as v1.23, v1.24, and v1.25.

However, v1.23 has already reached EoL, and v1.24 will reach EoL on 2023-07-28.

So we can support `v1.24-v1.26` or `v1.25-v1.27` in the next katib release.

ref: https://kubernetes.io/releases/

cc: @kubeflow/wg-automl-leads @anencore94 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

The training operator will support the `v1.25-v1.27` in the next release. 

ref: https://github.com/kubeflow/training-operator/issues/1803#issuecomment-1551201642

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

andreyvelich:
Thank you for creating this @tenzen-y.

I think, we agreed for these Kubernetes versions: `v1.25, v1.26, v1.27` for the next Katib release during our last [AutoML and Training Community Meeting](https://docs.google.com/document/d/1MChKfzrKAeFRtYqypFbMXL6ZIc_OgijjkvbqmwRV-64/edit#heading=h.vqsljon7kcug).

tenzen-y:
> Thank you for creating this @tenzen-y.
> 
> I think, we agreed for these Kubernetes versions: `v1.25, v1.26, v1.27` for the next Katib release during our last [AutoML and Training Community Meeting](https://docs.google.com/document/d/1MChKfzrKAeFRtYqypFbMXL6ZIc_OgijjkvbqmwRV-64/edit#heading=h.vqsljon7kcug).

Thank you for letting me know :)
I agree with the supported list.

PR #2189, opened 2023-08-02
Replace grpc_health_probe with the built-in gRPC container probe feature

tenzen-y:
The built-in gRPC container probe has been enabled as a default since K8s v1.24.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe

So we should replace grpc_health_probe with a built-in one.

Fixes #

**Checklist:**

### DIFF
--- a/cmd/db-manager/v1beta1/Dockerfile
+++ b/cmd/db-manager/v1beta1/Dockerfile
@@ -2,7 +2,6 @@
 FROM golang:alpine AS build-env
 
 ARG TARGETARCH
-ENV GRPC_HEALTH_PROBE_VERSION v0.4.15
 
 WORKDIR /go/src/github.com/kubeflow/katib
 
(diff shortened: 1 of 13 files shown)


cmd/db-manager/v1beta1/Dockerfile
```diff
@@ -2,7 +2,6 @@
 FROM golang:alpine AS build-env
 
 ARG TARGETARCH
-ENV GRPC_HEALTH_PROBE_VERSION v0.4.15
```
andreyvelich, line 5: @tenzen-y @gaocegege @johnugeorge I am wondering, do we still need HealthServicer for our Suggestions ?
https://github.com/kubeflow/katib/blob/master/cmd/suggestion/hyperopt/v1beta1/main.py#L30

tenzen-y: @andreyvelich I think we need HealthServer. What reasons do you think we need not HealthServer?

andreyvelich: Yeah, basically this prob triggers our HealthServer that we expose in our Suggestions/DB-manager: https://github.com/grpc/grpc/blob/master/doc/health-checking.md.

andreyvelich (commented):
/lgtm
/hold for review

gaocegege:
Then we could only support clusters above 1.24, if I understand it correctly.

But LGTM.

tenzen-y:
> Then we could only support clusters above 1.24, if I understand it correctly.

@gaocegege Thanks for the review. Yes, that's right. We're ok with supporting only clusters above v1.24 since the K8s v1.25 is required for the next katib version as a minimum required version: https://github.com/kubeflow/katib/issues/2155

gaocegege:
SGTM

PR #2190, opened 2023-08-02
Upgrade Go version to v1.20

tenzen-y:
I upgraded Go version to v1.20.

Fixes #

**Checklist:**

### DIFF
--- a/pkg/db/v1beta1/mysql/mysql.go
+++ b/pkg/db/v1beta1/mysql/mysql.go
@@ -17,11 +17,8 @@ limitations under the License.
 package mysql
 
 import (
-	crand "crypto/rand"
 	"database/sql"
 	"fmt"
-	"math/big"
-	"math/rand"
 	"os"
 	"time"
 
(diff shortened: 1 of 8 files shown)


pkg/db/v1beta1/mysql/mysql.go
```diff
@@ -59,26 +56,12 @@ func getDbName() string {
...
-	seed, err := crand.Int(crand.Reader, big.NewInt(1<<63-1))
-	if err != nil {
-		return nil, fmt.Errorf("RNG initialization failed: %v", err)
-	}
-	// We can do the following instead, but it creates a locking issue
-	//d.rng = rand.New(rand.NewSource(seed.Int64()))
-	rand.Seed(seed.Int64())
-
-	return d, nil
-}
```
tenzen-y, line 74: We need to replace the `rand.Seed()` function with another way since the `rand.Seed()` function is deprecated, as shown below.

> Deprecated: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access.

https://pkg.go.dev/math/rand#Seed

However, I don't think we need to set the global random seed to the same value every run since we don't use random values anywhere, as I can see. So, I removed this.

@andreyvelich @johnugeorge WDYT?

andreyvelich: I am not sure why did we set these values.
@gaocegege @johnugeorge Do you have any context ?

tenzen-y: @andreyvelich As I can see, I can not find any context for this random number by git blame.
It seems that this was copied from https://github.com/mlkube/katib/blob/0bed55aae3cfddf255e6d297e31d1fedbe5dca4a/db/interface.go#L65-L67 to here.

andreyvelich (approved):
/lgtm
/hold for review

gaocegege (approved):

gaocegege:
/lgtm

Issue #2191, opened 2023-08-02
Enable resource specification for trial containers in the `tune` method of the SDK 


droctothorpe:
/kind feature

**Describe the solution you'd like**
Currently, the only way to specify cpu and memory limits and requests for trial containers is by submitting a manifest directly. For example: [here](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/hp-tuning/random.yaml#L65-L68). This feature is absent from the SDK's `tune` method. It would be great if the SDK could support this capability. 

**Anything else you would like to add:**
About to submit a PR to this effect.

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

PR #2192, opened 2023-08-02
[SDK] Enable resource specification for trial containers

droctothorpe:
Currently, the only way to specify cpu and memory limits and requests for trial containers is by submitting a manifest directly. For example: [here](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/hp-tuning/random.yaml#L65-L68).

This PR adds resource specification capabilities to the SDK's `tune` method to achieve parity with direct manifest submission. `resources` defaults to `None`, which aligns with what `kubernetes.client.V1ResourceRequirements` defaults to, so when `resources` is not specified, the behavior is unmodified.

Fixes https://github.com/kubeflow/katib/issues/2191

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -16,7 +16,7 @@ import inspect
 import multiprocessing
 import textwrap
 import time
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import grpc
 import kubeflow.katib.katib_api_pb2 as katib_api_pb2
@@ -144,6 +144,7 @@ class KatibClient(object):
         max_trial_count: int = None,
         parallel_trial_count: int = None,
         max_failed_trial_count: int = None,
+        resources_per_trial: Union[dict, client.V1ResourceRequirements, None] = None,
         retain_trials: bool = False,
         packages_to_install: List[str] = None,
         pip_index_url: str = "https://pypi.org/simple",
@@ -177,6 +178,24 @@ class KatibClient(object):
                 values check this doc: https://www.kubeflow.org/docs/components/katib/experiment/#configuration-spec.
             parallel_trial_count: Number of Trials that Experiment runs in parallel.
             max_failed_trial_count: Maximum number of Trials allowed to fail.
+            resources_per_trial: A parameter that lets you specify how much
+            resources each trial container should have. You can either specify a
+            kubernetes.client.V1ResourceRequirements object (documented here:
+            https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1ResourceRequirements.md)
+            or a dictionary that includes one or more of the following keys:
+            `cpu`, `memory`, or `gpu` (other keys will be ignored). Appropriate
+            values for these keys are documented here:
+            https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/.
+            For example:
+                {
+                    "cpu": "1",
+                    "gpu": "1",
+                    "memory": "2Gi",
+                }
+            Please note, `gpu` specifies a resource request with a key of
+            `nvidia.com/gpu`, i.e. an NVIDIA GPU. If you need a different type
+            of GPU, pass in a V1ResourceRequirement instance instead, since it's
+            more flexible. This parameter is optional and defaults to None.
             retain_trials: Whether Trials' resources (e.g. pods) are deleted after Succeeded state.
             packages_to_install: List of Python packages to install in addition
                 to the base image packages. These packages are installed before
@@ -280,6 +299,15 @@ class KatibClient(object):
                 + exec_script
             )
 
+        if isinstance(resources_per_trial, dict):
+            if "gpu" in resources_per_trial:
+                resources_per_trial["nvidia.com/gpu"] = resources_per_trial.pop("gpu")
+
+            resources_per_trial = client.V1ResourceRequirements(
+                requests=resources_per_trial,
+                limits=resources_per_trial,
+            )
+
         # Create Trial specification.
         trial_spec = client.V1Job(
             api_version="batch/v1",
@@ -297,6 +325,7 @@ class KatibClient(object):
                                 image=base_image,
                                 command=["bash", "-c"],
                                 args=[exec_script],
+                                resources=resources_per_trial,
                             )
                         ],
                     ),
@@ -640,7 +669,6 @@ class KatibClient(object):
         namespace = namespace or self.namespace
 
         for _ in range(round(timeout / polling_interval)):
-
             # We should get Experiment only once per cycle and check the statuses.
             experiment = self.get_experiment(name, namespace, apiserver_timeout)
 
@@ -1175,7 +1203,6 @@ class KatibClient(object):
         )
 
         with katib_api_pb2.beta_create_DBManager_stub(channel) as client:
-
             try:
                 # When metric name is empty, we select all logs from the Katib DB.
                 observation_logs = client.GetObservationLog(


sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -147,6 +147,7 @@ def tune(
         retain_trials: bool = False,
         packages_to_install: List[str] = None,
         pip_index_url: str = "https://pypi.org/simple",
+        resources: Optional[client.V1ResourceRequirements] = None,
```
andreyvelich, line 150: Should we try to be more Data Scientists friendly here ?

I believe, users might not be familiar how V1ResourceRequirements should be set and how [limits and requests work in Kubernetes](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits).

Maybe we could define this argument like `resources_per_trial`, and allow user to simply set num of CPU/GPU if they require it:
```py
resources_per_trial = {"cpu": "1", "gpu: "3", "memory": "1Gb"}
```
We can convert those resources to Kubernetes Pod containers `limits`.
Also, We can allow to define `resources_per_trial` as V1ResourceRequirements object for users who are familiar with `limit` and `requests` in Kubernetes.
WDYT @droctothorpe @johnugeorge @gaocegege @tenzen-y @anencore94 ?

andreyvelich (commented):
Thank you for this great improvement @droctothorpe!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: I wondered the same thing! 🙃

I went back and forth on it for a bit before concluding that wrapping / redefining the k8s interface would introduce operational overhead that we might prefer to avoid.

Simplifications comes at the cost of configurability. Users may want to configure both `requests` and `limits` discretely, for example.

The overwhelming majority of users won't need this, and the few who do are probably smart enough to instantiate a `V1ResourceRequirements` object.

Happy to refactor either way. Just sharing some thoughts on the topic.

andreyvelich: That make sense.
That is why I think that we can make argument `resources_per_trial: object` and accept it as dictionary: 
`{"gpu": "3"}` or as `V1ResourceRequirements`. Thanks to Python nature.

droctothorpe: Sounds good. Would just specify type as `resources_per_trial: Union[dict, V1ResourceRequirements, None] = None`.

andreyvelich: `Union` even better, Thanks! 
@gaocegege @johnugeorge @tenzen-y @a9p @anencore94 What are your thoughts on this ?

tenzen-y: SGTM on `resources_per_trial: Union[dict, V1ResourceRequirements, None] = None`.

droctothorpe: We made the requested changes @andreyvelich @tenzen-y. The one thing we weren't 100% sure about was what key to use for the gpu resource request. Internally, we rely on `nvidia.com/gpu`, and that's what we saw when going through the Python Kubernetes client issues. Just want to make sure that it works for your use case too.

andreyvelich: I think, we can keep `nvidia.com/gpu` as a default resource for GPU initially.
In the future, we can introduce `gpu_type` parameter to our KatibClient to users can modify this value (e.g. for AMD GPU `amd.com/gpu`).

droctothorpe: Sounds good. AMD users can pass `V1ResourceRequirements` anyway so it's not like they're blocked.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -280,6 +295,23 @@ def tune(
...
+                "memory": "256Mi",
+            }
+            if "gpu" in resources_per_trial:
+                resources_per_trial["nvidia.com/gpu"] = resources_per_trial.pop("gpu")
+            requests.update(resources_per_trial)
+
+            resources = client.V1ResourceRequirements(
+                requests=requests,
+                limits=requests,
+            )
```
andreyvelich, line 311: I think, you can remove `requests` and only add `limits`.
Requests are not necessary and users always can use V1ResourceRequirements if they need it.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -280,6 +295,23 @@ def tune(
                 + exec_script
             )
 
+        resources = client.V1ResourceRequirements()
+        if isinstance(resources_per_trial, dict):
+            requests = {
+                "cpu": "200m",
+                "memory": "256Mi",
+            }
+            if "gpu" in resources_per_trial:
```
andreyvelich, line 304: Do you need to pop cpu and memory values from dict as well?

andreyvelich: Oh, I missed that!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: WDYT @johnugeorge @tenzen-y @droctothorpe ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -280,6 +295,23 @@ def tune(
...
+                resources_per_trial["nvidia.com/gpu"] = resources_per_trial.pop("gpu")
+            requests.update(resources_per_trial)
+
+            resources = client.V1ResourceRequirements(
+                requests=requests,
+                limits=requests,
+            )
+        else:
+            resources = resources_per_trial
+
```
andreyvelich, line 314: I think, we can simplify this.
Just re-configure `resources_per_trial` variable if `isinstance(resources_per_trial, dict):`
After that, just assign this variable to Pod Template Container.
In that case, we can avoid `else` statement.
Does it make sense @droctothorpe ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: Let me preface by saying that if any of this is stating the obvious, I humbly apologize.

Requests **guarantee** resources. Limits do not. If you don't specify requests, there's no guarantee that you'll receive the required resources.

Here's an analogy to clarify the distinction.

Let's say that you make a dinner reservation for 4 at your favorite restaurant. Those 4 seats are now guaranteed.

If you only made reservations for 2 and arrived with 4, there's no guarantee that the restaurant can provide 4 seats.

Requests are like restaurant reservations in that sense.

Now let's say that you reserved 4 seats and 4 more friends decide to join last minute.

When you arrive at the restaurant with a total of 8 people, the restaurant may or may not be able to accommodate the extra 4. It depends on how many open seats there are. The restaurant is like a Kubernetes node in this analogy. Nodes often have unused resources, and if pods don't set limits or have limits that exceed requests, the Kubernetes scheduler will let a pod consume the un-utilized resources, much like your 4 extra friends taking up 4 extra seats in the restaurant. 

Limits prevent this behavior. If a restaurant is really crowded, they might tell your 4 extra friends to get lost. Limits don't guarantee a reservation, they just restrict a pod from consuming additional, unreserved resources beyond a specified threshold.

One of the benefit of not specifying limits is that it can improve node resource utilization. One of the drawbacks is that it makes your executions less consistent: performance varies based on how much unused resources are available on the cluster.

All of which is to say that, IMO, requests are much more important than limits, and it's probably advisable to have them match in this context -- they actually match by default if resources are not specified. A user who wants to disable limits can always just use `V1ResourceRequirements`.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: No worries! I wasn't familiar with the `update` method until @shipengcheng1230 suggested that we use it during our pair programming session yesterday! 😅

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: Not sure I follow, sorry, @andreyvelich! Mind clarifying?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -182,6 +183,20 @@ def tune(
                 to the base image packages. These packages are installed before
                 executing the objective function.
             pip_index_url: The PyPI url from which to install Python packages.
+            resources_per trial: A parameter that lets you specify how much
```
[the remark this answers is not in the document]
droctothorpe: Done. Squashed commits (to simplify the merge) and added you as co-author.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: Do you mean something like this? i.e. reassign `resources_per_trial` and eliminate the `else`?

```python
        resources = client.V1ResourceRequirements()
        if isinstance(resources_per_trial, dict):
            requests = {
                "cpu": "200m",
                "memory": "256Mi",
            }
            if "gpu" in resources_per_trial:
                resources_per_trial["nvidia.com/gpu"] = resources_per_trial.pop("gpu")
            requests.update(resources_per_trial)

            resources_per_trial = client.V1ResourceRequirements(
                requests=requests,
                limits=requests,
            )
```

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Great analogy @droctothorpe!
In general, I agree with you.
It seems that looking at this doc, if `requests` are not set, Kubernetes make `request=limits`: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits
But in that case, maybe it worth to specify `limits` and `requests` together to not confuse user.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Yeah, exactly!

After that, we can just do this:
```python
client.V1Container(
  ...,
  resources= resources_per_trial,
)
                          
```

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -280,6 +296,23 @@ def tune(
                 + exec_script
             )
 
+        resources = client.V1ResourceRequirements()
+        if isinstance(resources_per_trial, dict):
+            requests = {
+                "cpu": "200m",
+                "memory": "256Mi",
+            }
```
andreyvelich, line 304: Also, I think we should avoid setting any default resources on behalf of the user.
If user sets only `gpu` in `resources_per_trial` dictionary, we should not set CPU and Memory, and resource allocation will be handle by Kubernetes.
WDYT @droctothorpe ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: > If you specify a limit for a resource, but do not specify any request, and no admission-time mechanism has applied a default request for that resource, then Kubernetes copies the limit you specified and uses it as the requested value for the resource.
TIL!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: Agreed that would be preferable. I have to check what K8s does when you pass a `V1ResourceRequirement` with one key set but not the others. Will test and report back.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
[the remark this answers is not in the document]
andreyvelich: @droctothorpe Did you commit this change ?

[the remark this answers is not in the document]
droctothorpe: Slipped through. Just did. Thanks!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: Looks like K8s fills in the gaps! Great call, @andreyvelich. Simplifies the logic too. Changes pushed up. I think that covers all the comments. Standing by for next steps.

Here are the validation results:

`tune` call:

```python
katib_client.tune(
    name=name,
    objective=objective,
    parameters=parameters,
    objective_metric_name="result",
    max_trial_count=12,
    base_image=image,
    namespace=namespace,
    algorithm_name="random",
    resources_per_trial={
        "cpu": 2
    }
)
```

Experiment spec:
```yaml
              resources:
                limits:
                  cpu: 2
                requests:
                  cpu: 2
```

Pod spec:
```yaml
    resources:
      limits:
        cpu: "2"
        memory: 256Mi
      requests:
        cpu: "2"
        memory: 256Mi
```

Once more with feeling, but specifying memory instead of CPU:

`tune` call:

```python
katib_client.tune(
    name=name,
    objective=objective,
    parameters=parameters,
    objective_metric_name="result",
    max_trial_count=12,
    base_image=image,
    namespace=namespace,
    algorithm_name="random",
    resources_per_trial={
        "memory": "2Gi"
    }
)
```

Experiment spec:
```yaml
              resources:
                limits:
                  memory: 2Gi
                requests:
                  memory: 2Gi
```

Pod spec:
```yaml
    resources:
      limits:
        cpu: 200m
        memory: 2Gi
      requests:
        cpu: 200m
        memory: 2Gi
```

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -177,6 +178,21 @@ def tune(
                 values check this doc: https://www.kubeflow.org/docs/components/katib/experiment/#configuration-spec.
             parallel_trial_count: Number of Trials that Experiment runs in parallel.
             max_failed_trial_count: Maximum number of Trials allowed to fail.
+            resources_per_trial: A parameter that lets you specify how much
+            resources each trial container should have. You can either specify a
+            kubernetes.client.V1ResourceRequirements object (documented here:
+            https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1ResourceRequirements.md)
+            or a dictionary that includes one or more of the following keys:
+            `cpu`, `memory`, or `gpu` (other keys will be ignored). Appropriate
```
tenzen-y, line 186: @droctothorpe Can you mention that `gpu` means NVIDIA GPU, and if users want to use other GPUs, users must pass the values as a client.V1ResourceRequirements form?

droctothorpe: Good call. Thanks for the feedback. Done!

tenzen-y: @droctothorpe As I can see, the commit isn't pushed to this PR. Can you check if the commit is pushed?

droctothorpe: Doh! Forgot to `git add` 🙃 . Thanks, @tenzen-y! Fixed.

andreyvelich (commented):
Thank you for this great feature @droctothorpe!
/lgtm
/assign @tenzen-y @johnugeorge @gaocegege

tenzen-y (commented):
@droctothorpe Thanks!
/lgtm
/approve

PR #2198, opened 2023-08-05
Bug: Wait for the certs to be mounted inside the container

tenzen-y:
To avoid the below error the cert-generator should wait for the certs to be mounted inside the container:

```shell
{"level":"error","ts":"2023-08-05T18:14:44Z","logger":"entrypoint","msg":"Unable to run the manager","error":"open /tmp/cert/tls.crt: no such file or directory","stacktrace":"main.main\n\t/go/src/github.com/kubeflow/katib/cmd/katib-controller/v1beta1/main.go:164\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
```

Currently, the cert-generator immediately sends data to the `certsReady` channel once the certs are generated.
However, since completed to generate certs doesn't mean the mounted secret is updated with certs, the cert-generator should wait for the mounted secret to be updated by kubelet.

ref: https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod.

Fixes #

**Checklist:**

### DIFF
--- a/pkg/certgenerator/v1beta1/generator.go
+++ b/pkg/certgenerator/v1beta1/generator.go
@@ -26,12 +26,15 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -67,9 +70,33 @@ func (c *CertGenerator) Start(ctx context.Context) error {
 	if err := c.generate(ctx); err != nil {
 		return err
 	}
+	klog.Info("Waiting for certs to get ready.")
+	if err := wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
+		Duration: time.Second,
+		Factor:   2,
+		Jitter:   1,
+		Steps:    10,
+		Cap:      time.Minute * 5,
+	}, ensureCertMounted); err != nil {
+		return err
+	}
 	// Sending an empty data to a certsReady means it starts to register controllers to the manager.
 	c.certsReady <- struct{}{}
 	return nil
+}
+
+// ensureCertMounted ensures that the generated certs are mounted inside the container.
+func ensureCertMounted(context.Context) (bool, error) {
+	certFile := filepath.Join(consts.CertDir, serverCertName)
+	if _, err := os.Stat(certFile); err != nil {
+		return false, nil
+	}
+	keyFile := filepath.Join(consts.CertDir, serverKeyName)
+	if _, err := os.Stat(keyFile); err != nil {
+		return false, nil
+	}
+	klog.Info("Succeeded to be mounted certs inside the container.")
+	return true, nil
 }
 
 func (c *CertGenerator) NeedLeaderElection() bool {
--- a/pkg/certgenerator/v1beta1/generator.go
+++ b/pkg/certgenerator/v1beta1/generator.go
@@ -56,11 +56,11 @@ type CertGenerator struct {
 	namespace          string
 	webhookServiceName string
 	webhookSecretName  string
+	fullServiceDomain  string
 	kubeClient         client.Client
 	certsReady         chan struct{}
 
-	certs             *certificates
-	fullServiceDomain string
+	certs *certificates
 }
 
 var _ manager.Runnable = &CertGenerator{}
@@ -109,8 +109,13 @@ func AddToManager(mgr manager.Manager, config configv1beta1.CertGeneratorConfig,
 		namespace:          consts.DefaultKatibNamespace,
 		webhookServiceName: config.WebhookServiceName,
 		webhookSecretName:  config.WebhookSecretName,
-		kubeClient:         mgr.GetClient(),
-		certsReady:         certsReady,
+		fullServiceDomain: strings.Join([]string{
+			config.WebhookServiceName,
+			consts.DefaultKatibNamespace,
+			"svc",
+		}, "."),
+		kubeClient: mgr.GetClient(),
+		certsReady: certsReady,
 	})
 }
 
@@ -126,8 +131,6 @@ func (c *CertGenerator) generate(ctx context.Context) error {
 		return fmt.Errorf("%w: %v", errCertCheckFail, err)
 	}
 	if !certExist {
-		c.fullServiceDomain = strings.Join([]string{c.webhookServiceName, c.namespace, "svc"}, ".")
-
 		if err = c.createCert(); err != nil {
 			return fmt.Errorf("%w: %v", errCreateCertFail, err)
 		}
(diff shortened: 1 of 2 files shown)


pkg/certgenerator/v1beta1/generator.go
```diff
@@ -67,11 +70,34 @@ func (c *CertGenerator) Start(ctx context.Context) error {
 	if err := c.generate(ctx); err != nil {
 		return err
 	}
+	if err := wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
```
johnugeorge, line 73: Is there a requirement for a timeout? How does the user get to know that it is being waited for certs to be ready?

tenzen-y: > Is there a requirement for a timeout?

Yes, timeout is 5 minutes.

> How does the user get to know that it is being waited for certs to be ready?

The controller pod doesn't ready until the cert is ready.

```shell
NAME                                READY   STATUS              RESTARTS   AGE
katib-controller-687798c75b-mtlz7   0/1     Running             0          4s

status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2023-08-13T11:48:12Z"
    message: 'containers with unready status: [katib-controller]'
    reason: ContainersNotReady
    status: "False"
    type: ContainersReady
controller@sha256:86697de94ac8dfbe53082beb96f1ea4ffea2bac982a2332719a99224e17335c7
    lastState: {}
    name: katib-controller
    ready: false
    restartCount: 0
    started: true
    state:
      running:
        startedAt: "2023-08-13T11:48:14Z"
```

johnugeorge: I meant, how will the user know that Katib controller is waiting for certs to get ready?

tenzen-y: The controller doesn't say, `Waiting for certs to get ready`. So, there are two ways that users can indirectly know whether the controller waits for certs to get ready:

1. As shown above, confirmation pod condition.
2. Creating the Experiments. If the controller waits for certs, users can get the following error:

```
Error from server (InternalError): error when creating "../../testdata/valid-experiment.yaml": Internal error occurred: failed calling webhook "defaulter.experiment.katib.kubeflow.org": could not get REST client: unable to load root certificates: unable to parse bytes as PEM block
```

Adding a log, `Waiting for certs to get ready`, to L72 in the file might be better.

johnugeorge: Adding a log is better because user can understand the reason behind the issue

tenzen-y: Added.

pkg/certgenerator/v1beta1/generator.go
```diff
@@ -64,26 +68,54 @@ var _ manager.Runnable = &CertGenerator{}
...
 func (c *CertGenerator) Start(ctx context.Context) error {
-	if err := c.generate(ctx); err != nil {
+	if c.enableGenerateCerts {
+		if err := c.generate(ctx); err != nil {
+			return err
+		}
+	}
+	// We need to wait for certs to get ready even though using cert-manager.
+	klog.Info("Waiting for certs to get ready.")
+	if err := wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
```
tenzen-y, line 78: Even though we use cert-manager, the controller must wait for certs to be injected into the Secret resource.

andreyvelich, line 78: @johnugeorge @tenzen-y Does it look confusing that we have this timeout in our generator package even if Katib Cert Generator doesn't used ?
Maybe we can add this timeout to the [`setupControllers` section before we adding the Webhook to the manager](https://github.com/kubeflow/katib/blob/69e9dfea1dea28dc1f5807226f7425efb31b71b6/cmd/katib-controller/v1beta1/main.go#L167-L170) ?

We can rename `certsReady` channel to `certsGenerated` to be more precise.

pkg/certgenerator/v1beta1/generator.go
```diff
@@ -64,26 +68,54 @@ var _ manager.Runnable = &CertGenerator{}
...
+// ensureCertMounted ensures that the generated certs are mounted inside the container.
+func ensureCertMounted(context.Context) (bool, error) {
+	certFile := filepath.Join(consts.CertDir, serverCertName)
+	if _, err := os.Stat(certFile); err != nil {
+		return false, nil
+	}
+	keyFile := filepath.Join(consts.CertDir, serverKeyName)
+	if _, err := os.Stat(keyFile); err != nil {
+		return false, nil
+	}
```
andreyvelich, line 101: Should we add the log line here as well ? E.g.
```
klog.Info("Public key file %v doesn't exist in the container yet", certFile)
klog.Info("Private key file %v doesn't exist in the container yet", keyFile)
```

pkg/certgenerator/v1beta1/generator.go
```diff
@@ -64,26 +68,54 @@ var _ manager.Runnable = &CertGenerator{}
...
-		webhookServiceName: config.WebhookServiceName,
-		webhookSecretName:  config.WebhookSecretName,
-		kubeClient:         mgr.GetClient(),
-		certsReady:         certsReady,
+		namespace:           consts.DefaultKatibNamespace,
+		webhookServiceName:  config.WebhookServiceName,
+		webhookSecretName:   config.WebhookSecretName,
+		kubeClient:          mgr.GetClient(),
+		certsReady:          certsReady,
+		enableGenerateCerts: config.Enable,
```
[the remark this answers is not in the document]
tenzen-y: It makes sense. Thanks!

pkg/certgenerator/v1beta1/generator.go
tenzen-y: I think locating this timeout here would be better since this timeout is a function for the webhook certs.
However, I agree with your confusion. So renaming the `certgenerator` with `certmanager` or `certcontroller` might be better to clarify the purpose of this component.

@andreyvelich WDYT?

tenzen-y: I think `certcontroller` would be better since `certmanager` conflicts context with [`cert-manager`](https://cert-manager.io/) :-/

pkg/certgenerator/v1beta1/generator.go
tenzen-y: I hesitate to add the log since the logs will be output in large quantities and the controller's logs will be dirty.
So output logs every 30 sec or 15 sec might be better. 
@andreyvelich WDYT?

pkg/certgenerator/v1beta1/generator.go
andreyvelich: If we rename it to `certcontroller` how we can deal with Katib Config Settings ?
If user wants the enable cert generator they set the following setting:
```
  certGenerator:
    enable: true
```

pkg/certgenerator/v1beta1/generator.go
andreyvelich: Sure, that also works, if we can print them every 15 sec.

pkg/certgenerator/v1beta1/generator.go
tenzen-y: So, the `certcontroller` will start only in the standalone installation.

andreyvelich: I think, if we are going to remove this timeout from Kubeflow installation, we need to name it as `certgenerator` since it is only internal logic that we have when Katib `certGenerator` is enabled.
Let's keep it as `certgenerator` for now and use this timeout only when our internal Cert Generator is used.
Does it sound good @tenzen-y ?

tenzen-y: @andreyvelich That sounds good to me. Thanks for the great suggestion!

tenzen-y: Done.

tenzen-y:
@andreyvelich I addressed your suggestions. Please take another look.

andreyvelich:
Thank you @tenzen-y!
/lgtm
/hold for review
/assign @johnugeorge

PR #2199, opened 2023-08-08
[SDK] Raise more human-readable name conflict exception

droctothorpe:
If you call `KatibClient.tune` using an experiment name that already exists, you get a giant traceback with the relevant / helpful message not being at the bottom. We had a user struggle to debug this on their own.

This PR simply raises a more helpful exception in the event of a 409 (name conflict) response from the API.

**Checklist:**

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -105,7 +105,11 @@ class KatibClient(object):
             raise TimeoutError(
                 f"Timeout to create Katib Experiment: {namespace}/{experiment.metadata.name}"
             )
-        except Exception:
+        except Exception as e:
+            if hasattr(e, "status") and e.status == 409:
+                raise Exception(
+                    f"A Katib Experiment with the name {namespace}/{experiment.metadata.name} already exists."
+                )
             raise RuntimeError(
                 f"Failed to create Katib Experiment: {namespace}/{experiment.metadata.name}"
             )


sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -105,7 +105,11 @@ def create_experiment(
             raise TimeoutError(
                 f"Timeout to create Katib Experiment: {namespace}/{experiment.metadata.name}"
             )
-        except Exception:
+        except Exception as e:
+            if hasattr(e, "status") and e.status == 409:
+                raise Exception(
+                    f"A Katib Experiment with the name {namespace}/{experiment.metadata.name} already exists."
+                )
```
tenzen-y, line 112: I think instead of checking the error, we should check if the requested Experiment already exists in the cluster before trying to create the Experiment in L97-L103.

cc: @andreyvelich

andreyvelich: I think, the downside of this that we require to call Kubernetes API server twice to always get Katib Experiment before creating this.
Would it be performance issue ?
On the other hand, KFP checks if Experiment exists before creating this: https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L473-L475.

What are your thoughts @tenzen-y @droctothorpe @johnugeorge ?

tenzen-y: > Would it be performance issue ?

It might be better to improve the performance as much possible as. Although, I think the performance wouldn't be much worse even if we call the api-server twice.

> On the other hand, KFP checks if Experiment exists before creating this: https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L473-L475.

I'm ok with the way If we adopt a way, the same as KFP since I want to avoid depending status code.

droctothorpe: Addressed your feedback, @tenzen-y. Apologies for the delay.

Good point, @andreyvelich. FWIW, the performance impact is negligible, `tune` isn't called repeatedly in a way where performance is a concern, IMO the code is a bit more clean and clear with the explicit check, and aligning with the KFP SDK is a nice plus.

I could go either way and am happy to implement either pattern.

droctothorpe: I decoupled the commits just in case we want to revert to the exception handling approach 🙃 .

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -93,6 +93,14 @@ def create_experiment(
 
         namespace = namespace or self.namespace
 
+        if experiment.metadata.name in [
+            experiment.metadata.name
+            for experiment in self.list_experiments(namespace=namespace)
```
andreyvelich, line 98: Should [we use `get_experiment()`](https://github.com/kubeflow/katib/blob/bf1ce98a832c4f76ede1afe1b54fee0bac92723c/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L361) API instead ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: @droctothorpe @tenzen-y What kind of pros and cons we have between:
- Calling API Server before creating Experiment
- Checking the Error code ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: The problem with `get_experiment` is that if the experiment doesn't exist, the API returns a 404, in which case we have to check for a 404 as well, at which point we might as well just check for a 409 in the create like we did originally, heh.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
droctothorpe: #### `list_experiment` first
- PRO: Cleaner code.
- PRO: Aligned with the KFP SDK.
- CON: Extra API call every time `tune` is called.

#### Handle 409
- PRO: [EAFP](https://docs.python.org/2/glossary.html#term-eafp).
- PRO: Less API calls.

andreyvelich: @droctothorpe @tenzen-y If we know that 409 code always represents that CR is already created do we have any other concerns why we don't want to follow this logic to check error code ?

For example, in the future we can handle other error codes (e.g. 401 means that user is not authorized to perform API call).

droctothorpe: Good point! Pushed up a revert commit. Happy to squash all 3 or they can be squashed at merge. Awaiting further input.

andreyvelich: @tenzen-y @johnugeorge What do you think about it ?

droctothorpe: Bump 🙃

andreyvelich (commented):
Thank you for updating this @droctothorpe 
/lgtm
/approve

andreyvelich:
@droctothorpe Please can you rebase your PR, it should fix the E2Es.

droctothorpe:
> @droctothorpe Please can you rebase your PR, it should fix the E2Es.

Squashed and rebased @andreyvelich. Thank you!

tenzen-y:
Thank you!
/lgtm

PR #2203, opened 2023-08-13
Skip to inject the metrics-collector pods to the katib controller

### LINKED ISSUE #2069: Add Label to Disable Katib Webhooks
In this PR: https://github.com/kubeflow/katib/pull/2018#issuecomment-1330713221, I proposed to introduce label for disabling Katib Webhooks (validator, defaulter, mutator). For example: `katib.kubeflow.org/webhooks: disabled`.
Let's discuss if that would be useful for the users with large-scale environment.

Currently, if user's namespace has `katib.kubeflow.org/metrics-collector-injection: enabled` label, Katib Mutation Webhook runs for every Pod in that namespace. That might increase latency in the Kubernetes API server. Some users might want to use Katib Experiments and run other pods in their namespaces without Webhook execution.

What do you think @gaocegege @johnugeorge @tenzen-y @anencore94 @terrytangyuan ?

/kind discussion

---

tenzen-y:
I introduced the label to skip injecting the metrics-collector to the katib system components.

Fixes #2069

**Checklist:**

### DIFF
--- a/manifests/v1beta1/components/controller/controller.yaml
+++ b/manifests/v1beta1/components/controller/controller.yaml
@@ -15,6 +15,7 @@ spec:
     metadata:
       labels:
         katib.kubeflow.org/component: controller
+        katib.kubeflow.org/metrics-collector-injection: disabled
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/port: "8080"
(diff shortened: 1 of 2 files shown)


tenzen-y:
I discussed offline with @andreyvelich and @johnugeorge.

Let me summarize. The replicaset controller occasionally faces the following errors since the webhooks are registered to the kube-apiserver before the controller pod is started. So we set the label to avoid injecting the metrics-collector pod only into the controller pod. 

Also in the future, we will switch to [`AdmissionWebhookMatchConditions`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions) feature to process requests more accurately once the feature gate is enabled as a default.

andreyvelich (commented):
Thank you @tenzen-y!
/lgtm
/hold for review

PR #2221, opened 2023-08-22
Run Stale Action Every 5th Hour

andreyvelich:
Currently, Stale Actions process a few issues/PRs per run every day: https://github.com/kubeflow/katib/actions/runs/5941243672

I changed Stale action to run every 5 hour to process more stale issues/PRs.
That should help us to track more issues and don't hit GitHub REST API limits: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions
Also, I changed `operations-per-run` to 60.

WDYT @tenzen-y @johnugeorge ?

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -7,7 +7,7 @@ name: Mark stale issues and pull requests
 
 on:
   schedule:
-    - cron: "0 16 * * *"
+    - cron: "0 */5 * * *"
 
 jobs:
   stale:


.github/workflows/stale.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
       - uses: actions/stale@v5
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+          operations-per-run: 60
```
tenzen-y, line 23: Can we keep using the default value (30) here? My concern is that by increasing this value to 60, other jobs failed to run due to the GitHub API limit.

WDYT?

andreyvelich: Sure, sounds good. Let's keep the default value for now since we are going to run this action every 5th hour.

tenzen-y: If any problem occurs, re-thinking the parameters would be great :)

tenzen-y:
Thanks!
/lgtm

PR #2153, opened 2023-05-10
Add Katib ROADMAP 2022/2023

andreyvelich:
I created initial draft for Katib ROADMAP 2022/2023.
I tried to consolidated features that we implemented in 2022, since we didn't update the ROADMAP in 2022 (if I missed some significant features that we've done, please let me know).

Please let me know if you have better ideas how to name headers.

Let's discuss the items during next month and finalise the ROADMAP.

/cc @johnugeorge @gaocegege @tenzen-y @anencore94 @kimwnasptd @elenzio9 @orfeas-k @apo-ger @d-gol @shaowei-su @zhixian82 @fischor @keisuke-umezawa @jbottum @DnPlas @juliusvonkohout 

cc @kubeflow/wg-training-leads

PR #2222, opened 2023-08-24
Update Ubuntu to 22.04 for E2E Tests

andreyvelich:
I updated Ubuntu version to `22.04` for our E2E tests to fix errors.
For more context follow this thread: https://github.com/kubeflow/katib/pull/2153#discussion_r1304485202

### DIFF
--- a/.github/workflows/build-and-publish-images.yaml
+++ b/.github/workflows/build-and-publish-images.yaml
@@ -22,7 +22,7 @@ on:
 jobs:
   build-and-publish:
     name: Publish Image
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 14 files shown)


tenzen-y (commented):
Thank you for this!

Should we use the Ubuntu 22.04 in other tests as well?
For example: https://github.com/kubeflow/katib/blob/373f6e6d7dbbe112f124f5781e511c88ad3899c4/.github/workflows/build-and-publish-images.yaml#L25

andreyvelich:
> Thank you for this!
> 
> Should we use the Ubuntu 22.04 in other tests as well? For example:
> 
> https://github.com/kubeflow/katib/blob/373f6e6d7dbbe112f124f5781e511c88ad3899c4/.github/workflows/build-and-publish-images.yaml#L25

Good point, let me keep 22.04 everywhere.

tenzen-y:
Thanks!
/lgtm

PR #2227, opened 2023-09-19
[SDK] Add 'algorithm_settings' in client tune

### LINKED ISSUE #2225: Add function of tune, a function supported by Katibclient on python sdk
/kind feature

Add function of tune, a function supported by Katibclient on python sdk
Please support algorithm settings as an argument from the KatibClient.tune function
I think that the algorithm_setting value should be entered at the moment of defining the Experiment.Spec.algorithm inside the tune function.
like this
` experiment.spec.algorithm = models.V1beta1AlgorithmSpec(
            algorithm_name=algorithm_name,
            algorithm_settings=algorithm_settings
        )`
please support it.

---

shipengcheng1230:
The `algorithm_settings` cannot be set in the `tune` method of the python SDK, therefore many details [here](https://www.kubeflow.org/docs/components/katib/experiment/#search-algorithms-in-detail) cannot be immediately used if not submitting the manifest directly. 

This PR adds the capability to do it. By default, `algorithm_settings` is set to `None` which aligns with the initiation of `V1beta1AlgorithmSpec`. So when not setting `algorithm_settings` the behavior is not changed.

Fixes #2225

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -141,6 +141,7 @@ class KatibClient(object):
         base_image: str = constants.BASE_IMAGE_TENSORFLOW,
         namespace: Optional[str] = None,
         algorithm_name: str = "random",
+        algorithm_settings: Union[dict, List[models.V1beta1AlgorithmSetting], None] = None,
         objective_metric_name: str = None,
         additional_metric_names: List[str] = [],
         objective_type: str = "maximize",
@@ -172,6 +173,8 @@ class KatibClient(object):
             base_image: Image to use when executing the objective function.
             namespace: Namespace for the Experiment.
             algorithm_name: Search algorithm for the HyperParameter tuning.
+            algorithm_settings: Settings for the search algorithm given.
+                For available fields, check this doc: https://www.kubeflow.org/docs/components/katib/experiment/#search-algorithms-in-detail.
             objective_metric_name: Objective metric that Katib optimizes.
             additional_metric_names: List of metrics that Katib collects from the
                 objective function in addition to objective metric.
@@ -232,8 +235,12 @@ class KatibClient(object):
             experiment.spec.objective.goal = objective_goal
 
         # Add Algorithm to the Katib Experiment.
+        if isinstance(algorithm_settings, dict):
+            algorithm_settings = [models.V1beta1AlgorithmSetting(name=str(k), value=str(v)) for k, v in algorithm_settings.items()]
+
         experiment.spec.algorithm = models.V1beta1AlgorithmSpec(
-            algorithm_name=algorithm_name
+            algorithm_name=algorithm_name,
+            algorithm_settings=algorithm_settings,
         )
 
         # Add Trial budget to the Katib Experiment.


sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -141,6 +141,7 @@ def tune(
         base_image: str = constants.BASE_IMAGE_TENSORFLOW,
         namespace: Optional[str] = None,
         algorithm_name: str = "random",
+        algorithm_settings: Union[dict, list[models.V1beta1AlgorithmSetting], None] = None,
```
andreyvelich, line 144: I am wondering should we discuss idea that @droctothorpe suggested here: https://github.com/kubeflow/katib/issues/2225#issuecomment-1719896866.

E.g. we can create separate Object for every Suggestion that Katib supports and add algorithm settings, search space, other features for this suggestion: `katib.suggestion.optuna`, `katib.suggestion.skopt`. 
Different Suggestions might have different supported features that we can add to these classes.
For example: @DeukWoongWoo asked for callable search space support for Optuna: https://github.com/kubeflow/katib/issues/2228

I understand that will add maintain overhead, but user will get better experience.
We need to discuss what is the easiest way for the user to get access for Katib Suggestion features.

What are your thoughts @johnugeorge @tenzen-y @droctothorpe @shipengcheng1230 @DeukWoongWoo @a9p @c-bata @anencore94 ?

droctothorpe: We are building similar configuration objects internally for stuff like dask and pytorch. Code completion, type hints, and robust docstrings make managing complex configurations a lot easier. All for it.

DeukWoongWoo: @andreyvelich 
I totally agree completely with the statement: "We need to discuss what is the easiest way for the user to get access to Katib Suggestion features." However, I think it's important to note that this PR is primarily aimed at enabling the setting of algorithm settings in katib_client, which is not currently possible as mentioned in the ["search-algorithms-in-detail" guide.](https://www.kubeflow.org/docs/components/katib/experiment/#search-algorithms-in-detail)
It's a bit unclear whether what you're trying to say is discussing features that will be added, or if you're focused on discussing how users will be able to access input into algorithm settings as this PR.

shipengcheng1230: Hey @andreyvelich, as @DeukWoongWoo suggested, do you feel comfortable to let this PR just enable the specification and have future one to validate and give suggestions to the input of algorithm settings?

andreyvelich: Thanks everyone for your comments!
>We are building similar configuration objects internally for stuff like dask and pytorch.

@droctothorpe Would you mind share some details here ? Is there something that we can adopt with Katib too ?

>It's a bit unclear whether what you're trying to say is discussing features that will be added, or if you're focused on discussing how users will be able to access input into algorithm settings as this PR.

I just want to discuss how to streamline experience for users to use Suggestions that Katib supports (e.g. Optuna, Hyperopt, etc.). Currently, it's not clear which underlying framework we use for HP tuning algorithms until user goes to [Search Algorithm in Detail guide](https://www.kubeflow.org/docs/components/katib/experiment/#search-algorithms-in-detail).
Also, there are overlaps in supported algorithms between different frameworks: https://www.kubeflow.org/docs/components/katib/katib-config/#suggestion-settings.
E.g. `tpe` is supported by Hyperopt and Optuna (I assume that 2 frameworks might have different features that we can use in Katib + Kubernetes).

As an user experience example, in Ray tune they wrap every supported framework in a separate module [(e.g. `ray.tune.search.optuna`)](https://docs.ray.io/en/latest/tune/api/doc/ray.tune.search.optuna.OptunaSearch.html)  that might have custom settings.
Wondering what would be the optimal experience for Katib users.

 > do you feel comfortable to let this PR just enable the specification and have future one to validate and give suggestions to the input of algorithm settings?

 
Yeah, sure, I don't want to block this PR, just to hear your thoughts.


cc @tenzen-y @johnugeorge @gaocegege

droctothorpe: > @droctothorpe Would you mind share some details here ? Is there something that we can adopt with Katib too ?

Happy to elaborate. Imagine something like this:

```python
from typing import Literal, Optional

class BayesianOptimizationSettings:
    def __init__(
        base_estimator: Literal["GP", "RF", "ET", "GBRT"] = "GP",
        n_initial_points: int = 10,
        acq_func: str = "gp_hedge",
        acq_optimizer: Literal["sampling", "lbfgs", "auto"] = "auto",
        random_state: Optional[int] = None
    ):
      self.base_estimator = base_estimator
      # etc.
```

This would include comprehensive docstrings, code completion, and input validation, so you wouldn't have to reference an external resource outside of the SDK. It could optionally leverage ipywidgets to do something like this:



You could then instantiate the config object, edit properties on the fly (not just at instantiation time) with the benefit of code completion, type hints, and docstrings, then pass the configuration object to the `tune` method.

Just a thought. FWIW, it's a bigger lift (and a much bigger operational burden) than this PR and should probably not be a condition for merging it.

andreyvelich: Thanks for sharing, actually, that is similar to what I proposed: https://github.com/kubeflow/katib/pull/2227#discussion_r1338805802, but using Suggestion instead of Algorithm.
Do we want to create separate issue to track this improvement @droctothorpe @shipengcheng1230 ?

andreyvelich, line 144: Should we use `List` similar as here: https://github.com/kubeflow/katib/blob/2798d300e2323b0cddcc64f2e55a3142df126585/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L146C34-L146C38 ?

shipengcheng1230: Fixed!

andreyvelich (approved):
Thanks for the update @shipengcheng1230!
/lgtm
/approve

tenzen-y:
@kubeflow/wg-automl-leads Could you restart failed CI jobs? It seems that some jobs failed due to issues regarding reachability to the container registry.

PR #2235, opened 2023-10-24
[SDK] Add `env` & `env_from` in client tune

shipengcheng1230:

The `env` and `env_from` cannot be directly set in the client `tune` method. Recently I found myself need to set up proxy environment variables for cross-region data access. Doing so inside the existing python objective function seems less desirable than having them set through these two specifications. 

This PR enables this capability. By default, these two are set to None which aligns with [V1Container](https://github.com/kubernetes-client/python/blob/36cfbe68a509d9b9d33395b22b6fa94d7d46c30f/kubernetes/client/models/v1_container.py#L89). Therefore, behavior is not modified when not setting them.

~Also, there was a missing of `cmaes` package when testing optuna suggestion. Therefore I added it in the optuna requirements.txt.~ (Rebased the master change)

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -25,7 +25,6 @@ from kubeflow.katib.api_client import ApiClient
 from kubeflow.katib.constants import constants
 from kubeflow.katib.utils import utils
 from kubernetes import client, config
-from kubernetes.client.models import V1EnvVar, V1EnvFromSource
 
 
 class KatibClient(object):
@@ -141,8 +140,7 @@ class KatibClient(object):
         parameters: Dict[str, Any],
         base_image: str = constants.BASE_IMAGE_TENSORFLOW,
         namespace: Optional[str] = None,
-        env: Union[Dict[str, str], List[V1EnvVar], None] = None,
-        env_from: Union[V1EnvFromSource, List[V1EnvFromSource], None] = None,
+        env_per_trial: Optional[Union[Dict[str, str], List[Union[client.V1EnvVar, client.V1EnvFromSource]]]] = None,
         algorithm_name: str = "random",
         algorithm_settings: Union[dict, List[models.V1beta1AlgorithmSetting], None] = None,
         objective_metric_name: str = None,
@@ -175,13 +173,12 @@ class KatibClient(object):
                 objective function.
             base_image: Image to use when executing the objective function.
             namespace: Namespace for the Experiment.
-            env: Environment variable(s) to be attached to each trial container. You can either specifiy
-                a list of kubernetes.client.models.V1EnvVar (documented here:
-                https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1EnvVar.md) or a dictionary  
-                corresponding to the environment variable name and value pair(s).
-            env_from: Source(s) of environment variables to be populated in each trial container. You can either specify
-                a kubernetes.client.models.V1EnvFromSource (documented here:
-                https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1EnvFromSource.md) or a list of such a type.
+            env_per_trial: Environment variable(s) to be attached to each trial container. 
+                You can specify a dictionary as a mapping object representing the environment variables. 
+                Otherwise, you can specify a list, in which the element can either be a kubernetes.client.models.V1EnvVar (documented here: 
+                https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1EnvVar.md)
+                or a kubernetes.client.models.V1EnvFromSource (documented here:
+                https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1EnvFromSource.md)
             algorithm_name: Search algorithm for the HyperParameter tuning.
             algorithm_settings: Settings for the search algorithm given.
                 For available fields, check this doc: https://www.kubeflow.org/docs/components/katib/experiment/#search-algorithms-in-detail.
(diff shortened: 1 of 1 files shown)


sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -172,6 +175,8 @@ def tune(
                 objective function.
             base_image: Image to use when executing the objective function.
             namespace: Namespace for the Experiment.
+            env: Environment variable(s) to be attached to each trial container.
+            env_from: Source(s) of environment variables to be populated in each trial container.
```
tenzen-y, line 179: Can we mention the types like `resources_per_trial`?

tenzen-y (commented):
This idea looks good.

/assign @andreyvelich

cmd/suggestion/optuna/v1beta1/requirements.txt
```diff
@@ -2,3 +2,4 @@ grpcio>=1.41.1
 protobuf>=3.19.5, <=3.20.3
 googleapis-common-protos==1.53.0
 optuna>=3.0.0
+cmaes>=0.10.0
```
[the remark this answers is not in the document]
shipengcheng1230: It seems the `cmaes` is listed as an optional depedency. And in the [GitHub CI](https://github.com/kubeflow/katib/actions/runs/6693916899/job/18186107558?pr=2235) it says
```
ModuleNotFoundError: No module named 'cmaes'
```

I think a change like from `optuna>=3.0.0` to `optuna[cmaes]>=3.0.0` is probably needed for the test to pass.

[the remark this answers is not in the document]
tenzen-y: Oh, I see. Thanks for the clarifying. Let's specify `cmaes>=0.10.0`.

tenzen-y (commented):
Otherwise lgtm.

cmd/suggestion/optuna/v1beta1/requirements.txt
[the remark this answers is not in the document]
andreyvelich: @shipengcheng1230 @tenzen-y Did optuna recently change their dependancies to make `cmaes` optional ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -140,6 +141,8 @@ def tune(
         parameters: Dict[str, Any],
         base_image: str = constants.BASE_IMAGE_TENSORFLOW,
         namespace: Optional[str] = None,
+        env: Union[Dict[str, str], List[V1EnvVar], None] = None,
+        env_from: Union[V1EnvFromSource, List[V1EnvFromSource], None] = None,
```
andreyvelich, line 145: Should we keep the name associated with Trials ? E.g. user could mislead this as environment for Algorithm Service.
E.g. name it as `trials_env`.
Also, I think we can allow to set `Dict[str, str]`,  `List[V1EnvVar]`, or `List[V1EnvFromSource]` in this parameter.

WDYT @shipengcheng1230 @tenzen-y @johnugeorge ?

cmd/suggestion/optuna/v1beta1/requirements.txt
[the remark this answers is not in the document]
shipengcheng1230: This commit moves it: https://github.com/optuna/optuna/commit/78159b516ebb694f5fa8a03072a2b54e9c53ef33

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
shipengcheng1230: I like `env_per_trial` which is more descriptive and consistent with `resources_per_trial`!

andreyvelich: Yeah, I think so @shipengcheng1230, we can accept multiple types for `env_per_trial` parameter.
What do you think about this name @tenzen-y @johnugeorge ?

shipengcheng1230: Here is the tentative implementation 12eea643d8ba49fd28e9be354941f80687b842e1

tenzen-y: `env_per_trial` looks great!

tenzen-y:
Can you rebase this PR since we fixed the optuna dependency issue.

shipengcheng1230:
Cool, done rebase and let me know what else is needed!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -140,6 +141,7 @@ def tune(
         parameters: Dict[str, Any],
         base_image: str = constants.BASE_IMAGE_TENSORFLOW,
         namespace: Optional[str] = None,
+        env_per_trial: Union[Dict[str, str], List[Union[V1EnvVar, V1EnvFromSource]], None] = None,
```
andreyvelich, line 144: Should we make this type Optional and import libs from Kubernetes client, similar to [other Kubernetes modules](https://github.com/kubeflow/katib/blob/bab4e4d63a4a61340a749b54c809e2d71af78a3a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L351) ?
```suggestion
        env_per_trial: Optional[Union[Dict[str, str], List[Union[client.V1EnvVar, client.V1EnvFromSource]]]] = None,
```

andreyvelich:
Sorry for the late reply @shipengcheng1230, I left one small comment.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
shipengcheng1230: Thanks for the great suggestion, I wasn't aware that those `client.models` are also importable from the `client` level!

tenzen-y (commented):
/approve

I'll leave lgtm on @andreyvelich.

andreyvelich:
Thank you for the update @shipengcheng1230!
/lgtm

PR #2240, opened 2023-11-01
Fix Optuna Validation for CMA-ES

andreyvelich:
Since Optuna recently moved `cmaes` to optional dependencies, our CI is failing: https://github.com/optuna/optuna/pull/4901.
I bumped Optuna to 3.3.0 version to unblock it, we can keep specific version of Optuna for now to avoid such errors.
If we need to relax the version, we can discuss the long-term plan on how to identify such errors in advance.

Also, I noticed that we have incorrect validation for `cmaes` in Optuna service.
We need to check that Search Space has more than 2 dimensional continues parameters instead of checking Algorithm Settings, similar to Goptuna: https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/goptuna/service.go#L188-L196

/assign @tenzen-y @johnugeorge @c-bata @shipengcheng1230

### DIFF
--- a/cmd/suggestion/optuna/v1beta1/requirements.txt
+++ b/cmd/suggestion/optuna/v1beta1/requirements.txt
@@ -1,4 +1,4 @@
 grpcio>=1.41.1
 protobuf>=3.19.5, <=3.20.3
 googleapis-common-protos==1.53.0
-optuna>=3.0.0
+optuna==3.3.0
--- a/pkg/suggestion/v1beta1/optuna/service.py
+++ b/pkg/suggestion/v1beta1/optuna/service.py
@@ -28,7 +28,6 @@ logger = logging.getLogger(__name__)
 
 
 class OptunaService(api_pb2_grpc.SuggestionServicer, HealthServicer):
-
     def __init__(self):
         super(OptunaService, self).__init__()
         self.lock = threading.Lock()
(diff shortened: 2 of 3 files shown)


tenzen-y (commented):
Thank you! Can we do cherry-pick this commit into the latest release branch?

/lgtm

andreyvelich:
> Thank you! Can we do cherry-pick this commit into the latest release branch?
> 
> /lgtm

Actually, I am not sure if we need to do another Release to fix this. `multivariate-tpe` is still working for Optuna service, and we use Goptuna for `cmaes` by default: https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-standalone/katib-config.yaml#L39
So, the issue will present only for users who want to use `cmaes` from Optuna suggestion.

WDYT @tenzen-y ?

tenzen-y:
> > Thank you! Can we do cherry-pick this commit into the latest release branch?
> > /lgtm
> 
> Actually, I am not sure if we need to do another Release to fix this. `multivariate-tpe` is still working for Optuna service, and we use Goptuna for `cmaes` by default: https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-standalone/katib-config.yaml#L39 So, the issue will present only for users who want to use `cmaes` from Optuna suggestion.
> 
> WDYT @tenzen-y ?

I don't think it is valuable to cut a new patch release v0.16.1. However, I think we should make Dockerfiles buildable on the latest release branch (release-0.16).

WDYT?

andreyvelich:
Great, tests were succeeded.

andreyvelich:
>I don't think it is valuable to cut a new patch release v0.16.1. However, I think we should make Dockerfiles buildable and the image should work fine on the latest release branch (release-0.16).

@tenzen-y But how user can find out about this fix if we are not going to update the Changelog and cut the new release ?

tenzen-y:
> > I don't think it is valuable to cut a new patch release v0.16.1. However, I think we should make Dockerfiles buildable and the image should work fine on the latest release branch (release-0.16).
> 
> @tenzen-y But how user can find out about this fix if we are not going to update the Changelog and cut the new release ?

@andreyvelich My suggestions aim to keep the latest branch runnable without the costs of a new patch release.
Even if without cherry-pick, v0.16.0 would work fine as you said.

So, I wanted to just say that we should keep the latest release branch proper regardless of whether we cut a new patch release.

Issue #2243, opened 2023-11-14
Support experiments with generate_name feature


fensterwetter:
/kind feature

**Describe the solution you'd like**
The Python katib client (again) supports experiments with the `generate_name` feature.

**Anything else you would like to add:**
After upgrading from kubeflow-katib 0.14.0 to 0.16.0, I noticed that the `create_experiment` function in `katib_client.py` now implicitly makes the assumption that the `experiment` object contains an attribute called `.metadata.name`. However, generally `name` can be missing in favour of `generate_name` which is then treated as a prefix which gets an additional suffix from the server. In version 0.14.0, the server return was used to access the resulting final name. At least from kubeflow-katib version 0.15.0 onwards, the server return is no longer captured and the original experiment object is used to access the `name` attribute.

I was wondering if the change was a conscious decision or rather a side effect that could be reverted. I am also happy to collaborate in case this is something that makes sense for katib to support.

Thanks
Thomas

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

tenzen-y:
That makes sense. We should get the generated name if the name is empty here:

https://github.com/kubeflow/katib/blob/fc6beec8358f1c4a50d86de63c5906fce2eba373/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L104-L114

cc: @andreyvelich

bharathk005:
@tenzen-y I can work on this if this feature is still needed.

tenzen-y:
> @tenzen-y I can work on this if this feature is still needed.

Yes, I believe that this is still needed. Feel free to open the PR.

bharathk005:
@tenzen-y 
Do we need to handle this scenario in the backend as well?

https://github.com/kubeflow/katib/blob/fc6beec8358f1c4a50d86de63c5906fce2eba373/pkg/ui/v1beta1/backend.go#L130

tenzen-y:
> katib/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py

@bharathk005 That backend is only for Katib UI. So, we can improve only SDK.

PR #2272, opened 2024-03-04
Python SDK - Generate Name functionality for creating experiments.

bharathk005:
Adds generate name functionality to Python SDK while creating experiments.
If metadata.name and metadata.generateName are missing, the code appends a new prefix called "exp-" to generateName.
This makes the server generate a new name with the prefix.

Fixes #
#2243


bharathk005:
Please provide feedback on the approach. 
I will make changes and do more testing based on this.

tenzen-y:
@bharathk005 Could you sign to the DCO?

andreyvelich (commented):
Thank you for updating the SDK @bharathk005!
I think, the point from @fensterwetter was that our SDK should not block to use `generateName` parameter while using `create_experiment` API.
So [here](https://github.com/kubeflow/katib/blob/db68f0d1f8c8f6e92f705db75d0d4237c7d8ff06/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L111-L124) when we rise Exception or print that Experiment has been created, SDK will fail.
I think, we can that `name` or `generate_name` is exist before running `create_namespaced_custom_object` and then use this later in the code.

bharathk005:
@tenzen-y  and @andreyvelich  thanks for the suggestions. 
Please review the change and let me know.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -93,29 +93,35 @@ def create_experiment(
 
         namespace = namespace or self.namespace
 
+        if 'name' in experiment.metadata and experiment.metadata.name:
```
andreyvelich, line 96: I think, it will raise Exception if `experiment.metadata.name` is not set.
We might need to have something like this:
```python
         experiment_name = None

        if type(experiment) == models.V1beta1Experiment:
            if experiment.metadata.name is not None:
                experiment_name = experiment.metadata.name
            elif experiment.metadata.generate_name is not None:
                experiment_name = experiment.metadata.generate_name
        elif "name" in experiment["metadata"]:
            experiment_name = experiment["metadata"]["name"]
        elif "generateName" in experiment["metadata"]:
            experiment_name = experiment["metadata"]["generateName"]
        if experiment_name is None:
            raise ValueError("Experiment must have name or generateName")
```
Any thoughts @droctothorpe @tenzen-y @johnugeorge @bharathk005 @fensterwetter

bharathk005: @andreyvelich  thanks for the review.
I have added some changes based on your suggestions. Test cases pass on my machine.

Regarding the type assertion, the function already expects a "models.V1beta1Experiment" type:

https://github.com/kubeflow/katib/blob/1039f4b1bd37ee7b2e0e5e94394d25dfdbd3dd05/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L80

Please let me know if we still need to check for the type.

tenzen-y: > I think, it will raise Exception if `experiment.metadata.name` is not set. We might need to have something like this:
> 
> ```python
>          experiment_name = None
> 
>         if type(experiment) == models.V1beta1Experiment:
>             if experiment.metadata.name is not None:
>                 experiment_name = experiment.metadata.name
>             elif experiment.metadata.generate_name is not None:
>                 experiment_name = experiment.metadata.generate_name
>         elif "name" in experiment["metadata"]:
>             experiment_name = experiment["metadata"]["name"]
>         elif "generateName" in experiment["metadata"]:
>             experiment_name = experiment["metadata"]["generateName"]
>         if experiment_name is None:
>             raise ValueError("Experiment must have name or generateName")
> ```
> 
> Any thoughts @droctothorpe @tenzen-y @johnugeorge @bharathk005 @fensterwetter

Additionally, I would suggest to add the Exception if both `name` and `generateName` are specified.
@andreyvelich WDYT?

andreyvelich: Won't Kubernetes Python client throw an error when we try to create custom resources in that case ?

andreyvelich: >Regarding the type assertion, the function already expects a "models.V1beta1Experiment" type:

That's true, but since Python is not strong typed language, users can still send `dict` as Experiment. Otherwise, we need to check `type(experiment)` and through exception. Do we want to give users functionality to pass dict with `create_experiment` API @droctothorpe @tenzen-y @johnugeorge ?

tenzen-y: > Won't Kubernetes Python client throw an error when we try to create custom resources in that case ?

I imaged such like 102-103:

https://github.com/kubeflow/katib/blob/5be7150ac760c800bdab156a01dff290ec67d3bd/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L102-L103

So, my concerns have gone away. Thanks.

tenzen-y: > That's true, but since Python is not strong typed language, users can still send dict as Experiment. Otherwise, we need to check type(experiment) and through exception. Do we want to give users functionality to pass dict with create_experiment API @droctothorpe @tenzen-y @johnugeorge ?

I agree with the necessary for type assertion, but I think this is PR to support name generation.
So, I'm ok with working on it in another PR.

bharathk005: @tenzen-y @andreyvelich thanks for the comments.
1. Type assertion
2. Throwing error if both name and generateName are provided
Are these the changes that need to be implemented?

bharathk005: @tenzen-y @andreyvelich Thanks for your time. I have pushed a commit with the changes mentioned in this thread. Please review and let me know.

andreyvelich:
Thank you for the updates @bharathk005!
/lgtm

andreyvelich:
> @andreyvelich is this part of 0.17?

Sure, sorry that we haven't merged this PR yet.
Please can you try to rebase it ? The tests might be fixed.

bharathk005:
> > @andreyvelich is this part of 0.17?
> 
> Sure, sorry that we haven't merged this PR yet. @bharathk005 Please can you try to rebase it ? The tests might be fixed.

@andreyvelich Done. Please validate.

andreyvelich (approved):
Thanks for the update @bharathk005!
/lgtm
/approve

PR #2278, opened 2024-03-10
Upgrade Python version to 3.11

tenzen-y:
I upgraded the Python version to 3.11 since Python 3.10 no longer fixes bugs, according to [this](https://devguide.python.org/versions/#supported-versions).

Also, I upgraded some libraries to switch Python 3.11 supported version:

- numpy 1.25.2: https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

Additionally, we keep the library and Python version for the skopt since some dependencies error occurs due to stopped maintenance: #2280.

Fixes #

**Checklist:**


tenzen-y:
/hold

We need to upgrade other library versions first.

manifests/v1beta1/installs/katib-standalone/katib-config.yaml
```diff
@@ -45,7 +45,7 @@ runtime:
       image: docker.io/kubeflowkatib/suggestion-enas:latest
       resources:
         limits:
-          memory: 200Mi
+          memory: 400Mi
```
tenzen-y, line 48: After I upgraded the Python version, I faced OOM killed in the ENAS suggestion service.

tenzen-y:
Ready for the review. PTAL.

andreyvelich (commented):
Thank you @tenzen-y!
A few questions:
1. Do we want to add support for Python 3.11 in Katib SDK: https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/setup.py#L61 ?
2. Do we want to add this change before the next Katib release ?

tenzen-y:
> Thank you @tenzen-y! A few questions:
> 
> 1. Do we want to add support for Python 3.11 in Katib SDK: https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/setup.py#L61 ?
> 2. Do we want to add this change before the next Katib release ?

@andreyvelich It's a good question. 

1. I agree with you. If so, should we remove Python 3.7 from the list as Python 3.7 is no longer supported? 
2. Adding support in this PR would be better since we can verify if our SDK with Python 3.11 can work for a while by adding CI: https://github.com/kubeflow/katib/blob/679e6fb8f8768b4d23d640ea56103a20fdb027fe/.github/workflows/e2e-test-darts-cifar10.yaml#L24

andreyvelich:
>I agree with you. If so, should we remove Python 3.7 from the list as Python 3.7 is no longer supported?

How strict we want to be with our SDK versions since it is user facing interface and sometimes users might be on the out-of-date environments ?
For example, Kubernetes Python client still supports Python 3.6 and 3.7: https://github.com/kubernetes-client/python/blob/master/setup.py#L80.

However, I prefer approach which [Jupyter Lab](https://github.com/jupyterlab/jupyterlab/blob/main/pyproject.toml#L13) and [Jupyter Server](https://github.com/jupyter-server/jupyter_server/blob/main/pyproject.toml#L24) follow where they add minimal Python version that is supported by today.

Could you give your feedback on this @kubeflow/wg-training-leads @droctothorpe @shipengcheng1230 @andreafehrman @harrisonfritz

tenzen-y:
> How strict we want to be with our SDK versions since it is user facing interface and sometimes users might be on the out-of-date environments ? For example, Kubernetes Python client still supports Python 3.6 and 3.7: https://github.com/kubernetes-client/python/blob/master/setup.py#L80.
> 
> However, I prefer approach which [Jupyter Lab](https://github.com/jupyterlab/jupyterlab/blob/main/pyproject.toml#L13) and [Jupyter Server](https://github.com/jupyter-server/jupyter_server/blob/main/pyproject.toml#L24) follow where they add minimal Python version that is supported by today.

I'm ok with either way, but I believe that we should define deprecation or support policies for SDK with Python version.
For example, we decided to support 3 Kubernets minor versions.

So, should we keep supported Python versions in this PR, and then open a dedicated issue and discuss the Python version?
This means that I should not add Python 3.11 to SDK and should not remove Python 3.7 from SDK.

@andreyvelich WDYt?

andreyvelich:
>So, should we keep supported Python versions in this PR, and then open a dedicated issue and discuss the Python version?
This means that I should not add Python 3.11 to SDK and should not remove Python 3.7 from SDK.

That's fine I believe. Let's create dedicated issue to discuss long-term support for SDK Python versions.

tenzen-y:
> > So, should we keep supported Python versions in this PR, and then open a dedicated issue and discuss the Python version?
> > This means that I should not add Python 3.11 to SDK and should not remove Python 3.7 from SDK.
> 
> That's fine I believe. Let's create dedicated issue to discuss long-term support for SDK Python versions.

I created: #2283

tenzen-y:
@andreyvelich So, can we merge this?

andreyvelich:
Yeah, let's merge it!
/lgtm

PR #2290, opened 2024-03-19
Updated dockerfiles for grpc installation for ppc64le

aditijadhav38:
This PR targets to get the katib/suggestions built on PowerPC.
This PR is continuation of PR https://github.com/kubeflow/katib/pull/2262


cmd/earlystopping/medianstop/v1beta1/Dockerfile
```diff
@@ -7,7 +7,10 @@ ENV PYTHONPATH ${TARGET_DIR}:${TARGET_DIR}/pkg/apis/manager/v1beta1/python
 
 RUN if [ "${TARGETARCH}" = "ppc64le" ] || [ "${TARGETARCH}" = "arm64" ]; then \
     apt-get -y update && \
-    apt-get -y install gfortran libopenblas-dev liblapack-dev && \
+    apt-get -y install gfortran libopenblas-dev liblapack-dev libssl-dev pkg-config gcc g++ && \
```
mkumatag, line 10: we shouldn't be installing packages like `libssl-dev pkg-config gcc g++` for arm64 if not required

cmd/earlystopping/medianstop/v1beta1/Dockerfile
```diff
@@ -7,7 +7,10 @@ ENV PYTHONPATH ${TARGET_DIR}:${TARGET_DIR}/pkg/apis/manager/v1beta1/python
...
+    if [ "${TARGETARCH}" = "ppc64le" ]; then \
+      GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=1 pip3 install grpcio>=1.41.1; \
```
mkumatag, line 12: why are we calling pip install here with specific version here? anyways we are calling the pip install below which does install the relevant version. Just that we need this `GRPC_PYTHON_BUILD_SYSTEM_OPENSSL` set to 1 only for the ppc64le platform.

mkumatag (changes requested):

cmd/earlystopping/medianstop/v1beta1/Dockerfile
aditijadhav38: @mkumatag Its the version specified in requirements.txt. Hence I have added same version here. But it will install relevant version only.

aditijadhav38: @mkumatag Updated dockerfile and have set flat `GRPC_PYTHON_BUILD_SYSTEM_OPENSSL` to 1 when arch is `ppc64le`

aditijadhav38:
@tenzen-y Could you please review the changes..

tenzen-y:
> updated dockerfiles for grpc builds for powerPC compilation~ #2262

I'm still wondering if we should not support ppc64le based on this discussion: https://github.com/kubeflow/katib/pull/2262#pullrequestreview-1934843268

Anyway, we should resolve that discussion before we move this forward.

aditijadhav38:
> > updated dockerfiles for grpc builds for powerPC compilation~ #2262
> 
> I'm still wondering if we should not support ppc64le based on this discussion: [#2262 (review)](https://github.com/kubeflow/katib/pull/2262#pullrequestreview-1934843268)
> 
> Anyway, we should resolve that discussion before we move this forward.

For this, a comment has been already added https://github.com/kubeflow/katib/pull/2262#issuecomment-1998212429. Please check. Could you please add comment in that discussion if you have any more queries.

lehrig:
@tenzen-y please refer to https://github.com/kubeflow/katib/pull/2262#issuecomment-1998212429 - does that help? As mentioned, we are happy to help maintaining, especially if anything architecture-specific comes up...

PR #2296, opened 2024-04-01
CI: Enable parallel mode for the coveralls

tenzen-y:
We sometimes face the following error when we retry CI. So, I enabled the parallel mode to avoid the error.

```shell
bad response status from coveralls: 422
{"message":"Can't add a job to a build that is already closed. Build 8475719809 is closed. See docs.coveralls.io/parallel-builds","error":true}
Error: Error: The process '/home/runner/work/_actions/shogo82148/actions-goveralls/v1/bin/goveralls_linux_amd64' failed with exit code 1
```

https://github.com/kubeflow/katib/actions/runs/8475719809/job/23237738755?pr=2285#step:5:29

Docs: https://github.com/shogo82148/actions-goveralls?tab=readme-ov-file#parallel-job-example

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/test-go.yaml
+++ b/.github/workflows/test-go.yaml
@@ -59,9 +59,19 @@ jobs:
         with:
           path-to-profile: coverage.out
           working-directory: ${{ env.GOPATH }}/src/github.com/kubeflow/katib
+          parallel: true
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


tenzen-y:
I will create this PR again since cla/google seems not to return any responses.

Issue #2141, opened 2023-04-03
Replace the tool of code generator for the gRPC API


tenzen-y:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
Currently, we are generating codes for the gRPC API from the protocol buffers definition with [`znly/protoc`](https://hub.docker.com/r/znly/protoc/).

However, the image doesn't seem to be maintained since the latest image was published 4 years ago.

So, we should replace the toll with another one.

follow-up: #2140 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

tenzen-y:
cc: @andreyvelich @johnugeorge

tenzen-y:
Maybe, we can migrate the tool to the [`buf`](https://buf.build/).

johnugeorge:
Agree. We should try it out

shivas1516:
From the description provided in this issue, I understand that the existing code generation tool 'znly/protoc' is not maintained properly, and there's a need to migrate to another tool, which has been mentioned in this conversation as Buf. As this issue is marked as a good first issue and aligns with my skill level, I'm keen to contribute to resolving it. 

I'm seeking guidance and recommendations to get started on this task. Your assistance and response are invaluable to me. Looking forward to your input!

tenzen-y:
> From the description provided in this issue, I understand that the existing code generation tool 'znly/protoc' is not maintained properly, and there's a need to migrate to another tool, which has been mentioned in this conversation as Buf. As this issue is marked as a good first issue and aligns with my skill level, I'm keen to contribute to resolving it. 
> 
> 
> 
> I'm seeking any guidance or recommendations to get started on this task. Your assistance and response are invaluable to me. Looking forward to your input!

@shivas1516 Your understanding is correct.
Feel free to open a PR. We’d be happy to review it.

PR #2317, opened 2024-04-29
Disable gRPC code generation in release script

andreyvelich:
I disabled gRPC code generation for our release script since `znly/protoc` image doesn't work on ARM right now.
We don't have any changes in gRPC APIs for Katib 0.17 release, so it should be ok.

We should prioritise migration to Buf in this issue: https://github.com/kubeflow/katib/issues/2141

### DIFF
--- a/pkg/apis/manager/health/build.sh
+++ b/pkg/apis/manager/health/build.sh
@@ -19,6 +19,9 @@ set -e
 
 cd "$(dirname "$0")"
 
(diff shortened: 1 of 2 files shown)


andreyvelich:
We sync offline with @tenzen-y, we can make our release by disable Rosetta emulation in Docker Desktop.
That will use qemu to run gRPC generation images.

PR #2328, opened 2024-05-15
[GSoC] KEP for Project 6: Push-based Metrics Collection for Katib

Electronic-Waste:

This PR converts the [GSoC Proposal](https://docs.google.com/document/d/1_C-0jVV5ZMgbl86YoJ19WLJlZACI-L_UD-9Nevq9Lo0/edit?usp=sharing) into KEP, which describes the details of implementing ["Project 6: Push-based Metrics Collection for Katib"](https://www.kubeflow.org/events/gsoc-2024/#project-6-push-based-metrics-collection-for-katib). There are some related issues discussing about this : 

- https://github.com/kubeflow/katib/issues/577
- https://github.com/kubeflow/katib/issues/2340

Fixes #

**Checklist:**


docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,77 @@
...
+
+As is mentioned above, we decided to add `metrics_collection_mechanism` to the tune function in Python SDK. Also, we have some changes to be made:
+
+1. Disable injection: set `katib.kubeflow.org/metrics-collector-injection` to `disabled` when the push-based way of metrics collection is adopted so as to disable the injection of the metrics collection sidecar container.
+
+2. Configure the way of metrics collection: set the configuration `spec.metricsCollectionSpec.collector.kind`(specify the way of metrics collection) to `NoneCollector`.
+
+### Code Injection in Webhook
+
+We decided to implement a code replacing function in Experiment Mutating Webhook. When `spec.metricsCollectionSpec.collector.kind` is set to `NoneCollector`, the code replacing function will recognize the metrics output lines (e.g. print, log.Info, e.t.c.) and replace them with push-based metrics collection code which will be discussed in the next section. It’s a better decision compared with offering users a `katib_client.push`-like interface, for that users can’t use a yaml file to define this operation.
```
andreyvelich, line 67: I am not sure if we should replace user code here. I would suggest, we should implement it using `katib_client.report_metrics("lr": 0.03)` or `katib_client.log_metrics("lr": 0.01")`. That will allow user to explicitly report metrics during training code to the Katib DB.
Also, we should always pass `KATIB_TRIAL_NAME` env variable to the Katib Trials, so training function knows where to report metrics to the Katib DB.

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Push-based Metrics Collection Code
+
+The push-based metrics collection code is a function making a grpc call to the persistent API to store training metrics. It will be injected to container args in the Experiment Mutating Webhook and then be called inside the Trial Worker Pod to push metrics to Katib DB.
```
andreyvelich, line 71: It would be nice to explain the gRPC call that we need to make to report metrics to Katib DB.

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,77 @@
...
+
+In the procedure of tuning hyperparameters, Metrics Collector, which is implemented as a sidecar container attached to each training container in the [current design](https://github.com/kubeflow/katib/blob/master/docs/proposals/metrics-collector.md), will collect training logs from Trials once the training is complete. Then, the Metrics Collector will parse training logs to get appropriate metrics like accuracy or loss and pass the evaluation results to the HyperParameter tuning algorithm.
+
+However, current implementation of Metrics Collector is pull-based, raising some [design problems](https://github.com/kubeflow/training-operator/issues/722#issuecomment-405669269) such as determining the frequency we scrape the metrics, performance issues like the overhead caused by too many sidecar containers, and restrictions on developing environments which must support sidecar containers. Thus, we should implement a new API for Katib Python SDK to offer users a push-based way to store metrics directly into the Kaitb DB and resolve those issues raised by pull-based metrics collection.
+
+![](../images/push-based-metrics-collection.png)
+Fig.1 Architecture of the new design
+
+## Goal
```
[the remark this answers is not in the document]
Electronic-Waste: Okay, I'll add non-goals to this KEP.

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,77 @@
...
+1. **A new parameter in Python SDK function `tune`**: allow users to specify the method of collecting metrics(push-based/pull-based).
+2. **A code injection function in mutating webhook**: recognize the metrics output lines and replace them with push-based metrics collection code.
+3. The final metrics of worker pods should be **pushed to Katib DB directly** in the push mode of metrics collection.
+
+## API
+
+### New Parameter in Python SDK Function `tune`
+
+We decided to add `metrics_collection_mechanism` to `tune` function in Python SDK.
```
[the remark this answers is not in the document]
Electronic-Waste: I agree with you. It can be extensible if we use dict as parameter type.

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,77 @@
...
+    max_failed_trial_count: int = None,
+    resources_per_trial: Union[dict, client.V1ResourceRequirements, None] = None,
+    retain_trials: bool = False,
+    packages_to_install: List[str] = None,
+    pip_index_url: str = "https://pypi.org/simple",
+    metrics_collection_mechanism: str = "pull", # The newly added parameter
+)
+```
+
+## Implementation
```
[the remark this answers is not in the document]
Electronic-Waste: Do you mean that Trial controller should have a response like StatusCode in http from Katib DB and process it in the further steps?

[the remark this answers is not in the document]
andreyvelich: I think, we should just verify if Trial controller will update Trial observation correctly in case of [`None` Metrics Collector.](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_util.go#L125-L133)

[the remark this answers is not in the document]
Electronic-Waste: Okay, I get your idea. Do you mean that Trial controller will report an error if we do not change it to be compatible with push-based metrics collection?

docs/proposals/push-based-metrics-collection.md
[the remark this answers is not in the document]
andreyvelich: I think, we should also discuss if we should rename Metrics Collector from `None` to `Push` or any other name.
Since it is not correct to call this metrics collector: `None`. cc @gaocegege

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: That's what I want to discuss with you in the upcoming weekly meeting.

docs/proposals/push-based-metrics-collection.md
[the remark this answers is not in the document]
andreyvelich: Currently, metrics collector inserts `unavailable` value for each metrics into DB if that metrics can't be found: https://github.com/kubeflow/katib/blob/master/pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go#L181-L193.
And since [`isObservationAvailable` returns false](https://github.com/kubeflow/katib/blob/master/pkg/apis/controller/trials/v1beta1/util.go#L94), we marked Trial with [Metrics Unavailable status](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_util.go#L69). We need to see how will it work for push-based Metrics collector.

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: We can implement it with `katib_client.report_metrics` and call this function when we create Trial with Python SDK. However, users sometimes apply a yaml file instead of using Python API. If we want to make the push-based metrics collection compatible with this scenario, we should make some changes to the user code, for example, importing katib library and calling `katib_client.report_metrics`. Is it a better choice? WDYT👀@andreyvelich

docs/proposals/push-based-metrics-collection.md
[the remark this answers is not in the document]
Electronic-Waste: What about reserving `None` and adding a new type called `NoSidecar`? cc👀 @andreyvelich @gaocegege .

Electronic-Waste:
ref issue: #2340

docs/proposals/push-based-metrics-collection.md
andreyvelich: >However, users sometimes apply a yaml file instead of using Python API. If we want to make the push-based metrics collection compatible with this scenario, we should make some changes to the user code

That's right, users can build docker image and submit YAML.
In that case, they need to install `kubeflow-katib` Python SDK and use `katib_client.report_metrics()` API to report metrics from their training code.
I don't think we should do any changes to the user code, tbh.
What do you think @Electronic-Waste @tenzen-y @johnugeorge ?

Electronic-Waste: Yeah... Maybe my expressions are ambiguous. I mean users may need to change their code(importing kubeflow-katib and using `katib_client.report_metrics()`). It seems that the change they'll make is not bothering and difficult. I agree with you and I'll modify the KEP soon. @andreyvelich

docs/proposals/push-based-metrics-collection.md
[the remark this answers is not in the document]
gaocegege: I agree. As for the name, Push may be better for me.

docs/proposals/push-based-metrics-collection.md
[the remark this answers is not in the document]
Electronic-Waste: Metrics collectors won't be injected into each trial pod when the push-based metrics collection is applied. Instead, training metrics will be pushed to Katib DB directly as users call `katib_client.report_metrics` for metrics collection. So, this issue will not happen in push-based metrics collection scenario. WDYT👀 @andreyvelich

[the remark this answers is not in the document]
andreyvelich: Yeah, I think that is right. Initially, I implemented this logic for pull-based metrics in this PR: https://github.com/kubeflow/katib/pull/1333
Since sometime Trials can be in succeeded state even when metrics collector sidecar container is still running.
Because some CRDs (e.g. Tekton Pipeline) don't wait for all sidecar to be complete.

For the push-based metrics it is not the case, since `katib_client.report_metrics()` call is part of user's training container execution.

We still need to make appropriate changes in the controller to make sure we insert `unavailable` value into Katib DB, if user doesn't report metric accidentally. What do you think @johnugeorge @tenzen-y @Electronic-Waste ?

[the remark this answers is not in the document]
Electronic-Waste: I agree with you. I'll investigate it.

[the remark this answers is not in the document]
Electronic-Waste: I noticed that the re-queue logic is implemented in the following lines:

```
// If observation is empty metrics collector doesn't finish.
// For early stopping metrics collector are reported logs before Trial status is changed to EarlyStopped.
if jobStatus.Condition == trialutil.JobSucceeded && instance.Status.Observation == nil {
	logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
	return errMetricsNotReported
}
```

I decide to add a if-else statement to distinguish pull-based and push-based metrics collection. In the push-based collection, the trial does not need to be requeued. Instead, we'll insert a `unavailable` value to Katib DB and change the status of trial to `MetricsUnavailable`. WDYT👀 @andreyvelich @tenzen-y @johnugeorge @gaocegege

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,195 @@
...
+    If you run this API in-cluster (e.g. from the Kubeflow Notebook) you can
+    use the default Katib DB Manager address: `katib-db-manager.kubeflow:6789`.
+
+    If you run this API outside the cluster, you have to port-forward the
+    Katib DB Manager before getting the Trial metrics: `kubectl port-forward svc/katib-db-manager -n kubeflow 6789`.
+    In that case, you can use this Katib DB Manager address: `localhost:6789`.
+
+    You can use `curl` to verify that Katib DB Manager is reachable: `curl <db-manager-address>`.
+
+    [!!!] Trial's namespace and name should always be passed into Katib Trials as env variable `KATIB_TRIAL_NAMESPACE` and `KATIB_TRIAL_NAME`.
```
andreyvelich, line 82: I think, we don't need to pass `KATIB_TRIAL_NAMESPACE` to the Trial since can always get the container namespace from `/var/run/secrets/kubernetes.io/serviceaccount/namespace`.
WDYT @Electronic-Waste @tenzen-y @johnugeorge ?

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,195 @@
...
+	return errMetricsNotReported
+}
+```
+1. Distinguish pull-based and push-based metrics collection
+
+We decide to add a if-else statement in the code above to distinguish pull-based and push-based metrics collection. In the push-based collection, the trial does not need to be requeued. Instead, we'll insert a unavailable value to Katib DB.
+
+2. Update the status of trial to `MetricsUnavailable`
+
+In the current implementation of pull-based metrics collection, trials will be re-queued when the metrics collector finds the `.Status.Observation` is empty. However, it's not compatible with push-based metrics collection because the forgotten metrics won't be reported in the new round of reconcile. So, we need to update its status in the function `UpdateTrialStatusCondition` in accomodation with the pull-based metrics collection.
```
andreyvelich, line 183: Please can you link the code location where we need to change `UpdateTrialStatusCondition`.
E.g. https://github.com/kubeflow/katib/blob/d69d04e77e00e84bfc3950e8efe0aaa297f9493c/pkg/controller.v1beta1/trial/trial_controller_util.go#L42

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: I agree with you Andrey.

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: Okay, I see.

Electronic-Waste: @andreyvelich The proposal has been modified. The code location:

https://github.com/kubeflow/katib/blob/d69d04e77e00e84bfc3950e8efe0aaa297f9493c/pkg/controller.v1beta1/trial/trial_controller_util.go#L69

We'll insert the `else if` clause just before this line.

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,196 @@
...
+    parallel_trial_count: int = None,
+    max_failed_trial_count: int = None,
+    resources_per_trial: Union[dict, client.V1ResourceRequirements, None] = None,
+    retain_trials: bool = False,
+    packages_to_install: List[str] = None,
+    pip_index_url: str = "https://pypi.org/simple",
+    # The newly added parameter metrics_collector_config.
+    # It specifies the config of metrics collector, for example, 
+    # metrics_collector_config={"kind": "Push"},
+    metrics_collector_config: Dict[str, Any] = None,
```
andreyvelich, line 63: Let's by default use the StdOut or Push metrics collector, so user doesn't need to configure it.

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,196 @@
...
+    Katib DB Manager service should be accessible while calling this API.
+
+    If you run this API in-cluster (e.g. from the Kubeflow Notebook) you can
+    use the default Katib DB Manager address: `katib-db-manager.kubeflow:6789`.
+
+    If you run this API outside the cluster, you have to port-forward the
+    Katib DB Manager before getting the Trial metrics: `kubectl port-forward svc/katib-db-manager -n kubeflow 6789`.
+    In that case, you can use this Katib DB Manager address: `localhost:6789`.
+
+    You can use `curl` to verify that Katib DB Manager is reachable: `curl <db-manager-address>`.
```
andreyvelich, line 80: I guess, this is not required since we always execute this API from the Trial, right ?

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Step 4. Get the best HyperParameters.
+print(katib_client.get_optimal_hyperparameters(name))
+```
+
+## Implementation
+
+### Add New Parameter in `tune`
+
+As is mentioned above, we decided to add `metrics_collector_config` to the tune function in Python SDK. Also, we have some changes to be made:
```
andreyvelich, line 139:
```suggestion
As mentioned above, we decided to add `metrics_collector_config` to the tune function in Python SDK. Also, we have some changes to be made:
```

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,196 @@
...
+
+1. Disable injection: set `katib.kubeflow.org/metrics-collector-injection` to `disabled` when the push-based way of metrics collection is adopted so as to disable the injection of the metrics collection sidecar container.
```
andreyvelich, line 141: Do we really need it and where we are planning to disable it ?
Should we just verify if metricsCollector type is Push here: https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/pod/inject_webhook.go#L102 ?

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,196 @@
...
+
+2. Configure the way of metrics collection: set the configuration `spec.metricsCollectionSpec.collector.kind`(specify the way of metrics collection) to `Push`.
+
+3. Rename metrics collector from `None` to `Push`: It's not correct to call push-based metrics collection `None`. We should modify related code to rename it.
+
+4. Write env variables into trial spec: set `KATIB_TRIAL_NAME` for `report_metrics` function to dial db manager.
```
andreyvelich, line 147:
```suggestion
4. Write env variables into Trial spec: set `KATIB_TRIAL_NAME` for `report_metrics` function to dial db manager.
```

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,196 @@
...
+// If observation is empty metrics collector doesn't finish.
+// For early stopping metrics collector are reported logs before Trial status is changed to EarlyStopped.
+if jobStatus.Condition == trialutil.JobSucceeded && instance.Status.Observation == nil {
+	logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
+	return errMetricsNotReported
+}
+```
+1. Distinguish pull-based and push-based metrics collection
+
+We decide to add a if-else statement in the code above to distinguish pull-based and push-based metrics collection. In the push-based collection, the trial does not need to be requeued. Instead, we'll insert a unavailable value to Katib DB.
```
andreyvelich, line 179:
```suggestion
We decide to add a if-else statement in the code above to distinguish pull-based and push-based metrics collection. In the push-based collection, the Trial does not need to be requeued. Instead, we'll insert a unavailable value to Katib DB.
```

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,196 @@
...
+
+2. Update the status of trial to `MetricsUnavailable`
+
+In the current implementation of pull-based metrics collection, trials will be re-queued when the metrics collector finds the `.Status.Observation` is empty. However, it's not compatible with push-based metrics collection because the forgotten metrics won't be reported in the new round of reconcile. So, we need to update its status in the function `UpdateTrialStatusCondition` in accomodation with the pull-based metrics collection. The following code will be insert into lines before [trial_controller_util.go#L69](https://github.com/kubeflow/katib/blob/7959ffd54851216dbffba791e1da13c8485d1085/pkg/controller.v1beta1/trial/trial_controller_util.go#L69)
```
andreyvelich, line 183:
```suggestion
In the current implementation of pull-based metrics collection, Trials will be re-queued when the metrics collector finds the `.Status.Observation` is empty. However, it's not compatible with push-based metrics collection because the forgotten metrics won't be reported in the new round of reconcile. So, we need to update its status in the function `UpdateTrialStatusCondition` in accommodation with the pull-based metrics collection. The following code will be insert into lines before [trial_controller_util.go#L69](https://github.com/kubeflow/katib/blob/7959ffd54851216dbffba791e1da13c8485d1085/pkg/controller.v1beta1/trial/trial_controller_util.go#L69)
```

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,196 @@
...
+
+
+```Golang
+else if instance.Spec.MetricCollector.Collector.Kind == "Push" && instance.Status.Obeservation == nil {
```
andreyvelich, line 187: I guess, we can skip this check: `instance.Status.Obeservation == nil` since we this condition will be met when Observation is not available given this: https://github.com/kubeflow/katib/blob/7959ffd54851216dbffba791e1da13c8485d1085/pkg/controller.v1beta1/trial/trial_controller_util.go#L48

andreyvelich (commented):
Thank you @Electronic-Waste!
I think, we almost ready with this proposal, just a few small comments.
Please take a look @tenzen-y @johnugeorge

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: I agree with you. It seems that the `instance.Status.Observation == nil ` is redundant.

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: I'll correct this spelling mistake.

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: I'll correct this spelling mistake.

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: I'll correct this spelling mistake.

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: Yes, I agree with you. I'll delete these comments.

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: I'll correct this grammar mistake.

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: Sorry for that, I think we do not really need it because there already exists some logic to handle this.

https://github.com/kubeflow/katib/blob/e6bd3e7b5bd95e0e77ce2f869e6b1eeb72d24180/pkg/webhook/v1beta1/pod/inject_webhook.go#L151

But we should not verify metricsCollector type in:

https://github.com/kubeflow/katib/blob/e6bd3e7b5bd95e0e77ce2f869e6b1eeb72d24180/pkg/webhook/v1beta1/pod/inject_webhook.go#L102

Because we need to add Katib Trial labels to the Pod metadata:

https://github.com/kubeflow/katib/blob/e6bd3e7b5bd95e0e77ce2f869e6b1eeb72d24180/pkg/webhook/v1beta1/pod/inject_webhook.go#L141

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: Okay, I'll use `StdOut` as the default metrics collector since `Push` metrics collector is under development.

docs/proposals/push-based-metrics-collection.md
andreyvelich: That's right, we should mutate `KATIB_TRIAL_NAME` to the training container env.

Electronic-Waste:
Sincere thanks for your time @andreyvelich . I modified the proposal according to your review.

Could you please review the updated version of the proposal? And We can discuss about it in the meeting today. If there are no extra reviews to make,  I can convert the proposal to PRs in this week.

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,184 @@
...
+else if instance.Spec.MetricCollector.Collector.Kind == "Push" {
+    ... // Update the status of this Trial to `MetricsUnavailable` and output the reason.
+}
+```
+
+### Collection of Final Metrics
+
+The final metrics of worker pods should be pushed to Katib DB directly in the push mode of metrics collection.
+
+\#WIP
```
andreyvelich, line 184: We can remove it.
```suggestion
```

andreyvelich (commented):
Thank you for these updates @Electronic-Waste!
It looks good.
/lgtm

docs/proposals/push-based-metrics-collection.md
Electronic-Waste: Okay, I'll remove it.

docs/proposals/push-based-metrics-collection.md
```diff
@@ -0,0 +1,184 @@
...
+1. Implement authentication model for Katib DB to push metrics.
+
+2. Support pushing data to different types of storage system(prometheus, self-defined interface etc.)
+
+
+## API
+
+### New Parameter in Python SDK Function `tune`
+
+We decided to add `metrics_collection_mechanism` to `tune` function in Python SDK.
```
nsingl00, line 36: It should be `metrics_collector_config` right?

Electronic-Waste: Yes, I forgot to change it. Thank you!

andreyvelich (commented):
I think, we can merge this proposal.
We can create followup PRs if we need to modify it.
Thanks again for driving this @Electronic-Waste!
/lgtm
/approve

PR #2338, opened 2024-05-30
[SDK]Support Docker image as objective in the tune API

### LINKED ISSUE #2326: [SDK] Support Docker image as objective in the `tune` API
Ref discussion: https://github.com/kubeflow/website/pull/3723#discussion_r1590261777.

Currently, user can only pass the training function as [objective in the `tune` API](https://github.com/kubeflow/katib/blob/af900202c67843650d80ff955e4e7fb7510cc68a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L156) in Katib Python SDK.

Similar to [`create_job` API in Training Python SDK](https://github.com/kubeflow/training-operator/blob/f23c5c8b0c87d93f61b539015883152a0872124f/sdk/python/kubeflow/training/api/training_client.py#L327), we should give user an ability to set objective as Docker image.

/area sdk
/good-first-issue

akhilsaivenkata:
**What this PR does / why we need it**: Supporting Docker image as an objective in the tune API.

Fixes #2326 

**Checklist:**


akhilsaivenkata:
Hi @andreyvelich , I made the changes to katib_client.py. Do we have any test cases that need to be updated or implemented for this change? I also want to test these new changes on my local machine. I ran "make check" & "make test" commands on my local machine and there were no failures. I wonder if you could see and suggest any action items that needs to taken up here.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -153,9 +153,9 @@ def tune(
         self,
         # TODO (andreyvelich): How to be consistent with other APIs (name) ?
         name: str,
-        objective: Callable,
+        objective: Union[Callable, str],
         parameters: Dict[str, Any],
-        base_image: str = constants.BASE_IMAGE_TENSORFLOW,
+        #base_image: str = constants.BASE_IMAGE_TENSORFLOW,
```
andreyvelich, line 158: I think, we should keep the `base_image`, since we use it when user set `objective` as train function.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -400,12 +407,12 @@ def tune(
         trial_template = models.V1beta1TrialTemplate(
             primary_container_name=constants.DEFAULT_PRIMARY_CONTAINER_NAME,
             retain=retain_trials,
-            trial_parameters=trial_params,
+            trial_parameters=trial_params if callable(objective) else [],
```
andreyvelich, line 410: `trial_parameters` still be required even when user sets Docker image.
You can check example here: https://github.com/kubeflow/katib/blob/master/examples/v1beta1/hp-tuning/random.yaml#L31-L36

andreyvelich (commented):
Thank you for doing this @akhilsaivenkata!
I left a few comments.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -283,65 +283,72 @@ def tune(
...
+
+            # Objective function might be defined in some indented scope
+            # (e.g. in another function). We need to dedent the function code.
+            objective_code = textwrap.dedent(objective_code)
+
+            # Iterate over input parameters.
+            input_params = {}
+            experiment_params = []
+            trial_params = []
+            base_image = constants.BASE_IMAGE_TENSORFLOW,
```
akhilsaivenkata, line 302: Yes @andreyvelich , i did keep the base_image here at this line, I have added it in the if block so this code change got mixed up with all the other lines

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
akhilsaivenkata: Sure @andreyvelich , I will revert this change and keep the trial_parameters.

akhilsaivenkata: Also, I would like to know if we need to write new unit test cases or change existing ones ?

andreyvelich: cc @tariq-hasan

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -400,12 +407,12 @@ def tune(
...
             primary_container_name=constants.DEFAULT_PRIMARY_CONTAINER_NAME,
             retain=retain_trials,
-            trial_parameters=trial_params,
+            trial_parameters=trial_params if callable(objective) else [],
             trial_spec=trial_spec,
         )
 
         # Add parameters to the Katib Experiment.
-        experiment.spec.parameters = experiment_params
+        experiment.spec.parameters = experiment_params if callable(objective) else []
```
Electronic-Waste, line 415: I think `parameters` field is also needed since `trial_paramaters` is still required.

WDTY👀 @akhilsaivenkata @andreyvelich

akhilsaivenkata: You are right @Electronic-Waste , I have just reverted these two code changes and pushed it now.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -395,8 +403,8 @@ def tune(
                             client.V1Container(
                                 name=constants.DEFAULT_PRIMARY_CONTAINER_NAME,
                                 image=base_image,
-                                command=["bash", "-c"],
-                                args=[exec_script],
+                                command=["bash", "-c"] if callable(objective) else None,
+                                args=[exec_script] if callable(objective) else None,
```
Electronic-Waste, line 407: Also I'm not sure if we can assign `None` to `command` and `args` here when we use Docker image as objective.

As @andreyvelich shows an [example](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/hp-tuning/random.yaml#L31-L36) for us, we sometimes need to pass `command` and `args` to the training container to execute python scripts with some parameters. 

Could you explain your idea in details so that I can understand more? WDYT👀 @akhilsaivenkata @andreyvelich

Electronic-Waste (commented):
It looks great. Thank you @akhilsaivenkata ! I left a question for you and @andreyvelich .

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: I think, initially we can just allow user to set `image` as `objective` without `command` and `args`.
Similar to how we allow create training job using `base_image` parameter: https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/api/training_client.py#L327C35-L327C45.

andreyvelich:
Hi @akhilsaivenkata, did you get a chance to finish this PR ?

mahdikhashan:
@andreyvelich is it fine if i start working on this issue?
cc: @akhilsaivenkata

akhilsaivenkata:
> @andreyvelich is it fine if i start working on this issue? cc: @akhilsaivenkata

Hi @mahdikhashan , thanks for offering to help with this PR! I've been out for a bit, but now that I'm back, I'd like to revisit this PR myself. A lot has changed since I originally submitted it, and I want to make sure it still fits with our current project goals.
Thanks for keeping an eye on it, and I'll definitely let you know if I need any help!"

akhilsaivenkata:
Hi @andreyvelich ,

As there are many changes to this tune API in meantime, Is it okay if we introduce a new parameter, say objective_docker_image, rather than overloading objective to handle both a callable and a string?

I think this flow would be good:

- **External Models & Datasets (unchanged):** If model_provider_parameters, dataset_provider_parameters, or trainer_parameters are set, we do the external approach.
- **Custom Objective (new logic)**:
    1. If objective is a callable, we inject Python code (the existing approach).
    2. If objective_docker_image is provided (and objective is None), we run the user’s Docker image directly.
    
**We would add simple validation:**
- If you don’t use external models, then you must specify one of objective or objective_docker_image.
- If you set both, that’s an error.

This approach keeps things explicit and avoids confusion about whether objective is a function or a Docker image. Please let me know your thoughts on whether this design aligns with this API.

andreyvelich:
@akhilsaivenkata Maybe we can use the `base_image` parameter for now similar to Kubeflow Training V1: https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/api/training_client.py#L374-L377?
Also, how are you planning to pass the Trail hyperparameters when only Docker image is set ?

akhilsaivenkata:
> @akhilsaivenkata Maybe we can use the `base_image` parameter for now similar to Kubeflow Training V1: https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/api/training_client.py#L374-L377?
> Also, how are you planning to pass the Trail hyperparameters when only Docker image is set ?

@andreyvelich , Yes we can just reuse the base_image parameter for consistency with Kubeflow Training V1.

Previously, when objective_func was provided, trial parameters like ${trialParameters.<param>} were dynamically injected into the function at runtime, either as environment variables or command-line arguments, ensuring seamless hyperparameter substitution.

If we use only a Docker image as input, I believe we should plan to inject these trial parameters as environment variables/cmd line arguments into the container. This would require users to ensure their training scripts inside their custom docker image retrieve these variables using methods like os.getenv. We also need to highlight this in the documentation to ensure user awareness.

What are your thoughts on this approach? This might be a bit overhead on the users end though..

andreyvelich:
@akhilsaivenkata Apologies for the very long review of this PR. We have a few users who want to see this feature in Katib SDK. Are you able to finalize this PR ?



/remove-lifecycle stale
/reopen

PR #2347, opened 2024-06-10
Introducing go-install-tool in the Makefile makes it easier to set up the development environment

forsaken628:
Current development environments use binary tools that are not isolated, which makes switching between projects troublesome and error-prone. We can improve this situation by introducing go-install-tool in the makefile.

alse see:
https://github.com/kubernetes-sigs/kubebuilder/blob/98d03bed2912b2f7adf73aeaf5f64fafa8fd927b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L263-L275

Fixes #

**Checklist:**


tenzen-y:
> @forsaken628 I apologies that we don't have time to review this improvement. Are we ready to review this PR ? @Electronic-Waste @helenxie-bit @tenzen-y Please can you help with review of this PR ?

I don't think that we should take this approach. Instead of this, we should enforce the fixed version tools to $REPO_DIR/bin directory so that we can use the same version in our local.

Example: https://github.com/kubeflow/training-operator/blob/cbd4ae7bb32164816278fc0e313effc14b5d2161/Makefile#L119-L125

PR #2350, opened 2024-06-10
Fix TestReconcileBatchJob

forsaken628:
Fix unstable tests
#1649

**Checklist:**

### DIFF
--- a/pkg/controller.v1beta1/trial/trial_controller_test.go
+++ b/pkg/controller.v1beta1/trial/trial_controller_test.go
@@ -176,6 +176,7 @@ func TestReconcileBatchJob(t *testing.T) {
 	}
 
 	t.Run(`Trial run with "Failed" BatchJob.`, func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
 		mockMC.msg = observationLogUnavailable
 		trial := newFakeTrialBatchJob()
 		batchJob := &batchv1.Job{}
@@ -234,6 +235,7 @@ func TestReconcileBatchJob(t *testing.T) {
 	})
 
 	t.Run(`Trail with "Complete" BatchJob and Available metrics.`, func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
 		mockMC.msg = observationLogAvailable
 		batchJob := &batchv1.Job{}
 		batchJobCompleteMessage := "BatchJob completed test message"
@@ -281,6 +283,7 @@ func TestReconcileBatchJob(t *testing.T) {
 	})
 
 	t.Run(`Trail with "Complete" BatchJob and Unavailable metrics.`, func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
 		mockMC.msg = observationLogUnavailable
 		// Create the Trial
 		trial := newFakeTrialBatchJob()
@@ -309,6 +312,7 @@ func TestReconcileBatchJob(t *testing.T) {
 	})
 
 	t.Run("Update status for empty Trial", func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
 		g.Expect(r.updateStatus(&trialsv1beta1.Trial{})).To(gomega.HaveOccurred())
 	})
 }
--- a/pkg/controller.v1beta1/trial/trial_controller_test.go
+++ b/pkg/controller.v1beta1/trial/trial_controller_test.go
@@ -281,6 +281,8 @@ func TestReconcileBatchJob(t *testing.T) {
 			case trial.IsFailed(), trial.IsKilled(), trial.IsEarlyStopped():
 				t.Log(time.Since(start), trial.UID, trial.Status)
 				panic("trial failed")
+			case trial.IsMetricsUnavailable():
+				t.Log(time.Since(start), trial.UID, "metrics unavailable", trial.Status.Conditions)
 			case trial.IsRunning():
 				t.Log(time.Since(start), trial.UID, "running")
 			case trial.IsCreated():
--- a/pkg/controller.v1beta1/trial/trial_controller_test.go
+++ b/pkg/controller.v1beta1/trial/trial_controller_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tenzen-y:
@forsaken628 Thank you for this contribution.
Could you explain the reason why the flakiness happened and what you fixed the root cause in the PR description?

forsaken628:
According to the previous logic，call GetTrialObservationLog will return observationLogAvailable once and then observationLogUnavailable multiple times.

I investigated the cause of the failure in the ci environment: in test 2, which should have returned observationLogAvailable, it actually returned observationLogUnavailable, indicating that GetTrialObservationLog was accidentally called once somewhere.

Instead of locating where the accident occurred, I rewrote mockManagerClient to always return the same reply before calling DeleteTrialObservationLog, which I think is consistent with the semantics of ManagerClient.

tenzen-y:
> According to the previous logic，call GetTrialObservationLog will return observationLogAvailable once and then observationLogUnavailable multiple times.
> 
> I investigated the cause of the failure in the ci environment: in test 2, which should have returned observationLogAvailable, it actually returned observationLogUnavailable, indicating that GetTrialObservationLog was accidentally called once somewhere.
> 
> Instead of locating where the accident occurred, I rewrote mockManagerClient to always return the same reply before calling DeleteTrialObservationLog, which I think is consistent with the semantics of ManagerClient.

Thank you for clarifying the root cause. It seems that you split out multiple test case into dedicated case based on the result of the `GetTrialObservationLog`. That sounds reasonable.

pkg/controller.v1beta1/trial/trial_controller_test.go
tenzen-y, line 1: TBH, we should reimplement these tests as ginkgo DDD style, but that is out of scope of this PR.
So, I leave such refactoring in the future.

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -179,137 +175,146 @@ func TestReconcileBatchJob(t *testing.T) {
 		},
 	}
 
-	mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogAvailable, nil).Times(1)
-	mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogUnavailable, nil).MinTimes(1)
-	mockManagerClient.EXPECT().DeleteTrialObservationLog(gomock.Any()).Return(nil, nil).AnyTimes()
+	t.Run(`Trial run with "Failed" BatchJob.`, func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
+		mockMC.msg = observationLogUnavailable
```
tenzen-y, line 180:
```suggestion
		mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogAvailable, nil).AnyTimes()
```

Instead of this original mock client, can we reuse the autogenerated mock clients like this?

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -179,137 +175,146 @@ func TestReconcileBatchJob(t *testing.T) {
...
+		// BatchJob can't be deleted because GC doesn't work in envtest and BatchJob stuck in termination phase.
+		// Ref: https://book.kubebuilder.io/reference/testing/envtest.html#testing-considerations.
+		g.Eventually(func() bool {
+			return errors.IsNotFound(c.Get(ctx, trialKey, &trialsv1beta1.Trial{}))
+		}, timeout).Should(gomega.BeTrue())
+	})
+
+	t.Run(`Trail with "Complete" BatchJob and Available metrics.`, func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
+		mockMC.msg = observationLogAvailable
```
tenzen-y, line 239:
```suggestion
		mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogAvailable, nil).AnyTimes()
```

the same question.

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -179,137 +175,146 @@ func TestReconcileBatchJob(t *testing.T) {
...
+
+		// Expect that Trial is deleted
+		g.Eventually(func() bool {
+			return errors.IsNotFound(c.Get(ctx, trialKey, &trialsv1beta1.Trial{}))
+		}, timeout).Should(gomega.BeTrue())
+	})
+
+	t.Run(`Trail with "Complete" BatchJob and Unavailable metrics.`, func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
+		mockMC.msg = observationLogUnavailable
```
tenzen-y, line 287:
```suggestion
		mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogUnAvailable, nil).AnyTimes()
```

same.

pkg/controller.v1beta1/trial/trial_controller_test.go
forsaken628: We need [WithOverridableExpectations](https://pkg.go.dev/go.uber.org/mock@v0.4.0/gomock#WithOverridableExpectations) here, but github.com/golang/mock/mock does not support this feature.
Or ["Each test should create a new Controller"](https://pkg.go.dev/github.com/golang/mock@v1.6.0/gomock#Controller).
In my opinion, gomock is not flexible enough, and it complicates a simple thing for this scenario.

forsaken628: Finally, I found another way.

tenzen-y: @forsaken628 I opened the issue to replace gomock with go.uber.com/gomock here: https://github.com/kubeflow/katib/issues/2356

As a separate PR, could you work on the issue?

tenzen-y (commented):
/lgtm
/approve

Thank you for the great contribution!

andreyvelich:
Thank you for this amazing contribution @forsaken628!
I really hope it will improve our unstable tests.

Issue #2283, opened 2024-03-12
Define the Python deprecation/supporting policies for the Katib SDK


tenzen-y:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
As we discussed in https://github.com/kubeflow/katib/pull/2278#issuecomment-1992328152, currently, we support the following Python versions for the SDK, but we don't have any policies to decide which versions we should drop/support the Python version. So, we should define the Python deprecation/supporting policies for the SDK.

https://github.com/kubeflow/katib/blob/5837b8a90eb571e1a03da2ea01277d614d426bef/sdk/python/v1beta1/setup.py#L56-L61

Additionally, I believe that it is possible to apply the same policies to the training-operator.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Appendix: 
  - JupyterLab's policies: https://github.com/jupyterlab/jupyterlab/blob/main/pyproject.toml#L13
  - Jupyter Server policies: https://github.com/jupyter-server/jupyter_server/blob/main/pyproject.toml#L24

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

tenzen-y:
cc: @kubeflow/wg-automl-leads

andreyvelich:
cc @kubeflow/wg-training-leads @droctothorpe @shipengcheng1230 @andreafehrman @harrisonfritz @kuizhiqing @yaobaiwei

droctothorpe:
I suggest either EOL or some small grace period after EOL just from a security perspective:
https://devguide.python.org/versions/

FWIW, pipelines still supports 3.7 😢 :
https://github.com/kubeflow/pipelines/blob/master/sdk/python/setup.py#L102

andreyvelich:
@tenzen-y Should we just update our [SDK readme](https://github.com/kubeflow/katib/tree/master/sdk/python/v1beta1#requirements) to say that we are going to follow Python EOL cycle: https://devguide.python.org/versions/ ?

tenzen-y:
> @tenzen-y Should we just update our [SDK readme](https://github.com/kubeflow/katib/tree/master/sdk/python/v1beta1#requirements) to say that we are going to follow Python EOL cycle: https://devguide.python.org/versions/ ?

SGTM

PR #2354, opened 2024-06-12
[SDK] Explain Python version support cycle

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/2283.
I added info about Python version support for Katib SDK.

/assign @tenzen-y @droctothorpe @johnugeorge


tenzen-y (commented):
Thank you!
@andreyvelich Could we select the same policy in the training operator as well?

/lgtm
/approve

PR #2325, opened 2024-05-06
Add unit test for `create_experiment` in the `katib_client` module

tariq-hasan:
**What this PR does / why we need it**:

- Unit test has been added for the `create_experiment` function as part of [this](https://github.com/kubeflow/katib/pull/2325/commits/1545807266cf37e6f82d6bb672d30e281ecb083c) commit.
- Code refactoring has been done as part of other commits in the PR.
- I plan to create a second PR to add unit test for the `tune` function once this is merged.

**Which issue(s) this PR fixes** _(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_:
Fixes #2184

**Test Plan**: The code for the Python tests has been verified as part of [this](https://github.com/tariq-hasan/katib/actions/runs/8964476302) CI workflow.

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/components/katib/) included if any changes are user facing

PR #2304, opened 2024-04-10
[SDK] Fix env per Trial parameter in tune API

andreyvelich:
I fixed the `env_per_trial` parameter assignment.
Currently, if user sets dict in `env_per_trial` we override it [to None in this condition.](https://github.com/kubeflow/katib/blob/7df05c23a55363ba7cbbd7b9859943562afd71f4/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L347-L351)

/assign @kubeflow/wg-training-leads @shipengcheng1230 @droctothorpe

PR #2360, opened 2024-06-17
[SDK] Fix empty list for env variables and numpy version

andreyvelich:
After this PR: https://github.com/kubeflow/katib/pull/2304, the `tune` API doesn't work correct.

@helenxie-bit and @quloos Identified bug when using `tune` API.
If user doesn't set `env_per_trial` parameter, the Experiment creation fails with this error:
```
"message":"admission webhook \"validator.experiment.katib.kubeflow.org\" denied the request:
invalid spec.trialTemplate: unable to convert: /spec/template/spec/containers/0/env - [] to Job
```

We should prioritise unit test PR for Katib SDK to help us detect invalid SDK: https://github.com/kubeflow/katib/pull/2325 cc @tariq-hasan 

/assign @johnugeorge @tenzen-y @helenxie-bit @quloos


andreyvelich:
@kubeflow/wg-training-leads It looks like `numpy` 2.0 was released yesterday: https://github.com/numpy/numpy/issues/24300.

Since `torchvision` just [installs the latest `numpy` version](https://github.com/pytorch/vision/blob/release/0.17/setup.py#L61), I am using  `numpy==1.26.0` version in our Trial images.

Otherwise, I see the following error from PyTorch:
```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
```

tenzen-y (commented):
/lgtm
/approve

Thank you!

helenxie-bit:
It works now! Thank you so much! @andreyvelich

PR #2361, opened 2024-06-18
Refactor updateStopRules and add tests

forsaken628:

Reorganize the scattered status. Add tests.
There should be a bug here. Step will be less than zero.
https://github.com/kubeflow/katib/blob/f8b8d8d48437d5724333048bb017730a884d706c/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L360-L365
Fixes #

**Checklist:**


andreyvelich:
>There should be a bug here. Step will be less than zero.

Thank you for refactoring these functions and add tests @forsaken628!
Please can you explain the use-case when step will be less than zero and why we don't catch this bug in our E2Es ?

PR #2363, opened 2024-06-19
Remove code generation from release script

andreyvelich:
I removed the code generation from our release script.

Currently, we have to clone Katib repo to the `$GOPATH/src/github.com/kubeflow/katib` to correctly run Go code generation. However, in the release script we clone the Katib repo to the [temp directory](https://github.com/kubeflow/katib/blob/0947169ab114dfe71f5492b0d094d4c2054052a9/scripts/v1beta1/release.sh#L43-L47) to avoid un-sync Git history.

We don't need to run `make generate` in our release script, since we have [the code generation check](https://github.com/andreyvelich/katib/blob/0947169ab114dfe71f5492b0d094d4c2054052a9/hack/verify-generated-codes.sh#L24) for each PR.

Also, we don't need to publish images with COMMIT tag during the release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,14 +115,12 @@ push-latest: generate
 	bash scripts/v1beta1/push.sh $(KATIB_REGISTRY) $(COMMIT)
 
 # Build and push Katib images for the given tag.
-push-tag: generate
+push-tag:
 ifeq ($(TAG),)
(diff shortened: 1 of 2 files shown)


tenzen-y (commented):
Thank you!

/lgtm
/approve

PR #2368, opened 2024-06-20
Remove default caBundle value

### LINKED ISSUE #2367: Explicit caBundle value makes using terraform to manage katib challenging
### What happened?

We use terraform to manage a lot of our clusters and so having the caBundle in the webhooks set to `Cg==` causes issues since it later gets replaced but that makes terraform unhappy since the cluster state no longer matches the spec.

The caBundle has been optional since Kubernetes v1.13 (see https://github.com/kubernetes/kubernetes/issues/69590) so having it unset is better than setting it to the newline char.

Unsetting the caBundle fixes the issue since terraform treats it as a computed field.

### Environment

Kubernetes version: v1.30.1-gke.1156000

Katib controller version: v0.16.0

### Impacted by this bug?

vihangm:
caBundle hasn't been a required key since k8s v1.13
Removing it makes it simpler to deploy katib with terraform.

Fixes #2367

**Checklist:**

### DIFF
--- a/manifests/v1beta1/components/webhook/webhooks.yaml
+++ b/manifests/v1beta1/components/webhook/webhooks.yaml
@@ -9,7 +9,6 @@ webhooks:
     admissionReviewVersions:
       - v1
     clientConfig:
-      caBundle: Cg==
       service:
         name: katib-controller
         namespace: kubeflow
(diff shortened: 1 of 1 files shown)


andreyvelich (approved):
Thank you for updating this @vihangm!
/lgtm
/approve

Issue #2340, opened 2024-06-03
[GSoC] Project 6: Push-based Metrics Collection for Katib


Electronic-Waste:
## Goal

The project aims to provide a Python SDK API interface for users to push metrics to Katib DB directly. 

The current implementation of Metrics Collector is pull-based, raising design problems such as determining the frequency at which we scrape the metrics, performance issues like the overhead caused by too many sidecar containers, and restrictions on developing environments that must support sidecar containers and admission webhooks. 

Thus, we decided to implement a new API for Katib Python SDK to offer users a push-based way to store metrics directly into the Kaitb DB and resolve those issues raised by pull-based metrics collection.

<img width="856" alt="push-based-metrics-collection" src="https://github.com/user-attachments/assets/28ea3eda-c09e-4df7-8414-6a8f306b2d01">

## What I did in GSoC Project & Ongoing Works

This issue tracks the progress of developing push-based metrics collection for katib during the GSoC coding phase.

I raised numerous PRs for the Katib and Training-Operator project. Some of them are related to my GSoC project, and others may contribute to the completeness of UTs, simplicity of dependency package, and the compatibility of UI component, etc. 

Also, I raised some issues not only to describe the problems and bugs I met during the coding period, but also to suggest the future enhancement direction for Katib and Training-Operator. 

PRs concerned with the project:

- [x] Convert GSoC proposal to KEP: https://github.com/kubeflow/katib/pull/2328
- [x] Add new parameter in `tune` function: https://github.com/kubeflow/katib/pull/2369
- [x] New interface `report_metrics` in Python SDK: https://github.com/kubeflow/katib/pull/2371
- [x] Compatibility changes in trial controller: https://github.com/kubeflow/katib/pull/2394
- [x] Perform unit tests and e2e tests: 
- https://github.com/kubeflow/katib/pull/2398 
- https://github.com/kubeflow/katib/pull/2399
- [x] Create documents for the new feature, including API specification, usage, examples, etc.
- https://github.com/kubeflow/website/pull/3844
- https://github.com/kubeflow/katib/pull/2437
- https://github.com/kubeflow/blog/pull/155

Other PRs:
- https://github.com/kubeflow/katib/pull/2391
- https://github.com/kubeflow/katib/pull/2401
- https://github.com/kubeflow/training-operator/pull/2196
- https://github.com/kubeflow/katib/pull/2418

Issues I raised:
- https://github.com/kubeflow/katib/issues/2340
[...]

PR #2369, opened 2024-06-23
[GSoC] Add New Parameter in `tune`

Electronic-Waste:

I add a new parameter `metrics_collector_config` to `tune` function. Design details: https://github.com/kubeflow/katib/blob/cc95ef03cb25df3d86a1a1c20c9c69bad17fce92/docs/proposals/push-based-metrics-collection.md#add-new-parameter-in-tune

Fixes #

**Checklist:**

### DIFF
--- a/pkg/apis/controller/common/v1beta1/common_types.go
+++ b/pkg/apis/controller/common/v1beta1/common_types.go
@@ -220,8 +220,8 @@ const (
 	CustomCollector CollectorKind = "Custom"
 
 	// When model training source code persists metrics into persistent layer
-	// directly, metricsCollector isn't in need, and its kind is "noneCollector"
-	NoneCollector CollectorKind = "None"
+	// directly, sidecar container isn't in need, and its kind is "pushCollector"
+	PushCollector CollectorKind = "Push"
 
 	MetricsVolume = "metrics-volume"
 )
--- a/pkg/webhook/v1beta1/pod/inject_webhook.go
+++ b/pkg/webhook/v1beta1/pod/inject_webhook.go
@@ -147,8 +147,8 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 		return mutatedPod, nil
 	}
 
-	// If Metrics Collector in None, skip the mutation.
-	if trial.Spec.MetricsCollector.Collector.Kind == common.NoneCollector {
+	// If Metrics Collector is Push, skip the mutation.
+	if trial.Spec.MetricsCollector.Collector.Kind == common.PushCollector {
 		return mutatedPod, nil
 	}
 
--- a/pkg/webhook/v1beta1/pod/inject_webhook.go
+++ b/pkg/webhook/v1beta1/pod/inject_webhook.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"path/filepath"
 	"strconv"
@@ -139,6 +140,25 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 
 	// Add Katib Trial labels to the Pod metadata.
 	mutatePodMetadata(mutatedPod, trial)
+
+
+	// Pass env variable KATIB_TRIAL_NAME to training containers using fieldPath.
+	for idx := range mutatedPod.Spec.Containers {
+		if mutatedPod.Spec.Containers[idx].Env == nil {
+			mutatedPod.Spec.Containers[idx].Env = []v1.EnvVar{}
+		}
+		mutatedPod.Spec.Containers[idx].Env = append(
+			mutatedPod.Spec.Containers[idx].Env, 
+			v1.EnvVar{
+				Name: consts.EnvTrialName,
+				ValueFrom: &v1.EnvVarSource{
+					FieldRef: &v1.ObjectFieldSelector{
+						FieldPath: fmt.Sprintf("metadata.labels['%s']", consts.LabelTrialName),
+					},
+				},
+			},
+		)
+	}
 
 	// Do the following mutation only for the Primary pod.
 	// If PrimaryPodLabel is not set we mutate all pods which are related to Trial job.
--- a/pkg/webhook/v1beta1/pod/inject_webhook.go
+++ b/pkg/webhook/v1beta1/pod/inject_webhook.go
@@ -148,7 +148,7 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 			mutatedPod.Spec.Containers[index].Env = []v1.EnvVar{}
 		}
 		mutatedPod.Spec.Containers[index].Env = append(
-			mutatedPod.Spec.Containers[index].Env, 
+			mutatedPod.Spec.Containers[index].Env,
 			v1.EnvVar{
 				Name: consts.EnvTrialName,
 				ValueFrom: &v1.EnvVarSource{
--- a/pkg/webhook/v1beta1/pod/inject_webhook.go
+++ b/pkg/webhook/v1beta1/pod/inject_webhook.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"net/http"
 	"path/filepath"
 	"strconv"
@@ -141,26 +140,11 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 	// Add Katib Trial labels to the Pod metadata.
 	mutatePodMetadata(mutatedPod, trial)
 
-	// Pass env variable KATIB_TRIAL_NAME to the primary container using fieldPath.
-	index := getPrimaryContainerIndex(mutatedPod.Spec.Containers, trial.Spec.PrimaryContainerName)
-	if index >= 0 {
-		if mutatedPod.Spec.Containers[index].Env == nil {
-			mutatedPod.Spec.Containers[index].Env = []v1.EnvVar{}
-		}
-		mutatedPod.Spec.Containers[index].Env = append(
-			mutatedPod.Spec.Containers[index].Env,
-			v1.EnvVar{
-				Name: consts.EnvTrialName,
-				ValueFrom: &v1.EnvVarSource{
-					FieldRef: &v1.ObjectFieldSelector{
-						FieldPath: fmt.Sprintf("metadata.labels['%s']", consts.LabelTrialName),
-					},
-				},
-			},
-		)
-	} else {
-		return nil, fmt.Errorf("Unable to find primary container %v in mutated pod containers %v",
-			trial.Spec.PrimaryContainerName, pod.Spec.Containers)
+	// Add env variables to the Pod's primary container.
+	// We add this function because of push-based metrics collection function `report_metrics` in Python SDK.
+	// Currently, we only pass the Trial name as env variable `KATIB_TRIAL_NAME` to the training container.
+	if err := mutatePodEnv(mutatedPod, trial); err != nil {
+		return nil, err
 	}
 
 	// Do the following mutation only for the Primary pod.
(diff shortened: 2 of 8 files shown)


Electronic-Waste:
ref issue: #2340

Electronic-Waste:
PTAL👀 @andreyvelich @johnugeorge

pkg/apis/controller/common/v1beta1/common_types.go
```diff
@@ -220,8 +220,8 @@ const (
 	CustomCollector CollectorKind = "Custom"
 
 	// When model training source code persists metrics into persistent layer
-	// directly, metricsCollector isn't in need, and its kind is "noneCollector"
-	NoneCollector CollectorKind = "None"
+	// directly, sidecar container isn't in need, and its kind is "pushCollector"
+	PushCollector CollectorKind = "Push"
```
tenzen-y, line 224: Do you have any consensus on removing existing fields with @andreyvelich and @johnugeorge ?
In fact, removing the existing field means losing backward compatibility.
In general, the behavior is allowed us only when we upgrade API version (v1 -> v2).

andreyvelich: Yes, but I don't have strong concerns with removing None metrics collector.
I think, this parameter was never used by Katib users since it is not possible to run Katib Experiments when metrics are not collected.

johnugeorge: Since it is never used, I think, it is ok.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -369,6 +373,19 @@ def tune(
...
+            collector=models.V1beta1CollectorSpec(kind=metrics_collector_config["kind"])
+        )
+        if metrics_collector_config["kind"] == "Push":
+            trial_params.append(
+                models.V1beta1TrialParameterSpec(name="trialName", reference="${{trialSpec.Name}}")
+            )
+            env.append(
+                client.V1EnvVar(name="KATIB_TRIAL_NAME", value="${{trialParameters.trialName}}")
+            )
+
```
andreyvelich, line 388: Do we want to always pass KATIB_TRIAL_NAME to all Katib Trials ?
I feel that this parameter is very useful for different use-cases and we should always pass it to the training container.
For example, when user wants to export trained model and model prefix = Trial name.

Electronic-Waste: Yeah, it seems that it's worthwhile to always pass `KATIB_TRIAL_NAME` to Katib Trials. 

I'll modify the code and add some comment in the doc string.

andreyvelich: @Electronic-Waste What if we use `fieldPath` for `KATIB_TRIAL_NAME` env as follows:
```
name: RANK
valueFrom:
  fieldRef:
    fieldPath: metadata.labels['katib.kubeflow.org/trial']
```

Like in this [JobSet example](https://github.com/kubernetes-sigs/jobset/blob/main/examples/pytorch/cnn-mnist/mnist.yaml#L27-L30).

I believe, we already mutate Trial name [to the Trial's Pod labels.](https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/pod/utils.go#L278).

Electronic-Waste: @andreyvelich Okay, I see. I'll commit new changes to this branch soon.

Electronic-Waste:
@andreyvelich I add my code here: https://github.com/kubeflow/katib/blob/db17214cf02c8d4eb9dafc21a15ffb7af4596015/pkg/webhook/v1beta1/pod/inject_webhook.go#L140-L142


```Golang
// Pass env variable KATIB_TRIAL_NAME to training containers using fieldPath.
for idx := range mutatedPod.Spec.Containers {
	if mutatedPod.Spec.Containers[idx].Env == nil {
		mutatedPod.Spec.Containers[idx].Env = []v1.EnvVar{}
	}
	mutatedPod.Spec.Containers[idx].Env = append(
		mutatedPod.Spec.Containers[idx].Env, 
		v1.EnvVar{
			Name: consts.EnvTrialName,
			ValueFrom: &v1.EnvVarSource{
				FieldRef: &v1.ObjectFieldSelector{
					FieldPath: fmt.Sprintf("metadata.labels['%s']", consts.LabelTrialName),
				},
			},
		},
	)
}
```

PTAL👀. Thank you for your review!

andreyvelich:
I don't think, we should mutate `KATIB_TRIAL_NAME` env to all containers in the Pod.
So maybe, similar as in [`wrapWorkerContainer` function](https://github.com/kubeflow/katib/blob/38114593d609bf30d511100a59cde1a2327216aa/pkg/webhook/v1beta1/pod/utils.go#L152-L156), we should only add this environment to the primary container.

andreyvelich:
Please rebase your PR as well @Electronic-Waste

andreyvelich:
@Electronic-Waste Additionally, I think we forgot to add this line to the post-gen script of Katib SDK:
```python
# Import Katib report metrics functions
from kubeflow.katib.api.report_metrics import report_metrics
```

Similar to how we include [other imports to the`__init__.py`](https://github.com/kubeflow/katib/blob/master/hack/gen-python-sdk/post_gen.py#L42-L43).
Since this file is auto-generated by code-gen, we need to update it via post-gen script.

@Electronic-Waste Please update this script, otherwise after we re-generate the SDK, your import will be deleted.

Electronic-Waste:
@andreyvelich Done! I've rebased the branch, modified the code to pass env variable only to the primary container and added the importing code to the post-gen script of Katib SDK.

Electronic-Waste:
@andreyvelich Thank you for your detailed review!

pkg/webhook/v1beta1/pod/inject_webhook.go
```diff
@@ -140,15 +141,37 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 	// Add Katib Trial labels to the Pod metadata.
 	mutatePodMetadata(mutatedPod, trial)
 
+	// Pass env variable KATIB_TRIAL_NAME to the primary container using fieldPath.
```
andreyvelich, line 144: @Electronic-Waste Additionally, since we always pass the Trial name to the training container, we should not allow to pass Trial name and namespace [via template generator](https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/experiment/manifest/generator.go#L150-L153)

pkg/webhook/v1beta1/pod/inject_webhook.go
```diff
@@ -140,15 +141,37 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
...
+	index := getPrimaryContainerIndex(mutatedPod.Spec.Containers, trial.Spec.PrimaryContainerName)
+	if index >= 0 {
```
andreyvelich, line 146: I am also wondering, do we really need to set `primaryPodLabels` and `primaryContainerName` when metrics collector is Push ? @Electronic-Waste @johnugeorge @tenzen-y ?

pkg/webhook/v1beta1/pod/inject_webhook.go
```diff
@@ -140,15 +141,37 @@ func (s *SidecarInjector) Mutate(pod *v1.Pod, namespace string) (*v1.Pod, error)
 	// Add Katib Trial labels to the Pod metadata.
 	mutatePodMetadata(mutatedPod, trial)
 
+	// Pass env variable KATIB_TRIAL_NAME to the primary container using fieldPath.
+	index := getPrimaryContainerIndex(mutatedPod.Spec.Containers, trial.Spec.PrimaryContainerName)
```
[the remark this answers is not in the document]
Electronic-Waste: I agree with you. I'll wrap the code into `mutatePodEnv` function and add some unit tests for it.

pkg/webhook/v1beta1/pod/inject_webhook.go
Electronic-Waste: I guess it's not friendly to users. Since junior users are unfamiliar with that, they may be puzzled about why they can't pass Trial's namespace and name to the training container. It's more straightforward for them to simply define trialParameter `trialName` and use it in the pod env as an [example](https://github.com/kubeflow/katib/blob/master/examples/v1beta1/trial-template/trial-metadata-substitution.yaml) suggests.

Also, it's more backward-compatible to just allow users passing Trial name and namespace via triallParameter and pod env. WDYT👀 @andreyvelich @tenzen-y @johnugeorge .

pkg/webhook/v1beta1/pod/inject_webhook.go
Electronic-Waste: I think we don't need to set `primaryPodLabels` and `primaryContainerName` for push-based metrics collector. 

But I guess it's not worthwhile to add more if-else branches since it will sacrifice readability of code. If we prohibit specifying `primaryPodLabel` and `primaryContainerName` for the Trial, users will find that it's not similar to other collectors' specification and start figuring out why throughout docs, github and stackoverflow, which will definitely do harm to their experience of using Katib.

So I suggest using unified interface to maintain backward compatibility and users' experience.

andreyvelich: >users will find that it's not similar to other collectors' specifications when using kubectl to apply Experiment

Actually, I think I am wrong. We also use the `primaryContainerName` to mutate Suggestion's volume to the Trial: https://github.com/kubeflow/katib/blob/master/pkg/webhook/v1beta1/pod/inject_webhook.go#L334.
So we might need to keep it for `push` metrics collector as well.

Electronic-Waste:
@andreyvelich I've wrapped the code into `mutatePodEnv` and added some unit tests for it. PTAL👀

pkg/webhook/v1beta1/pod/inject_webhook_test.go
```diff
@@ -25,7 +25,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/onsi/gomega"
+	"google.golang.org/protobuf/testing/protocmp"
```
andreyvelich, line 31: I am wondering, do we really need to use  protocmp? 
It looks like for some unit test we just compare 2 objects: https://github.com/kubeflow/katib/blob/master/pkg/util/v1beta1/katibconfig/config_test.go#L173
cc @tenzen-y @Electronic-Waste

andreyvelich (commented):
Thank you @Electronic-Waste!
just a small comment from me.

pkg/webhook/v1beta1/pod/inject_webhook_test.go
Electronic-Waste: @andreyvelich Yeah, I find we can also pass the unit tests without using `protocmp`.

I use `protocmp` because of the former test cases I wrote in: https://github.com/kubeflow/katib/blob/0b4e7c178094a4eb84c2ef0cb3be1c6fdf2e3928/pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector_test.go#L324-L325

Now I have deleted it in `inject_webhook_test.go`.

pkg/webhook/v1beta1/pod/inject_webhook.go
andreyvelich: >Since junior users are unfamiliar with that, they may be puzzled about why they can't pass Trial's namespace and name to the training container. 

Since we always pass KATIB_TRIAL_NAME to the Trial's Pods, users will always be able to read this environment in their training code, isn't ?
However, this can be useful for other part of the spec: What if user wants to generate different volume which contains Trial name.

@Electronic-Waste Please can you an create an issue to track this discussion ? I think, we should discuss the future of Trial metadata parameters.

andreyvelich (commented):
I think, we should be ready to merge it.
Thanks to your contribution @tenzen-y 🎉 
/lgtm
/approve

PR #2371, opened 2024-06-24
[GSoC] New Interface `report_metrics` in Python SDK

Electronic-Waste:

I add a new interface `report_metrics` in Python SDK. Design details: https://github.com/kubeflow/katib/blob/4161cff0faf7700598489f1fbaa5346b948623e2/docs/proposals/push-based-metrics-collection.md#new-interface-report_metrics-in-python-sdk

Fixes #

**Checklist:**

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
@@ -13,11 +13,10 @@
 # limitations under the License.
 
 import os
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any, Dict
 
 import grpc
-import pytz
 import kubeflow.katib.katib_api_pb2 as katib_api_pb2
 from kubeflow.katib.constants import constants
 from kubeflow.katib.utils import utils
@@ -29,7 +28,7 @@ def report_metrics(
 ):
     """Push Metrics Directly to Katib DB
 
-    [!!!] Trial name should always be passed into Katib Trials as env variable `KATIB_TRIAL_NAME`.
+    Katib always pass Trial name as env variable `KATIB_TRIAL_NAME` to the training container.
 
     Args:
         metrics: Dict of metrics pushed to Katib DB.
@@ -56,7 +55,7 @@ def report_metrics(
     
     with katib_api_pb2.beta_create_DBManager_stub(channel) as client:
         try:
-            timestamp = datetime.now(pytz.UTC).isoformat(timespec="nanoseconds")
+            timestamp = datetime.now(timezone.utc).strftime(constants.RFC3339_FORMAT)
             client.ReportObservationLog(
                 request=katib_api_pb2.ReportObservationLogRequest(
                     trial_name=name,
@@ -76,3 +75,4 @@ def report_metrics(
             raise RuntimeError(
                 f"Unable to push metrics to Katib DB for Trial {namespace}/{name}. Exception: {e}"
             )
+        
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
@@ -1,0 +1,78 @@
+# Copyright 2024 The Kubeflow Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from datetime import datetime
+from typing import Any, Dict
+
+import grpc
+import pytz
+import kubeflow.katib.katib_api_pb2 as katib_api_pb2
+from kubeflow.katib.constants import constants
+from kubeflow.katib.utils import utils
+
+def report_metrics(
+    metrics: Dict[str, Any],
+    db_manager_address: str = constants.DEFAULT_DB_MANAGER_ADDRESS,
+    timeout: int = constants.DEFAULT_TIMEOUT,
+):
+    """Push Metrics Directly to Katib DB
+
+    [!!!] Trial name should always be passed into Katib Trials as env variable `KATIB_TRIAL_NAME`.
+
+    Args:
+        metrics: Dict of metrics pushed to Katib DB.
+            For examle, `metrics = {"loss": 0.01, "accuracy": 0.99}`.
+        db-manager-address: Address for the Katib DB Manager in this format: `ip-address:port`.
+        timeout: Optional, gRPC API Server timeout in seconds to report metrics.
+    
+    Raises:
+        ValueError: The Trial name is not passed to environment variables.
+        RuntimeError: Unable to push Trial metrics to Katib DB.
+    """
+
+    namespace = utils.get_current_k8s_namespace()
+    name = os.getenv("KATIB_TRIAL_NAME")
+    if name is None:
+        raise ValueError(
+            "The Trial name is not passed to environment variables"
+        )
+
+    db_manager_address = db_manager_address.split(":")
+    channel = grpc.beta.implementations.insecure_channel(
+        db_manager_address[0], int(db_manager_address[1])
+    )
+    
+    with katib_api_pb2.beta_create_DBManager_stub(channel) as client:
+        try:
+            timestamp = datetime.now(pytz.UTC).isoformat(timespec="nanoseconds")
+            client.ReportObservationLog(
+                request=katib_api_pb2.ReportObservationLogRequest(
+                    trial_name=name,
+                    observation_logs=katib_api_pb2.ObservationLog(
+                        metric_logs=[
+                            katib_api_pb2.MetricLog(
+                                time_stamp=timestamp,
+                                metric=katib_api_pb2.Metric(name=name,value=value)   
+                            )
+                            for name, value in metrics.items()
+                        ]
+                    )
+                ),
+                timeout=timeout,
+            )
+        except Exception as e:
+            raise RuntimeError(
+                f"Unable to push metrics to Katib DB for Trial {namespace}/{name}. Exception: {e}"
+            )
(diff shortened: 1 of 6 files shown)


Electronic-Waste:
ref issue: #2340

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from datetime import datetime
+from typing import Any, Dict
+
+import grpc
+import pytz
```
andreyvelich, line 20: Is it possible to not use `pytz` for date-time format ? E.g. we can add the format to the SDK constants.
That will help us to reduce number of dependencies for Katib SDK.

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2024 The Kubeflow Authors.
```
andreyvelich, line 1: One idea might be to add this function to Katib utils: https://github.com/kubeflow/katib/blob/1d9be470dbf9499c6a94070572956ff844b1c66e/sdk/python/v1beta1/kubeflow/katib/utils/utils.py

And then add this function to [the `__init__.py file`](https://github.com/kubeflow/katib/blob/1d9be470dbf9499c6a94070572956ff844b1c66e/sdk/python/v1beta1/kubeflow/katib/__init__.py#L72-L75):

```py
from kubeflow.katib.utils.utils import report_metrics
```

So users can just use it as follows:
```python

katib.report_metrics()
```


That will help us to reduce number of files.
WDYT @Electronic-Waste @johnugeorge @droctothorpe ?

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+from typing import Any, Dict
+
+import grpc
+import pytz
+import kubeflow.katib.katib_api_pb2 as katib_api_pb2
+from kubeflow.katib.constants import constants
+from kubeflow.katib.utils import utils
+
+def report_metrics(
+    metrics: Dict[str, Any],
```
andreyvelich, line 26: Isn't the value for metrics should be always `str` ?

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+    db_manager_address: str = constants.DEFAULT_DB_MANAGER_ADDRESS,
+    timeout: int = constants.DEFAULT_TIMEOUT,
+):
+    """Push Metrics Directly to Katib DB
+
+    [!!!] Trial name should always be passed into Katib Trials as env variable `KATIB_TRIAL_NAME`.
```
andreyvelich, line 32: Should we modify this message to tell that Katib will always send Trial name as env variable: `KATIB_TRIAL_NAME` to the training container ?

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+                            for name, value in metrics.items()
+                        ]
+                    )
+                ),
+                timeout=timeout,
+            )
+        except Exception as e:
+            raise RuntimeError(
+                f"Unable to push metrics to Katib DB for Trial {namespace}/{name}. Exception: {e}"
+            )
```
andreyvelich, line 78: Please add the new line.

andreyvelich (commented):
Thank you for adding this @Electronic-Waste!
I added a few comments.

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
Electronic-Waste: Okay, I'll do that.

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
Electronic-Waste: I agree with you. We've already passed Trial name into the training container for users.

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
Electronic-Waste: It can also be `int` or `float`. I think it will be more convenient for users.

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
Electronic-Waste: I think we should put `report_metrics` in the `api` dir because it's a user interface rather than a util function.

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
Electronic-Waste: Yes, I will import `timezone` from `datetime` package to accomplish that.

sdk/python/v1beta1/kubeflow/katib/api/katib_report_metrics.py
andreyvelich: Ok, than should we just call this file `report_metrics.py` and also add it to the `__init__.py` file as I mentioned above?

Electronic-Waste: Yeah, I agree with that.

Electronic-Waste:
@andreyvelich I made the following changes as you requested after the meeting yesterday:
- A validation function for metrics value in `utils.py`
- Check the values in `dict[str, Any]` in `report_metrics`
- Rename `katib_report_metrics.py` to `report_metrics.py`

Can you take a look when you are free? Thanks a lot!

sdk/python/v1beta1/kubeflow/katib/api/report_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+    Katib always pass Trial name as env variable `KATIB_TRIAL_NAME` to the training container.
+
+    Args:
+        metrics: Dict of metrics pushed to Katib DB.
+            For examle, `metrics = {"loss": 0.01, "accuracy": 0.99}`.
+        db-manager-address: Address for the Katib DB Manager in this format: `ip-address:port`.
+        timeout: Optional, gRPC API Server timeout in seconds to report metrics.
+    
+    Raises:
+        ValueError: The Trial name is not passed to environment variables.
```
andreyvelich, line 40: I guess, if metrics have incorrect format we also return `ValueError`. It would be nice to add it to the doc string.

sdk/python/v1beta1/kubeflow/katib/api/report_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+                        ]
+                    )
+                ),
+                timeout=timeout,
+            )
+        except Exception as e:
+            raise RuntimeError(
+                f"Unable to push metrics to Katib DB for Trial {namespace}/{name}. Exception: {e}"
+            )
+
```
andreyvelich, line 85: Please add the new line.

andreyvelich (commented):
Thanks @Electronic-Waste, I just added a few nits.

sdk/python/v1beta1/kubeflow/katib/api/report_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+from kubeflow.katib.utils import utils
+
+def report_metrics(
+    metrics: Dict[str, Any],
+    db_manager_address: str = constants.DEFAULT_DB_MANAGER_ADDRESS,
+    timeout: int = constants.DEFAULT_TIMEOUT,
+):
+    """Push Metrics Directly to Katib DB
+
+    Katib always pass Trial name as env variable `KATIB_TRIAL_NAME` to the training container.
```
andreyvelich, line 31:
```suggestion
    Katib always passes Trial name as env variable `KATIB_TRIAL_NAME` to the training container.
```

sdk/python/v1beta1/kubeflow/katib/api/report_metrics.py
Electronic-Waste: Sorry, I might not have understood what you mean. I've added a new line(L85) after the last line of code. Do you mean that I need to add another new blank line?

sdk/python/v1beta1/kubeflow/katib/api/report_metrics.py
Electronic-Waste: Okay, I'll add this scenario to the doc string.

sdk/python/v1beta1/kubeflow/katib/api/report_metrics.py
Electronic-Waste: I'll correct this spelling error soon.

sdk/python/v1beta1/kubeflow/katib/api/report_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+                    )
+                ),
+                timeout=timeout,
+            )
+        except Exception as e:
+            raise RuntimeError(
+                f"Unable to push metrics to Katib DB for Trial {namespace}/{name}. Exception: {e}"
+            )
+        
+
```
andreyvelich, line 87:
```suggestion
            )

```

sdk/python/v1beta1/kubeflow/katib/api/report_metrics.py
andreyvelich: Oh, I see. I think, it complains because you have empty spaces in that lines.
You can try to format your code with `black` to fix it.
Otherwise, I just added the suggestion that you need to make.

Electronic-Waste: Okay, I see. I've never met with this issue before. 
Thank you for your patient response and time @andreyvelich !
Btw, I think your suggestions are correct and valuable. So I usually re-think my implementation and modify it.

andreyvelich (commented):
Thank you @Electronic-Waste!
/lgtm

andreyvelich (approved):
I think, we should be ready to merge it.
/approve

Issue #2339, opened 2024-06-01
Tracking Issue: Implementation of Tuning API in Katib for LLMs


helenxie-bit:
This issue tracks the progress of developing the LLM Hyperparameters Tuning API in Katib. The API aims to provide an easy-to-use interface for tuning the hyperparameters of large language models, leveraging the capabilities of Katib and Training Operator.

- [x] https://github.com/kubeflow/katib/pull/2333
- [ ] Setup development environment
- [ ] Develop the initial version of the API with basic endpoints
- [ ] Conduct unit tests for basic endpoints
- [ ] Develop extended API functionalities
- [ ] Perform unit tests for new functionalities
- [ ] Conduct integration tests
- [ ] Implement authentication mechanisms and perform security testing
- [ ] Finalize the codebase
- [ ] Create documentation for the API, including usage instructions, code examples, etc
- [ ] Create API specification

Please let me know if you have any suggestions @andreyvelich @johnugeorge

andreyvelich:
Hi @helenxie-bit, are we ready to close this issue ?

helenxie-bit:
> Hi [@helenxie-bit](https://github.com/helenxie-bit), are we ready to close this issue ?

Thank you for following up! Should I close this once the PRs for the e2e test and user guide are merged?

andreyvelich:
Sure, let's close it once we finish with E2Es and user guides.

andreyvelich:
We can close it since the E2E has been merged.

PR #2393, opened 2024-07-21
[GSoC] Update `tune` API for LLM hyperparameters optimization

helenxie-bit:
This PR implements the initial functionality for LLM hyperparameter optimization in the `tune` API.

Fixes #

**Checklist:**
- For a detailed explanation of this API, please refer to the proposal: https://github.com/kubeflow/katib/pull/2333


sdk/python/v1beta1/kubeflow/katib/constants/constants.py
```diff
@@ -56,3 +58,26 @@
...
+)
+
+STORAGE_INITIALIZER_VOLUME = client.V1Volume(
+    name=STORAGE_INITIALIZER,
+    persistent_volume_claim=client.V1PersistentVolumeClaimVolumeSource(
+        claim_name=STORAGE_INITIALIZER
+    ),
+)
+
+TRAINER_TRANSFORMER_IMAGE = "" # Need to be built using the `trainer` file
```
helenxie-bit, line 83: Based on the [trainer](https://github.com/kubeflow/training-operator/tree/master/sdk/python/kubeflow/trainer) in the Training Operator, I updated the Python script to properly process the parameters and obtain metrics for hyperparameter optimization. Therefore, we need to build a new image for `TRAINER_TRANSFORMER_IMAGE` based on the updated code for the `tune` API. Currently, this is left blank. @andreyvelich Can you guide me on how to push the image to Kubeflow's repository?

sdk/python/v1beta1/kubeflow/trainer/hf_llm_optimization.py
```diff
@@ -0,0 +1,196 @@
...
+        logger.info("-" * 40)
+        trainer.data_collator = DataCollatorForLanguageModeling(
+            tokenizer,
+            pad_to_multiple_of=8,
+            mlm=False,
+        )
+
+    # Train the model.
+    train_results = trainer.train()
+    print(f"train_loss={train_results.training_loss}")
```
helenxie-bit, line 138: To be implemented: Currently, it only supports optimizing hyperparameters based on `train_loss`. Additional metrics need to be implemented so that hyperparameters can be optimized according to the `objective_metric_name` and `additional_metric_names` defined by the user.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -176,23 +187,39 @@ def tune(
         packages_to_install: List[str] = None,
```
helenxie-bit, line 187: To be implemented: Update `resources_per_trial` to enable PyTorch distributed training.

helenxie-bit:
For the test example, please refer to [this example](https://github.com/kubeflow/katib/blob/1526a9d7846afdc4bb7b52c0c126d487add9b1fe/docs/proposals/llm-hyperparameter-optimization-api.md#example) in the proposal.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -188,23 +199,39 @@ def tune(
...
+        - Custom objective function: Specify the `objective` parameter to define your own objective function. The `base_image` parameter will be used to execute the objective function. `trainer_parameters` should be a dictionary to define the search space for these parameters.
 
         Args:
             name: Name for the Experiment.
+            model_provider_parameters: Parameters for the model provider in the Storage Initializer.
+                For example, HuggingFace model name and Transformer type for that model, like: AutoModelForSequenceClassification. This argument must be the type of `kubeflow.storage_initializer.hugging_face.HuggingFaceModelParams`.
```
nsingl00, line 209: Can we add the example for S3 as well here?

helenxie-bit: Thank you for your suggestions! Do you mean add a use example of S3?

andreyvelich:
Ref: https://github.com/kubeflow/katib/issues/2339

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -164,9 +168,16 @@ def tune(
...
-        objective: Callable,
-        parameters: Dict[str, Any],
-        base_image: str = constants.BASE_IMAGE_TENSORFLOW,
+        model_provider_parameters: Optional[Any] = None,
+		dataset_provider_parameters: Optional[Any] = None,
+		storage_config: Optional[Dict[str, Optional[Union[str, List[str]]]]] = {
+            "size": constants.PVC_DEFAULT_SIZE,
+            "storage_class": None,
+            "access_modes": constants.PVC_DEFAULT_ACCESS_MODES,
+        },
```
andreyvelich, line 177: Please can you format your code with `black` ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -164,9 +168,16 @@ def tune(
         self,
         # TODO (andreyvelich): How to be consistent with other APIs (name) ?
         name: str,
-        objective: Callable,
-        parameters: Dict[str, Any],
-        base_image: str = constants.BASE_IMAGE_TENSORFLOW,
+        model_provider_parameters: Optional[Any] = None,
+		dataset_provider_parameters: Optional[Any] = None,
```
andreyvelich, line 172: Since we only can set the specific types for model and dataset providers should we have one of Union type here ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: We don't support S3 as model provider today.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -164,9 +168,16 @@ def tune(
         self,
         # TODO (andreyvelich): How to be consistent with other APIs (name) ?
         name: str,
-        objective: Callable,
-        parameters: Dict[str, Any],
```
andreyvelich, line 168: I think, we should keep `parameters` for now. So user will have 2 options:
- Set `objective + parameters + base_image` or
- Set `model_provider_parameters + dataset_provider_parameters + trainer_parameters`.

I think, migration of `parameter` to something else, should be out of scope of this change.

WDYT @helenxie-bit @johnugeorge @tenzen-y @deepanker13 ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -259,6 +286,24 @@ def tune(
             RuntimeError: Failed to create Katib Experiment.
         """
 
+        print(
+            "Thank you for using `tune` API for LLMs hyperparameters optimization. This feature is in alpha stage Kubeflow community is looking for your feedback. Please share your experience via #kubeflow-katib Slack channel or Kubeflow Katib GitHub."
+        )
+        
+        if (
+            ((model_provider_parameters is not None) and (dataset_provider_parameters is not None)) == (objective is not None)
+        ):
```
andreyvelich, line 295: I think, we should validate `base_image` also.
What do you think about this:
```py
 if
 (model_provider_parameters is not None or dataset_provider_parameters is not None or trainer_parameters is not None) and
 (objective is not None or base_image is not None or parameters is not None):
```

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -411,9 +450,173 @@ def tune(
...
+                    HuggingFaceModelParams
+                from kubeflow.storage_initializer.hugging_face import \
+                    HuggingFaceTrainerParams
+                from kubeflow.storage_initializer.s3 import S3DatasetParams
+                import peft
+                import transformers
+            except ImportError:
+                raise ImportError(
+                    "Tune API dependencies not installed. "
+                    + "Run: pip install -U 'kubeflow-training[huggingface]' "
```
andreyvelich, line 479: Should it be `pip install -U 'kubeflow-katib[huggingface]'`

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -411,9 +450,173 @@ def tune(
...
+                )
+            
+            # Create PVC for the Storage Initializer.
```
andreyvelich, line 482: Please add TODO that this should be moved to the Controller.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -411,9 +450,173 @@ def tune(
...
+            try:
+                self.core_api.create_namespaced_persistent_volume_claim(
+                    namespace=namespace,
+                    body=utils.get_pvc_spec(
+                        pvc_name=constants.STORAGE_INITIALIZER,
```
andreyvelich, line 487: Should we just have pvc_name == experiment_name, in that case every Katib Experiment will have a separate PVC.
I know that, we don't do this for Training Operator, but we should fix it there as well.
WDYT @johnugeorge @helenxie-bit @deepanker13 ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -411,9 +450,173 @@ def tune(
...
+                metadata=models.V1ObjectMeta(
+                    annotations={"sidecar.istio.io/inject": "false"}
+                ),
+                spec=client.V1PodSpec(
+                    restart_policy="Never",
+                    containers=[container_spec],
+                    init_containers=[init_container_spec],
+                    volumes=[constants.STORAGE_INITIALIZER_VOLUME],
+                ),
+            )
```
andreyvelich, line 612: If we can, we should create helper function and re-use it in Training Operator + Katib.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -411,9 +450,173 @@ def tune(
...
+
+        # Create Trial specification.
+        trial_spec = client.V1Job(
```
andreyvelich, line 615: This should be done only when `tune` is used with objective.

sdk/python/v1beta1/kubeflow/katib/constants/constants.py
```diff
@@ -60,3 +62,30 @@
...
+    name=STORAGE_INITIALIZER,
+    mount_path=INIT_CONTAINER_MOUNT_PATH,
+)
+
+STORAGE_INITIALIZER_VOLUME = client.V1Volume(
+    name=STORAGE_INITIALIZER,
+    persistent_volume_claim=client.V1PersistentVolumeClaimVolumeSource(
+        claim_name=STORAGE_INITIALIZER
+    ),
+)
```
andreyvelich, line 87: we should re-use it from Training SDK

sdk/python/v1beta1/setup.py
```diff
@@ -69,4 +69,7 @@
         "Topic :: Software Development :: Libraries :: Python Modules",
     ],
     install_requires=REQUIRES,
+    extras_require={
+		"huggingface": ["kubeflow-training[huggingface]"],
```
shivaylamba, line 73: shouldn't it be kubeflow-katib?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: I currently use `trainer_parameters` for both options to avoid user confusion about which one to use for each option. I think it also makes sense if we keep `parameters` and make explanations about that.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich Yes, I agree. I think we need to validate two things:
1) Ensure that only one option is chosen.
2) Verify that all necessary parameters are provided for the selected option.

We can revise the validation for the first point according to your suggestions and add validation for the second point under each option. Does this sound okay to you?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Thank you for pointing it out! I will fix that.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Yes, I think it will be better. I will adjust that.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Yes, of course. I will do that after I made all the changes.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Got it! Already added.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I think it will be better than using a constant pvc name. I will make adjustment here. Should I also fix it in Training Operator?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I think we can reuse [get_container_spec](https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/utils/utils.py#L186) and [get_pod_template_spec](https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/utils/utils.py#L236) from the Training Operator. However, I have a small question. In Katib's PodSpec, we set the `restart_policy` to "Never," but in the Training Operator, we don't set it explicitly, so the default value is "Always." Can we maintain consistency between Katib and the Training Operator, or is there a specific reason for this difference?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich Sorry, I didn't understand the difference. Since the `V1beta1TrialTemplate` requires a `trial_spec`, if we only provide this when `tune` is used with an objective, what should we do when `tune` is used with external models and datasets?

sdk/python/v1beta1/kubeflow/katib/constants/constants.py
helenxie-bit: Got it! I have also checked other constants from the Training Operator and reused them from the Training SDK.

sdk/python/v1beta1/setup.py
helenxie-bit: I think it needs to be "kubeflow-training[huggingface]" because we need to reuse assets, such as the `storage-initializer`, from the Training SDK.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: I'm not sure about this. @andreyvelich Do you have any idea about that?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: As we discussed on the call, we will keep `parameters` to avoid breaking changes for `tune` API in this PR.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Yes, please submit PR to change it in the Training Operator as well!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: I think, `restart_policy=Never` is the only requirements for Job: https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-template, not for `PyTorchJob`

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: >what should we do when tune is used with external models and datasets?

In that case, Trial spec is a PyTorchJob, like in this example: https://github.com/kubeflow/katib/blob/master/examples/v1beta1/kubeflow-training-operator/pytorchjob-mnist.yaml

sdk/python/v1beta1/setup.py
andreyvelich: Yes, that's right, it should be `kubeflow-training[huggingface]`

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: That makes sense. I keep `parameters` in `tune` API, and when users set a custom objective function, they will use `parameters` to define the search space for hyperparameters.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I have reformatted the code with Black, but I noticed that some comments are too long and do not wrap automatically. Do I need to adjust it too? If so, is there a specific tool I can use for this?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I found that `base_image` is set to "constants.BASE_IMAGE_TENSORFLOW" by default in the original API. Therefore, we cannot validate `base_image` here unless we set its default value to "None". However, I'm concerned this might cause unexpected changes for existing users. So, I have adjusted the validation here as follows:
```python
if
 (model_provider_parameters is not None or dataset_provider_parameters is not None or trainer_parameters is not None) and
 (objective is not None or parameters is not None):
```
And I have added the validation for `base_image` under the custom objective function option:
```python
if objective is not None:
    if not base_image or not parameters:
        raise ValueError("One of the required parameters is None.")
```
How do you like this?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I have made adjustments here and also submitted a PR to the Training Operator. Please review it when you have time: https://github.com/kubeflow/training-operator/pull/2187

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I have reused `get_container_spec` and `get_pod_template_spec` for the external models and datasets option. To enable this, a few parameters need to be added to these functions. Please review this PR:  https://github.com/kubeflow/training-operator/pull/2197. 
Additionally, I'm considering whether reusing them under custom objective function option to keep consistency. However, since users only need to install `kubeflow-training[huggingface]` when using external models and datasets, it might be better not to reuse them there. What are your thoughts?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I understand. I am considering adding PyTorchJob as an additional option for setting `resources_per_trial` for both the custom objective function and external models and datasets. I plan to implement the PyTorchJob functionality in a separate PR. How does that sound?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Yes, please adjust them to follow the max line width

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Sure, it looks good.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: Yes, let's do that after we support LLM trainer within `tune` API.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Got it! I have reformatted it with max line length of 79.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich Sorry I think I have misunderstood your meaning, so I initially tried to reuse functions from the Training Operator. Now, I've added some helper functions in `utils.py` for reuse, which are based on the helper functions in the Training Operator. Please review.

andreyvelich:
@droctothorpe @Ygnas Please can you help us to see why are we getting the error on the pre-commit check ?
It looks like black conflicts with isort for some reason.

Ygnas:
> @droctothorpe @Ygnas Please can you help us to see why are we getting the error on the pre-commit check ? It looks like black conflicts with isort for some reason.

@andreyvelich  Black compatible configurations says that isort should be set to profile black to make them fully compatible https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort, but that would format them in a different way then google profile that is used at the moment. Not sure if you would be fine switching to the black profile?

droctothorpe:
Looking into this. KFP uses yapf instead of black ([source](https://github.com/kubeflow/pipelines/blob/master/.pre-commit-config.yaml#L35)). Switching to yapf for consistency or adjusting the isort profile to black should fix this issue. Any preference between the two approaches, @andreyvelich?

andreyvelich:
I see, that makes sense, thanks for sharing. We've been using `black` for Katib + Training SDK for quite some time.
Let's try to change the profile from google to black for now. @helenxie-bit It should fix the issue as you said, right ?

In the future, we can migrate to `yapf` if that is more preferable formatter for us.

helenxie-bit:
> I see, that makes sense, thanks for sharing. We've been using `black` for Katib + Training SDK for quite some time. Let's try to change the profile from google to black for now. @helenxie-bit It should fix the issue as you said, right ?
> 
> In the future, we can migrate to `yapf` if that is more preferable formatter for us.

@andreyvelich Yes, changing isort profile to "black" can fix the problem.

andreyvelich:
@helenxie-bit Please can you update the profile to black, so we can merge this PR ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -30,6 +30,12 @@
 
 logger = logging.getLogger(__name__)
 
+if TYPE_CHECKING:
```
andreyvelich, line 33: Why do you want to use TYPE_CHECKING for these checks ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -393,36 +420,268 @@ def tune(
                     )
 
         # Add metrics collector to the Katib Experiment.
-        # Up to now, we only support parameter `kind`, of which default value
-        # is `StdOut`, to specify the kind of metrics collector.
+        # Up to now, We only support parameter `kind`, of which default value is
```
andreyvelich, line 423:
```suggestion
        # Up to now, we only support parameter `kind`, of which default value is
```

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -393,36 +420,268 @@ def tune(
...
+            source=models.V1beta1SourceSpec(
+                filter=models.V1beta1FilterSpec(
+                    metrics_format=[
+                        # For example: train_loss=0.846
+                        r"([\w|-]+)\s*=\s*([+-]?\d*(\.\d+)?([Ee][+-]?\d+)?)",
+                        # For example: 'train_loss':0.846
+                        r"'([\w|-]+)'\s*:\s*([+-]?\d*(\.\d+)?([Ee][+-]?\d+)?)",
+                    ]
+                )
+            ),
```
andreyvelich, line 438: Should we modify the metrics collector format only when user sets trainer_parameters, dataset_parameters, and model_parameters ? We don't want to modify it when user sets `objective`.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -393,36 +420,268 @@ def tune(
...
+                )
+                from kubeflow.training.constants.constants import (
+                    STORAGE_INITIALIZER_VOLUME_MOUNT,
+                )
+                from kubeflow.training.constants.constants import (
+                    TRAINER_TRANSFORMER_IMAGE,
+                )
+            except ImportError:
+                raise ImportError(
+                    "Tune API dependencies not installed. "
```
andreyvelich, line 520:
```suggestion
                    "LLM dependencies for Tune API are not installed. "
```

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -393,36 +420,268 @@ def tune(
...
+                    json.dumps(
+                        model_provider_parameters.__dict__, cls=utils.SetEncoder
+                    ),
+                    "--dataset_provider",
+                    dp,
+                    "--dataset_provider_parameters",
+                    json.dumps(dataset_provider_parameters.__dict__),
+                ],
+                volume_mounts=[STORAGE_INITIALIZER_VOLUME_MOUNT],
+            )
```
andreyvelich, line 639: I thought, we discussed before that we are going to re-use helper functions from Training SDK .
Am I missing something ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -393,36 +420,268 @@ def tune(
...
+
+            # Iterate over input parameters.
+            experiment_params = []
+            trial_params = []
+
+            training_args = trainer_parameters.training_parameters
+            for (
+                p_name,
+                p_value,
+            ) in trainer_parameters.training_parameters.to_dict().items():
```
andreyvelich, line 574: Given that loop for lora_config (e.g. L601) and training_parameters are almost identical, can you create helper function in Katib SDK util to reduce code duplication please ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -393,36 +420,268 @@ def tune(
...
+                ),
+            )
+
+            pod_spec = utils.get_pod_template_spec(
+                containers=[container_spec],
+                init_containers=[init_container_spec],
+                volumes=[storage_initializer_volume],
+                restart_policy="Never",
+            )
+
```
andreyvelich, line 678: We are still creating Job instead of PyTorchJob in the Trial template.
Did you push your latest changes @helenxie-bit ?

helenxie-bit:
> @helenxie-bit Please can you update the profile to black, so we can merge this PR ?

Yes, I will do that.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I attempted to create a PR without adding this, but it failed the flake8 check in pre-commit check because the `HuggingfaceModelParams`, `HuggingFaceDatasetParams`, `S3DatasetParams`, and `HuggingFaceTrainerParams` classes are used in the definition of the `tune` function but were not defined or imported at that point. Therefore, I had to add the TYPE_CHECKING thing to pass the flake8 check. Do you have any alternative suggestions?

Here is the detailed message of the error:
```shell
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:178:45: F821 undefined name 'HuggingFaceModelParams'
sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:180:19: F821 undefined name 'HuggingFaceDatasetParams'
sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:180:47: F821 undefined name 'S3DatasetParams'
sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:182:38: F821 undefined name 'HuggingFaceTrainerParams'
```

helenxie-bit: @andreyvelich Without adding these imports here, the flake8 check fails because the `tune` API references the following undefined names:
```shell
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:178:45: F821 undefined name 'HuggingFaceModelParams'
sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:180:19: F821 undefined name 'HuggingFaceDatasetParams'
sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:180:47: F821 undefined name 'S3DatasetParams'
sdk/python/v1beta1/kubeflow/katib/api/katib_client.py:182:38: F821 undefined name 'HuggingFaceTrainerParams'
```
To resolve this, I used TYPE_CHECKING to avoid actually importing these modules, while still passing the flake8 check, as these modules are only required when users choose to work with external models and datasets. Do you have any alternative suggestions?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Got it! I will change according to your suggestions.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich Apologies for any confusion, and I may have misunderstood your point. Initially, I attempted to reuse helper functions from the Training SDK, but I later realized it might be easier to maintain if we define our own helper functions in Katib, as the parameters for creating containers and pods differ slightly. Do you think it would be better to reuse functions from the Training SDK?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I’ve pushed my changes in a separate PR: https://github.com/kubeflow/katib/pull/2409. I’m not sure I fully understood your suggestions, so could you please refer to this comment: https://github.com/kubeflow/katib/pull/2393#discussion_r1724681935?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: @helenxie-bit Can we follow the same as for Training Operator and remove fmt check for it: https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/api/training_client.py#L170-L185

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: >: 1) use a custom objective function, or 2) use external models and datasets.

@helenxie-bit When user sets external model and dataset, we should use PyTorchJob as Trial template, not Job.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: It looks better! I will modify according to this.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Got it! I will update it as soon as possible.

helenxie-bit:
@andreyvelich I’ve updated the code based on your suggestion. Please take a look when you have time, thanks!

In the meantime, during the pre-commit check, it appears that many unrelated files failed due to the recent update of the isort profile to black. What would be the best way to handle these files?

andreyvelich:
>In the meantime, during the pre-commit check, it appears that many unrelated files failed due to the recent update of the isort profile to black. What would be the best way to handle these files?

Please can you create a separate PR to switch profile to black and run pre-commit to resolve errors ?

Ygnas:
@andreyvelich I can open that PR with extra pkg dir added to the black and flake8, if that would be fine?

andreyvelich:
> @andreyvelich I can open that PR with extra pkg dir added to the black and flake8, if that would be fine?

Sure, that would be great @Ygnas!

Ygnas:
@andreyvelich @helenxie-bit Here it is #2413

helenxie-bit:
@Ygnas Thank you for creating the PR!

@andreyvelich I have rerun all the tests, and it is now ready to merge. Please review it when you have time and let me know if there are any issues. Thank you!

johnugeorge:
Can you resolve the conflict?

helenxie-bit:
> Can you resolve the conflict?

@johnugeorge I have resolved the conflict and please review when you have time. Thank you!

helenxie-bit:
@andreyvelich Thank you for your suggestions! I have removed properties of "TrainerResources" as they are not necessary at the moment. We can add them in future PRs when we need to implement validation logic.

@andreyvelich @johnugeorge Please review this PR when you have time and let me know if there are any further problems. Thank you!

sdk/python/v1beta1/kubeflow/katib/types/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+# flake8: noqa
+
+
+from __future__ import absolute_import
+
+# Import types into type package.
+from kubeflow.katib.types.trainer_resources import TrainerResources
+
+# Import Kubernetes models.
+from kubernetes.client import *
```
andreyvelich, line 12: Do you need this file ?

If users should be able to run this: `from kubeflow.katib import TrainerResources`, just update this script: https://github.com/kubeflow/katib/blob/master/hack/gen-python-sdk/post_gen.py#L43 with this line:
```
 lines.append("from kubeflow.katib.types.trainer_resources import TrainerResources\n")
 ```

Comment (helenxie-bit) [reply sdk/python/v1beta1/kubeflow/katib/types/__init__.py]: Thank you for your suggestions! I've updated the code accordingly.

**Comment by helenxie-bit** [comment]
@andreyvelich When I pushed my latest changes, it was strange that only three of the end-to-end tests succeeded, while the others failed due to a Katib deployment issue. And the main reason seems to be related to the following error:
```shell
error: timed out waiting for the condition on pods/katib-mysql-77b9495867-q6txk
NAME                                 READY   STATUS    RESTARTS      AGE
katib-controller-dbc9cc-bmtf4        1/1     Running   0             2m
katib-db-manager-67b8c998f4-mmljn    1/1     Running   1 (59s ago)   2m
katib-mysql-77b9495867-q6txk         0/1     Pending   0             2m
training-operator-86d756f697-5scdr   1/1     Running   0             2m2s
```
```shell
Containers:
  katib-mysql:
    Image:      mysql:8.0.29
    Port:       3306/TCP
    Host Port:  0/TCP
    Args:
      --datadir
      /var/lib/mysql/datadir
    Liveness:   exec [/bin/bash -c mysqladmin ping -u root -p${MYSQL_ROOT_PASSWORD}] delay=10s timeout=1s period=5s #success=1 #failure=10
    Readiness:  exec [/bin/bash -c mysql -D ${MYSQL_DATABASE} -u root -p${MYSQL_ROOT_PASSWORD} -e 'SELECT 1'] delay=10s timeout=1s period=5s #success=1 #failure=10
    Startup:    exec [/bin/bash -c mysqladmin ping -u root -p${MYSQL_ROOT_PASSWORD}] delay=0s timeout=1s period=15s #success=1 #failure=60
    Environment:
      MYSQL_ROOT_PASSWORD:         <set to the key 'MYSQL_ROOT_PASSWORD' in secret 'katib-mysql-secrets'>  Optional: false
      MYSQL_ALLOW_EMPTY_PASSWORD:  true
      MYSQL_DATABASE:              katib
    Mounts:
      /var/lib/mysql from katib-mysql (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-mgm9j (ro)
Conditions:
  Type           Status
  PodScheduled   False 
Volumes:
  katib-mysql:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  katib-mysql
    ReadOnly:   false
  kube-api-access-mgm9j:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   BestEffort
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type     Reason            Age   From               Message
  ----     ------            ----  ----               -------
  Warning  FailedScheduling  2m    default-scheduler  0/1 nodes are available: pod has unbound immediate PersistentVolumeClaims. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling..
```
I have run my code locally, and everything works fine. I also tried reverting the changes and pushing the code to the CI/CD pipelines from before the latest update, but the same error occurred. I suspect this might be due to resource limitations or a Minikube configuration issue.
Could you please help me look into this problem? I would greatly appreciate it!

andreyvelich:
@tenzen-y Any thoughts ?
@helenxie-bit Can you try to update the minikube version ?
https://github.com/kubeflow/katib/blob/master/.github/workflows/template-setup-e2e-test/action.yaml#L40C13-L40C35

@kubeflow/wg-training-leads Do you remember why we are not using Kind cluster for Katib E2Es ? Is it because we need PVC with ReadWriteMany for PBT Suggestion ? 
https://github.com/kubeflow/training-operator/blob/master/.github/workflows/integration-tests.yaml#L75-L79

helenxie-bit:
> @tenzen-y Any thoughts ? @helenxie-bit Can you try to update the minikube version ? https://github.com/kubeflow/katib/blob/master/.github/workflows/template-setup-e2e-test/action.yaml#L40C13-L40C35
> 
> @kubeflow/wg-training-leads Do you remember why we are not using Kind cluster for Katib E2Es ? Is it because we need PVC with ReadWriteMany for PBT Suggestion ? https://github.com/kubeflow/training-operator/blob/master/.github/workflows/integration-tests.yaml#L75-L79

@andreyvelich I have updated the Minikube version, and it is working well now. Thank you very much! Please review the latest changes when you have time :smiley:

tenzen-y:
> @kubeflow/wg-training-leads Do you remember why we are not using Kind cluster for Katib E2Es ? Is it because we need PVC with ReadWriteMany for PBT Suggestion ?
https://github.com/kubeflow/training-operator/blob/master/.github/workflows/integration-tests.yaml#L75-L79

Yes, the reason is RWX PersistentVolume. KinD does not support RWX PV.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -302,137 +424,347 @@ def tune(
...
+                    kind=metrics_collector_config["kind"]
+                ),
+                source=models.V1beta1SourceSpec(
+                    filter=models.V1beta1FilterSpec(
+                        metrics_format=[
+                            r"'([\w|-]+)'\s*:\s*([+-]?\d*(\.\d+)?([Ee][+-]?\d+)?)",
+                        ]
+                    )
+                ),
+            )
```
andreyvelich, line 759: Since we support only StdOut metrics collector for Trainer right now, I think you can remove this part.

```suggestion
            experiment.spec.metrics_collector_spec = models.V1beta1MetricsCollectorSpec(
                source=models.V1beta1SourceSpec(
                    filter=models.V1beta1FilterSpec(
                        metrics_format=[
                            r"'([\w|-]+)'\s*:\s*([+-]?\d*(\.\d+)?([Ee][+-]?\d+)?)",
                        ]
                    )
                ),
            )
```

sdk/python/v1beta1/setup.py
```diff
@@ -85,4 +85,7 @@
         "Topic :: Software Development :: Libraries :: Python Modules",
     ],
     install_requires=REQUIRES,
+    extras_require={
+        "huggingface": ["kubeflow-training[huggingface]"],
```
andreyvelich, line 89: Should we use the specific version of Training SDK here ?
```suggestion
        "huggingface": ["kubeflow-training[huggingface]==1.8.0"],
```

andreyvelich (commented):
Thanks for this @helenxie-bit!
I left a few comments.
/assign @kubeflow/wg-training-leads @deepanker13

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -162,13 +165,23 @@ def create_experiment(
                     )
                 )
 
+    # fmt: off
```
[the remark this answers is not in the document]
helenxie-bit: @andreyvelich Thank you for your suggestion! I have removed the code that disables formatting for the entire function. 

Is it necessary to disable formatting when checking imports? Since the import statements here have already been adjusted according to the pre-commit requirements, there are no errors here.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -184,29 +197,79 @@ def tune(
         max_trial_count: int = None,
         parallel_trial_count: int = None,
         max_failed_trial_count: int = None,
-        resources_per_trial: Union[dict, client.V1ResourceRequirements, None] = None,
+        resources_per_trial: Union[
+            dict, client.V1ResourceRequirements, TrainerResources, None
+        ] = None,
```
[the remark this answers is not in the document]
helenxie-bit: Thank you for your careful review! I have updated accordingly.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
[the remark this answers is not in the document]
andreyvelich: Then, we should be fine, thank you!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -263,6 +351,40 @@ def tune(
             RuntimeError: Failed to create Katib Experiment.
         """
 
+        print(
+            "Thank you for using `tune` API for LLM hyperparameter optimization. This feature "
+            "is in the alpha stage. Kubeflow community is looking for your feedback. Please "
+            "share your experience via #kubeflow-katib Slack channel or the Kubeflow Katib "
+            "GitHub."
+        )
```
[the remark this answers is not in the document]
helenxie-bit: Thank you for your careful review! I have updated accordingly.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Thank you for your careful review! I have updated accordingly.

sdk/python/v1beta1/setup.py
helenxie-bit: Yeah, I agree it will be better to specify version here. I have updated accordingly.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -302,137 +424,347 @@ def tune(
...
+            # Add objective code to the execute script.
+            exec_script = exec_script.format(objective_code=objective_code)
+
+            # Install Python packages if that is required.
+            if packages_to_install is not None:
+                exec_script = (
+                    utils.get_script_for_python_packages(
+                        packages_to_install, pip_index_url
+                    )
+                    + exec_script
```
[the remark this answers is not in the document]
helenxie-bit: Got it! I agree that it will make the code clearer. I'll implement it as soon as possible.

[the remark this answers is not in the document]
helenxie-bit: @andreyvelich I have added the helper function `get_exec_script_from_objective` in utils.py:
https://github.com/kubeflow/katib/blob/c2df96757ce2fb39b5debf60ad7ed1873ce31ae5/sdk/python/v1beta1/kubeflow/katib/utils/utils.py#L212
And I have updated the `parameter_substitution` function so that it can also be reused with custom objective functions:
https://github.com/kubeflow/katib/blob/c2df96757ce2fb39b5debf60ad7ed1873ce31ae5/sdk/python/v1beta1/kubeflow/katib/utils/utils.py#L146
Please review when you have time and let me know if you have any further suggestions.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: @andreyvelich I have removed this part and moved the following code before the two options to make it clearer (it was previously inside the custom objective function option):
https://github.com/kubeflow/katib/blob/c2df96757ce2fb39b5debf60ad7ed1873ce31ae5/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L418
Do you think this makes sense?

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
```diff
@@ -129,3 +132,132 @@ class FakeResponse:
...
+        if isinstance(obj, type):
+            return obj.__name__
+        return json.JSONEncoder.default(self, obj)
+
+
+def parameter_substitution(
+    parameters: Union[Dict[str, Any], "TrainingArguments", "LoraConfig"],  # noqa: F821
+    experiment_params: List[models.V1beta1ParameterSpec],
+    trial_params: List[models.V1beta1TrialParameterSpec],
+):
```
andreyvelich, line 150: @helenxie-bit Can you also add the return of this function ? 
```
-> Dict[str,str]

File: sdk/python/v1beta1/kubeflow/katib/utils/utils.py [hunk 31]
```diff
@@ -129,3 +132,132 @@ class FakeResponse:
...
+
+    return parameters
+
+
+def get_exec_script_from_objective(
+    objective: Callable,
+    input_params: Dict[str, Any] = None,
+    packages_to_install: Optional[List[str]] = None,
+    pip_index_url: str = "https://pypi.org/simple",
+):
```
andreyvelich, line 217: Similar as above, we should add return here.

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
```diff
@@ -129,3 +132,132 @@ class FakeResponse:
...
+            return obj.__name__
+        return json.JSONEncoder.default(self, obj)
+
+
+def parameter_substitution(
+    parameters: Union[Dict[str, Any], "TrainingArguments", "LoraConfig"],  # noqa: F821
+    experiment_params: List[models.V1beta1ParameterSpec],
+    trial_params: List[models.V1beta1TrialParameterSpec],
+):
+    if isinstance(parameters, dict):
```
andreyvelich, line 151: I think, if this function doesn't share any common code, it makes sense to separate it between 2 helper function.
In that case, you don't need to introduce if/else statement here.
For example, you can create 2 functions:
```
get_trial_substitutions_from_dict()
get_trial_substitutions_from_trainer()
```
WDYT @helenxie-bit ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: I think, that movement is not necessary since you override it for LLM option here: https://github.com/kubeflow/katib/blob/c2df96757ce2fb39b5debf60ad7ed1873ce31ae5/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L697-L705.
So, I think, you can move that code just after this check: https://github.com/kubeflow/katib/blob/c2df96757ce2fb39b5debf60ad7ed1873ce31ae5/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L426

helenxie-bit: Got it! I will move it after this check.

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
helenxie-bit: Got it! I will add it.

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
helenxie-bit: Got it! I will add it.

sdk/python/v1beta1/kubeflow/katib/utils/utils.py
helenxie-bit: Yeah, it makes sense. I will adjust accordingly.

helenxie-bit:
@andreyvelich Thank you for your comments! I have made the updates accordingly. Please review them when you have time.

sdk/python/v1beta1/kubeflow/katib/types/trainer_resources.py
```diff
@@ -0,0 +1,25 @@
...
+        self._num_workers = None
+        self._num_procs_per_worker = None
+        self._resources_per_worker = None
+
+        if num_workers is not None:
+            self.num_workers = num_workers
+        if num_procs_per_worker is not None:
+            self.num_procs_per_worker = num_procs_per_worker
+        if resources_per_worker is not None:
+            self.resources_per_worker = resources_per_worker
```
deepanker13, line 25: Do we need lines from 16 to 25?
can we just do the following without the if conditions
```
self.num_workers = num_workers
self.num_procs_per_worker = num_procs_per_worker
self.resources_per_worker = resources_per_worker
```

andreyvelich (commented):
Thanks for this great contribution @helenxie-bit 🎉 
/lgtm
/assign @deepanker13 @johnugeorge @tenzen-y 
Please take a look at the final changes

sdk/python/v1beta1/kubeflow/katib/types/trainer_resources.py
helenxie-bit: Thank you for your suggestions! I have adjusted accordingly.

sdk/python/v1beta1/kubeflow/katib/types/trainer_resources.py
```diff
@@ -0,0 +1,25 @@
...
+        resources_per_worker=None,
+        local_vars_configuration=None,
+    ):
+        if local_vars_configuration is None:
+            local_vars_configuration = Configuration()
+        self.local_vars_configuration = local_vars_configuration
+
+        self._num_workers = None
+        self._num_procs_per_worker = None
+        self._resources_per_worker = None
```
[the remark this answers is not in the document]
helenxie-bit: Yeah, I have adjusted accordingly. Thank you!

PR #2394, opened 2024-07-24
[GSoC] Compatibility Changes in Trial Controller

Electronic-Waste:

I made some compatibility changes to the Trial Controller. Design details: https://github.com/kubeflow/katib/blob/master/docs/proposals/push-based-metrics-collection.md#compatibility-changes-in-trial-controller

Fixes #

**Checklist:**

### DIFF
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -248,12 +248,11 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1beta1.Trial) error {
 		// If observation is empty, metrics collector doesn't finish.
 		// For early stopping scenario, metrics collector will report logs before Trial status is changed to EarlyStopped.
 		// We need to requeue reconcile when the Trial is succeeded, metrics collector's type is not `Push`, and metrics are not reported.
-		if jobStatus.Condition == trialutil.JobSucceeded && instance.Status.Observation == nil {
-			if instance.Spec.MetricsCollector.Collector == nil ||
-				instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector {
-				logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
-				return errMetricsNotReported
-			}
+		if jobStatus.Condition == trialutil.JobSucceeded && 
+			instance.Status.Observation == nil && 
+			instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector {
+			logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
+			return errMetricsNotReported
 		}
 
 		// Update Trial job status only
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -248,11 +248,12 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1beta1.Trial) error {
 		// If observation is empty, metrics collector doesn't finish.
 		// For early stopping scenario, metrics collector will report logs before Trial status is changed to EarlyStopped.
 		// We need to requeue reconcile when the Trial is succeeded, metrics collector's type is not `Push`, and metrics are not reported.
-		if jobStatus.Condition == trialutil.JobSucceeded &&
-			instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector &&
-			instance.Status.Observation == nil {
-			logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
-			return errMetricsNotReported
+		if jobStatus.Condition == trialutil.JobSucceeded && instance.Status.Observation == nil {
+			if instance.Spec.MetricsCollector.Collector == nil ||
+				instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector {
+				logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
+				return errMetricsNotReported
+			}
 		}
 
 		// Update Trial job status only
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -58,8 +58,8 @@ var (
 	log = logf.Log.WithName(ControllerName)
 	// errMetricsNotReported is the error when Trial job is succeeded but metrics are not reported yet
 	errMetricsNotReported = fmt.Errorf("metrics are not reported yet")
-	// errReportMetricsFailed is the error when metrics are reported but not successful yet
-	errReportMetricsFailed = fmt.Errorf("failed to report metrics")
+	// errReportMetricsFailed is the error when `unavailable` metrics value can't be inserted to the Katib DB.
+	errReportMetricsFailed = fmt.Errorf("failed to report unavailable metrics")
 )
 
 // Add creates a new Trial Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -46,6 +46,7 @@ import (
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/consts"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/trial/managerclient"
 	trialutil "github.com/kubeflow/katib/pkg/controller.v1beta1/trial/util"
+	commonapiv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/common/v1beta1"
 )
 
 const (
@@ -244,9 +245,12 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1beta1.Trial) error {
 			}
 		}
 
-		// If observation is empty metrics collector doesn't finish.
-		// For early stopping metrics collector are reported logs before Trial status is changed to EarlyStopped.
-		if jobStatus.Condition == trialutil.JobSucceeded && instance.Status.Observation == nil {
+		// If observation is empty, metrics collector doesn't finish.
+		// For early stopping scenario, metrics collector will report logs before Trial status is changed to EarlyStopped.
+		// We need to requeue reconcile when the Trial is succeeded, metrics collector's type is not `Push`, and metrics are not reported.
+		if jobStatus.Condition == trialutil.JobSucceeded && 
+			instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector && 
+			instance.Status.Observation == nil {
 			logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
 			return errMetricsNotReported
 		}
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -58,6 +58,8 @@ var (
 	log = logf.Log.WithName(ControllerName)
 	// errMetricsNotReported is the error when Trial job is succeeded but metrics are not reported yet
 	errMetricsNotReported = fmt.Errorf("metrics are not reported yet")
+	// errReportMetricsFailed is the error when metrics are reported but not successful yet
+	errReportMetricsFailed = fmt.Errorf("failed to report metrics")
 )
 
 // Add creates a new Trial Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
@@ -180,7 +182,7 @@ func (r *ReconcileTrial) Reconcile(ctx context.Context, request reconcile.Reques
 	} else {
 		err := r.reconcileTrial(instance)
 		if err != nil {
-			if err == errMetricsNotReported {
+			if err == errMetricsNotReported || err == errReportMetricsFailed {
 				return reconcile.Result{
 					RequeueAfter: time.Second * 1,
 				}, nil
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -248,8 +248,8 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1beta1.Trial) error {
 		// If observation is empty, metrics collector doesn't finish.
 		// For early stopping scenario, metrics collector will report logs before Trial status is changed to EarlyStopped.
 		// We need to requeue reconcile when the Trial is succeeded, metrics collector's type is not `Push`, and metrics are not reported.
-		if jobStatus.Condition == trialutil.JobSucceeded && 
-			instance.Status.Observation == nil && 
+		if jobStatus.Condition == trialutil.JobSucceeded &&
+			instance.Status.Observation == nil &&
 			instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector {
 			logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
 			return errMetricsNotReported
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -259,7 +259,7 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1beta1.Trial) error {
 		//    if job has succeeded and if observation field is available.
 		//    if job has failed
 		// This will ensure that trial is set to be complete only if metric is collected at least once
-		r.UpdateTrialStatusCondition(instance, deployedJob.GetName(), jobStatus)
+		return r.UpdateTrialStatusCondition(instance, deployedJob.GetName(), jobStatus)
 	}
 	return nil
 }
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -18,13 +18,14 @@ package trial
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/spf13/viper"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -153,7 +154,7 @@ func (r *ReconcileTrial) Reconcile(ctx context.Context, request reconcile.Reques
 	original := &trialsv1beta1.Trial{}
 	err := r.Get(ctx, request.NamespacedName, original)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			// Object not found, return.  Created objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers.
 			return reconcile.Result{}, nil
@@ -277,7 +278,7 @@ func (r *ReconcileTrial) reconcileJob(instance *trialsv1beta1.Trial, desiredJob 
 	deployedJob.SetGroupVersionKind(gvk)
 	err = r.Get(context.TODO(), types.NamespacedName{Name: desiredJob.GetName(), Namespace: desiredJob.GetNamespace()}, deployedJob)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			if instance.IsCompleted() {
 				return nil, nil
 			}
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -42,11 +42,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	commonapiv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/common/v1beta1"
 	trialsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/trials/v1beta1"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/consts"
 	"github.com/kubeflow/katib/pkg/controller.v1beta1/trial/managerclient"
 	trialutil "github.com/kubeflow/katib/pkg/controller.v1beta1/trial/util"
-	commonapiv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/common/v1beta1"
 )
 
 const (
@@ -248,8 +248,8 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1beta1.Trial) error {
 		// If observation is empty, metrics collector doesn't finish.
 		// For early stopping scenario, metrics collector will report logs before Trial status is changed to EarlyStopped.
 		// We need to requeue reconcile when the Trial is succeeded, metrics collector's type is not `Push`, and metrics are not reported.
-		if jobStatus.Condition == trialutil.JobSucceeded && 
-			instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector && 
+		if jobStatus.Condition == trialutil.JobSucceeded &&
+			instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector &&
 			instance.Status.Observation == nil {
 			logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
 			return errMetricsNotReported
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -182,7 +182,7 @@ func (r *ReconcileTrial) Reconcile(ctx context.Context, request reconcile.Reques
 	} else {
 		err := r.reconcileTrial(instance)
 		if err != nil {
-			if err == errMetricsNotReported || err == errReportMetricsFailed {
+			if errors.Is(err, errMetricsNotReported) || errors.Is(err, errReportMetricsFailed) {
 				return reconcile.Result{
 					RequeueAfter: time.Second * 1,
 				}, nil
--- a/pkg/controller.v1beta1/trial/trial_controller_test.go
+++ b/pkg/controller.v1beta1/trial/trial_controller_test.go
@@ -301,16 +301,17 @@ func TestReconcileBatchJob(t *testing.T) {
 
 		// Expect that Trial status is succeeded with "false" status and "metrics unavailable" reason.
 		// Metrics unavailable because GetTrialObservationLog returns "unavailable".
-		g.Eventually(func() bool {
-			if err = c.Get(ctx, trialKey, trial); err != nil {
-				return false
-			}
-			return trial.IsMetricsUnavailable() &&
-				len(trial.Status.Observation.Metrics) > 0 &&
-				trial.Status.Observation.Metrics[0].Min == consts.UnavailableMetricValue &&
-				trial.Status.Observation.Metrics[0].Max == consts.UnavailableMetricValue &&
-				trial.Status.Observation.Metrics[0].Latest == consts.UnavailableMetricValue
-		}, timeout).Should(gomega.BeTrue())
+		g.Eventually(func(g gomega.Gomega) {
+			g.Expect(c.Get(ctx, trialKey, trial)).Should(gomega.Succeed())
+			g.Expect(trial.IsMetricsUnavailable()).Should(gomega.BeTrue())
+			g.Expect(trial.Status.Observation.Metrics).ShouldNot(gomega.HaveLen(0))
+			g.Expect(trial.Status.Observation.Metrics[0]).Should(gomega.BeComparableTo(commonv1beta1.Metric{
+			  Name: objectiveMetric,
+			  Min: consts.UnavailableMetricValue,
+			  Max: consts.UnavailableMetricValue,
+			  Latest: consts.UnavailableMetricValue,
+			}))
+		}, timeout).Should(gomega.Succeed())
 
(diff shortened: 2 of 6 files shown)


Electronic-Waste:
@andreyvelich @johnugeorge PTAL👀 if you are available. Thanks!

ref issue: #2340

pkg/controller.v1beta1/trial/trial_controller.go
```diff
@@ -244,11 +245,15 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1beta1.Trial) error {
...
 
-		// If observation is empty metrics collector doesn't finish.
-		// For early stopping metrics collector are reported logs before Trial status is changed to EarlyStopped.
+		// If observation is empty, metrics collector doesn't finish.
+		// For early stopping scenario, metrics collector will report logs before Trial status is changed to EarlyStopped.
+		// We need to requeue reconcile when the Trial is succeeded, metrics collector's type is not `Push`, and metrics are not reported.
 		if jobStatus.Condition == trialutil.JobSucceeded && instance.Status.Observation == nil {
-			logger.Info("Trial job is succeeded but metrics are not reported, reconcile requeued")
-			return errMetricsNotReported
+			if instance.Spec.MetricsCollector.Collector == nil ||
```
andreyvelich, line 252: Do we have a use-case when collector is nil ? I think, we add the StdOut as a default MC: https://github.com/kubeflow/katib/blob/4b477e82d56433389a70da2ffc54088ebc9c449d/pkg/apis/controller/experiments/v1beta1/experiment_defaults.go#L131

pkg/controller.v1beta1/trial/trial_controller.go
```diff
@@ -244,11 +245,15 @@ func (r *ReconcileTrial) reconcileTrial(instance *trialsv1beta1.Trial) error {
...
+				instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector {
```
andreyvelich, line 253: Can you just add this condition to the L251 ? E.g.
```golang
if jobStatus.Condition == trialutil.JobSucceeded && instance.Status.Observation == nil && instance.Spec.MetricsCollector.Collector.Kind != commonapiv1beta1.PushCollector
```

pkg/controller.v1beta1/trial/trial_controller_util.go
```diff
@@ -70,6 +70,14 @@ func (r *ReconcileTrial) UpdateTrialStatusCondition(instance *trialsv1beta1.Tria
 			msg := "Metrics are not available"
 			reason := TrialMetricsUnavailableReason
 
+			// If the type of metrics collector is Push, We should insert an unavailable value to Katib DB
+			if instance.Spec.MetricsCollector.Collector != nil &&
```
andreyvelich, line 74: This might not be needed.

andreyvelich, line 74: We might need to add an integration test for it: https://github.com/kubeflow/katib/blob/4b477e82d56433389a70da2ffc54088ebc9c449d/pkg/controller.v1beta1/trial/trial_controller_test.go
cc @tenzen-y

andreyvelich: Additionally, please try to run the example with Push collector when you don't insert objective metrics and verify that Trial has been moved into `MetricsUnavailable` status.

andreyvelich (commented):
Thank you for this @Electronic-Waste!
cc @kubeflow/wg-automl-leads

pkg/controller.v1beta1/trial/trial_controller.go
Electronic-Waste: @andreyvelich It seems that we pass a nil collector in the testcase: https://github.com/kubeflow/katib/blob/51b246fa1c23db130c373fa9749897a0d0b334ab/pkg/controller.v1beta1/trial/trial_controller_test.go#L102-L124

And I got seg fault in CI/CD when I did not add nil condition judgement: 

截屏2024-08-02 14 59 35

pkg/controller.v1beta1/trial/trial_controller_util.go
Electronic-Waste: > We might need to add an integration test for it: https://github.com/kubeflow/katib/blob/4b477e82d56433389a70da2ffc54088ebc9c449d/pkg/controller.v1beta1/trial/trial_controller_test.go
cc @tenzen-y

@andreyvelich If I understand correctly, do you mean we need to add an integration test for Push Collector?

> Additionally, please try to run the example with Push collector when you don't insert objective metrics and verify that Trial has been moved into MetricsUnavailable status.

SGTM.

Electronic-Waste: @andreyvelich I tried to run the example with `report_metrics` commented. And the result shows that all the Trials have been moved into `MetricsUnavailable` status.

截屏2024-08-02 16 36 35

andreyvelich: >@andreyvelich If I understand correctly, do you mean we need to add an integration test for Push Collector?

Yes, just to verify that appropriate functional calls are happening.

pkg/controller.v1beta1/trial/trial_controller.go
Electronic-Waste: I'll investigate into this question and fix CI/CD.

pkg/controller.v1beta1/trial/trial_controller.go
Electronic-Waste: Okay for that.

pkg/controller.v1beta1/trial/trial_controller_util.go
Electronic-Waste: Okay, I'll implement that in the next few commits.

pkg/controller.v1beta1/trial/trial_controller.go
Electronic-Waste: @andreyvelich I find the nil condition occurred due to: https://github.com/kubeflow/katib/blob/51b246fa1c23db130c373fa9749897a0d0b334ab/pkg/controller.v1beta1/trial/trial_controller_test.go#L425-L445

which omitted the `MetricsCollector` field when creating Trials. I have added it now.

pkg/controller.v1beta1/trial/trial_controller_util.go
Electronic-Waste: The integration test for Push Collector has been added now. Since there are no differences for MCs in successful scenario, I only add a test case in which the batch job is completed but no metrics are reported to the Katib DB.

Does it meet our need? WDYT👀 @andreyvelich

pkg/controller.v1beta1/trial/trial_controller.go
andreyvelich: Amazing, thank you!

pkg/controller.v1beta1/trial/trial_controller_util.go
andreyvelich: Yes, I think, it should be good. Thank you!

pkg/controller.v1beta1/trial/trial_controller_util.go
```diff
@@ -162,6 +169,23 @@ func (r *ReconcileTrial) updateFinalizers(instance *trialsv1beta1.Trial, finaliz
 	}
 }
 
+func (r *ReconcileTrial) reportUnavailableMetrics(instance *trialsv1beta1.Trial) error {
```
andreyvelich, line 172: Please can you verify if user doesn't report any results for `additionalMetricNames` it inserts the `unavailable` metrics to the observation.
For the pull-based metrics, we are doing this as part of this script: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_util.go#L170-L176.
So the observation (objective metric is `result`) might look as follows:
```
  observation:
    metrics:
    - latest: "123"
      max: "123"
      min: "123"
      name: result
    - latest: unavailable
      max: unavailable
      min: unavailable
      name: metrics1
    - latest: unavailable
      max: unavailable
      min: unavailable
      name: metrics2
```

Electronic-Waste: Okay, I'll investigate into this.

tenzen-y:
> /assign @johnugeorge @tenzen-y Please review it when you have time.

ACK

pkg/controller.v1beta1/trial/trial_controller_util.go
Electronic-Waste: @andreyvelich  The result shows that Katib will insert the `unavailable` metrics into the observation when the user doesn't report any results for `additionalMetricsNames`. This logic is not implemented by Push Collector.

截屏2024-08-15 17 40 51

截屏2024-08-15 17 41 15

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -290,14 +290,47 @@ func TestReconcileBatchJob(t *testing.T) {
...
+		g.Eventually(func() bool {
+			return errors.IsNotFound(c.Get(ctx, trialKey, &trialsv1beta1.Trial{}))
+		}, timeout).Should(gomega.BeTrue())
+	})
+
+	t.Run(`Trail with "Complete" BatchJob and Unavailable metrics(Push MC).`, func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
+		gomock.InOrder(
+			mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogUnavailable, nil).MinTimes(1),
+			mockManagerClient.EXPECT().ReportTrialObservationLog(gomock.Any(), gomock.Any()).Return(nil, nil).MinTimes(1),
```
tenzen-y, line 329: What happens if this reporting fails?

tenzen-y (commented):
Basically, lgtm

pkg/controller.v1beta1/trial/trial_controller_util.go
```diff
@@ -70,6 +70,13 @@ func (r *ReconcileTrial) UpdateTrialStatusCondition(instance *trialsv1beta1.Tria
 			msg := "Metrics are not available"
 			reason := TrialMetricsUnavailableReason
 
+			// If the type of metrics collector is Push, We should insert an unavailable value to Katib DB
+			if instance.Spec.MetricsCollector.Collector.Kind == commonv1beta1.PushCollector {
+				if err := r.reportUnavailableMetrics(instance); err != nil {
+					logger.Error(err, "Failed to insert unavailable value to Katib DB")
```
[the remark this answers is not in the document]
Electronic-Waste: Yeah, I agree with you since retry reconciliation is better for inserting metrics into Katib DB. I'll modify the code.

pkg/controller.v1beta1/trial/trial_controller_test.go
Electronic-Waste: It will fail to insert `Unavailable` to Katib DB. I haven't considered this scenario. Do you mean I need to add another UT for the failed case?

andreyvelich: If we can the unit test for it, that would be great @Electronic-Waste!

Electronic-Waste: @andreyvelich I'll add the UT when I have the bandwidth this two weeks.

pkg/controller.v1beta1/trial/trial_controller_util.go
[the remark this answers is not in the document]
Electronic-Waste: @tenzen-y The retrying reconciliation logic has been added now. PTAL👀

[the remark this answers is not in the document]
Electronic-Waste: @andreyvelich I've added `errReportMetricsFailed` to describe the error caused by failing to report unavailable metrics to Katib DB as you suggested in the meeting today. Please let me know if there are any changes I need to make further.

pkg/controller.v1beta1/trial/trial_controller.go
```diff
@@ -57,6 +58,8 @@ var (
 	log = logf.Log.WithName(ControllerName)
 	// errMetricsNotReported is the error when Trial job is succeeded but metrics are not reported yet
 	errMetricsNotReported = fmt.Errorf("metrics are not reported yet")
+	// errReportMetricsFailed is the error when metrics are reported but not successful yet
+	errReportMetricsFailed = fmt.Errorf("failed to report metrics")
```
andreyvelich, line 62:
```suggestion
	// errReportMetricsFailed is the error when `unavailable` metrics value can't be inserted to the Katib DB.
	errReportMetricsFailed = fmt.Errorf("failed to report unavailable metrics")
```

andreyvelich: @Electronic-Waste Just a small nit from my side and we can merge it.

Electronic-Waste: @andreyvelich I've committed your suggestions :)

andreyvelich:
Thank you for this @Electronic-Waste!
/lgtm
/approve

tenzen-y (commented):
/hold

pkg/controller.v1beta1/trial/trial_controller_util.go
```diff
@@ -70,6 +70,15 @@ func (r *ReconcileTrial) UpdateTrialStatusCondition(instance *trialsv1beta1.Tria
 			msg := "Metrics are not available"
 			reason := TrialMetricsUnavailableReason
 
+			// If the type of metrics collector is Push, We should insert an unavailable value to Katib DB.
+			// We would retry reconciliation if some error occurs while we report unavailable metrics.
+			if instance.Spec.MetricsCollector.Collector.Kind == commonv1beta1.PushCollector {
+				if err := r.reportUnavailableMetrics(instance); err != nil {
+					logger.Error(err, "Failed to insert unavailable value to Katib DB")
+					return errReportMetricsFailed
```
tenzen-y, line 78:
```suggestion
					return fmt.Errorf("%w: %w", errReportMetricsFailed, err)
```

We should wrap the error.

pkg/controller.v1beta1/trial/trial_controller.go
```diff
@@ -179,7 +182,7 @@ func (r *ReconcileTrial) Reconcile(ctx context.Context, request reconcile.Reques
 	} else {
 		err := r.reconcileTrial(instance)
 		if err != nil {
-			if err == errMetricsNotReported {
+			if err == errMetricsNotReported || err == errReportMetricsFailed {
```
tenzen-y, line 185:
```suggestion
			if errors.Is(err, errMetricsNotReported) || errors.Is(err, errReportMetricsFailed) {
```

We should not directly compare errors.

pkg/controller.v1beta1/trial/trial_controller_util.go
Electronic-Waste: LGTM

pkg/controller.v1beta1/trial/trial_controller.go
Electronic-Waste: Yeah, I agree with you.

Electronic-Waste:
@tenzen-y I've updated the code as you suggested. Please let me know of any modifications I need to make :)

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -290,14 +290,82 @@ func TestReconcileBatchJob(t *testing.T) {
...
+		// Delete the Trial
+		g.Expect(c.Delete(ctx, trial)).NotTo(gomega.HaveOccurred())
+
+		// Expect that Trial is deleted
+		g.Eventually(func() bool {
+			return errors.IsNotFound(c.Get(ctx, trialKey, &trialsv1beta1.Trial{}))
+		}, timeout).Should(gomega.BeTrue())
+	})
+
+	t.Run(`Trail with "Complete" BatchJob and Unavailable metrics(Push MC).`, func(t *testing.T) {
```
tenzen-y, line 325: https://github.com/kubeflow/katib/actions/runs/10598019897/job/29371352434?pr=2394#step:4:132

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -290,14 +290,82 @@ func TestReconcileBatchJob(t *testing.T) {
...
+		// Delete the Trial
+		g.Expect(c.Delete(ctx, trial)).NotTo(gomega.HaveOccurred())
+
+		// Expect that Trial is deleted
+		g.Eventually(func() bool {
+			return errors.IsNotFound(c.Get(ctx, trialKey, &trialsv1beta1.Trial{}))
+		}, timeout).Should(gomega.BeTrue())
+	})
+
+	t.Run(`Trail with "Complete" BatchJob and Unavailable metrics(Push MC, failed once).`, func(t *testing.T) {
```
tenzen-y, line 358: https://github.com/kubeflow/katib/actions/runs/10598019897/job/29371352434?pr=2394#step:4:138

tenzen-y (commented):
Some test cases are so flaky.
@Electronic-Waste Could you investigate them?

Electronic-Waste:
@tenzen-y Okay,I'll investigate it.

Electronic-Waste:
@tenzen-y @andreyvelich  I think it's difficult for me to figure out how to solve this error.

By adding logs to the test file, I found that the error occurred because the reconciliation is sometimes not triggered when `c.Delete` is called, thus giving rise to the failure to remove finalizers. As a result, the former Trial can't be deleted correctly and the next UT will report `object is being deleted: trials.kubeflow.org "test-trial" already exists` while creating a new Trial.

But I don't know why the deletion sometimes works and sometimes not. Could you please help me figure out the reason for it? Or we can open a new issue to discuss this :)

REF: https://github.com/kubeflow/katib/actions/runs/10635118286/job/29484008031?pr=2394
截屏2024-08-30 23 58 32

REF: https://github.com/kubeflow/katib/actions/runs/10634460384/job/29482355521
截屏2024-08-31 00 08 55

andreyvelich:
@Electronic-Waste Do you use different Trial name for each test case ?

Electronic-Waste:
> @Electronic-Waste Do you use different Trial name for each test case ?

@andreyvelich I use the same name for each test case. But I think it makes no difference to use different names because the failure of deleting Trial is due to the missing of reconciliation.

Electronic-Waste:
@tenzen-y @andreyvelich I updated the order of UT and it just turned out to be useful to resolve the flasky error.

I don't know why but it just works🤣.

andreyvelich (commented):
I think we can merge it.
/lgtm
/assign @tenzen-y 
Please take a look if that looks good to you.

tenzen-y:
> @tenzen-y @andreyvelich I updated the order of UT and it just turned out to be useful to resolve the flasky error.
> 
> I don't know why but it just works🤣. Maybe the error results from some tricky bugs in `envtest`.

@Electronic-Waste It seems that there are still flaky situations in the same test cases: 

- https://github.com/kubeflow/katib/actions/runs/10654478578/job/29576729814?pr=2394#step:4:129
- https://github.com/kubeflow/katib/actions/runs/10654478578/job/29576845852?pr=2394#step:4:127
- https://github.com/kubeflow/katib/actions/runs/10654478578/job/29576845852?pr=2394#step:4:127

1/4 trials succeeded, otherwise failed.

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -290,14 +290,82 @@ func TestReconcileBatchJob(t *testing.T) {
...
 		g := gomega.NewGomegaWithT(t)
 		gomock.InOrder(
 			mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogUnavailable, nil).MinTimes(1),
+			mockManagerClient.EXPECT().ReportTrialObservationLog(gomock.Any(), gomock.Any()).Return(nil, nil).MinTimes(1),
 			mockManagerClient.EXPECT().DeleteTrialObservationLog(gomock.Any()).Return(nil, nil),
 		)
-		// Create the Trial
-		trial := newFakeTrialBatchJob()
+		// Create the Trial with Push MC
+		trial := newFakeTrialBatchJob(commonv1beta1.PushCollector)
```
andreyvelich, line 301: Try to use different Trial name for every test case, maybe it will help.

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -290,14 +290,82 @@ func TestReconcileBatchJob(t *testing.T) {
...
+			return errors.IsNotFound(c.Get(ctx, trialKey, &trialsv1beta1.Trial{}))
+		}, timeout).Should(gomega.BeTrue())
+	})
+
+	t.Run(`Trail with "Complete" BatchJob and Unavailable metrics(Push MC, failed once).`, func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
+		gomock.InOrder(
+			mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogUnavailable, nil).MinTimes(1),
+			mockManagerClient.EXPECT().ReportTrialObservationLog(gomock.Any(), gomock.Any()).Return(nil, errReportMetricsFailed).MinTimes(1),
+			mockManagerClient.EXPECT().GetTrialObservationLog(gomock.Any()).Return(observationLogUnavailable, nil).MinTimes(1),
```
andreyvelich, line 363: Why do you verify that GetTrialObservationLog will be called twice ?

pkg/controller.v1beta1/trial/trial_controller_test.go
Electronic-Waste: Okay, I'll try that.

pkg/controller.v1beta1/trial/trial_controller_test.go
Electronic-Waste: Because Trial Controller will retry reconciliation when the `ReportTrialObservationLog` func fails. And then `GetTrialObservationLog` will be called again. 

I want to verify that `GetTrialObservationLog` is called after `ReportTrialObservationLog` fails.

pkg/controller.v1beta1/trial/trial_controller_test.go
Electronic-Waste: It does not work...

Electronic-Waste:
@andreyvelich @tenzen-y I've fixed the flaky error by separating the UTs and adjusting the `EXPECT()` mock clause!

AFAIK, the flaky error is caused by the uncertain triggering times of the reconciliation, thus giving rise to the uncertainty of the times we call the function. The most annoying issue was that we must call `GetObservationLog` and `ReportObservationLog` in turn and we didn't know the times they were triggered when we use Push MC. 

So I reserved some `EXPECT()` function in the `gomock.InOrder` and added some new `.AnyTimes()` clause outside it. Please let me know if there are any modifications need to be made to this PR. Thanks!

Electronic-Waste:
@kubeflow/wg-automl-leads it seems that the coverage reports have some accidents. Could you please re-rerun these test cases and check them again?

andreyvelich:
@Electronic-Waste I think, you can re-trigger tests by add this comment: `/rerun-all`

Electronic-Waste:
Thanks @andreyvelich
I think what I mean is running those few failed test cases. Sorry for not realizing that I can re-run them all without you. I'll re-rerun them on my own in the future.

andreyvelich:
Not sure why coveralls fails tho.
@Electronic-Waste Please can you try to investigate it ?
cc @kubeflow/wg-training-leads

Electronic-Waste:
@andreyvelich AFAIK It fails sometimes. It may turn normal in a few hours.

Electronic-Waste:
> @andreyvelich There is an issue describing it: https://github.com/lemurheavy/coveralls-public/issues/1716
>
> It seems that the coverage report will fail if we rerun the old CI build: https://github.com/lemurheavy/coveralls-public/issues/1716#issuecomment-1593414065

@tenzen-y  FYR, the coverage report will fail if we rerun the old CI build. 

The way I can think of to examine these is to rerun all CI tests.

Electronic-Waste:
@tenzen-y  The `/rerun-all` command only restarts `Go Test`. We need to restart all CI builds since the bug report is:

```
{"message":"Can't add a job to a build that is already closed. Build 10789114390 is closed. See docs.coveralls.io/parallel-builds","error":true}
```

Electronic-Waste:
I made some tiny changes in the comment lines. It will retrigger all CI builds:)

Maybe this can help you check the robustness of the UTs in `Go Test` @tenzen-y :)

Electronic-Waste:
> @andreyvelich @tenzen-y I've fixed the flaky error by separating the UTs and adjusting the `EXPECT()` mock clause!
> 
> 
> 
> AFAIK, the flaky error is caused by the uncertain triggering times of the reconciliation, thus giving rise to the uncertainty of the times we call the function. The most annoying issue was that we must call `GetObservationLog` and `ReportObservationLog` in turn and we didn't know the times they were triggered when we use Push MC. 
> 
> 
> 
> So I reserved some `EXPECT()` function in the `gomock.InOrder` and added some new `.AnyTimes()` clause outside it. Please let me know if there are any modifications need to be made to this PR. Thanks!

@tenzen-y I'm sure that the flaky error has been addressed now. May I ask whether you need to check it again or not? I can trigger the CI builds again by making more tiny changes since the coverage report only works when we retrigger all CI builds.

tenzen-y:
> > @andreyvelich @tenzen-y I've fixed the flaky error by separating the UTs and adjusting the `EXPECT()` mock clause!
> > AFAIK, the flaky error is caused by the uncertain triggering times of the reconciliation, thus giving rise to the uncertainty of the times we call the function. The most annoying issue was that we must call `GetObservationLog` and `ReportObservationLog` in turn and we didn't know the times they were triggered when we use Push MC.
> > So I reserved some `EXPECT()` function in the `gomock.InOrder` and added some new `.AnyTimes()` clause outside it. Please let me know if there are any modifications need to be made to this PR. Thanks!
> 
> @tenzen-y I'm sure that the flaky error has been addressed now. May I ask whether you need to check it again or not? I can trigger the CI builds again by pushing more tiny changes since the coverage report only works when we retrigger all CI builds.

Thank you for driving this! Throughout this 2 week CI result, I am sure that we succeeded to get rid of flakiness root causes.

https://github.com/kubeflow/katib/actions/workflows/test-go.yaml

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -134,59 +168,21 @@ func TestReconcileBatchJob(t *testing.T) {
 	viper.Set(consts.ConfigTrialResources, trialResources)
 	g.Expect(add(mgr, recFn)).NotTo(gomega.HaveOccurred())
 
-	// Start test manager.
-	wg := &sync.WaitGroup{}
-	wg.Add(1)
+	// Start test manager
+	mgrCtx, cancel := context.WithCancel(context.TODO())
```
tenzen-y, line 172:
```suggestion
	mgrCtx, cancel := context.WithCancel(context.TODO())
	t.Cleanup(cancel)
```

Instead of calling cancel function in line 369, could you enforce to call cancel after this test is finished?

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -326,6 +364,9 @@ func TestReconcileBatchJob(t *testing.T) {
 		g := gomega.NewGomegaWithT(t)
 		g.Expect(r.updateStatus(&trialsv1beta1.Trial{})).To(gomega.HaveOccurred())
 	})
+
+	// Stop the test manager
+	cancel()
```
tenzen-y, line 369:
```suggestion
```

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -290,14 +287,55 @@ func TestReconcileBatchJob(t *testing.T) {
...
+		g.Eventually(func() bool {
+			if err = c.Get(ctx, trialKey, trial); err != nil {
+				return false
+			}
+			return trial.IsMetricsUnavailable() &&
+				len(trial.Status.Observation.Metrics) > 0 &&
+				trial.Status.Observation.Metrics[0].Min == consts.UnavailableMetricValue &&
+				trial.Status.Observation.Metrics[0].Max == consts.UnavailableMetricValue &&
+				trial.Status.Observation.Metrics[0].Latest == consts.UnavailableMetricValue
+		}, timeout).Should(gomega.BeTrue())
```
tenzen-y, line 312:
```suggestion
		g.Eventually(func(g gomega.Gomega) {
			g.Expect(c.Get(ctx, trialKey, trial)).Should(gomega.Succeeded())
			g.Expect(trial.IsMetricsUnavailable()).Should(gomega.BeTrue)
			g.Expect(trial.Status.Observation.Metrics).ShouldNot(gomega.HaveLen(0))
			g.Expect(trial.Status.Observation.Metrics[0]).Should(gomega.BeComparableTo(commonv1beta1.Metric{
			  Name: METRCS_NAME,
			  Min: consts.UnavailableMetricValue,
			  Max: consts.UnavailableMetricValue,
			  Latest: consts.UnavailableMetricValue,
			}))
		}, timeout).Should(gomega.Succeeded())
```

Could you use the propagated gomega instance to improve debuggability?

tenzen-y: I guess you just copy and paste here, but we should select the ideal approach in the new codes.
If you can replace the not ideally codes with ideal forms in followup PR, it would be great.

tenzen-y (commented):
The most things lgtm

pkg/controller.v1beta1/trial/trial_controller_test.go
Electronic-Waste: Sure! I think it looks better to me.

pkg/controller.v1beta1/trial/trial_controller_test.go
Electronic-Waste: Yeah, I think the propagated gomega instance looks better. 
I'll raise a new issue describing it and open a PR to it later :)

tenzen-y (commented):
That was great improvments!
Thank you for doing this!

/lgtm
/approve

Electronic-Waste:
Thank you for your detailed review @tenzen-y!

This PR is holding now. Can you remove the `hold` label so that this PR can be merged?

tenzen-y:
> Thank you for your detailed review @tenzen-y!
> 
> This PR is holding now. Can you remove the `hold` label so that this PR can be merged?

Sure.
/hold cancel

PR #2401, opened 2024-08-01
[Test] Refactor `inject_webhook_test.go` according to the Developer Guide

### LINKED ISSUE #2388: Refactor `/pkg/webhook/v1beta1/pod/inject_webhook_test.go` according to Developer Guide.

Refactor `/pkg/webhook/v1beta1/pod/inject_webhook_test.go` to obey rules defined in the [Developer Guide](https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md).

### Why is this needed?

Currently, testcases in `/pkg/webhook/v1beta1/pod/inject_webhook_test.go` are defined in slices and use `reflect.Equal` for comparison. It's not recommended according to the Developer Guide. We should:

1. Use [cmp.Diff](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) instead of reflect.Equal, to provide useful comparisons.

2. Define test cases as maps instead of slices to avoid dependencies on the running order. Map key should be equal to the test case name.

### Love this feature?

Electronic-Waste:

I refactored `/pkg/webhook/v1beta1/pod/inject_webhook_test.go` with rule described in [Developer Guide](https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md):

1. Use [cmp.Diff](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) instead of reflect.Equal, to provide useful comparisons.
2. Define test cases as maps instead of slices to avoid dependencies on the running order. Map key should be equal to the test case name.

Fixes #2388

**Checklist:**

### DIFF
--- a/pkg/webhook/v1beta1/pod/inject_webhook_test.go
+++ b/pkg/webhook/v1beta1/pod/inject_webhook_test.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
-	"reflect"
 	"sync"
 	"testing"
 	"time"
@@ -32,7 +31,6 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/webhook/v1beta1/pod/inject_webhook_test.go
```diff
@@ -510,23 +496,6 @@ func TestGetMetricsCollectorArgs(t *testing.T) {
...
-			mCSpec: common.MetricsCollectorSpec{
-				Collector: &common.CollectorSpec{
-					Kind: common.StdOutCollector,
-				},
-			},
-			earlyStoppingRules: earlyStoppingRules,
-			katibConfig:        configv1beta1.MetricsCollectorConfig{},
-			name:               "Trial with invalid Experiment label name. Suggestion is not created",
-			err:                true,
 		},
```
tenzen-y, line 499: What is the reason that we need to remove this case?

Electronic-Waste: Sorry,I might delete this test case by accident.

Electronic-Waste:
I've added the missing UT which was deleted by accident now. PTAL👀 @tenzen-y  if you have time.

tenzen-y (commented):
Thanks!
/lgtm
/approve

PR #2407, opened 2024-08-14
Enhance pre-commit hooks with flake8 and black

Ygnas:
This PR implements the flake8 linter and black formatting as a pre-commit hook to ensure code quality and adherence to style guidelines. 

Fixes #

**Checklist:**

**`Not sure if I there is a need to exclude something else in the pre-commit config??.
 If it should be please let me know`**

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+max-line-length = 100
+# E203 is ignored to avoid conflicts with Black's formatting, as it's not PEP 8 compliant
+extend-ignore = W503, E203
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,16 @@ repos:
       - id: isort
         name: isort
         entry: isort --profile google
+  - repo: https://github.com/psf/black
+    rev: 24.2.0
+    hooks:
+      - id: black
+        files: (sdk|examples)/.*
+  - repo: https://github.com/pycqa/flake8
+    rev: 7.1.1
+    hooks:
+      - id: flake8
+        files: (sdk|examples)/.*
 exclude: |
   (?x)^(
     .*zz_generated.deepcopy.*|
--- a/examples/v1beta1/kubeflow-pipelines/mpi-job-horovod.py
+++ b/examples/v1beta1/kubeflow-pipelines/mpi-job-horovod.py
@@ -21,10 +21,12 @@
 
 # This Experiment is similar to this:
 # https://github.com/kubeflow/katib/blob/master/examples/v1beta1/kubeflow-training-operator/mpijob-horovod.yaml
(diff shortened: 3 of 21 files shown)


.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+max-line-length = 100
+extend-ignore = W503, E203
```
tenzen-y, line 3:
```suggestion
extend-ignore = W503
```

Instead of ignoring this error, could you fix those errors?

tenzen-y (commented):
Thanks for doing this. 
Basically, lgtm

.flake8
Ygnas: Sure I will, there is a small mistake in my flake8 fix as well, wrong indentation. Will fix that as well

Ygnas: Black adds whitespace before : in slices, and flake8 complains about it.

Recomended configs for black:
[Compatible configuration files can be found here](https://github.com/psf/black/blob/main/docs/compatible_configs/)

This behaviour may raise E203 whitespace before ':' warnings in style guide enforcement tools like Flake8. Since E203 is not PEP 8 compliant, you should tell Flake8 to ignore these warnings. [More here](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#e203).

WDYT?

tenzen-y: Oh, that makes sense.
Could you add a comment here on why we must ignore E203?

tenzen-y (commented):
Thank you!
/lgtm
/approve

.pre-commit-config.yaml
```diff
@@ -11,6 +11,16 @@ repos:
...
+  - repo: https://github.com/psf/black
+    rev: 24.2.0
+    hooks:
+      - id: black
+        files: (sdk|examples)/.*
+  - repo: https://github.com/pycqa/flake8
+    rev: 7.1.1
+    hooks:
+      - id: flake8
+        files: (sdk|examples)/.*
```
andreyvelich, line 23: @Ygnas Please can create PR to add the `pkg` directory as well ?
We have some Python files also: https://github.com/kubeflow/katib/tree/master/pkg/suggestion/v1beta1, https://github.com/kubeflow/katib/tree/master/pkg/earlystopping/v1beta1/medianstop

tenzen-y: Good point.

helenxie-bit:
@Ygnas Thank you for your contribution! I'm trying to update a PR (https://github.com/kubeflow/katib/pull/2393), and the new pre-commit check is working. However, it looks like there's an issue with the compatibility between the isort and black checks. Specifically, I've run "pre-commit run --all-files" locally, but the isort and black checks keep failing for "katib_client.py". It seems that this is because the pre-commit configuration is using the `--profile google` argument with isort, which has a different style than black. Could you please take a look into this issue?

PR #2251, opened 2024-01-03
Install typing-extensions v4.6.3 for Optuna

tenzen-y:
**What this PR does / why we need it**:
We need to install typing-extensions v4.6.0 or newer to resolve the following CI error:

> E   ImportError: cannot import name 'TypeAliasType' from 'typing_extensions' (/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/typing_extensions.py)

https://github.com/kubeflow/katib/actions/runs/7400396607/job/20133921603?pr=2250#step:4:594

The root cause is that the sqlalchemy on which the optuna depends requires typing-extensions v4.6.0 or newer:

Optuna dependencies: https://github.com/optuna/optuna/blob/v3.3.0/pyproject.toml#L40

sqlalchemy dependencies: https://github.com/sqlalchemy/sqlalchemy/blob/4871a2f72f5f0e514fe81a8caa17a14f9fc51300/setup.cfg#L38-L39

**Which issue(s) this PR fixes** _(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_:
Fixes #

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/components/katib/) included if any changes are user facing

PR #2412, opened 2024-08-21
[GSOC] `hyperopt` suggestion service logic update

### LINKED ISSUE #2374: [GSOC] Project 8: Support various `Parameter Distribution` in Katib
## Motivation
Currently, [Katib](https://github.com/kubeflow/katib) is limited to supporting only uniform distribution for integer, float, and categorical hyperparameters. By introducing additional distributions, Katib will become more flexible and powerful in conducting hyperparameter optimization tasks.
A Data Scientist requires Katib to support multiple hyperparameter distributions, such as log-uniform, normal, and log-normal, in addition to the existing uniform distribution. For instance, learning rates often benefit from a log-uniform distribution because small values can significantly impact performance. Similarly, normal distributions are useful for parameters that are expected to vary around a central value.
## Goal
The project aims to enhance Katib's Experiment APIs by introducing support for various hyperparameter distributions beyond the current uniform distribution. 
Currently, the Experiment APIs support `parameterType` and `feasibleSpace`. To provide greater flexibility, we expanded the `feasibleSpace` to include a `Distribution` field, allowing users to specify the type of distribution for each parameter.

This Distribution field is designed to be mutually exclusive, meaning it can be left empty, enabling users to tune hyperparameters without specifying a distribution if desired.

### Preparation
- [x] Converted the GSoC proposal into a KEP: #2334
### Development
- [x] Added `Distribution` field to the `FeasibleSpace` object: #2377 
- [x] Updated gRPC APIs to include the `Distribution` field: #2397
- [x] Modified Suggestion client to handle the new Distribution field: #2403 
- [x] Updated validation webhook to support both ParameterType and Distribution: #2404 
- [x] Implemented logic updates for handling new distributions in `hyperopt`suggestion service: #2412
- [ ] Implemented logic updates for handling new distributions in `optuna`suggestion service
- [ ] Implemented logic updates for handling new distributions in `goptuna`suggestion service
### Documentation and Ongoing Work
- [x] Write e2e tests for `normal` and `lognormal` distributions in `hyperopt`: #2412 
- [ ] Extend the Python SDK to include support for the `Distribution` field.
- [ ] Update Katib documentation to reflect new API changes and usage guidelines.
- [ ] Create examples demonstrating the use of different parameter distributions.

## What I learnt during the project
- One of the key hurdles was integrating multiple new distributions into the Katib Experiment APIs while ensuring backward compatibility. Modifying the `feasibleSpace` to include a `Distribution` field required careful consideration of API design and how different suggestion services, like Optuna and Hyperopt, would handle these changes.

- Effective communication is key. The weekly syncs with @andreyvelich and @tenzen-y were extremely valuable, and having a clear agenda beforehand helped me get my questions answered efficiently and stay on track.

- I learned the importance of end-to-end (E2E) tests. By implementing these tests, I gained valuable insights into how the different components of Katib interact and how critical it is to verify that new features, like the distribution support.

## In the End
I am grateful to Google for the opportunity to work on this project and to the Kubeflow community for their support and guidance. Special thanks to my project mentors, @andreyvelich, @tenzen-y, and @terrytangyuan  for your invaluable insights and encouragement. Your mentorship was pivotal in achieving the project's goals.

I hope the enhancements made will significantly benefit users and advance Katib’s capabilities in hyperparameter optimization. Looking forward to continue to contribute to Kubeflow.

shashank-iitbhu:

Fixes #2374 

**Checklist:**


pkg/apis/manager/v1beta1/api.proto
```diff
@@ -101,11 +101,11 @@ enum ParameterType {
  * Distribution types for HyperParameter.
  */
 enum Distribution {
-    UNIFORM = 0;
-    LOG_UNIFORM = 1;
-    NORMAL = 2;
-    LOG_NORMAL = 3;
-    DISTRIBUTION_UNKNOWN = 4;
+    DISTRIBUTION_UNKNOWN = 0;
```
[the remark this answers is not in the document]
tenzen-y:
```suggestion
    DISTRIBUTION _UNSPECIFIED = 0;
```

I would like to select the `UNSPECIFIED` suffix here.
Please see: https://google.aip.dev/126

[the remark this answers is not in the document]
andreyvelich: Make sense, @tenzen-y should we rename other gRPC parameters to UNSPECIFIED ?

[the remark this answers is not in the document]
tenzen-y: Changing released gRPC, it indicates losing backward compatibility.
So, I would like to keep using the existing API for released gRPC, WDYT?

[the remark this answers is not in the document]
andreyvelich: Since these gRPC API is not exposed to the end-users, do you still think that we should not change the existing APIs ?
It only affects users who build their own Suggestion service.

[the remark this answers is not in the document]
tenzen-y: > Since these gRPC APIs are not exposed to the end-users, do you still think that we should not change the existing APIs ?

Almost correct. Additionally, when users keep using the removed Suggestion Services like the Chocolate Suggestion, users face the same problem.

So, can we collect feedback on the dedicated issue outside of here?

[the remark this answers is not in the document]
tenzen-y: SGTM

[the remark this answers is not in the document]
shashank-iitbhu: > Sure, let's followup on this in the issue, and rename it after few months if we don't get any feedback. @shashank-iitbhu Please can you create an issue to track it ?

Sure, I will create a separate issue to track the renaming of other gRPC parameters to UNSPECIFIED.

pkg/suggestion/v1beta1/hyperopt/base_service.py
```diff
@@ -63,13 +65,38 @@ def create_hyperopt_domain(self):
         hyperopt_search_space = {}
         for param in self.search_space.params:
             if param.type == INTEGER:
-                hyperopt_search_space[param.name] = hyperopt.hp.quniform(
-                    param.name, float(param.min), float(param.max), float(param.step)
-                )
-            elif param.type == DOUBLE:
-                hyperopt_search_space[param.name] = hyperopt.hp.uniform(
+                hyperopt_search_space[param.name] = hyperopt.hp.uniformint(
```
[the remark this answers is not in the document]
shashank-iitbhu: Distributions like `uniform` `quniform` `loguniform` `normal` etc return float values. They are designed to sample from a range of values that can take any real number (float), which might not make sense if we're looking for an integer value.
Although we can definitely add support for these distributions when parameter is int also. Should we do this?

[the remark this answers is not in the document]
andreyvelich: @tenzen-y @kubeflow/wg-training-leads @shashank-iitbhu Should we round this `float` value to `int` if user wants to use this distribution and int parameter type ?

examples/v1beta1/hp-tuning/hyperopt-distribution.yaml
```diff
@@ -0,0 +1,60 @@
...
+        min: "0.01"
+        max: "0.05"
+        step: "0.01"
+        distribution: "uniform"
+    - name: momentum
+      parameterType: double
+      feasibleSpace:
+        min: "0.5"
+        max: "0.9"
+        distribution: "logUniform"
```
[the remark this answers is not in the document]
shashank-iitbhu: done

pkg/suggestion/v1beta1/hyperopt/base_service.py
```diff
@@ -63,13 +64,66 @@ def create_hyperopt_domain(self):
...
+                            float(param.max),
+                            float(param.step),
+                        )
+                    else:
+                        hyperopt_search_space[param.name] = hyperopt.hp.loguniform(
+                            param.name, float(param.min), float(param.max)
+                        )
+                elif param.distribution == api_pb2.NORMAL:
+                    mu = (float(param.min) + float(param.max)) / 2
+                    sigma = (float(param.max) - float(param.min)) / 6
```
shashank-iitbhu, line 97: I followed this [article](https://soc.utah.edu/sociology3112/normal-distribution.php) to determine the value of `sigma` from `min` and `max`.

pkg/suggestion/v1beta1/hyperopt/base_service.py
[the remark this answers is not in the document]
tenzen-y: > @tenzen-y @kubeflow/wg-training-leads @shashank-iitbhu Should we round this float value to int if user wants to use this distribution and int parameter type ?

SGTM
Users can specify the double parameter type if they want to compute more exactly.
But, documentation of this restriction for int parameter type would be better.

pkg/suggestion/v1beta1/hyperopt/base_service.py
tenzen-y, line 97:
```suggestion
                    // We consider the normal distribution based on the range of ±3 sigma.
                    sigma = (float(param.max) - float(param.min)) / 6
```

shashank-iitbhu:
@tenzen-y I have added two new parameters, `weight_decay` and `dropout_rate`, to the Hyperopt example and passed them to `mnist.py`, but I haven't used them in the `Net` class yet in the `train` and `test` functions. If you check the logs for this e2e test, the maximum value of the loss metrics is an enormously large number. I can't figure out what I'm missing. Also tested this locally.

examples/v1beta1/hp-tuning/hyperopt-distribution.yaml
```diff
@@ -0,0 +1,69 @@
...
+      feasibleSpace:
+        min: "0.001"
+        max: "1"
+        distribution: "uniform"
+    - name: batch_size
+      parameterType: int
+      feasibleSpace:
+        min: "32"
+        max: "64"
+        distribution: "logNormal"
```
shashank-iitbhu, line 36: @tenzen-y Testing the `logNormal` distribution using `batch_size`.

shashank-iitbhu:
@tenzen-y 
https://github.com/kubeflow/katib/blob/867c40a1b0669446c774cd6e770a5b7bbf1eb2f1/pkg/suggestion/v1beta1/hyperopt/base_service.py#L265-L266
here the `float` values sampled from the distribution get converted to `int` for `INT` parameter type.

pkg/suggestion/v1beta1/hyperopt/base_service.py
```diff
@@ -62,14 +64,76 @@ def create_hyperopt_domain(self):
...
+                        )
+                    else:
+                        hyperopt_search_space[param.name] = hyperopt.hp.normal(
+                            param.name,
+                            mu,
+                            sigma,
+                        )
+                elif param.distribution == api_pb2.LOG_NORMAL:
+                    log_min = math.log(float(param.min))
+                    log_max = math.log(float(param.max))
```
tenzen-y, line 119: Shouldn't we use the fixed value when the min and max are scalers the same as Nevergrad, right?

test/unit/v1beta1/suggestion/test_hyperopt_service.py
```diff
@@ -176,6 +177,42 @@ def test_get_suggestion(self):
...
+                            parameter_type=api_pb2.DOUBLE,
+                            feasible_space=api_pb2.FeasibleSpace(
+                                max="10", min="5", list=[], step="0.8", distribution=api_pb2.NORMAL)
+                        ),
+                        api_pb2.ParameterSpec(
+                            name="param-10",
+                            parameter_type=api_pb2.DOUBLE,
+                            feasible_space=api_pb2.FeasibleSpace(
+                                max="10", min="5", list=[], step="0.8", distribution=api_pb2.LOG_NORMAL)
                         )
```
tenzen-y, line 216: Could you add test cases that contain the integer parameter types?

pkg/suggestion/v1beta1/hyperopt/base_service.py
shashank-iitbhu: Yeah we can, but that was an edge case when `min` and `max` are not defined in case of nevergrad.
```python
    elif isinstance(param, (p.Log, p.Scalar)):
        if (param.bounds[0][0] is None) or (param.bounds[1][0] is None):
            if isinstance(param, p.Scalar) and not param.integer:
                return hp.lognormal(label=param_name, mu=0, sigma=1)
``` 
For example,
```
    - name: batch_size
      parameterType: int
      feasibleSpace:
        min: "32"
        max: "64"
        distribution: "logNormal"
``` 
The above parameter will be sampled out from this graph:
Screenshot 2024-09-22 at 8 48 49 PM
where `u=3.8123` and `sigma=0.3465` are calculated by putting `min=32` and `max=64` in our code. and `E(X)` represents the mean which is `48` in our case.

tenzen-y: That makes sense.
In that case, could you address the cases where min or max is not specified, as well as nevergrad?

https://github.com/facebookresearch/nevergrad/blob/a2006e50b068fe598e0f3d7dab9c9bcf6cf97e00/nevergrad/optimization/externalbo.py#L61-L64

test/unit/v1beta1/suggestion/test_hyperopt_service.py
```diff
@@ -176,6 +177,72 @@ def test_get_suggestion(self):
...
+                            name="param-14",
+                            parameter_type=api_pb2.INT,
+                            feasible_space=api_pb2.FeasibleSpace(
+                                max="10", min="5", list=[], distribution=api_pb2.LOG_UNIFORM)
+                        ),
+                        api_pb2.ParameterSpec(
+                            name="param-15",
+                            parameter_type=api_pb2.INT,
+                            feasible_space=api_pb2.FeasibleSpace(
+                                max="100", min="5", list=[], step="0.8", distribution=api_pb2.NORMAL)
```
tenzen-y, line 245:
```suggestion
                                max="100", min="5", list=[], distribution=api_pb2.NORMAL)
```

We should verify the case where the step is not specified and distribution is Normal.

test/unit/v1beta1/suggestion/test_hyperopt_service.py
```diff
@@ -176,6 +177,72 @@ def test_get_suggestion(self):
...
+                            parameter_type=api_pb2.DOUBLE,
+                            feasible_space=api_pb2.FeasibleSpace(
+                                max="10", min="5", list=[], distribution=api_pb2.UNIFORM)
+                        ),
+                        api_pb2.ParameterSpec(
+                            name="param-9",
+                            parameter_type=api_pb2.DOUBLE,
+                            feasible_space=api_pb2.FeasibleSpace(
+                                max="10", min="5", list=[], step="0.8", distribution=api_pb2.NORMAL)
+                        ),
```
tenzen-y, line 210: Could you add a case where the distribution is Normal, and step is not specified?

test/unit/v1beta1/suggestion/test_hyperopt_service.py
```diff
@@ -176,6 +177,72 @@ def test_get_suggestion(self):
...
+                        api_pb2.ParameterSpec(
+                            name="param-10",
+                            parameter_type=api_pb2.DOUBLE,
+                            feasible_space=api_pb2.FeasibleSpace(
+                                max="10", min="5", list=[], step="0.8", distribution=api_pb2.LOG_NORMAL)
+                        ),
```
tenzen-y, line 216: Could you add a case where the distribution is LogNormal and step is not specified?

test/unit/v1beta1/suggestion/test_hyperopt_service.py
```diff
@@ -176,6 +177,72 @@ def test_get_suggestion(self):
                             parameter_type=api_pb2.DOUBLE,
                             feasible_space=api_pb2.FeasibleSpace(
                                 max="5", min="1", list=[])
+                        ),
```
tenzen-y, line 180: Could you add a case where the distribution is None, and the step is specified?

pkg/suggestion/v1beta1/hyperopt/base_service.py
tenzen-y: @shashank-iitbhu This is still pending.

test/unit/v1beta1/suggestion/test_hyperopt_service.py
tenzen-y: This is still pending as well.

pkg/suggestion/v1beta1/hyperopt/base_service.py
shashank-iitbhu: https://github.com/kubeflow/katib/blob/867c40a1b0669446c774cd6e770a5b7bbf1eb2f1/pkg/webhook/v1beta1/experiment/validator/validator.go#L287-L290

The webhook validator requires `feasibleSpace.max` or `feasibleSpace.min` to be specified.

tenzen-y: But when either min or max is empty, this validation does not reject the request, right?
So, shouldn't we implement the special case in the Suggestion Service?

shashank-iitbhu: Yes, the validation webhook does not reject the request when either min or max is empty. But I created an example where:
```
    - name: batch_size
      parameterType: int
      feasibleSpace:
        min: "32"
        distribution: "logNormal"
``` 
For this, the experiment is being created but the suggestion service is not sampling out any value hence the trials are not running, though handled this case (when either min or max are not specified) in `pkg/suggestion/v1beta1/hyperopt/base_service.py`.
Do we need to check `experiment_defaults.go` file?
https://github.com/kubeflow/katib/blob/867c40a1b0669446c774cd6e770a5b7bbf1eb2f1/pkg/apis/controller/experiments/v1beta1/experiment_defaults.go

pkg/suggestion/v1beta1/hyperopt/base_service.py
```diff
@@ -62,14 +64,75 @@ def create_hyperopt_domain(self):
...
+                            param.name,
+                            math.log(float(param.min)),
+                            math.log(float(param.max)),
+                            float(param.step),
+                        )
+                    else:
+                        hyperopt_search_space[param.name] = hyperopt.hp.loguniform(
+                            param.name,
+                            math.log(float(param.min)),
+                            math.log(float(param.max)),
```
andreyvelich, line 97: Do we need to take logarithm from the values before passing them to the `loguniform` ?

pkg/suggestion/v1beta1/hyperopt/base_service.py
```diff
@@ -62,14 +64,75 @@ def create_hyperopt_domain(self):
...
+                        )
+                elif param.distribution == api_pb2.NORMAL:
+                    mu = (float(param.min) + float(param.max)) / 2
```
andreyvelich, line 100: Please can you add the comment before this line on why we do this.

pkg/suggestion/v1beta1/hyperopt/base_service.py
andreyvelich: Maybe we should add this article to the comments. WDYT @tenzen-y @johnugeorge ?

tenzen-y: I do not want to depend on the individual article. Instead of that, it would be better to add an actual mathematical description here as a comment.

pkg/suggestion/v1beta1/hyperopt/base_service.py
shashank-iitbhu: We convert `param.min` and `param.max` to logarithmic scale as when optimizing, the sampled value is constrained to the interval `[exp(low), exp(high)]`. Refer to `hp.loguniform` documentation [here](http://hyperopt.github.io/hyperopt/getting-started/search_spaces/#parameter-expressions).

Also [here](https://github.com/facebookresearch/nevergrad/blob/c25cfd7b0aa72749331754ba5a20b67fa2615dc7/nevergrad/optimization/externalbo.py#L66-L68) in the `nevergrad` implementation the upper and lower bounds were converted to logarithmic scale for `hp.loguniform`

@andreyvelich @tenzen-y

pkg/suggestion/v1beta1/hyperopt/base_service.py
```diff
@@ -62,14 +64,75 @@ def create_hyperopt_domain(self):
...
-                    param.name, float(param.min), float(param.max), float(param.step)
-                )
-            elif param.type == DOUBLE:
-                hyperopt_search_space[param.name] = hyperopt.hp.uniform(
-                    param.name, float(param.min), float(param.max)
-                )
+            if param.type in [INTEGER, DOUBLE]:
+                if param.distribution == api_pb2.UNIFORM or param.distribution is None:
+                    if param.step:
+                        hyperopt_search_space[param.name] = hyperopt.hp.quniform(
```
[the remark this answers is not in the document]
shashank-iitbhu: oh yes, missed this, should use `quniformint` here for `INT`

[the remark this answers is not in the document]
shashank-iitbhu: `quniformint` is documented in the documentation but unavailable in the hyperopt package. Please check here on hyperopt github. https://github.com/hyperopt/hyperopt/blob/master/hyperopt/hp.py

andreyvelich (commented):
lgtm, just a small comment @shashank-iitbhu

andreyvelich:
@shashank-iitbhu Also, please rebase your PR to fix E2Es

shashank-iitbhu:
@andreyvelich Python Tests are failing due to an incompatible version of `typing-extensions`
```
    from typing_extensions import TypeIs as _TypeIs
E   ImportError: cannot import name 'TypeIs' from 'typing_extensions' (/opt/anaconda3/envs/katib/lib/python3.11/site-packages/typing_extensions.py)
==================================================================== short test summary info =====================================================================
ERROR sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
``` 
@tenzen-y This is related to #2251 
Could this be the root cause?
```torch 2.6.0 requires typing-extensions>=4.10.0, but you have typing-extensions 4.6.3 which is incompatible.```

andreyvelich:
@shashank-iitbhu @Electronic-Waste @helenxie-bit @mahdikhashan please can you help us to resolve the problems with Python dependencies ?

helenxie-bit:
> @andreyvelich Python Tests are failing due to an incompatible version of `typing-extensions`
> 
> ```
>     from typing_extensions import TypeIs as _TypeIs
> E   ImportError: cannot import name 'TypeIs' from 'typing_extensions' (/opt/anaconda3/envs/katib/lib/python3.11/site-packages/typing_extensions.py)
> ==================================================================== short test summary info =====================================================================
> ERROR sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
> ```
> 
> @tenzen-y This is related to #2251 Could this be the root cause? `torch 2.6.0 requires typing-extensions>=4.10.0, but you have typing-extensions 4.6.3 which is incompatible.`

It looks like that might be the cause, since `TypeIs` was introduced in version 4.10.0 according to the `typing-extensions` documentation.
@shashank-iitbhu Could we try updating the version in the Makefile by setting `pip install typing-extensions==4.10.0` to see if it resolves the issue?
https://github.com/kubeflow/katib/blob/28e466e1b84478c5e35542e44eff725ffd6ad09e/Makefile#L166

helenxie-bit:
> @shashank-iitbhu @Electronic-Waste @helenxie-bit @mahdikhashan please can you help us to resolve the problems with Python dependencies ?

@andreyvelich Shall I open an issue and a PR about this?

helenxie-bit:
> > > @shashank-iitbhu @Electronic-Waste @helenxie-bit @mahdikhashan please can you help us to resolve the problems with Python dependencies ?
> > 
> > 
> > @andreyvelich Shall I open an issue and a PR about this?
> 
> Yes @helenxie-bit , updating the version to `typing-extensions==4.10.0` resolves the issue, tested this locally.

Great! Ok, let me raise an issue and PR about this.

Electronic-Waste (changes requested):
LGTM! Just a few comments with typo.

Electronic-Waste:
Now, https://github.com/kubeflow/katib/pull/2504 has been merged. Could you please rebase this PR so that it can fix the error in Python Test? @shashank-iitbhu

PR #2423, opened 2024-09-05
[GSoC] Add unit tests for `tune` API

helenxie-bit:
This PR adds unit tests for `tune` API.

Fixes #

**Checklist:**


helenxie-bit:
\area gsoc

helenxie-bit:
Ref: https://github.com/kubeflow/katib/issues/2339

test/unit/v1beta1/tune-api/test_tune_api.py
```diff
@@ -0,0 +1,890 @@
+import unittest
```
andreyvelich, line 1: @helenxie-bit Let's add the unit tests under [the `katib_client_test.py` file](https://github.com/kubeflow/katib/blob/e4f7922ac448257485dcd6377d2b4aa663ba0d15/sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py#L295) since we previously agreed to place unit test files near the actual files.
You can use the tests data for `tune` API and define the fixture, similar to other SDK APIs.

andreyvelich (commented):
Thank you for doing this @helenxie-bit!

test/unit/v1beta1/tune-api/test_tune_api.py
helenxie-bit: @andreyvelich Thank you! I have followed this instruction and updated unit tests for `tune` API. Please review it when you have time.

helenxie-bit:
@andreyvelich The unit test for `tune` API is ready. I have also fixed some errors in this PR. Please have a look when you have time. Thanks!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -415,7 +415,10 @@ class name in this argument.
             experiment.spec.max_failed_trial_count = max_failed_trial_count
 
         # If users choose to use a custom objective function.
-        if objective is not None:
+        if objective is not None or parameters is not None:
+            if objective is None or base_image is None or parameters is None:
+                raise ValueError("One of the required parameters is None")
+
```
andreyvelich, line 421: Why do you need this if you have the check here: https://github.com/kubeflow/katib/blob/e68fe38f5f994526146defc38afb66a6d79cd0aa/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L366-L374

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -57,6 +57,27 @@ def get_observation_log_response(*args, **kwargs):
...
 
+def create_namespaced_persistent_volume_claim_response(*args, **kwargs):
+    if kwargs.get("namespace") == "pvc creation failed":
+        raise Exception("PVC creation failed")
+    else:
+        return {"metadata": {"name": "tune_test"}}
+
+
+def list_namespaced_persistent_volume_claim_response(*args, **kwargs):
+    if args[0] == "pvc creation failed":
```
andreyvelich, line 68: Should you check namespace like you did in the `create_namespaced_persistent_volume_claim_response` func ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -57,6 +57,27 @@ def get_observation_log_response(*args, **kwargs):
         )
 
 
+def create_namespaced_persistent_volume_claim_response(*args, **kwargs):
+    if kwargs.get("namespace") == "pvc creation failed":
```
andreyvelich, line 61: Please can you create constant for namespace that you want to use to check the PVC creation error ?
Like what we do in Training Operator: https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/api/training_client_test.py#L29

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +291,217 @@ def create_experiment(
...
+        {
+            "name": "tune_test",
+            "model_provider_parameters": "invalid",
+            "dataset_provider_parameters": HuggingFaceDatasetParams(),
+            "trainer_parameters": HuggingFaceTrainerParams(),
+        },
+        ValueError,
+    ),
+    (
+        "invalid dataset_provider_parameters",
```
andreyvelich, line 461: Do you also need test case for invalid trainer_parameters ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +291,217 @@ def create_experiment(
...
+            "trainer_parameters": HuggingFaceTrainerParams(),
+        },
+        RuntimeError,
+    ),
+    (
+        "valid flow with custom objective",
+        {
+            "name": "tune_test",
+            "namespace": "tune",
+            "objective": lambda x: x,
```
andreyvelich, line 486: Why this test case is valid if we test that parameters can be passed into objective function: https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/kubeflow/katib/utils/utils.py#L101 ?
E.g. we should be able to run this:
```py
objective({"param1": "value"})

File: sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py [hunk 7]
```diff
@@ -320,3 +562,65 @@ def test_get_trial_metrics(katib_client, test_name, kwargs, expected_output):
...
+
+@pytest.mark.parametrize("test_name,kwargs,expected_output", test_tune_data)
+def test_tune(katib_client, test_name, kwargs, expected_output):
+    """
+    test tune function of katib client
+    """
+    print("\n\nExecuting test:", test_name)
+
+    PYTORCHJOB_KIND = "PyTorchJob"
+    JOB_PARAMETERS = {
```
andreyvelich, line 575: Do you need this since you just use the default values from Training SDK: https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/constants/constants.py#L162?

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -320,3 +562,65 @@ def test_get_trial_metrics(katib_client, test_name, kwargs, expected_output):
...
+    ), patch(
+        "kubeflow.katib.utils.utils.get_trial_substitutions_from_trainer",
+        return_value={"param": "value"},
+    ), patch.object(
+        katib_client, "create_experiment", return_value=Mock()
+    ) as mock_create_experiment:
+        try:
+            katib_client.tune(**kwargs)
+            mock_create_experiment.assert_called_once()
+            assert expected_output == TEST_RESULT_SUCCESS
```
andreyvelich, line 623: Is there a way for you to validate which Experiment object we will create here: https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L715 ?
I think, that will allow you to make sure we create the correct Experiment object for various test cases.

andreyvelich (commented):
Thank you for doing this @helenxie-bit!
/assign @tenzen-y @johnugeorge @deepanker13 @droctothorpe @Electronic-Waste

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: Here, I added code to handle the case where users specify only "objective" or "parameters". This is not covered in the previous check.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: I initially tried to check the namespace in the same way, but it caused an error. The issue was that `create_namespaced_persistent_volume_claim_response` was called with keyword arguments:
```[python]
self.core_api.create_namespaced_persistent_volume_claim(
                    namespace=namespace,
                    body=training_utils.get_pvc_spec(
                        pvc_name=name,
                        namespace=namespace,
                        storage_config=storage_config,
                    ),
                )
```
However, `list_namespaced_persistent_volume_claim_response` was called with a positional argument:
```[python]
pvc_list = self.core_api.list_namespaced_persistent_volume_claim(
                    namespace
                )
```
I’ve now adjusted `list_namespaced_persistent_volume_claim_response` to use keyword arguments as well, for consistent namespace checking.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: Sure, I have created a constant for that.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: Previously, we didn’t include a type check for `trainer_parameters` in the `tune` API. You’re right that it’s important to do so. I’ve now added the type check for `trainer_parameters` in the tune API, along with the corresponding unit test.
https://github.com/kubeflow/katib/blob/b25f7ba37a60710191666b936ed47598c2de4ac2/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L603-L606
https://github.com/kubeflow/katib/blob/b25f7ba37a60710191666b936ed47598c2de4ac2/sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py#L473-L481

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: The objective function `"objective": lambda x: x` is an identity function, meaning it returns whatever input it receives unchanged. In the test case, This function will accept the input `{"param1": "value"}` and return it exactly as it was passed in.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: I added `JOB_PARAMETERS` here because it’s required in the following sections of the `tune` API:
https://github.com/kubeflow/katib/blob/b25f7ba37a60710191666b936ed47598c2de4ac2/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L637-L662
https://github.com/kubeflow/katib/blob/b25f7ba37a60710191666b936ed47598c2de4ac2/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L718-L724
I defined and patched it in the unit test to avoid importing `JOB_PARAMETERS` externally. This way, the test runs independently without needing to fetch `JOB_PARAMETERS` from outside the test environment.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: @andreyvelich I previously added a similar check in the following section of the code (which is deleted now):
https://github.com/kubeflow/katib/blob/e4f7922ac448257485dcd6377d2b4aa663ba0d15/test/unit/v1beta1/tune-api/test_tune_api.py#L487-L887
One concern I have is that when users want to optimize hyperparameters in LLMs, the container arguments will include the JSON transformation of `training_parameters` (from the `transformers.TrainingArguments` class) and `lora_config` (from the `peft.LoraConfig` class). Since these classes are defined outside our scope, any future changes to them could cause the actual results in the unit tests to no longer match the expected results, leading to test failures. I'm not sure if this concern is valid. Do you have any suggestions?

helenxie-bit:
@andreyvelich Thank you so much for reviewing this! I’ve replied to all your feedback, and please take a look when you have time.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
andreyvelich: Maybe you don't need to validate the all arguments ? Just make sure that arguments contain the hyperparameters that we want to optimize. E.g. `"${trialParameters.r}"`

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
andreyvelich: >I defined and patched it in the unit test to avoid importing JOB_PARAMETERS externally. This way, the test runs independently without needing to fetch JOB_PARAMETERS from outside the test environment.

@helenxie-bit Why we don't want to install `kubeflow-training` SDK as part of our unit tests and fetch values from there instead of re-define them in the Katib unit test ?

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
andreyvelich: make sense, thanks.

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
andreyvelich: I see, similar to how we have check here: https://github.com/kubeflow/katib/blob/e68fe38f5f994526146defc38afb66a6d79cd0aa/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L509-L513

helenxie-bit: Yes, exactly.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: Oh, I initially thought it would be better to avoid external dependencies in the unit tests, which is why I redefined those values. However, I can use this way (install `kubeflow-training` SDK as part of the unit test) if that’s preferred.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: OK got it! I will add that.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +293,227 @@ def create_experiment(
...
+        self.api_version = api_version
+        self.kind = kind
+        self.metadata = metadata
+        self.spec = spec
+        self.status = status
+
+
+test_tune_data = [
+    (
+        "not specify name",
```
Electronic-Waste, line 378: How about "missing name"?

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +293,227 @@ def create_experiment(
...
+        {
+            "name": None,
+            "objective": lambda x: x,
+            "parameters": {"param": "value"},
+        },
+        ValueError,
+    ),
+    (
+        "set both options",
```
Electronic-Waste, line 387: I guess, it would be better if you could explain what this error belongs to. Like `invalid hybrid parameters - objective and model_provider_paramters`.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +293,227 @@ def create_experiment(
...
+        {
+            "name": "tune_test",
+            "objective": lambda x: x,
+            "model_provider_parameters": HuggingFaceModelParams(),
+        },
+        ValueError,
+    ),
+    (
+        "set no option",
```
Electronic-Waste, line 396: What about `missing parameters`?

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +293,227 @@ def create_experiment(
...
+        {
+            "name": "tune_test",
+        },
+        ValueError,
+    ),
+    (
+        "set one option but missing parameters - only objective",
```
Electronic-Waste, line 403: Like I mentioned above, it would be better if you could specify the scope of this test case. Like `missing parameters in custom objective tuning - lack of parameters`

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +293,227 @@ def create_experiment(
...
+        {
+            "name": "tune_test",
+            "objective": lambda x: x,
+        },
+        ValueError,
+    ),
+    (
+        "set one option but missing parameters - only parameters",
```
Electronic-Waste, line 411: Like the suggestion above, for your reference :)

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +293,227 @@ def create_experiment(
...
+        {
+            "name": "tune_test",
+            "parameters": {"param": "value"},
+        },
+        ValueError,
+    ),
+    (
+        "set one option but missing parameters - only model_provider_parameters",
```
Electronic-Waste, line 419: Same as above, `missing parameters in external model tuning - lack of xxxx`

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +293,227 @@ def create_experiment(
...
+        {
+            "name": "tune_test",
+            "model_provider_parameters": HuggingFaceModelParams(),
+        },
+        ValueError,
+    ),
+    (
+        "set one option but missing parameters - only dataset_provider_parameters",
```
Electronic-Waste, line 427: Like the suggestion above, for your reference :)

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -270,6 +293,227 @@ def create_experiment(
...
+        {
+            "name": "tune_test",
+            "dataset_provider_parameters": HuggingFaceDatasetParams(),
+        },
+        ValueError,
+    ),
+    (
+        "set one option but missing parameters - only trainer_parameters",
```
Electronic-Waste, line 435: Like the suggestion above, for your reference :)

Electronic-Waste (commented):
Thanks for your great contribution @helenxie-bit !  I left some suggestions about the name of UTs for your reference 

I guess it will be more friendly for other developers to understand these UTs if we could specify the scope of each UTs:)

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: That sounds better. I will adjust it. Thank you!

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: Sounds better. Thanks!

helenxie-bit:
> Thanks for your great contribution @helenxie-bit ! I left some suggestions about the name of UTs for your reference
> 
> I guess it will be more friendly for other developers to understand these UTs if we could specify the scope of each UTs:)

@Electronic-Waste Thank you so much for your suggestions! I have incorporated your advice into the UTs.

helenxie-bit:
@andreyvelich I’ve added the `kubeflow-training` dependency to the unit tests, allowing values and parameters to be fetched from there instead of redefining them within the Katib unit tests. You can find the change here:
https://github.com/kubeflow/katib/blob/b24b44fc3c54a58173d007f4cf9f6a1311e677d3/test/unit/v1beta1/requirements.txt#L3
Additionally, I have added verification for the experiment object in two valid flow cases in the following section:
https://github.com/kubeflow/katib/blob/b24b44fc3c54a58173d007f4cf9f6a1311e677d3/sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py#L587-L639
Please review when you have time. Thanks!

helenxie-bit:
@andreyvelich Please review this PR when you have time :eyes:. I'll adjust the e2e tests for the tune API after this PR is merged, as it fixes some small bugs in the tune API.

helenxie-bit:
Since @mahdikhashan is currently working on the notebook with instructions for this API, and this PR also includes some bug fixes for the API, @andreyvelich, would it be better to separate the bug fixes into a standalone PR to make the review process more convenient? Please let me know your thoughts.

andreyvelich:
> Since @mahdikhashan is currently working on the notebook with instructions for this API, and this PR also includes some bug fixes for the API, @andreyvelich, would it be better to separate the bug fixes into a standalone PR to make the review process more convenient? Please let me know your thoughts.

Sure, please create separate PR for the bug fixes.

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
Electronic-Waste: Maybe valid case needs `objective_metric_name` field, since we use `StdOut` metrics collector by default.

And it would be better if we could check the existence of `objective_metric_name` parameter here:

https://github.com/kubeflow/katib/blob/93bee4dc25529a92320cf33634bb6920013c8281/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L388-L393

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
```diff
@@ -320,3 +566,78 @@ def test_get_trial_metrics(katib_client, test_name, kwargs, expected_output):
...
+                        V1beta1TrialParameterSpec(name="a", reference="a"),
+                    ]
+                    # Verify experiment_params
+                    assert experiment.spec.parameters == [
+                        V1beta1ParameterSpec(
+                            name="a",
+                            parameter_type="int",
+                            feasible_space=V1beta1FeasibleSpace(min="10", max="100"),
+                        ),
+                    ]
```
Electronic-Waste, line 609: It would be better if we also verify the `objective_metric_name` parameter.

Electronic-Waste (commented):
Basically LGTM! I left a few comments concerned with `objective_metric_name` for you.

Thanks for your great contribution! @helenxie-bit

sdk/python/v1beta1/kubeflow/katib/api/katib_client_test.py
helenxie-bit: Thank you for your suggestions! I will add that.

helenxie-bit:
@andreyvelich @Electronic-Waste This PR is ready for review. But I'm not sure why the ci test `Publusi Katib Core images/Publish Image (katib-ui, cmd/ui/v1beta1/Dockerfile)` kept failing due to network connectivity issue. Do you have any ideas?

andreyvelich (approved):
Thank you for this great contribution @helenxie-bit!
/lgtm
/approve

PR #2435, opened 2024-09-26
Server side apply

### LINKED ISSUE #2271: Update the kubernetes object's status with server-side apply
/kind feature

I'd like to replace the client-side apply with the server-side apply when the controller updates the any statuses like https://github.com/kubeflow/katib/blob/fc858d15dd41ff69166a2020efa200199063f9ba/pkg/controller.v1beta1/experiment/experiment_controller_status.go#L27.

Server-side applies allow us to avoid updating errors due to conflicts, and then it would reduce confusions for users. 

---

gjyotin305:
This PR adds the feature of server side apply, replacing the previously client side apply. Server-side applies allow us to avoid updating errors due to conflicts, and then it would reduce confusions for users.

Fixes #2271 

**Checklist:**

### DIFF
--- a/pkg/controller.v1beta1/experiment/experiment_controller_status.go
+++ b/pkg/controller.v1beta1/experiment/experiment_controller_status.go
@@ -20,12 +20,13 @@ import (
 	"context"
 
 	experimentsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/experiments/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type updateStatusFunc func(instance *experimentsv1beta1.Experiment) error
 
 func (r *ReconcileExperiment) updateStatus(instance *experimentsv1beta1.Experiment) error {
-	err := r.Status().Update(context.TODO(), instance)
+	err := r.Status().Update(context.TODO(), instance, &client.SubResourceUpdateOptions{UpdateOptions: client.UpdateOptions{FieldManager: "experiment-controller"}})
 	if err != nil {
 		return err
 	}


pkg/controller.v1beta1/experiment/experiment_controller_status.go
```diff
@@ -20,12 +20,13 @@ import (
...
 
 	experimentsv1beta1 "github.com/kubeflow/katib/pkg/apis/controller/experiments/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type updateStatusFunc func(instance *experimentsv1beta1.Experiment) error
 
 func (r *ReconcileExperiment) updateStatus(instance *experimentsv1beta1.Experiment) error {
-	err := r.Status().Update(context.TODO(), instance)
+	err := r.Status().Update(context.TODO(), instance, &client.SubResourceUpdateOptions{UpdateOptions: client.UpdateOptions{FieldManager: "experiment-controller"}})
```
andreyvelich, line 29: @kubeflow/wg-training-leads @gjyotin305 Can we also use the server side apply for other resources that we update in our controllers.
For instance:
Suggestion CR: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/suggestion/suggestion_controller_status.go#L36
Trial CR: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller_status.go#L45

@tenzen-y Do we need to update k8s version for server-side apply for Katib ?

andreyvelich (commented):
Thanks for this contribution @gjyotin305 and sorry for the late reply!

PR #2455, opened 2024-11-29
Remove Dropout layer from ENAS Trial container to fix E2E tests

andreyvelich:
This should fix E2E tests for ENAS example and unblock other PRs.

I removed the Dropout layer from the model constructor given that it fails for TensorFlow distributed training:
```
ValueError: creating distributed tf.Variable with aggregation=MEAN and 
a non-floating dtype is not supported, please use a different aggregation or dtype
```

/assign @kubeflow/wg-training-leads @Electronic-Waste @tariq-hasan


tenzen-y (commented):
Thanks!
/lgtm
/approve

PR #2465, opened 2024-12-12
chore: supporting the listen-address parameter on db-manager

caiofralmeida:
**What this PR does / why we need it**: I would like to be able to pass a different listen-address to handle other ports or even ipv6.

Fixes #

**Checklist:**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,9 +88,10 @@ Below is a list of command-line flags accepted by Katib controller:
 
 Below is a list of command-line flags accepted by Katib DB Manager:
 
-| Name            | Type          | Default | Description                                             |
-| --------------- | ------------- | ------- | ------------------------------------------------------- |
-| connect-timeout | time.Duration | 60s     | Timeout before calling error during database connection |
+| Name            | Type          | Default      | Description                                                         |
+| --------------- | ------------- | -------------| ------------------------------------------------------------------- |
+| connect-timeout | time.Duration | 60s          | Timeout before calling error during database connection             |
+| listen-address  | string        | 0.0.0.0:6789 | The network interface or IP address to receive incoming connections |
 
 ## Katib admission webhooks
 
--- a/cmd/db-manager/v1beta1/main.go
+++ b/cmd/db-manager/v1beta1/main.go
@@ -35,7 +35,7 @@ import (
 )
 
 const (
-	port                  = "0.0.0.0:6789"
+	defaultListenAddress  = "0.0.0.0:6789"
 	defaultConnectTimeout = time.Second * 60
 )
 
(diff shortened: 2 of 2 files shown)


tenzen-y (commented):
Thank you for making the address customizable.
This looks great to me.
/approve
/rerun-all

@kubeflow/wg-automl-leads How about this improvement?

andreyvelich (commented):
Thank you for this contribution @caiofralmeida 🎉 
@kubeflow/wg-automl-leads @caiofralmeida @Electronic-Waste @helenxie-bit Don't we want to use env variable to configure all settings for Katib DB Manager for now ?
Ideally, we should migrate to the Katib Config, but it could be done in the future.


The problem is that we have a few settings for Katib DB manager that need to configured via envs: https://www.kubeflow.org/docs/components/katib/user-guides/env-variables/#katib-db-manager

caiofralmeida:
@andreyvelich thanks for the heads up. I can implement the listen-address configuration as an environment variable as well (in another PR) what do you think?

We just need to define which of the settings should take precedence.

andreyvelich:
I would suggest that we migrate all Katib DB manager args to the envs as part of this PR:
`CONNECT_TIMEOUT` and `LISTEN_ADDRESS`
What do you think @caiofralmeida ?

caiofralmeida:
@andreyvelich I agree with refactoring ARGs to environment variables, but I think that making this change in this PR will generate a braking change. I would feel more comfortable if we already had a way to override the listen address (through args) remove an impediment on our side and implement the PR by refactoring all args to environment variables, already doing the necessary tests, and declaring that ARGs will no longer be supported.

andreyvelich:
Actually, it is a great point @caiofralmeida, I agree with you!
Ideally, we should migrate Katib DB manager settings to the Katib config all together, but we can do it later.
https://www.kubeflow.org/docs/components/katib/user-guides/katib-config/
If you could create tracking issue for it, that would be great!

caiofralmeida:
> Actually, it is a great point @caiofralmeida, I agree with you! Ideally, we should migrate Katib DB manager settings to the Katib config all together, but we can do it later. https://www.kubeflow.org/docs/components/katib/user-guides/katib-config/ If you could create tracking issue for it, that would be great!

Nice, I'm going to create a tracking issue right now. Thanks!

andreyvelich:
@caiofralmeida Please can you re-push your commit to trigger GitHub actions ?
Since the action was triggered more than 1 month ago, GitHub doesn't allow me to trigger them.

caiofralmeida:
> @caiofralmeida Please can you re-push your commit to trigger GitHub actions ? Since the action was triggered more than 1 month ago, GitHub doesn't allow me to trigger them.

For sure, done @andreyvelich :)

andreyvelich (commented):
/lgtm
/approve

caiofralmeida:
@andreyvelich I've created the [issue](https://github.com/kubeflow/katib/issues/2500). I have not found any template that fits the intention of the issue.

Issue #2390, opened 2024-07-17
[Discussion] The future of Trial metadata parameters.


Electronic-Waste:
As we discuss in issue #2369 , we choose to always pass the Trial name as env variable `KATIB_TRIAL_NAME` to the training container. And we implemented this feature as function `mutatePodEnv` and used it here:

https://github.com/kubeflow/katib/blob/206fe1c1065826ae173a4baea002bb34b483aba8/pkg/webhook/v1beta1/pod/inject_webhook.go#L140-L142

Here are some discussions about it: https://github.com/kubeflow/katib/pull/2369#discussion_r1677885344 https://github.com/kubeflow/katib/pull/2369#discussion_r1678669508 https://github.com/kubeflow/katib/pull/2369#discussion_r1681061505

We use this issue to track the discussion. Any suggestions are welcome!

mickvangelderen:
In my project I would like to know:

```
KATIB_EXPERIMENT_NAME
KATIB_TRIAL_NAME
```

These may be generated by K8S. It would be nice to have:

```
KATIB_NAMESPACE
```

as well, even though that value may be known.

For my project, I also need

```
KATIB_BASE_URL
```

but I suppose Katib does not necessarily know it's external-facing base URL and so I can inject that value into the container envs myself.

Electronic-Waste:
Could you please pin this issue so that more people can join the discussion? @andreyvelich 
/remove-good-first-issue
/kind discussion

Electronic-Waste:
I think it's a good idea to pass corresponding experiment name and the namespace to the env of Pod! As for the url, I'm not sure if we need to pass it to the pod. 

WDYT👀 @kubeflow/wg-automl-leads @tariq-hasan @helenxie-bit

andreyvelich:
We can get Kubernetes namespace from this location on Pod:
```
/run/secrets/kubernetes.io/serviceaccount/namespace
```

Electronic-Waste:
@andreyvelich Oh, I see. Maybe we need to add the experiment name to the env of Pod. I'll raise a PR to address that.

PR #2488, opened 2025-01-15
Add KATIB_EXPERIMENT_NAME to pod env.

### LINKED ISSUE #2473: Expose Experiment metadata to Trial Template

We can access metadata from the trialSpec in the template as [documented here](https://www.kubeflow.org/docs/components/katib/user-guides/trial-template/#use-metadata-in-trial-template). 

I would like to obtain the experiment name (which is generated by K8S for me) and so I would like to have something similar to `trialSpec.Name` but for `experimentSpec.Name`. 

With my limited understanding of go, it seems like the [relevant code is here](https://github.com/kubeflow/katib/blob/9531372530043cdf4b5de1b3efb5f0c1043c5141/pkg/controller.v1beta1/experiment/manifest/generator.go#L109-L197). 

For now I can work around this by making an assumption on how the trial name is generated from the experiment name, but it would be nice to not have to make that assumption.

As a related question, can we make katib provide certain env vars by default (`KATIB_EXPERIMENT_NAME`, `KATIB_TRIAL_NAME`, and `KATIB_NAMESPACE` for example) to the trial programs? This would eliminate the need for the  additional experiment spec configuration. 

### Why is this needed?

I let K8S generate the experiment name. In order to obtain the experiment name, I currently have to query the experiment name from the trial name at run time in my trial program or make assumptions on how the trial name is generated from the experiment name which I would rather not. 

My use case is having consistent naming between katib experiments and trials and MLflow experiments and runs. I want to set the MLFlow experiment and run names based on the katib experiment and trial names. 

### Love this feature?

Electronic-Waste:

Users sometimes may need experiment metadata for their customized applications like: https://github.com/kubeflow/katib/issues/2473

As we discussed in https://github.com/kubeflow/katib/issues/2390, we can get namespace of experiment from: 

```
/var/run/secrets/kubernetes.io/serviceaccount/namespace
```

So we only need to pass experiment name as `KATIB_EXPERIMENT_NAME` to the Pod Env just like `KATIB_TRIAL_NAME`:)

Fixes #2473
Also related to: https://github.com/kubeflow/katib/issues/2390

**Checklist:**

/cc @kubeflow/wg-automl-leads @mickvangelderen @helenxie-bit @tariq-hasan

### DIFF
--- a/pkg/webhook/v1beta1/pod/utils.go
+++ b/pkg/webhook/v1beta1/pod/utils.go
@@ -313,7 +313,7 @@ func mutatePodEnv(pod *v1.Pod, trial *trialsv1beta1.Trial) error {
 		)
 		return nil
 	} else {
-		return fmt.Errorf("unable to find primary container %v in mutated pod containers %v",
+		return fmt.Errorf("Unable to find primary container %v in mutated pod containers %v",
 			trial.Spec.PrimaryContainerName, pod.Spec.Containers)
 	}
 }
--- a/pkg/webhook/v1beta1/pod/utils.go
+++ b/pkg/webhook/v1beta1/pod/utils.go
@@ -291,7 +291,7 @@ func mutatePodEnv(pod *v1.Pod, trial *trialsv1beta1.Trial) error {
 			pod.Spec.Containers[index].Env = []v1.EnvVar{}
 		}
 
-		// Pass env variable KATIB_TRIAL_NAME to the primary container using fieldPath
+		// Pass env variable KATIB_TRIAL_NAME and KATIB_EXPERIMENT_NAME to the primary container using fieldPath
 		pod.Spec.Containers[index].Env = append(
 			pod.Spec.Containers[index].Env,
 			v1.EnvVar{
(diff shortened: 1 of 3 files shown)


pkg/webhook/v1beta1/pod/utils.go
```diff
@@ -302,10 +302,18 @@ func mutatePodEnv(pod *v1.Pod, trial *trialsv1beta1.Trial) error {
 					},
 				},
 			},
+			v1.EnvVar{
+				Name: consts.EnvExperimentName,
+				ValueFrom: &v1.EnvVarSource{
+					FieldRef: &v1.ObjectFieldSelector{
+						FieldPath: fmt.Sprintf("metadata.labels['%s']", consts.LabelExperimentName),
```
Electronic-Waste, line 309: This label exists in Trial Pod. Here is an example:

```
Name:             tune-experiment-lhs5mrsh-9xqwd
Namespace:        kubeflow
Priority:         0
Service Account:  default
Start Time:       Wed, 15 Jan 2025 06:29:28 +0000
Labels:           batch.kubernetes.io/controller-uid=859791df-d3da-404d-841c-86eca6c9148f
                  batch.kubernetes.io/job-name=tune-experiment-lhs5mrsh
                  controller-uid=859791df-d3da-404d-841c-86eca6c9148f
                  job-name=tune-experiment-lhs5mrsh
                  katib.kubeflow.org/experiment=tune-experiment
                  katib.kubeflow.org/trial=tune-experiment-lhs5mrsh
Annotations:      sidecar.istio.io/inject: false
Status:           Running
...
```

mickvangelderen: Oh so we can already reference `metadata.labels['katib.kubeflow.org/experiment']` in the spec?

Electronic-Waste: @mickvangelderen  We could not access `metadata.labels['katib.kubeflow.org/experiment']` directly since it's not included in `.trialSpec.Labels` field. The label lies in `.Labels` field in Trial and Pod.

And also, I agree with you that it will be more convenient if we could also pass `KATIB_NAMESPACE` to env of Pod by default. WDTY👀 @kubeflow/wg-automl-leads

Electronic-Waste: @mickvangelderen And it will also be okay to pass namespace and trial name as env variables like this: https://github.com/kubeflow/katib/blob/fc858d15dd41ff69166a2020efa200199063f9ba/examples/v1beta1/trial-template/trial-metadata-substitution.yaml#L77-L81

PR #2497, opened 2025-01-18
[SDK] Update `tune` API

helenxie-bit:
This PR includes the following updates to the `tune` API:
- Enhanced Parameter Validation: Adds additional checks for input parameters to ensure better error handling and robustness.
- Bug Fix for `num_labels`: Resolves an issue where the `num_labels` parameter was not being processed correctly.
- Training Operator SDK Update: Updates the SDK version to include critical changes required for the trainer.

Fixes #

**Checklist:**

### DIFF
--- a/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
+++ b/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
@@ -416,7 +416,7 @@ class KatibClient(object):
 
         # If users choose to use a custom objective function.
         if objective is not None or parameters is not None:
-            if not objective or not base_image or not parameters:
+            if not objective or not parameters:
                 raise ValueError("One of the required parameters is None")
             # Add metrics collector to the Katib Experiment.
             # Up to now, we only support parameter `kind`, of which default value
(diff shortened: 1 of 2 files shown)


sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
```diff
@@ -415,7 +415,9 @@ class name in this argument.
             experiment.spec.max_failed_trial_count = max_failed_trial_count
 
         # If users choose to use a custom objective function.
-        if objective is not None:
+        if objective is not None or parameters is not None:
+            if not objective or not base_image or not parameters:
+                raise ValueError("One of the required parameters is None")
```
andreyvelich, line 420: Why do we need this validation if by default the base_image is the TensorFlow image ?

andreyvelich (commented):
Thanks for the fix @helenxie-bit!

sdk/python/v1beta1/kubeflow/katib/api/katib_client.py
helenxie-bit: You're right, the check for `base_image` is unnecessary. My intention was to maintain consistency, which is why I included it. However, to avoid any confusion, I’ll go ahead and remove it. Thanks for pointing this out!

Electronic-Waste (approved):
LGTM. Thanks for this! @helenxie-bit

andreyvelich (commented):
Thanks for the update @helenxie-bit!
/lgtm
/approve

PR #2521, opened 2025-03-02
Upgrade Kubernetes Version to 1.32.2

### LINKED ISSUE #2516: Support Kubernetes v1.32

We should support the latest k8s v1.32 release https://github.com/kubernetes/kubernetes/releases/tag/v1.32.2.

### Why is this needed?

Kubeflow Trainer has decided to support Kubernetes v1.32: https://github.com/kubeflow/trainer/issues/2434

And some dependencies like the JobSet API depend on it.

### Love this feature?

Electronic-Waste:

We should support the latest k8s v1.32 release https://github.com/kubernetes/kubernetes/releases/tag/v1.32.2.

Kubeflow Trainer has decided to support Kubernetes v1.32 https://github.com/kubeflow/trainer/issues/2434. We'd better upgrade the Kubernetes version to maintain consistency with Trainer.

/cc @kubeflow/wg-automl-leads @helenxie-bit @mahdikhashan @Doris-xm @truc0 

Fixes #2516 

**Checklist:**


pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -49,14 +51,18 @@ import (
 )
 
 const (
-	namespace       = "default"
-	batchJobName    = "test-job"
-	objectiveMetric = "accuracy"
-	timeout         = time.Second * 10
+	namespace             = "default"
+	succeededBatchJobName = "test-job-succeeded"
+	failedBatchJobName    = "test-job-failed"
```
Electronic-Waste, line 56: We need to separate the `Trial run with "Failed" BatchJob.` test and other tests. That's because:

- If a Job is succeeded, we need to set `Complete` and `SuccessCriteriaMet` Conditions in the Job status
- If a Job is failed, we need to set `Failed` and `FailureTarget` Conditions in the Job status

And we can't mutate these completion conditions, or we will get:

```
Invalid value: cannot disable the terminal Failed=True condition, status.conditions: Invalid value: cannot disable the terminal FailureTarget=True condition,
```

So we need to separate these UTs with two different Job Names.

pkg/controller.v1beta1/trial/trial_controller_test.go
```diff
@@ -238,28 +253,69 @@ func TestReconcileBatchJob(t *testing.T) {
...
+		batchJobCompleteMessage := "BatchJob completed test message"
+		batchJobCompleteReason := "BatchJob completed test reason"
+		g.Eventually(func(g gomega.Gomega) {
+			g.Expect(c.Get(ctx, succeededBatchJobKey, batchJob)).Should(gomega.Succeed())
+			// TODO(Electronic-Waste): Remove this condition when K8s 1.30 is no longer supported.
+			// SuccessPolicy is available in K8s 1.31 and later. If we set it in K8s 1.30, it will be ignored.
+			// And when we set the status with `SuccessCriteriaMet`, it will report error:
+			// "Invalid value: cannot set SuccessCriteriaMet=True for Job without SuccessPolicy".
+			// REF: https://kubernetes.io/docs/concepts/workloads/controllers/job/#success-policy
+			isK8sVersion130 := strings.Contains(os.Getenv("KUBEBUILDER_ASSETS"), "1.30")
```
Electronic-Waste, line 284: We need to remove this when we do not plan to support Kubernetes v1.30.

Related CI error: https://github.com/kubeflow/katib/actions/runs/13613616097/job/38053846460

Electronic-Waste (commented):
This PR is ready for the review. PTAL if you have time👀

@kubeflow/wg-automl-leads @helenxie-bit @mahdikhashan @Doris-xm @truc0

Makefile
```diff
@@ -33,7 +33,7 @@ fmt:
 
 lint:
 ifndef HAS_LINT
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.57.2
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.60.1
```
Electronic-Waste, line 36: Update this because support for Go 1.23 starts from v1.60.1

https://golangci-lint.run/product/changelog/#v1601

.github/workflows/test-go.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
       fail-fast: false
       matrix:
         # Detail: `setup-envtest list`
-        kubernetes-version: ["1.29.3", "1.30.0", "1.31.0"]
+        kubernetes-version: ["1.29.3", "1.30.0", "1.31.0", "1.32.0"]
```
mahdikhashan, line 70: do we need to have 1.32.2 here as well?

Electronic-Waste:
> @Electronic-Waste Could you finalize this PR please ?

Sure. I'll resolve the conflict.

> Maybe we can directly upgrade to k8s 1.34 like in Trainer: https://github.com/kubeflow/trainer/pull/2804

I don't think it's a good practice to directly upgrade k8s version to 1.34. We need to gradually upgrade the version so that we can roll back to the older version if we faced with incompatibility:)

WDYT? @andreyvelich

andreyvelich:
>I don't think it's a good practice to directly upgrade k8s version to 1.34. We need to gradually upgrade the version so that we can roll back to the older version if faced with incompatibility:)

Sure, we can split it between 2 PRs, but in the next Katib release I want us to support same k8s version as in Trainer.

andreyvelich (commented):
@Electronic-Waste Are we ready to merge this ?

Electronic-Waste:
@andreyvelich I think we're ready to merge this.

andreyvelich (commented):
/lgtm
/approve
@Electronic-Waste Please create the followup PR to support k8s 1.33

PR #2596, opened 2025-12-09
Add seccomp profile to UI component security context

juliusvonkohout:


juliusvonkohout:
Probably fixed by https://github.com/kubeflow/manifests/pull/3303

PR #2618, opened 2026-02-07
feat: Add configurable security context support to Katib controller pods

jsonmp-k8:
## Summary
- Add explicit, per-component security context fields to the KatibConfig API (`PodSecurityContext` for suggestions, `SecurityContext` for early stopping and metrics collector containers)
- Apply security contexts in the composer (suggestion deployments) and webhook (metrics collector sidecars), with backward compatibility for the legacy `InjectSecurityContext` flag
- Update all 6 `katib-config.yaml` manifests with hardened defaults (runAsNonRoot, drop ALL capabilities, seccomp RuntimeDefault)

## Test plan
- [x] `go test ./pkg/apis/config/v1beta1/...` — SecurityContext fields survive defaulting round-trip
- [x] `go test ./pkg/controller.v1beta1/suggestion/composer/...` — Deployment created with PodSecurityContext and early stopping SecurityContext
- [x] `go test ./pkg/webhook/v1beta1/pod/...` — Config SecurityContext takes precedence over legacy flag; fallback works; nil case preserved

🤖 Generated with [Claude Code](https://claude.com/claude-code)


andreyvelich (commented):
/ok-to-test

PR #2620, opened 2026-02-16
fix: inject Katib labels early to support PodDefaults for trial pods

### LINKED ISSUE #2548: PodDefaults over trial pods not being applied accordingly
### What happened?

I have two pod defaults using `matchExpressions` for the following two labels:
- `katib.kubeflow.org/trial` -> To be applied over trial pods
- `katib.kubeflow.org/experiment` -> To be applied on the above an the algorithm pod

For some reason the poddefaults are note being applied as expected. For instance if I create a pod with these labels it works as expected but with the pods created for the experiment it does not behave as expected.

In the escenario described above, the trial pod does not have any podefault. If I change the matchexpression to check for another label, for instance, `job-name`, that poddefault gets applied, but still the `katib.kubeflow.org/experiment` one is not. On the other hand on the algorithm pod (the one that makes the suggestions), the experiment label is detected and the poddfeault applied (this is on both scenarios, before and after changing the trial label in the poddefault).

Both the poddefaults should be applied to the trial pod.

### Environment

Kubernetes version:
```bash
$ kubectl version
Client Version: v1.32.3
Kustomize Version: v5.5.0
Server Version: v1.31.5+k3s1
```
Katib controller version:
```bash
docker.io/kubeflowkatib/katib-controller:v0.17.0% 
```
Katib Python SDK version:
```bash
$ pip show kubeflow-katib
Name: kubeflow-katib
Version: 0.18.0
Author: Kubeflow Authors
Location: /home/grudloff/projects/mlops-example-katib/.venv/lib/python3.10/site-packages
Required-by: mlops-example-katib
```

### Impacted by this bug?

Priyanshu-u07:
When a trial pod is created, Katib labels like `katib.kubeflow.org/trial` are added too late . They get added by a webhook that runs after the PodDefaults webhook has already finished checking. Because of this timing issue, PodDefaults that depend on Katib labels are never applied to trial pods.

This PR fixes it by adding the labels directly into the pod template before the pod is created, so PodDefaults can see them in time.

**Changes:**

- Added InjectLabelsToPodTemplate utility in pkg/controller.v1beta1/util/unstructured.go that injects labels into Pod templates across workload types (Job, CronJob, Training Operator replicaSpecs)
- Updated generator.go to inject experiment labels into the trial template early
- Updated trial_controller.go to inject trial name labels into the workload before creation
- Added unit tests covering all supported workload types, including edge case where no initial labels exist

Fixes #2548

### DIFF
--- a/pkg/controller.v1beta1/experiment/manifest/generator.go
+++ b/pkg/controller.v1beta1/experiment/manifest/generator.go
@@ -103,6 +103,11 @@ func (g *DefaultGenerator) GetRunSpecWithHyperParameters(experiment *experiments
 	runSpec.SetName(trialName)
 	runSpec.SetNamespace(trialNamespace)
 
+	// Inject Experiment labels into the RunSpec's Pod template.
+	if err := util.InjectLabelsToPodTemplate(runSpec, util.TrialLabels(experiment)); err != nil {
+		return nil, err
+	}
+
 	return runSpec, nil
 }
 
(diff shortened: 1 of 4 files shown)


Priyanshu-u07:
@andreyvelich Could you please take a look when you get a chance?
Thanks!

PR #2629, opened 2026-03-01
fix(controller): Correct error formatting in trial controller package

### LINKED ISSUE #2628: Incorrect error formatting in trial controller package
### What happened?

Three error formatting issues in `pkg/controller.v1beta1/trial/`:

1. `trial_controller_util.go:229` uses `%e` (scientific notation for floats) instead of `%w`. Produces garbage output like `&{%!e(string=connection refused)}` on timestamp parse errors.

2. `trial_controller.go:61-63`  sentinel errors use `fmt.Errorf()` instead of `errors.New()`. These are compared with `errors.Is()` so they should be immutable sentinels

3. `trial/util/job_util.go:82,103`  uses `%v` instead of `%w`, breaking the error chain for `errors.Is()` and `errors.As()` callers

Errors should use correct format verbs and follow Go error handling conventions.

### Environment

Spotted while reading through the trial controller code

### Impacted by this bug?

ruskaruma:
  Fixes three error formatting issues in `pkg/controller.v1beta1/trial/`:                                                                                                                                                              
                                                                                                                                                                                                                                       
  1. `trial_controller_util.go:229` uses `%e` (scientific notation for floats) instead of `%w`, producing garbled output like `&{%!e(string=...)}` on timestamp parse errors.                                                          
  2. `trial_controller.go:61-63` uses `fmt.Errorf()` for sentinel errors that are compared with `errors.Is()`. Changed to `errors.New()` for immutability.                                                                             
  3. `trial/util/job_util.go:82,103` uses `%v` instead of `%w`, which breaks the error chain for `errors.Is()`/`errors.As()` callers.                                                                                                  
                                                                                                                                                                                                                                       
  Fixes #2628                                                                                                                                                                                                                          
                                                                                                                                                                                                                                       
  **Checklist:**                                                                                                                                                                                                                       
                                                                                                                                                                                                                                       
    - N/A: internal error handling only, no user facing changes

### DIFF
--- a/pkg/controller.v1beta1/trial/trial_controller.go
+++ b/pkg/controller.v1beta1/trial/trial_controller.go
@@ -58,9 +58,9 @@ const (
 var (
 	log = logf.Log.WithName(ControllerName)
 	// errMetricsNotReported is the error when Trial job is succeeded but metrics are not reported yet
-	errMetricsNotReported = fmt.Errorf("metrics are not reported yet")
+	errMetricsNotReported = errors.New("metrics are not reported yet")
 	// errReportMetricsFailed is the error when `unavailable` metrics value can't be inserted to the Katib DB.
-	errReportMetricsFailed = fmt.Errorf("failed to report unavailable metrics")
+	errReportMetricsFailed = errors.New("failed to report unavailable metrics")
 )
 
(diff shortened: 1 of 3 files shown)


ruskaruma:
@andreyvelich, can you please review this?

PR #2666, opened 2026-05-26
Fail loudly on unsupported Optuna parameter distributions

### LINKED ISSUE #2665: Optuna suggestion service silently ignores NORMAL and LOG_NORMAL distributions
### What happened?

The Optuna suggestion service maps only `UNIFORM` and `LOG_UNIFORM` distributions for `int` and `double` parameters. In `pkg/suggestion/v1beta1/optuna/base_service.py`, `_get_optuna_search_space()` has branches for `UNIFORM`/`None` and `LOG_UNIFORM`, but no branch and no error for `NORMAL` or `LOG_NORMAL`:

```python
if param.type == INTEGER:
    if param.distribution in [api_pb2.UNIFORM, None]:
        ...
    elif param.distribution == api_pb2.LOG_UNIFORM:
        ...
    # NORMAL / LOG_NORMAL: no branch, param is never added
elif param.type == DOUBLE:
    if param.distribution in [api_pb2.UNIFORM, None]:
        ...
    elif param.distribution == api_pb2.LOG_UNIFORM:
        ...
    # NORMAL / LOG_NORMAL: no branch, param is never added
```

When a parameter declares `distribution: normal` or `distribution: lognormal`, it falls through every branch and is left out of the returned `search_space` dict. Optuna then never samples that parameter, so the hyperparameter is silently dropped from tuning. No error is raised and no warning is logged, so the experiment runs to completion with results that quietly ignore that parameter.

The `Distribution` enum in `api.proto` defines all four (`UNIFORM`, `LOG_UNIFORM`, `NORMAL`, `LOG_NORMAL`), and the `hyperopt` suggestion service already handles `NORMAL` and `LOG_NORMAL` (`pkg/suggestion/v1beta1/hyperopt/base_service.py`). So the distributions are valid input that one backend honors and the other silently discards.

Optuna has no native normal or log-normal distribution (it offers uniform, log-uniform via the `log` flag, and categorical), so the Optuna backend cannot honor `NORMAL`/`LOG_NORMAL`. Given that, it should fail loudly with a clear error naming the parameter and the unsupported distribution, rather than silently dropping the parameter and producing misleading results.

### Environment

Katib version (master, `pkg/suggestion/v1beta1/optuna/base_service.py`)

### Impact

A misconfigured experiment (Optuna algorithm + a `normal`/`lognormal` parameter) appears to run normally but silently tunes a smaller search space than the user defined. The omission is invisible without reading the suggestion-service internals.

### Suggested fix

Add an explicit `else` in the `INTEGER` and `DOUBLE` branches of `_get_optuna_search_space()` that raises a `ValueError` naming the parameter and distribution. A follow-up could implement actual `NORMAL`/`LOG_NORMAL` support if a faithful Optuna mapping is agreed on.

saivedant169:

The Optuna suggestion service maps only `UNIFORM` and `LOG_UNIFORM` for `int` and `double` parameters. In `_get_optuna_search_space()` (`pkg/suggestion/v1beta1/optuna/base_service.py`), a parameter declaring `NORMAL` or `LOG_NORMAL` matched neither branch and was left out of the returned search space. Optuna then never sampled that parameter, so the hyperparameter was silently dropped from tuning with no error and no log line. The experiment completed with results that quietly ignored the parameter.

The `Distribution` enum in `api.proto` defines all four values, and the `hyperopt` suggestion service already handles `NORMAL` and `LOG_NORMAL`. So these are valid inputs that one backend honors and the Optuna backend silently discarded.

Optuna has no native normal or log-normal distribution (it offers uniform, log-uniform via the `log` flag, and categorical), so the backend cannot faithfully honor `NORMAL`/`LOG_NORMAL`. This PR adds an explicit `else` branch for `int` and `double` that raises a `ValueError` naming the parameter and the unsupported distribution, turning a silent correctness bug into an actionable error. Implementing real support can be a follow-up if a faithful mapping is agreed on.

Fixes #2665

**Checklist:**

- [x] Unit tests added that prove the fix is effective
- [x] `make pytest` style local run passes (`pytest test/unit/v1beta1/suggestion/test_optuna_service.py`: 29 passed)

**Testing**:

Added `TestOptunaUnsupportedDistribution`, a parametrized regression test covering `NORMAL` and `LOG_NORMAL` for both `int` and `double` parameters. It asserts `get_suggestions` raises `ValueError`. The four cases fail on master (`DID NOT RAISE`, confirming the silent drop) and pass with this change. The existing 25 Optuna tests continue to pass. `black`, `isort --profile black`, and `flake8` are clean on both changed files.

### DIFF
--- a/pkg/suggestion/v1beta1/optuna/base_service.py
+++ b/pkg/suggestion/v1beta1/optuna/base_service.py
@@ -129,6 +129,13 @@ class BaseOptunaService(object):
                         log=True,
                         step=1,
                     )
+                else:
+                    raise ValueError(
+                        f"Unsupported distribution "
+                        f"{api_pb2.Distribution.Name(param.distribution)} for parameter "
+                        f"{param.name}. The Optuna suggestion service supports only UNIFORM "
+                        f"and LOG_UNIFORM for int parameters."
+                    )
 
             elif param.type == DOUBLE:
                 if param.distribution in [api_pb2.UNIFORM, None]:
@@ -147,6 +154,13 @@ class BaseOptunaService(object):
                         high=float(param.max),
                         log=True,
                         step=None,
+                    )
+                else:
+                    raise ValueError(
+                        f"Unsupported distribution "
+                        f"{api_pb2.Distribution.Name(param.distribution)} for parameter "
+                        f"{param.name}. The Optuna suggestion service supports only UNIFORM "
+                        f"and LOG_UNIFORM for double parameters."
                     )
 
             elif param.type in [CATEGORICAL, DISCRETE]:
--- a/test/unit/v1beta1/suggestion/test_optuna_service.py
+++ b/test/unit/v1beta1/suggestion/test_optuna_service.py
@@ -18,6 +18,12 @@ import pytest
 import utils
 
 from pkg.apis.manager.v1beta1.python import api_pb2
+from pkg.suggestion.v1beta1.internal import constant
+from pkg.suggestion.v1beta1.internal.search_space import (
+    HyperParameter,
+    HyperParameterSearchSpace,
+)
+from pkg.suggestion.v1beta1.optuna.base_service import BaseOptunaService
 from pkg.suggestion.v1beta1.optuna.service import OptunaService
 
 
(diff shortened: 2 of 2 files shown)


saivedant169:
Gentle bump on this. It is a small change to the Optuna suggestion service: an unsupported parameter distribution (NORMAL or LOG_NORMAL) currently falls through without a clear error, and this makes it raise a ValueError naming the distribution instead, with a test. CI is green and the commit is DCO signed. @andreyvelich @kramaranya could one of you take a look when you have a moment? Happy to rebase if it has drifted.

andreyvelich (commented):
Thanks for this @saivedant169!
/lgtm
/approve
I would suggest you to take a look at OptimizationJob KEP: https://github.com/kubeflow/trainer/pull/3565
This API will replace the Katib project in the future versions.

PR #1765, opened 2022-01-01
[WIP] Support JSON format logs in `file-metrics-collector`

tenzen-y:
**What this PR does / why we need it**:
Support JSON format logs in `file-metrics-collector`.

**Which issue(s) this PR fixes** _(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_:
Fixes #1744 

**Checklist:**

- [x] [Docs](https://www.kubeflow.org/docs/components/katib/) included if any changes are user facing

PR #2672, opened 2026-06-13
fix(metricscollector): parse fractional UNIX timestamps in JSON logs

immanuwell:
What this PR does / why we need it

`file-metricscollector` supports JSON logs, but numeric UNIX timestamps with fractions were parsed wrong. The fractional part was treated like raw nanoseconds, so `1638422847.28721` became `2021-12-02T05:27:27.000028721Z` instead of `2021-12-02T05:27:27.28721Z`.

This patch keeps the fractional seconds when building the RFC3339 timestamp. String timestamps and integer UNIX timestamps stay the same. This is a small fix, but it is pretty easy to hit on the JSON log path.

Related to #944
Follow-up to #1765

How to reproduce

1. Run `go test ./pkg/metricscollector/v1beta1/file-metricscollector -count=1`
2. Before this patch, the JSON fixture in `TestCollectObservationLog` stored `1638422847.28721` as `2021-12-02T05:27:27.000028721Z`
3. With this patch, it stores `2021-12-02T05:27:27.28721Z`

Checklist:

### DIFF
--- a/pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go
+++ b/pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go
@@ -217,17 +217,23 @@ func parseTimestamp(timestamp interface{}) string {
 		}
 
 		stringTimestamp = strconv.FormatFloat(floatTimestamp, 'f', -1, 64)
-		t := strings.Split(stringTimestamp, ".")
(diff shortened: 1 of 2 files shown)


immanuwell:
@anencore94 @Electronic-Waste @gaocegege please take a look

PR #2679, opened 2026-06-22
Validate positive range for log-scale distribution parameters

### LINKED ISSUE #2678: Suggestion services crash on log-distribution parameters with non-positive min/max
### What happened

If a parameter uses a log-scale distribution (`logUniform` or `logNormal`) and its `feasibleSpace.min` (or `max`) is `0` or negative, the suggestion service crashes on the first `GetSuggestions` call with a math domain error.

In `pkg/suggestion/v1beta1/hyperopt/base_service.py`, `create_hyperopt_domain` calls `math.log(float(param.min))` and `math.log(float(param.max))` directly for `LOG_UNIFORM` and `LOG_NORMAL` (lines 95, 102, 128). `math.log` of a non-positive value raises `ValueError: math domain error`, which propagates straight out of the gRPC handler. The skopt service has the same problem, since it builds every `DOUBLE` parameter with a hardcoded `log-uniform` prior.

The experiment webhook validator does not catch this. `validateParameters` only checks that the distribution name is one of the allowed values and that `min`/`max` are non-empty. It never checks that the range is positive, so an experiment with `min: "0"` and `distribution: logUniform` is admitted and then crashes the suggestion pod.

### What you expected to happen

A log-scale distribution is only defined for a strictly positive range. The experiment should be rejected at admission with a clear message, the same way other invalid parameter specs are, instead of crashing the suggestion service at runtime.

### How to reproduce it

Create an experiment with a parameter like:

```yaml
- name: lr
  parameterType: double
  feasibleSpace:
    min: "0"
    max: "10"
    distribution: logUniform
```

The suggestion service fails on the first `GetSuggestions` with `ValueError: math domain error`.

### Environment

Katib master (33c8ff4).

I have a fix for the validator side and will open a PR.

saivedant169:

Fixes #2678.

A parameter with a `logUniform` or `logNormal` distribution and a non-positive `min` or `max` is currently admitted by the webhook and then crashes the suggestion service at `math.log` (hyperopt `create_hyperopt_domain`, and skopt's hardcoded log-uniform prior) with `ValueError: math domain error`.

`validateParameters` now rejects `logUniform` / `logNormal` parameters whose `min` or `max` parses to a value `<= 0`, with a clear field error, so the experiment is refused at admission instead of crashing the suggestion pod. This matches how the validator already rejects other invalid `feasibleSpace` values, and follows the same fail-at-admission approach as #2666.

The check only fires when the value parses as a float and is `<= 0`, so non-numeric min/max are left to the existing handling.

## Test

Added a `TestValidateParameters` case (non-positive min for `logUniform`).

```
go test ./pkg/webhook/v1beta1/experiment/validator/ -run TestValidateParameters
ok  github.com/kubeflow/katib/pkg/webhook/v1beta1/experiment/validator
```

/kind bug

### DIFF
--- a/pkg/webhook/v1beta1/experiment/validator/validator.go
+++ b/pkg/webhook/v1beta1/experiment/validator/validator.go
@@ -289,6 +289,19 @@ func (g *DefaultValidator) validateParameters(parameters []experimentsv1beta1.Pa
 					allErrs = append(allErrs, field.Required(parametersPath.Index(i).Child("feasibleSpace").Child("max"),
 						fmt.Sprintf("feasibleSpace.max or feasibleSpace.min must be specified for parameterType: %v", param.ParameterType)))
 				}
+				if param.FeasibleSpace.Distribution == experimentsv1beta1.DistributionLogUniform ||
+					param.FeasibleSpace.Distribution == experimentsv1beta1.DistributionLogNormal {
+					// Log-scale distributions are only defined for a strictly positive range,
+					// so a non-positive min or max would crash the suggestion service at math.log.
+					if v, err := strconv.ParseFloat(param.FeasibleSpace.Min, 64); err == nil && v <= 0 {
+						allErrs = append(allErrs, field.Invalid(parametersPath.Index(i).Child("feasibleSpace").Child("min"),
+							param.FeasibleSpace.Min, fmt.Sprintf("feasibleSpace.min must be greater than 0 for distribution: %v", param.FeasibleSpace.Distribution)))
+					}
+					if v, err := strconv.ParseFloat(param.FeasibleSpace.Max, 64); err == nil && v <= 0 {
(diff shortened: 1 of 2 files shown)


saivedant169:
@andreyvelich this one has been open about a week without a look. The red checks are the flaky ones (the trial-controller `TestReconcileBatchJob` gomock timing test and the e2e version-matrix timeouts), not the validator change, which passes. Since I can't trigger a re-run myself, could you kick the CI and take a look when you have a moment? It's a small kind/bug fix (positive-range validation for logUniform/logNormal params).

saivedant169:
@andreyvelich @tenzen-y gentle bump on this one, it has been open about three weeks. The validator change is small and the red checks are the known flaky e2e ones. Also still needs an /ok-to-test from a trusted user.

andreyvelich (commented):
Sorry for the delay @saivedant169!
/lgtm
/approve

PR #2683, opened 2026-07-04
config: add restricted PSS contexts for suggestions

danish9039:

This PR makes the built-in Katib suggestion deployments compatible with Kubernetes restricted Pod Security Standards by using existing `runtime.suggestions[*].securityContext` support in the install configs.

It also enables `init.controller.injectSecurityContext` so Katib metrics collector init containers get restricted-compatible security contexts. No API/schema fields or Trial template mutation are added.

OpenShift install note: `katib-openshift` keeps the restricted fields but omits a fixed `runAsUser` so OpenShift SCC UID allocation can choose the namespace-assigned UID.

Validation:

- `KUBEBUILDER_ASSETS="$(/home/danish/go/bin/setup-envtest use 1.34.0 -p path)" go test -count=1 ./pkg/util/v1beta1/katibconfig ./pkg/controller.v1beta1/suggestion/composer`
- `kustomize build` for all six install overlays plus `katib-with-kubeflow`
- `git diff --check`

Related to kubeflow/manifests#3487

**Checklist:**


danish9039:
Two notes from testing this alongside the downstream restricted Pod Security Standards work in kubeflow/community-distribution#3487.

- **The red `e2e` checks here are not caused by this change.** The same jobs fail on unrelated dependabot pull requests that touch only `pkg/ui/v1beta1/frontend` JavaScript dependencies and cannot affect suggestion runtimes — #2702 fails `v1.31.3`, `v1.33.1` and `v1.34.0`; #2697 fails all four; #2700 fails `v1.34.0`. Worth treating as pre-existing rather than a regression in this pull request.
- **The `katib-openshift` variant correctly omits `runAsUser`** while the other installs set `1000`. That is the right split: on OpenShift the security context constraint assigns a UID from the namespace range, so a hard-coded `1000` would be rejected, whereas on plain Kubernetes the kubelet needs a numeric identity to satisfy `runAsNonRoot: true`.

On the fixed `runAsUser: 1000` in the non-OpenShift installs: it works today, but only because the images have no OCI `User` and therefore run as root unless told otherwise. I confirmed every suggestion image plus `earlystopping-medianstop` and `file-metrics-collector` declares an empty `User`. Their Dockerfiles already end with `chgrp -R 0 ${TARGET_DIR} && chmod -R g+rwX ${TARGET_DIR}`, the arbitrary-UID pattern — and that pattern only works because `runAsUser` without `runAsGroup` yields GID `0`, which I verified on a restricted-enforcing cluster (`uid=1000 gid=0(root) groups=0(root)`).

The durable fix would be a numeric `USER 1000:0` in the images themselves. The group-0 groundwork is already there, so it is a one-line addition per Dockerfile, and it would let every manifest drop `runAsUser` — matching what kserve/kserve#5855 and the merged kubeflow/notebooks#1229 do. Restricted only requires `runAsNonRoot: true`; `runAsUser` is not part of the profile. Happy to raise that as a follow-up if it is useful.

One thing I could not verify and that is worth a second look: whether all writable paths work under the selected identity for the `pbt` suggestion specifically, since it mounts a `ReadWriteMany` `persistentVolumeClaim`. Unlike `emptyDir`, PVC ownership is not automatically world-writable, so that path may need `fsGroup`.

PR #2685, opened 2026-07-09
Cast resample_probability before range check in PBT validation

### LINKED ISSUE #2684: PBT ValidateAlgorithmSettings crashes with TypeError on resample_probability
Reraising a small validation crash in the PBT suggestion service, in the same spirit as #2666 and #2679 (fail with a clear message instead of an internal crash).

### What happened

PBT's `ValidateAlgorithmSettings` checks `resample_probability` against the `[0, 1]` range without casting it:

https://github.com/kubeflow/katib/blob/master/pkg/suggestion/v1beta1/pbt/service.py#L67-L70

Algorithm setting values arrive as strings, so `0 <= settings["resample_probability"] <= 1` raises `TypeError: '<=' not supported between instances of 'int' and 'str'`. Validation then fails with an internal `UNKNOWN` error rather than a clear `INVALID_ARGUMENT` message.

The adjacent `n_population` and `truncation_threshold` checks cast first (`int(...)` and `float(...)`); only `resample_probability` is missing the cast.

### How to reproduce

Submit a PBT experiment that sets `resample_probability` (to any value). The webhook validation crashes instead of returning a validation message.

### Fix

Cast the value with `float()` before the range check, matching the adjacent settings.

saivedant169:

PBT's `ValidateAlgorithmSettings` compares `resample_probability` against the `[0, 1]` range without casting it, but algorithm setting values arrive as strings, so `0 <= settings["resample_probability"] <= 1` raises `TypeError` and validation crashes with an internal `UNKNOWN` error instead of a clear message. The adjacent `n_population` and `truncation_threshold` checks cast their values first; this brings `resample_probability` in line with them.

- Cast `resample_probability` with `float()` before the range check.
- Add a PBT service unit test covering `ValidateAlgorithmSettings`. The service had no unit test before. The test fails on the current code (the valid case returns `UNKNOWN` from the `TypeError`) and passes with the fix.

Same fail-loudly spirit as #2666 and #2679.

## Which issue(s) this PR fixes

Fixes #2684

## Checklist

- [x] Unit tests added and passing locally


saivedant169:
Hi @andreyvelich, when you have a moment, could you take a look at this one? It is the same validation-hardening line as #2666 and #2679, it fixes the PBT `TypeError` crash from #2684 by casting `resample_probability` before the range check. Thanks!

andreyvelich (commented):
Thanks @saivedant169!
/lgtm
/approve

PR #2689, opened 2026-07-16
Honor parameter distribution in Skopt and Goptuna suggestion services

### LINKED ISSUE #2688: skopt and goptuna suggestion services silently ignore parameter distributions
### What happened?

The `skopt` and `goptuna` suggestion services build their search spaces from `parameter.type` only and never read `parameter.distribution`, so the declared distribution is discarded and a hardcoded one is used instead.

**skopt** (`pkg/suggestion/v1beta1/skopt/base_service.py`, `create_optimizer`):

- INTEGER params build `skopt.space.Integer(min, max, name=...)`, which defaults to `prior="uniform"`.
- DOUBLE params build `skopt.space.Real(min, max, "log-uniform", name=...)`, a hardcoded `log-uniform` prior.

So every double parameter is sampled log-uniformly regardless of its declared distribution. A parameter with `distribution: uniform` (or unset, the default) is silently sampled log-uniformly, and `normal` / `logNormal` are also forced to log-uniform.

**goptuna** (`pkg/suggestion/v1beta1/goptuna/converter.go`, `toGoptunaSearchSpace`): always builds `goptuna.UniformDistribution` / `IntUniformDistribution` keyed on `ParameterType`. `p.Distribution` is never read, even though `goptuna.LogUniformDistribution` is available, so a `logUniform` parameter is silently sampled uniformly.

The distribution reaches these services intact: the validator (`pkg/webhook/v1beta1/experiment/validator/validator.go`) accepts `Uniform`, `LogUniform`, `Normal`, `LogNormal` for any algorithm, `convertDistribution` (`pkg/controller.v1beta1/suggestion/suggestionclient/suggestionclient.go`) maps the CRD value to the correct enum, and `convert_parameter` (`pkg/suggestion/v1beta1/internal/search_space.py`) preserves it (only `""` / `DISTRIBUTION_UNSPECIFIED` become `None`). So these are valid, validated inputs that both backends ignore: skopt forces log-uniform for doubles, goptuna forces uniform.

This is the same class of bug that #2665 fixed for optuna in #2666, where hyperopt already honors these distributions and optuna now rejects unsupported ones with an error.

### Reproduction

**skopt** (scikit-optimize):

```python
import skopt
# create_optimizer builds this for every DOUBLE param (base_service.py:68-73):
print(skopt.space.Real(1e-4, 1e-1, "log-uniform", name="lr").prior)  # log-uniform
# a param declaring distribution: uniform (the default) should instead be:
print(skopt.space.Real(1e-4, 1e-1, name="lr").prior)                 # uniform
```

The `"log-uniform"` prior is hardcoded and `param.distribution` is never read, so a uniform double is sampled log-uniformly.

**goptuna** (calling `toGoptunaSearchSpace` with a DOUBLE param whose `Distribution` is `LOG_UNIFORM`):

```
param declared distribution = LOG_UNIFORM
toGoptunaSearchSpace built  = goptuna.UniformDistribution
result: distribution IGNORED, sampled uniform
```

An Experiment with `algorithm: skopt` (or a goptuna-backed `random` / `tpe` / `cmaes`) whose parameters set `distribution` runs without error, but the declared distribution is ignored.

Both backends should honor `parameter.distribution`. skopt supports it directly (`skopt.space.Real` / `Integer` take `prior="uniform"` or `"log-uniform"`) and goptuna has `UniformDistribution` and `LogUniformDistribution`, so `uniform` and `logUniform` can be mapped through. For `normal` / `logNormal`, which neither library samples natively, the service should fail loudly naming the parameter and the unsupported distribution, consistent with #2666, rather than silently substituting a different distribution.

### Environment

- Katib: master
- Suggestion services: skopt, goptuna

/kind bug

saivedant169:

The Skopt and Goptuna suggestion services built their search spaces from `parameter.type` only and never read `parameter.distribution`, so a declared distribution was silently discarded:

- **Skopt** (`skopt/base_service.py`) hardcoded a `"log-uniform"` prior for every double parameter and left integer parameters uniform, so a `distribution: uniform` double (the default) was sampled log-uniformly.
- **Goptuna** (`goptuna/converter.go`) always built `UniformDistribution` / `IntUniformDistribution`, so a `distribution: logUniform` parameter was sampled uniformly.

Both now honor the declared distribution. Because both libraries support uniform and log-uniform directly (`skopt.space.Real`/`Integer` take a `prior`, and goptuna has `LogUniformDistribution`), the distribution is mapped through rather than rejected. For `normal` / `logNormal`, which neither library samples natively (and `logUniform` for goptuna int parameters, which goptuna has no distribution for), the service now raises a clear error naming the parameter and the unsupported distribution, consistent with the Optuna service in #2666.

Note this changes the Skopt default for double parameters from log-uniform to uniform, which is the intended correction.

## Which issue(s) this PR fixes

Fixes #2688

## Testing

- Added Goptuna unit tests in `converter_test.go` (log-uniform maps to `LogUniformDistribution`; `normal` on a double and `logUniform` on an int return an error). All pass with `go test ./pkg/suggestion/v1beta1/goptuna/`.
- Added Skopt unit tests in `test_skopt_service.py` covering the distribution to prior mapping and the fail-loud path for unsupported distributions. These run under `make pytest-skopt`.
- Verified the Skopt mapping against the actual `create_optimizer`: a uniform double now yields `prior="uniform"`, a log-uniform double yields `prior="log-uniform"`, and a `normal` double raises `ValueError`.


saivedant169:
Hi @andreyvelich, whenever you get a window, mind taking a look at this one? Same distribution-handling line as #2666, just extended to the skopt and goptuna services, which were silently ignoring the declared distribution. No rush.

andreyvelich:
Thanks @saivedant169!
/ok-to-test
/lgtm
/approve

saivedant169:
Hi @andreyvelich, this one has been approved and is green apart from the `tune-example-llm-optimization` e2e, which times out with the trial still Running rather than failing, so it looks unrelated to this change. A `/retest` last week hit the same timeout. Could you re-run it when you get a moment, or let me know if there is anything you would like me to adjust on my end?

saivedant169:
@andreyvelich this one's still sitting on the flaky e2e, logs are from July. Mind kicking off a rerun?

PR #2706, opened 2026-08-25
Return no-op instead of SQL error when ObservationLog has no valid entries

### LINKED ISSUE #2705: db-manager : RegisterObservationLog fails with SQL syntax error when observation log has no valid entries (MySQL & Postgres)
### What happened?

/kind bug

When a metrics-collector sidecar reports an `ObservationLog` where every `MetricLog` entry is skipped — i.e., the log is empty, or every entry has an empty `TimeStamp` — `RegisterObservationLog` builds an invalid SQL query and the RPC fails.

In [`pkg/db/v1beta1/mysql/mysql.go`](https://github.com/kubeflow/katib/blob/master/pkg/db/v1beta1/mysql/mysql.go) (and the same pattern in `pkg/db/v1beta1/postgres/postgres.go`), the query is constructed as:

```go
sqlQuery := "INSERT INTO observation_logs (trial_name, time, metric_name, value) VALUES "
for _, mlog := range observationLog.MetricLogs {
    if mlog.TimeStamp == "" {
        continue // all entries can be skipped here
    }
    sqlQuery += "(?, ?, ?, ?),"
    // ...
}
sqlQuery = sqlQuery[0 : len(sqlQuery)-1] // assumes at least one row was appended
```

If no rows are appended, stripping the trailing character produces:
```sql
INSERT INTO observation_logs (trial_name, time, metric_name, value) VALUES
```

which fails `Prepare()` with a SQL syntax error (e.g., `Error 1064: You have an error in your SQL syntax near 'VALUES'`). The error propagates through the `db-manager` gRPC back to the metrics collector, so metrics reporting for that trial fails instead of being treated as a harmless no-op.

This is reachable in practice: the `StdOut` collector can emit entries with zero-time/missing timestamps (similar behavior observed in #1702).

### Proposed fix: Guard against the empty case before building/executing the statement, in both `mysql.go` and `postgres.go`:
```go
if len(values) == 0 {
    return nil
}
```

Alternatively, build placeholder groups into a slice and use `strings.Join(groups, ",")`, which removes the fragile string truncation entirely. Happy to submit a PR for either approach.

Reporting an observation log with no valid metric entries should succeed as a no-op (return `nil`), not fail the `ReportObservationLog` RPC with a SQL syntax error.

### Environment

Kubernetes version:
Any (bug is version-independent; code path exists on master)

Katib controller version:
master / latest (`pkg/db/v1beta1`)

Katib Python SDK version:
Not applicable — bug is in the Go db-manager backend (`pkg/db/v1beta1/mysql/mysql.go`, `pkg/db/v1beta1/postgres/postgres.go`)

### Impacted by this bug?

pujitha24:

`RegisterObservationLog` in `pkg/db/v1beta1/mysql/mysql.go` and `pkg/db/v1beta1/postgres/postgres.go` builds an `INSERT ... VALUES` query by appending a `(?, ?, ?, ?),` placeholder group per `MetricLog` entry that has a non-empty `TimeStamp`, then strips the trailing comma with `sqlQuery[0:len(sqlQuery)-1]`. When every entry in the `ObservationLog` is skipped (empty `MetricLogs`, or every entry has an empty `TimeStamp`), no placeholder groups are appended, and the truncation instead chops the trailing space off `"...VALUES "`, producing a malformed statement. `Prepare()` then fails with a SQL syntax error, and that error is returned as-is by the `ReportObservationLog` gRPC handler (`cmd/db-manager/v1beta1/main.go`), so a call that should be a harmless no-op fails the RPC instead. This is per-RPC only — it does not crash or otherwise affect the db-manager process itself, just that one `ReportObservationLog` call.

**Approach**: Add an early `if len(values) == 0 { return nil }` guard right after the loop that builds `values`, before the query string is truncated, in both `mysql.go` and `postgres.go`. This is the primary fix the issue proposed. `RegisterObservationLog`'s only caller passes its error straight through as the gRPC error, so returning `nil` when there is nothing to insert is a safe no-op with no skipped side effects.

**Validation**:
- Added `TestRegisterObservationLogNoValidEntries` to `pkg/db/v1beta1/mysql/mysql_test.go` and `pkg/db/v1beta1/postgres/postgres_test.go`, each registering an `ObservationLog` with a single `MetricLog` whose `TimeStamp` is empty (so it is skipped), asserting `RegisterObservationLog` returns `nil`.
- Confirmed these tests reproduce the exact bug: reverting only the two-line fix (keeping the new tests) makes both fail with `Prepare SQL statement failed: ... call to Prepare 'INSERT INTO observation_logs (trial_name, time, metric_name, value) VALUES' query was not expected` — sqlmock rejecting the malformed truncated query from the issue report. Re-applying the fix makes them pass.
- `go build ./pkg/db/...` — passes.
- `go test ./pkg/db/v1beta1/mysql/... ./pkg/db/v1beta1/postgres/...` — all tests pass, including the two new ones and the pre-existing `TestRegisterObservationLog`.
- `gofmt -l` on the changed files — clean.
- `go vet ./pkg/db/v1beta1/mysql/... ./pkg/db/v1beta1/postgres/...` — clean.
- `golangci-lint run ./pkg/db/v1beta1/mysql/... ./pkg/db/v1beta1/postgres/...` reports 7 pre-existing issues (errcheck on `stmt.Close()`/`rows.Close()`, staticcheck ST1005 on capitalized error strings); verified via `git stash` that all 7 are present identically on unmodified `master` and none are introduced by this change. Note: the locally available `golangci-lint` is v2.12.2, while this repo's `make lint` pins v1.64.7 — a version mismatch, disclosed here rather than resolved, since the flagged lines are all pre-existing and outside this diff.
- Did not run against a live MySQL or PostgreSQL instance; validation relies on this package's existing `sqlmock`-based unit test harness (`github.com/DATA-DOG/go-sqlmock`), the same harness the file's pre-existing `RegisterObservationLog` test uses.
- Branch is cut from `upstream/master` tip (`dec5030ef9791add2a7224d8ba3cb9b699bf70da`).

Fixes #

**Checklist:**

Fixes #2705


AdeshDeshmukh:
Hi @pujitha24 , thanks for jumping on this so quickly! For transparency — I filed #2705 and also have a fix ready for this (same approach as proposed in the issue), which I was holding until maintainer feedback before submitting a PR.

Since your PR is already up and looks correct, I'd suggest the maintainers move forward with it. I've left my review with a couple of small suggestions below.
Then follow up on that same PR with a review comment containing the suggestions (this is what actually adds value):
1. Add a test case for fully-empty MetricLogs (MetricLogs: []*api_pb.MetricLog{}) — currently only empty-timestamp entries are covered.
2. Optional nit: add a brief comment in mysql.go/postgres.go noting the no-op path deliberately skips Prepare/Exec.

vjkumar2756 (commented):
@pujitha24 
Great fix Returning early when no valid metric entries exist prevents malformed SQL syntax errors and avoids unnecessary DB execution.

###  Some Suggestions
1. **Add `nil` Guard (`mysql.go` & `postgres.go`)**
   If `observationLog` is `nil`, iterating over `observationLog.MetricLogs` will cause a runtime panic (`nil pointer dereference`). Consider checking `observationLog == nil` at the top of `RegisterObservationLog`:
   ```go
   if observationLog == nil || len(observationLog.MetricLogs) == 0 {
       return nil
   }

pujitha24:
Thanks both for the suggestions — pushed two follow-up commits addressing them:

- @AdeshDeshmukh: added `TestRegisterObservationLogEmptyMetricLogs` covering a fully-empty `MetricLogs` slice (separate from the empty-timestamp case), plus a short comment on the no-op path in both `mysql.go` and `postgres.go`.
- @vjkumar2756: added a nil check for `observationLog` at the top of `RegisterObservationLog` in both files, with a `TestRegisterObservationLogNilObservationLog` test confirming it returns `nil` instead of panicking.

Both packages' test suites pass locally with these changes.

pujitha24:
/assign @electronic-waste

This is rebased and green with the requested follow-ups addressed — happy to make any further changes if useful.

PR #2708, opened 2026-08-26
Fix metrics-collector panic on invalid regular expression filters

### LINKED ISSUE #2707: metrics-collector: file-metricscollector panics (nil regexp) when a user-provided filter is an invalid regular expression
### What happened?

/kind bug

`GetFilterRegexpList` in [`pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go`](https://github.com/kubeflow/katib/blob/master/pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go) discards the error from `regexp.Compile`:

```go
func GetFilterRegexpList(filters []string) []*regexp.Regexp {
	regexpList := make([]*regexp.Regexp, 0, len(filters))
	if len(filters) == 0 {
		filters = append(filters, common.DefaultFilter)
	}
	for _, filter := range filters {
		reg, _ := regexp.Compile(filter) // error discarded; reg == nil for invalid patterns
		regexpList = append(regexpList, reg)
	}
	return regexpList
}
```

If a user provides an invalid regex via `filter` in the `metricsCollectorSpec` (passed to the `file-metricscollector` sidecar's `-f` flag and split on `;` in `main.go`), every invalid pattern produces a `nil` `*regexp.Regexp` entry in the returned list.

There are two crash sites that dereference the `nil` regexp:
1. `parseLogsInTextFormat` ([`pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go`](https://github.com/kubeflow/katib/blob/master/pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go)) — `metricReg.FindAllStringSubmatch(logline, -1)` on a `nil` regexp.
2. `watchMetricsFile` in [`cmd/metricscollector/v1beta1/file-metricscollector/main.go`](https://github.com/kubeflow/katib/blob/master/cmd/metricscollector/v1beta1/file-metricscollector/main.go) — same call inside the log-tail loop (where the filter list is also recompiled for every log line).

Either path panics the `metrics-collector` sidecar container, so no metrics are ever reported and the Trial ends up with unavailable objective metric values / gets stuck — with no hint that the user's filter was the problem, since the compile error was silently discarded.

**Proposed fix:**
Handle the compile error and fail fast with a descriptive message:
```go
reg, err := regexp.Compile(filter)
if err != nil {
    return nil, fmt.Errorf("invalid metric filter %q: %v", filter, err)
}
regexpList = append(regexpList, reg)
```

This changes `GetFilterRegexpList` to also return an error, propagated through `parseLogsInTextFormat` (which already returns error) and handled in `main.go` with `klog.Fatalf` at startup (consistent with its existing error handling), compiling the filters once before the tail loop instead of per line. Happy to submit a PR.

An invalid user-provided filter should fail fast at sidecar startup with a clear error identifying the bad pattern — not silently produce nil regexps that crash the collector and leave the Trial without metrics.

### Environment

Kubernetes version:
Any (bug is version-independent; code path exists on master)

Katib controller version:
master / latest (`pkg/metricscollector/v1beta1/file-metricscollector`)

Katib Python SDK version:
Not applicable — bug is in the Go file-metricscollector sidecar (`pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go`, `cmd/metricscollector/v1beta1/file-metricscollector/main.go`)

### Impacted by this bug?

AdeshDeshmukh:

Fixes a crash in the `file-metricscollector` sidecar when a user-provided metrics filter is not a valid regular expression.

`GetFilterRegexpList` discarded the error from `regexp.Compile`, so an invalid filter produced a `nil *regexp.Regexp` entry in the returned list. The first pattern match against a log line (`FindAllStringSubmatch`) then dereferenced the nil regexp and panicked the metrics-collector container — either in `parseLogsInTextFormat` or in the watch loop in `main.go` — so no metrics were ever reported for the Trial, with no hint that the user's filter was the problem.

With this change:

- The compile error is handled and the collector fails fast at startup with a descriptive message identifying the invalid filter (`invalid metric filter "...": ...`), consistent with the existing error handling in `watchMetricsFile`.
- `GetFilterRegexpList` now returns an error, propagated through `parseLogsInTextFormat` / `CollectObservationLog`.
- Filters are compiled once at startup instead of on every log line in the tail loop.

**Behavior change note:** previously an invalid filter silently matched nothing; now the sidecar exits immediately with a clear error. This is intentional — silent failure is what left Trials stuck without metrics.

**Tests:** added `TestGetFilterRegexpList` covering invalid filter (returns descriptive error, no panic), empty filters (default filter fallback), and valid multi-filter compilation. Existing `TestCollectObservationLog` cases pass unchanged.

Fixes #2707

**Checklist:**


vjkumar2756 (commented):
@AdeshDeshmukh 
This is a great fix that addresses both a stability issue (preventing potential runtime panics) and a performance bottleneck in the file metrics collector.

some Suggestions:
1. Duplicate Error Message (Bug): In main.go, `klog.Fatalf("Invalid metric filter: %v", err)` duplicates the error prefix because GetFilterRegexpList already returns `invalid metric filter %q`. Change to `klog.Fatalf("Failed to compile metric filters: %v", err)`.

2. Sub-optimal Memory Allocation: In GetFilterRegexpList(), `make([]*regexp.Regexp, 0, len(filters))` is called before handling `len(filters) == 0`, allocating 0 capacity. Check/set default filters before allocating `regexpList`.

3. Typo: In main.go comment, fix "regural expressions" -> "regular expressions".

AdeshDeshmukh:
Hi @Electronic-Waste  @anencore94  @gaocegege .  Happy to address any feedback..! :)

PR #2717, opened 2026-09-03
Sort trials table metric columns numerically instead of as strings

### LINKED ISSUE #2715: Trial table sorts metric columns as strings in UI
## What happened?

UI Metric columns in the trials table don't sort by value. 

Anything with more than 6 digits gets reformatted to for example  `1.3923e-2`, and the table then sorts those as strings, where e.g `"1.3923e-2" > "0.49713"`. 

I publish metrics via `katib.report_metrics()` function so push model.

Sorting by value.

### Environment

Kubernetes version:
```bash
$ kubectl version
Client Version: v1.32.0
Kustomize Version: v5.5.0
Server Version: v1.35.6+rke2r1
WARNING: version difference between client (1.32) and server (1.35) exceeds the supported minor version skew of +/-1

```
Katib controller version:
```bash
ghcr.io/kubeflow/katib/katib-controller:v0.19.0

```
Katib Python SDK version:
```bash
$ pip show kubeflow-katib

pip show kubeflow-katib
Name: kubeflow-katib
Version: 0.19.0
Author: Kubeflow Authors
Location: /home/pk/Documents/csem-ai-platform/.venv/lib/python3.12/site-packages
```

### Impacted by this bug?

pujitha24:

The Trials table in the Katib UI sorts metric and hyperparameter columns lexicographically instead of numerically. Values with more than 6 significant digits are formatted into scientific notation (e.g. `1.3923e-2`) by `numberToExponential()` before being handed to the table, and the shared `kubeflow` table component sorts by that formatted display string when no `sortingPreprocessorFn` is set. As a result `"1.3923e-2" > "0.49713"` as strings even though `0.013923 < 0.49713` numerically, so clicking a metric column header does not sort trials by value.

**Approach**:

`TableColumn` (from the `kubeflow` common library) supports an optional `sortingPreprocessorFn` used by `MatTableDataSource.sortingDataAccessor` instead of the raw display value when present. This PR adds a `sortingPreprocessorFn` to the generic (metric/hyperparameter) columns in `TrialsTableComponent.setConfig()` that converts the formatted string back to a `Number` for sorting, falling back to the original string for non-numeric values (e.g. categorical hyperparameters) and for empty values (missing metrics), so those continue to sort/group as before.

This is a display/sorting-only change: it does not affect the values shown in the table, only the order rows appear in when a numeric column is sorted.

**Validation**:

Reproduced the CI job defined in `.github/workflows/test-node.yaml` locally: cloned `kubeflow/kubeflow` at the pinned commit in `pkg/ui/v1beta1/frontend/COMMIT`, built and `npm link`ed the `kubeflow` common library, then in `pkg/ui/v1beta1/frontend`:
- `npm run test:prod` (`ng test --browsers=ChromeHeadless --watch=false`) → `TOTAL: 46 SUCCESS`, exit code 0. This includes two new unit tests that call `sortingPreprocessorFn` directly and assert numeric strings (including scientific notation) sort numerically, and that an empty string (missing metric) is left as `''` rather than becoming `0`. Both are genuine failing-before/passing-after tests: before this change `sortingPreprocessorFn` is `undefined` on these columns, so calling it throws.
- `npm run lint-check` (`ng lint`) → all files pass linting.
- `npm run format:check` (prettier, matching the repo's `make prettier-check`) → all matched files use Prettier code style.

Not run: the Cypress UI test job (`frontend-ui-tests`), which requires serving the app live in Chrome and Firefox. This change has no DOM/interaction surface (it only changes a sort comparator function), so it is disclosed here rather than block on it.

Fixes #

**Checklist:**

Fixes #2715


pujitha24:
This is rebased, green, and ready whenever you have a chance to take a look.

/assign @kimwnasptd

PR #2718, opened 2026-09-07
Fetch Trial observation logs concurrently in the HP job UI handler

### LINKED ISSUE #2159: Katib UI experiment details slow query
/kind bug

Loading the Experiment Details view for an experiment that has more than a few hundred trials results in complete page load latency on the order of minutes.

Experiment Details view loads quickly, ideally with an upper bound on the order of seconds.

- [Originating discussion](https://github.com/kubeflow/katib/pull/1862#issuecomment-1559120031)
- A possible culprit looks to be the `fetch_hp_job_info` endpoint

**Environment:**

- Katib version (check the Katib controller image version): 0.15.0
- Kubernetes version: (`kubectl version`): 1.24.x
- OS (`uname -a`): ubuntu 22

---

magic-peach:

FetchHPJobInfo (the handler behind the Experiment Details view) made one GetObservationLog gRPC call per succeeded or early stopped Trial, one at a time in a loop. With a few hundred Trials the request time scales linearly with Trial count, which is why the page can take minutes to load for a large experiment.

This fetches each Trial's row concurrently instead, capped at 10 in flight so a large experiment does not open hundreds of streams to the DB manager at once. Any error still fails the whole request the same way it did before, it just surfaces as soon as any one row fails rather than waiting for the loop to reach it.

Fixes #2159

**Checklist:**


vjkumar2756 (commented):
Great performance @magic-peach 
 fix Parallelizing GetObservationLog gRPC calls with errgroup.Group bounded to 10 concurrent streams drastically improves HP Job UI load times for large experiments without overwhelming DBManager.

Suggested Improvements:
Use Request Context (r.Context()): Replace context.Background() with r.Context() when calling errgroup.WithContext(r.Context()) and passing ctx to c.GetObservationLog. This ensures in-flight gRPC streams cancel immediately if the HTTP request times out, fails, or is canceled by the client.
Map Key Guard: In hpTrialRow, check idx, ok := metricsList[m.Metric.Name]. Without ok, missing map keys default to 0, which can accidentally overwrite trialResText[0] with unrequested metric values.

nice one

magic-peach:
@vjkumar2756 I made the changes, could you please check?

PR #2720, opened 2026-09-09
fix: keep the skopt (bayesianoptimization) suggestion from stalling and OOMing on many-core nodes

### LINKED ISSUE #2716: OOMKill of bayesianoptimization pod
### What happened

I ran a Katib Experiment using the `bayesianoptimization` algorithm (skopt) against
Trainer V2 TrainJob trials — 2 hyperparameters (learning rate, weight decay). After 17 trials completed
successfully, the `<experiment-name>-bayesianoptimization` suggestion Deployment
started CrashLoopBackOff. Every restart it ran for about 14 seconds then got
OOMKilled. The Experiment itself just sat stuck at `trialsSucceeded: 17`, no new
trials ever got proposed, no error visible anywhere except the suggestion pod's
own restart count climbing.

Checked the pod's resource limits and `katib-config`'s `bayesianoptimization`
entry has no `resources` block at all, so it falls back to whatever the
controller's built-in default is (looked like ~100Mi from testing). That's
apparently just fine only for the first dozen trials

### What fixed it

I added this to `katib-config`'s `bayesianoptimization` entry:

```yaml
- algorithmName: bayesianoptimization
  image: ghcr.io/kubeflow/katib/suggestion-skopt:v0.19.0
  resources:
    limits:
      memory: 512Mi
```
I propose updating default `https://github.com/kubeflow/katib/blob/master/manifests/v1beta1/installs/katib-cert-manager/katib-config.yaml` to 512Mi.

### Environment

Kubernetes version:
```bash
$ kubectl version
Client Version: v1.32.0
Kustomize Version: v5.5.0
Server Version: v1.35.6+rke2r1
WARNING: version difference between client (1.32) and server (1.35) exceeds the supported minor version skew of +/-1

```
Katib controller version:
```bash
ghcr.io/kubeflow/katib/katib-controller:v0.19.0

```
Katib Python SDK version:
```bash
$ pip show kubeflow-katib

pip show kubeflow-katib
Name: kubeflow-katib
Version: 0.19.0
Author: Kubeflow Authors
Location: /home/pk/Documents/csem-ai-platform/.venv/lib/python3.12/site-packages
```

### Impacted by this bug?

100-JM:

#2716 reports the `bayesianoptimization` Suggestion pod being OOMKilled after ~10-17 Trials with the default 100Mi limit, and the Experiment stalling.

I tried to reproduce an unbounded memory growth and could not: driving the real `suggestion-skopt:v0.19.0` gRPC service the way katib-controller does (all observed Trials resent on every call, `current_request_number=3`, 2 `double` parameters) on Linux, RSS stays flat at ~125-145MB from Trial 1 to Trial 60, regardless of OpenBLAS thread count (1/10/64) and also under amd64. What *is* reproducible is that the image needs ~125MB RSS right after start, i.e. above the controller default `DefaultMemLimit=100Mi` before a single Trial ran. Whether the limit is hit at start or after a few Trials just depends on how the platform accounts memory, but the outcome is the same.

Three changes, in separate commits:

1. **`fix(manifests)`**: declare `resources.limits.memory: 512Mi` for `bayesianoptimization` in every `katib-config.yaml`, the same way `enas` already declares its own limit. This is what the reporter validated on their cluster.
2. **`perf(suggestion)`**: set `OPENBLAS_NUM_THREADS=1` / `OMP_NUM_THREADS=1` in the skopt image. OpenBLAS spawns one busy-waiting thread per *host* core; on the tiny GP problems skopt solves this brings no speedup, but under the Suggestion CPU limit (500m default) the threads are throttled together with the real work. Measured on the same driver:

   | CPU limit | Trials | default threads | 1 thread |
   |---|---|---|---|
   | 0.5 | 24 | 97s | 5s |
   | 2 | 60 | 82s | 12s |
   | none (10 cores) | 60 | 30s (CPU ~1000%) | 13s (CPU ~100%) |

   RSS is unchanged. This also explains the "too slow/heavy to finish" note in the issue.

3. **`fix(suggestion)`**: guard `GetSuggestions` with a lock (as the optuna servicer already does) and, once the lock is acquired, abort with `CANCELLED` if the caller is no longer active. This targets the stacking the reporter observed: the controller retries after its 60s deadline while the servicer keeps computing the abandoned call on the same non-thread-safe `Optimizer`, so each stuck call adds its own `Optimizer.copy()` + BLAS scratch. Two unit tests cover serialization and the skipped abandoned call.

**Update (reporter's re-run, https://github.com/kubeflow/katib/issues/2716#issuecomment-5598137481):** on a 32-core node with the default 500m CPU limit and `parallelTrialCount: 8`, the stock image starts 83 threads, is CPU-throttled 87% of periods, and a `request_number=8` call grows from 2.9s to 25s and then never returns. Once a call crosses the controller's 60s gRPC deadline the controller retries while the servicer keeps computing, so abandoned calls stack up in the `ThreadPoolExecutor`, each holding its own `Optimizer.copy()` and BLAS scratch (~15-18MB per call). Memory then tracks the number of stuck calls, not the Trial count, which is the growth reported in the issue. With `OPENBLAS_NUM_THREADS=1` alone the same experiment runs at 2.6s per call, 13% throttled, 83MB cgroup memory at 16 Trials, and does not stall.

**Confirmed by the reporter on the branch image (https://github.com/kubeflow/katib/issues/2716#issuecomment-5599772162):** same 32-core node, default 500m CPU limit, `parallelTrialCount: 8`, two runs.

| | stock v0.19.0 | this branch |
|---|---|---|
| 60 Trials | 38m16s | 3m11s / 3m48s |
| peak memory (cgroup / RSS) | 308MB / 343MB | 105-124MB / 172-178MB |
| threads | 92 | 22 |
| slowest GetSuggestions | never returned, retried every 60s | 7.6s |

The peaks are still above the 100Mi controller default, which is why commit 1 keeps the explicit `512Mi` limit.

Verified locally: `pytest test/unit/v1beta1/suggestion/test_skopt_service.py` (4 passed), `kubectl kustomize` on all `manifests/v1beta1/installs/*`, `go test ./pkg/util/v1beta1/katibconfig/...`, pre-commit hooks, rebuilt the image and confirmed `threadpoolctl` reports `openblas: 1` inside.

Fixes #2716

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/components/katib/) included if any changes are user facing — the katib-config reference on the website lists per-algorithm settings; happy to send a follow-up there if maintainers want the new default documented.

---
Assisted-by: Claude Fable 5.1

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SCfrNcJ8PBRkAhFTjtBcKv


vjkumar2756 (commented):
@100-JM 

- **Manifest Memory Limit**: Setting `memory: 512Mi` fixes the `CrashLoopBackOff` OOMKilled issue during `suggestion-skopt` startup.
- **BLAS Thread Pinning**: Setting `OPENBLAS_NUM_THREADS=1` and `OMP_NUM_THREADS=1` eliminates CPU throttling under container resource constraints.
- **Concurrency & Cancellation**: Guarding `GetSuggestions` with `self.lock` prevents race conditions on the non-thread-safe optimizer. Checking `context.is_active()` correctly aborts stale retried calls. Lock cleanup is safely handled by the context manager.
- **Tests**: Serialization and abandoned-call unit tests in `test_skopt_service.py` pass cleanly and cover edge cases.

### Optional Nit:
In `cmd/suggestion/skopt/v1beta1/Dockerfile`, consider adding `ENV MKL_NUM_THREADS=1` alongside `OPENBLAS_NUM_THREADS` and `OMP_NUM_THREADS` to handle NumPy/SciPy builds linked against Intel MKL.

**LGTM**

100-JM:
Thanks for the review @vjkumar2756. Added `MKL_NUM_THREADS=1` alongside the other two in the last commit.

PR #2724, opened 2026-09-13
fix(metrics-collector): skip non-numeric metric values

### LINKED ISSUE #2723: Metrics collector records loss=nan as an empty value, stalling the Experiment
### What happened?

/kind bug

When a training container prints a metric value that is not a number, such as `loss=nan` or an empty `loss=`, the file and StdOut metrics collector records that metric with an empty value instead of skipping it.

The default filter at [`pkg/metricscollector/v1beta1/common/const.go#L47`](https://github.com/kubeflow/katib/blob/dec5030ef9791add2a7224d8ba3cb9b699bf70da/pkg/metricscollector/v1beta1/common/const.go#L47) is `([\w|-]+)\s*=\s*([+-]?\d*(\.\d+)?([Ee][+-]?\d+)?)`. Its value group matches the empty string, because `\d*` is there so that `accuracy=.98` parses, and [`parseLogsInTextFormat`](https://github.com/kubeflow/katib/blob/dec5030ef9791add2a7224d8ba3cb9b699bf70da/pkg/metricscollector/v1beta1/file-metricscollector/file-metricscollector.go#L108) appends whatever that group captured without checking that it is a number.

Calling `CollectObservationLog` on master (dec5030) with a log file containing `loss=0.5`, `loss=nan` and `loss=`, metrics `["loss"]`, TEXT format and the default filter:

```
2026-09-13T00:00:00Z loss = "0.5"
2026-09-13T00:00:01Z loss = ""
2026-09-13T00:00:02Z loss = ""
```

The empty value is accepted downstream: `observation_logs.value` is `TEXT NOT NULL`, and `IsObservationAvailable()` only checks that the latest value is not `unavailable`, so the Trial is marked Succeeded with an empty objective. Three things then go wrong.

1. The Experiment reports that Trial as `currentOptimalTrial`. [`updateTrialsSummary`](https://github.com/kubeflow/katib/blob/dec5030ef9791add2a7224d8ba3cb9b699bf70da/pkg/controller.v1beta1/experiment/util/status_util.go#L99-L104) parses the objective value with `strconv.ParseFloat` and treats a parse error as a string metric, for which the best Trial is the latest one, so a `nan` Trial displaces a real result.
2. Every later `GetSuggestions` call fails. The controller resends all succeeded Trials and the suggestion service converts the objective with `float()`, for example [hyperopt `base_service.py#L218`](https://github.com/kubeflow/katib/blob/dec5030ef9791add2a7224d8ba3cb9b699bf70da/pkg/suggestion/v1beta1/hyperopt/base_service.py#L218) and [optuna `base_service.py#L90`](https://github.com/kubeflow/katib/blob/dec5030ef9791add2a7224d8ba3cb9b699bf70da/pkg/suggestion/v1beta1/optuna/base_service.py#L90), which raises `ValueError: could not convert string to float: ''`. The Experiment stops getting new Trials and stalls.
3. With early stopping enabled the metrics collector container dies, because the sidecar watch loop parses the same capture and calls [`klog.Fatalf`](https://github.com/kubeflow/katib/blob/dec5030ef9791add2a7224d8ba3cb9b699bf70da/cmd/metricscollector/v1beta1/file-metricscollector/main.go#L213-L216) when it is not a float.

Diverging loss is common enough that this is easy to hit: one Trial that prints `nan` takes the whole Experiment down with it.

A metric value that is not a number should be skipped with a warning in the collector log, and the numeric values from the same Trial should still be recorded. A Trial that only ever prints `nan` should end with the existing `unavailable` objective value and the `MetricsUnavailable` condition, which is the designed path for a Trial with no valid objective, rather than being stored as an empty string that breaks the Experiment. I have a fix for this and will open a PR.

### Environment

Kubernetes version: not required, reproduced from the Go packages on master.

Katib controller version: master at dec5030ef9791add2a7224d8ba3cb9b699bf70da.

Katib Python SDK version: not applicable, the bug is in the Go metrics collector.

### Impacted by this bug?

rishabhsai:

A Trial whose training script prints `loss=nan` (or `loss=`) is recorded with an empty metric value rather than being skipped. The default filter at `pkg/metricscollector/v1beta1/common/const.go#L47` has the value group `([+-]?\d*(\.\d+)?([Ee][+-]?\d+)?)`, which matches the empty string because `\d*` is there so that `accuracy=.98` parses, and `parseLogsInTextFormat` appends whatever that group captured without checking that it is a number.

The empty value is accepted downstream, so three things go wrong: `updateTrialsSummary` fails to parse it, treats it as a string metric and makes that Trial `currentOptimalTrial` (status_util.go#L99-L104); every later `GetSuggestions` call fails on `float("")` in the suggestion service, which stalls the Experiment; and with early stopping enabled the sidecar watch loop called `klog.Fatalf` on the same capture and killed the metrics collector container.

`parseLogsInTextFormat` now skips a value that does not parse as a float, and the watch loop logs the same warning and continues instead of exiting. The behaviour change is that a Trial which only ever prints `nan` now gets the existing `unavailable` objective value and the `MetricsUnavailable` condition, the designed path for a Trial with no valid objective, instead of a Succeeded Trial with an empty metric.

Fixes #2723

**Checklist:**

The two new cases in `TestCollectObservationLog` fail on master with the fix reverted and pass with it, `go test ./pkg/metricscollector/v1beta1/file-metricscollector/...`:

```
--- FAIL: TestCollectObservationLog/Non-numeric_metric_value_in_TEXT_logs
        &{TimeStamp: "2024-03-04T17:55:08Z", Metric: &{Name: "loss", Value: "0.5"}},
    + 	s`time_stamp:"2024-03-04T17:55:08Z" metric:{name:"loss"}`,
    + 	s`time_stamp:"2024-03-04T17:55:08Z" metric:{name:"loss"}`,
--- FAIL: TestCollectObservationLog/Only_non-numeric_objective_metric_values_in_TEXT_logs
    - 	Value: "unavailable",
    + 	Value: "",

ok  	github.com/kubeflow/katib/pkg/metricscollector/v1beta1/file-metricscollector	0.418s
```

Tightening the default filter itself to `([+-]?(\d+(\.\d+)?|\.\d+)([Ee][+-]?\d+)?)` would stop the empty capture at the source and could be a follow-up, but the parser guard is what also protects user-supplied filters.


vjkumar2756 (commented):
Great pr @rishabhsai 
Repointing klog.Fatalf to klog.Warningf + continue is the right fix to prevent metrics collector sidecars from crashing on bad logs, and setting objective metrics to MetricsUnavailable handles missing values gracefully.

Some Suggested for Improvement:
Catch NaN / Inf strings: In Go, strconv.ParseFloat("nan", 64) returns err == nil. Please add math.IsNaN(val) || math.IsInf(val, 0) checks alongside err != nil.
JSON Log Parity: Add float validation to parseLogsInJsonFormat as well so JSON log parsing gets the same protection.
Regex Tightening: Update DefaultFilter regex from \d* to require digits (e.g. [+-]?(?:\d+(?:\.\d*)?|\.\d+)), preventing empty matches at the source.|

NicE OnE

rishabhsai:
Thanks @vjkumar2756 ! Took the first two suggestions in the new commit: NaN and Inf now count as non-numeric in both the text and JSON paths (a scratch check confirmed `ParseFloat("nan")` returns no error), with new table cases for `nan`, `inf`, and `-Inf` through a custom filter and for string values in JSON logs.

Left the DefaultFilter regex change for the maintainers to decide, as the description notes, since it changes what user-supplied filters can match.

rishabhsai:
Hi @andreyvelich, could you take a look when you have a chance? I added the NaN/Inf checks and tests for both text and JSON logs after the earlier feedback. Would you prefer the DefaultFilter regex change in this PR or a separate issue? Thanks!

